PR #286, opened 2017-08-22
Fix secrets cuke encoding issue

### LINKED ISSUE #285: Fix secrets cuke encoding issue
One of our secrets cukes that deals with binary data is attempting to compare a UTF-8 encoded binary string with an ASCII-8BIT encoded string. This comparison fails, even though they have identical byte values.

jtuttle:
Closes #285

### DIFF
--- a/cucumber/api/features/step_definitions/authz_steps.rb
+++ b/cucumber/api/features/step_definitions/authz_steps.rb
@@ -82,7 +82,7 @@ Given(/^I create (\d+) secret values?$/) do |n|
 end
 
 Given(/^I create a binary secret value?$/) do
-  @value = Random.new.bytes(16).force_encoding(Encoding::UTF_8)
+  @value = Random.new.bytes(16)
   Secret.create resource_id: @current_resource.id, value: @value
 end
 
--- a/cucumber/api/features/step_definitions/authz_steps.rb
+++ b/cucumber/api/features/step_definitions/authz_steps.rb
@@ -82,7 +82,7 @@ Given(/^I create (\d+) secret values?$/) do |n|
 end
 
 Given(/^I create a binary secret value?$/) do
-  @value = Random.new.bytes(16)
+  @value = Random.new.bytes(16).force_encoding(Encoding::UTF_8)
   Secret.create resource_id: @current_resource.id, value: @value
 end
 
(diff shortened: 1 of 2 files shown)


cucumber/api/features/step_definitions/authz_steps.rb
```diff
@@ -82,7 +82,7 @@
 end
 
 Given(/^I create a binary secret value?$/) do
-  @value = Random.new.bytes(16)
+  @value = Random.new.bytes(16).force_encoding(Encoding::UTF_8)
```
apotterri: I don't think this follows the original intent of this step. The value of a Secret should be able to be *any* binary data, not just data that's representable as UTF-8 (see http://haacked.com/archive/2012/01/30/hazards-of-converting-binary-data-to-a-string.aspx/).

Could be that the problem is actually in the comparison step?

mizziness: Per my note in Slack, changing the actual comparison in the step gets:

```expected: "$)=b\xB1\xB9\x16H*\x03\xAF$\xBE\x86g\x8D"
           got: "$)=b\xB1\xB9\u0016H*\u0003\xAF$\xBE\x86g\x8D"```

Which means the encoding needs to be enforced one way or the other to actually compare the strings correctly.  I think this is okay considering the data is only being interpolated as UTF-8 for comparison purposes, but I could be wrong.

Thoughts?

dividedmind: Yes, binary is `Encoding::ASCII_8BIT`, which is exactly what you get from `Random#bytes`. That's also what the service should return.

apotterri: I don't think they should be compared as strings. I think they should be compared as arrays of bytes.

dividedmind:
I believe this might be an actual bug, and if so, this change hides it.

cucumber/api/features/step_definitions/authz_steps.rb
jtuttle: Ah yes I was considering doing an array of bytes comparison previously. That works as well. @dividedmind I think UTF-8 is RestClient's default way of returning binary data. Maybe we can change the encoding, though.

jtuttle: Check out this RestClient PR: https://github.com/rest-client/rest-client/pull/278

Looks like it used to always use `ASCII-8BIT` for binary data but they changed it to use `UTF-8` instead if the data falls within the appropriate character set. So this is likely an issue w/ RestClient if I'm reading this correctly.

PR #287, opened 2017-08-22
Fix incorrect RestClient encoding

### LINKED ISSUE #285: Fix secrets cuke encoding issue

jtuttle:
Closes #285

### DIFF
--- a/cucumber/api/features/step_definitions/response_steps.rb
+++ b/cucumber/api/features/step_definitions/response_steps.rb
@@ -53,5 +53,5 @@ end
 Then(/^the binary result is preserved$/) do
   expect(@result).to be
   expect(@result.headers[:content_type]).to include("application/octet-stream")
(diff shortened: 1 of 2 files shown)


cucumber/api/features/step_definitions/response_steps.rb
```diff
@@ -53,5 +53,5 @@
 Then(/^the binary result is preserved$/) do
   expect(@result).to be
   expect(@result.headers[:content_type]).to include("application/octet-stream")
```
apotterri: Can we check to make sure `content_type` is exactly `application/octet-stream`, here and in the step above? It shouldn't ever be anything else, but let's not give RestClient the chance to muck with the headers.

PR #291, opened 2017-08-24
Add recaptcha to hosted signup  4287

jtuttle:


docs/_includes/sign-up/sign-up-form.html
```diff
@@ -22,6 +22,11 @@ <h2 id='hosted-signup'>Sign Up to Evaluate Conjur</h2>
         <div class="help-block with-errors"></div>
     </div>
 
+    <div class="form-group required">
+      <div id="recaptcha" class="g-recaptcha" data-sitekey="6Lddji0UAAAAAAiy2M--2e_BNcPop7b_kbJKv5tp"></div>
```
jvanderhoof: Wish we could get this out of code :( Could we pass the site key in through at compilation time and store it in Conjur (or at least the `secrets.yml` file)?

docs/get-started/evaluate-conjur.md
```diff
@@ -2,6 +2,8 @@
 title: Evaluate Conjur
 layout: page
 section: get-started
+custom_js:
+- https://www.google.com/recaptcha/api.js
```
jvanderhoof: very nice!

docs/javascript/conjur-account.js
```diff
@@ -42,23 +42,48 @@ function displayAccountCredentials(email, account_id, api_key) {
   });
 }
 
+function displayError(formField, message) {
+  // Manually inserting the error here because there doesn't seem to
+  // be a way to make bootstrap-validator display an error on-demand.
+  var errorElement =
+    '<ul class="list-unstyled"> \
+      <li style="color:#a94442">' +
```
jvanderhoof: Can we use a class instead of an explicit color hex here?

docs/_includes/sign-up/sign-up-form.html
jtuttle: Ah yeah I suppose we could work it in at compile time, but not sure if it's worth it since someone who really wanted to get this value only needs to load our site and poke around in the source. There's a separate secret key value for calling the API for token verification. I got the impression that this one was supposed to be semi-public, a bit like the Google Analytics key.

docs/javascript/conjur-account.js
jtuttle: Yes I imagine we can! Will get that hooked up.

docs/_includes/sign-up/sign-up-form.html
jvanderhoof: I'm fine keeping it in here for now.

PR #314, opened 2017-08-29
Add a CloudFormation template for Conjur CE

### LINKED ISSUE #313: Conjur CE has a CloudFormation template for stack creation
As an AWS user, I'd like a CloudFormation template to bring up a useable Conjur CE stack. The stack should have an EC2 instance running Conjur CE, connected to an RDS (Postgres) instance

apotterri:
Closes #313 

Adds the Conjur CE CloudFormation template 


aws/cft.yml
```diff
@@ -0,0 +1,141 @@
...
+
+  ConjurAdminPassword:
+    Type: String
+    NoEcho: true
+    Description: Password for Conjur admin account
+
+  DBAdminUser:
+    Type: String
+    Description: RDS instance admin user
+    Default: conjuradmin
```
dustinmm80: This is often a sensitive parameter as well, can we mark it `NoEcho`?

aws/cft.yml
```diff
@@ -0,0 +1,141 @@
...
+    Type: AWS::EC2::KeyPair::KeyName
+    Description: AWS keypair to use
+
+  VpcId:
+    Type: AWS::EC2::VPC::Id 
+    Description: VPC in which to launch Conjur instances
+
+  VpcSubnetIds:
+    Type: "List<AWS::EC2::Subnet::Id>"
+    Description: Subnets in which to launch Conjur and Postgres. Specify two, in different Availability Zones.
```
dustinmm80: There are neat ways to provide drop-down choices here, but we can do that in another issue/PR.

aws/cft.yml
```diff
@@ -0,0 +1,141 @@
...
+          - KeyName
+          - VpcId
+          - VpcSubnetIds
+
+Resources:
+  ConjurInstance:
+    Type: "AWS::EC2::Instance"
+    Properties: 
+      ImageId: !Ref ImageId
+      InstanceType: t2.medium
```
dustinmm80: `InstanceType` should be a param with a correct default, and used here. 

We can add a list of options later on, once we figure that out.

aws/cft.yml
```diff
@@ -0,0 +1,141 @@
...
+      KeyName: !Ref KeyName
+      SecurityGroupIds: [!Ref ConjurSecurityGroup]
+      SubnetId: !Select [0, !Ref VpcSubnetIds]
+      UserData:
```
dustinmm80: This works for now, but we'll want to clean it up. Thanks for getting this going!

aws/cft.yml
```diff
@@ -0,0 +1,141 @@
...
+          /opt/conjur/bin/set_admin_password "$admin_api_key" "${ConjurAdminPassword}"
+
+
+  RDSInstance:
+    Type: "AWS::RDS::DBInstance"
+    Properties:
+      Engine: postgres
+      EngineVersion: 9.3.17
+      AllocatedStorage: 5
+      DBInstanceClass: db.t2.medium
```
dustinmm80: Is this the default size we recommend for DB? if so great, i would have thought larger.

aws/cft.yml
```diff
@@ -0,0 +1,141 @@
...
+      MasterUsername: !Ref DBAdminUser
+      MasterUserPassword: !Ref DBAdminPassword
+      DBSubnetGroupName: !Ref DBSubnetGroup
+      VPCSecurityGroups:
+        - !Ref PostgresSecurityGroup
+
+  ConjurSecurityGroup:
```
dustinmm80: For our CFN templates, I think it's better to put the security groups up top, because that is the part most consumers will want to tweak. Different `CidrIp` for example. `Resources` should be in SGs, EC2, DB order i think.

aws/cft.yml
```diff
@@ -0,0 +1,141 @@
+AWSTemplateFormatVersion: "2010-09-09"
```
dustinmm80: Can we name this something more descriptive than `cft.yml`? perhaps `<project-root>/cfn/conjur.yml` or something more semantic?

apotterri: Yes, will do.

aws/cft.yml
apotterri: The problem I had with marking this `NoEcho` is that I couldn't find a way to ask the user to enter it a second time, and then verify that the two matched.

Only being able to enter it once, and not being able to see it, seems like a recipe for rendering it unusable (when you fat-finger the password, like I always do).

aws/cft.yml
apotterri: I get a list of subnets in a drop-down here. Do you not?

aws/cft.yml
apotterri: Agreed. I picked a minimal one here, mostly as a place-holder.

aws/cft.yml
apotterri: What about this needs cleaning up?

aws/cft.yml
apotterri: I'm not aware of us making a recommendation yet. But, you really think 5gig is too small?

aws/cft.yml
apotterri: OK, will do.

aws/cft.yml
ryanprior: Maybe I'm prone to over-explanation here because of unfamiliarity, but I would try harder to make the file names and directory structure descriptive. ie `<project-root>/aws-cloudfoundation/conjur-template.yml`

PR #327, opened 2017-08-30
Split internal and external docker image pushes

ghost:
This paves the way to have releasers push the external images. Releasers are tied a different Conjur layer than executors and have access to external publishing creds.


dustinmm80:
Merging this.

kgilpin:
Releasers. Awesome!

PR #328, opened 2017-08-30
Builds conjur-apidocs image before using it in website.sh

ryanprior:
This is a proposal to fix a problem where, when the `conjur-apidocs` image needs to be built before being used in `website.sh`, the build logs were getting prepended to `api.html` instead of being logged.

### DIFF
--- a/website.sh
+++ b/website.sh
@@ -1,9 +1,7 @@
 #!/bin/bash -ex
 
-docker-compose build apidocs # make sure we build first so that the
-                             # build output doesn't make it into
-                             # `docs/_includes/api.html`
-docker-compose run --rm apidocs > docs/_includes/api.html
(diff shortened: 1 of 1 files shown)


dividedmind:
Looks good, if a bit fragile (there can be a race there somewhere if docker cleanup runs in between). Can we also add `-T` to `docker-compose run` for good measure (to disable TTY allocation)?

(NB, I'm starting to think that maybe using `docker-compose run` for tasks like this isn't what the tool was intended for -- especially given it allocates a TTY by default. Is there a straightforward way to get the image SHA from `docker-compose build` and use that for straight `docker run`? That would ensure nothing like that could happen. I feel maybe `docker-compose run` is too much magic.)

dividedmind:
Merged in 4e64ac3335fec8c32f93a4cd270c8a1233f32ee6.

PR #345, opened 2017-09-05
Email and organization is pushed to hubspot

jtuttle:


docs/javascript/conjur-account.js
```diff
@@ -56,7 +56,13 @@ function displayError(formField, message) {
 }
 
 $(document).ready(function() {
-  var accountCookie = getAccountCookie();
+  var ipAddress = null;
+
+  $.getJSON("http://ipv4.jsonip.com/", function (data) {
```
apotterri: This feels like it could be problematic. Are we ok with the IP address not getting passed to the form (e.g. because this requests fails or is really slow)?

Is there a reason not to use the client's IP address that we see on the server?

jtuttle: Since we're making a request from Conjur to CPanel, isn't the IP address of the request going to be for the Conjur host?

Brian (Sterling) says we can do without the IP address, so I'll just remove this.

apotterri: I'm not sure what you mean about making the request from Conjur to CPanel. 

The form on this page gets submitted to CPanel. The client's address will be available to that entrypoint. It could fill in the IP address as part of processing the request.

Having said that, though, if we don't need the IP address, it's probably better to get rid of it altogether.

apotterri:
Also, this PR is against `pre-revert-master`. Is that really what you want? Are you expecting this to go to a staging site before it goes live?

docs/javascript/conjur-account.js
jtuttle: Sounds good, removed.

jtuttle:
Ah nope, forgot we're on `master` now!

PR #374, opened 2017-09-13
An api key can be reset for an account 

jtuttle:
### Testing Instructions
1) modify the CPanel `dev/docker-compose.yml` entry for `possum_cpanel` as follows:
    - change `ports` value to `3000:3000`
1) modify the CPanel `dev/docker-compose.yml` entry for `possum` as follows:
    - change `image` to `conjur-dev` (run Conjur’s `build.sh` if you haven’t already)
    - add `entrypoint: conjurctl`
    - add `volumes` value `../../conjur:/src/conjur-server`
1) run `dev/start.sh`
1) follow Conjur README to set up conjur_docs container
1) visit eval page on localhost:4000
1) create a hosted Conjur account, take note of the email you used and the API key
1) use a cookie manager to delete your `accounts` cookie
1) refresh the page
1) sign up again w/ the same email address
1) observe the reset API key form appears
1) click reset
1) observe the new API key is displayed


app/controllers/accounts_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
   def create
     authorize :execute
 
-    api_key = Account.create account_name
+    api_key = Account.create account_name, params[:owner_id]
```
dividedmind: I think this should be `:owner` for consistency. It also probably needs a default (`current_user`?) here; in fact, I'm not sure it should be configurable at all.

jvanderhoof:
I think we should change the color of the `Cancel` link to something a bit more visible.  It looks really washed out on my screen:

screen shot 2017-09-15 at 2 09 36 pm

app/controllers/accounts_controller.rb
jtuttle: Oh that's true, I think using the current user would actually work quite nicely. I'll do that.

PR #399, opened 2017-09-29
A Code of Conduct exists for the Conjur project

ryanprior:
Adds a code of conduct to make clear our expectations for behavior.

https://opensource.guide/code-of-conduct/

Communities which maintain a friendly atmosphere and push out abusers are more sustainable and useful.

CONDUCT.md

We should have drills every so often to make sure that emails sent to the listed addresses are promptly addressed and that our response is appropriate.

Yes, this would be a good blog post.

### DIFF
--- a/CONDUCT.md
+++ b/CONDUCT.md
@@ -1,0 +1,131 @@
+# Conjur Code of Conduct
+
+Conjur exists to make DevOps easier, by making it easy for teams to communicate
+about and follow their security plans.
+
+Conjur is a piece of technology, but more importantly, it is a community.
+
+We believe that our mission is best served in an environment that is friendly,
+safe, and accepting; free from intimidation or harassment.
+
+Towards this end, certain behaviors and practices will not be tolerated.
+
+## tl;dr
+
+* Be respectful.
+* We're here to help: <abuse@conjur.org>
+* Abusive behavior is never tolerated.
+* Violations of this code may result in swift and permanent expulsion from the
+  Conjur community.
+
+## Scope
+
+We expect all members of the Conjur community, including paid and unpaid agents,
+administrators, users, and customers of CyberArk, to abide by this Code of
+Conduct at all times in all Conjur community venues, online and in person, and
+in one-on-one communications pertaining to Conjur affairs.
+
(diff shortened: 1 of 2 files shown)


izgeri:
@ryanprior just stopping by to note this repo now has a CHANGELOG - please merge in master and add an entry to the changelog following the guidelines in README before submitting this for a PR

CONDUCT.md
```diff
@@ -0,0 +1,131 @@
+# Conjur Code of Conduct
+
+Conjur exists to make DevOps easier, by making it easy for teams to communicate
```
izgeri: How about "to simplify DevOps" or something so that we don't use easy twice in one sentence?

CONDUCT.md
```diff
@@ -0,0 +1,131 @@
...
+behalf of CyberArk. (collectively, the "Service"). It also applies to behavior
+in the context of the Conjur Open Source project communities, including but not
+limited to public GitHub repositories, Slack and IRC channels, social media,
+mailing lists, and public events.
+
+This Code of Conduct is in addition to, and does not in any way nullify or
+invalidate, any other terms or conditions related to use of the Service.
+
+The definitions of various subjective terms such as "discriminatory", "hateful",
+or "confusing" will be decided at the sole discretion of the Conjur abuse team.
```
izgeri: this is the only place that "confusing" is referenced in the document, and I'm not sure how it's relevant. if we may ban people for being confusing, I think that's worth expounding on. if including "confusing" on this list isn't meant to imply that someone could be banned for being "confusing", then I think this line could use some rewording.

izgeri:
This closely follows the npm code of conduct. I like how explicit the `Unacceptable Behavior` section is in the [Citizen Code of Conduct](http://citizencodeofconduct.org/). Could we integrate that with our doc?

We are also really vague on what will happen if someone is said to be violating the CoC. The Rust CoC is much clearer on this. Why not integrate something like their `Moderation` section into our doc?

ryanprior:
I added an "unacceptable behavior" section. I think we should hold off on a "moderation" or similar section until we have some formal process in place for that. We don't have the same formal structure as Rust, eg with IRC channel mods and mailing list admins and so on.

PR #403, opened 2017-09-29
Proof read

kgilpin:


docs/integrations/puppet.md
```diff
@@ -27,9 +27,9 @@ See the module page on Puppet Forge or README.md on GitHub for complete document
 
 ### Establishing machine identity
 
-In most cases we recommend bootstrapping Conjur machine identities with
+For managing servers and VMs (but not containers) we recommend bootstrapping Conjur machine identities with
```
dividedmind: This begs the question "what about containers, then"? IMHO there should be a link here, or the wording left vague.

dustinmm80: I think this is fine since it's on the puppet page. puppet + containers isn't too common. Once we have a page or pages on containers we can link to it from this page.

dustinmm80:
Merging this, thanks Kevin!

PR #407, opened 2017-10-03
Clipboard button copies $ on install conjur page

### LINKED ISSUE #406: Clipboard button copies $ on install conjur page
The problem is that using `` ```sh-session`` to define the shell block does not set the `data-lang` attribute of the resulting HTML. This causes the block to be treated as a plain-text when performing the clipboard copy and the `$` is included.

jtuttle:
Closes #406 

#### Test steps
1) Set up docs container on localhost using README instructions
1) Confirm that copying from shell blocks on install conjur page with clipboard button does not include `$` characters.

### DIFF
--- a/docs/get-started/install-conjur.md
+++ b/docs/get-started/install-conjur.md
@@ -21,24 +21,24 @@ official Conjur containers on DockerHub.
 
 1. In your terminal, download the Conjur quick-start configuration:
 
-   ```sh-session
-   $ curl -o docker-compose.yml https://www.conjur.org/get-started/docker-compose.quickstart.yml
-   ```
+{% highlight shell %}
+$ curl -o docker-compose.yml https://www.conjur.org/get-started/docker-compose.quickstart.yml
+{% endhighlight %}
 
 1. pull all the required Docker images from DockerHub
 
    `docker-compose` can do this for you automatically:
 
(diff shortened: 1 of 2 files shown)


docs/get-started/install-conjur.md
```diff
@@ -21,24 +21,24 @@ official Conjur containers on DockerHub.
 
 1. In your terminal, download the Conjur quick-start configuration:
 
-   ```sh-session
-   $ curl -o docker-compose.yml https://www.conjur.org/get-started/docker-compose.quickstart.yml
-   ```
+{% highlight shell %}
```
ryanprior: Is this a stopgap? eg should we never use "sh-session" fences in documentation, or can we introduce code to make clipboard work correctly with those?

ryanprior:
the test instructions say to set up the docs container following the README, but the files changed don't touch the readme or the docs container- did you miss something?

docs/get-started/install-conjur.md
jtuttle: Yes, you can consider this a stopgap for now, just a quick way to fix it as it's urgent.

jtuttle:
There are no changes to the README, I only referenced it as the place where you can find instructions for setting up a test container with which to test my install page changes.

PR #415, opened 2017-10-05
#361 - Fixing page width issues on mobile views.

### LINKED ISSUE #361: Fix page width on tablet views
Screen widths smaller than 775px are scrolling horizontally. (API docs, Install pages, etc)
conjur_600px_bp

mizziness:
Closes #361 

This PR fixes the styling of the width and display of copy boxes, and fixes the page width issue on tablets and smaller.

There is an issue on some pages in mobile widths where content is pushed off the screen and is often inaccessible. 

These changes are very small and entirely in the css, so just give a look at the syntax and see if it makes sense.

Build this branch's version of the doc container by following the instructions in the README inside /docs/.

Before:
image

After:
image

https://jenkins.conjur.net/job/cyberark--conjur/job/361-code-block-page-width/


jtuttle:
The scrollbars are hidden on OSX until you actually try to scroll. Not sure if there's a way to make them appear all the time (or if we'd even want to), but worth considering. I believe it goes against Apple's design philosophy to do so. 😄

docs/_sass/_elements.scss
```diff
@@ -152,14 +152,15 @@
...
-	float: right;
+	position: absolute;
 	width: 30px;
 	height: 30px;
 	background: url("../../img/copy-to-clipboard.svg");
 	background-repeat: no-repeat;
 	background-position: center;
 	border: none;
 	cursor: pointer;
+  right: 24px;
```
izgeri: It looks like this file is using tabs for indentation, and that you've used spaces here (I see the same pattern below, too). Can this be updated to be more consistent?

docs/_sass/_global.scss
```diff
@@ -81,6 +81,13 @@ img {
 	flex: 1;
 }
 
+@media (max-width: 767px) {
+  .container {
```
izgeri: In this file it's just the one line that has spaces instead of tabs - can you please update this as well?

mizziness:
@izgeri  I went through and converted the tabs to 2 spaces - at some point, we'll want to do this in all files, but that's going to be a PITA.

PR #417, opened 2017-10-06
updating tutorial account names to consistently be myorg - removing i…

izgeri:
…nstances of dev and keg

also updated version to be 1 in response when loading root policy -- this is what I got as a response when I ran through this tutorial, please verify that this is also the case for you (I can see why it might be 2, but that's not what I got)


izgeri:
Handles https://github.com/cyberark/waffle/issues/2

ryanprior:
The API docs use `mycorp`, so if we're standardizing we might want to move those over too.

PR #420, opened 2017-10-10
fix the numbering for /getting-started/install-conjur

doodlesbykumbi:
fix the numbering for section steps at `/getting-started/install-conjur`


typaulhus:
I see that the numbering is changed in the .md, but it’s still showing as 1.1.1. on the page. Looks like the code blocks are closing the <ol> (when I remove them to check, the numbering works.) I wonder if the <ol> needs to be hardcoded, instead of letting markdown generate it?

doodlesbykumbi:
@typaulhus should work now

PR #439, opened 2017-10-12
Missing hubspot token is submitted as empty string

### LINKED ISSUE #438: Missing HubSpot token is submitted as empty string
Currently the HubSpot token is being submitted to the Conjur API from CPanel as the string "null". We instead want it to actually be submitted as an empty URL param.

jtuttle:
Closes #438.

Prevents hosted Conjur signup from submitting the string "null" for the hutk param.

I would just throw a console.log in conjur-account.js and confirm that the URL params are getting created with a blank string for hutk instead of "null".

### DIFF
--- a/docs/javascript/conjur-account.js
+++ b/docs/javascript/conjur-account.js
@@ -78,13 +78,13 @@ function submitAccountSignup() {
   }
 
   var hutk = getCookieValue("hubspotutk");
-  
+
   var payload =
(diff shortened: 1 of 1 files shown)


jtuttle:
On second thought, I think we should just remove this as we're not using HubSpot tracking right now. Can work it in later if we decide to. Making a new issue to that effect.

PR #440, opened 2017-10-13
Conjur uses rubygems CLI

### LINKED ISSUE #431: Conjur uses RubyGems CLI

jtuttle:
Closes #431 

Update Conjur to use the CLI gem from RubyGems instead of the possum git branch.

I tested by running `dev/start.sh` then manually installing the 6.0.0 CLI gem on the client container and executing CLI commands from there.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -64,7 +64,7 @@ group :development, :test do
   gem 'cucumber'
   gem 'aruba'
   gem 'rake_shared_context'
-  gem 'conjur-cli', '= 6.0.0.rc1'
+  gem 'conjur-cli', '~> 6.0'
   gem 'rails_layout'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -64,7 +64,7 @@ group :development, :test do
   gem 'cucumber'
   gem 'aruba'
   gem 'rake_shared_context'
-  gem 'conjur-cli', github: 'conjurinc/cli-ruby', branch: 'possum'
+  gem 'conjur-cli', '~> 6.0'
```
dustinmm80: Can we use `~> 6` instead? That way we don't have to update it again when we hit `6.1.0`

jtuttle: Fixed.

PR #467, opened 2017-10-24
conjurctl has a command to retrieve a user's API key

### LINKED ISSUE #464: conjurctl has a command to retrieve a user's API key

jtuttle:
Closes #464

### DIFF
--- a/bin/conjur-cli.rb
+++ b/bin/conjur-cli.rb
@@ -176,6 +176,20 @@ $ conjurctl account create myorg
       exec "rake account:delete[#{account}]"
     end
   end
+
+  cgrp.desc "Retrieve an account's API key"
+    cgrp.arg_name 'account'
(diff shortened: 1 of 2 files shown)


jtuttle:
Ah yeah you need to `conjurctl role retrieve-key "!:user:admin"` because I guess the ! is a special character in bash. Kind of annoying but hopefully this is something people will copy/paste from the docs. Not sure there's much we can do about that short of choosing a different special char to use for root stuff...

I don't think there's a good way to test `conjurctl` stuff. Maybe there's a way using backticks to run a shell command...or at the very least we could test the method that's getting called. I'll look into it a bit when I can.

izgeri:
merged because was able to get it to work with `conjurctl role retrieve-key '!:user:admin'` and its functionality appears to be tested elsewhere.

PR #470, opened 2017-10-25
Enhances conjurctl behavior

ryanprior:
* `role retrieve-key` can handle one or more arguments, errors if any retrievals fail
* `policy load` can handle one or more arguments, errors if any loads fail
* retrieve-key prints keys to stdout instead of stderr
* incorrect arguments to a command shows help (use strict GLI arguments)

In reviewing the recent addition of "retrieve-key" to conjurctl, I noticed a few things I'm inclined to change about how conjurctl works.

1. Use the dev environment instructions in readme.
1. `conjurctl role retrieve-key [some-role] [some-other-role]` and notice this retrieves each role's key
1. `conjurctl role retrieve-key` and notice this prints help for "role retrieve-key"
1. `conjurctl data-key generate lol wat` and notice this prints help for "data-key generate"

https://jenkins.conjur.net/job/cyberark--conjur/job/enhance-conjurctl-behavior/

### DIFF
--- a/bin/conjur-cli.rb
+++ b/bin/conjur-cli.rb
@@ -9,7 +9,7 @@ version File.read(File.expand_path("../VERSION", File.dirname(__FILE__)))
 # Attempt to connect to the database.
 def connect
   require 'sequel'
-  
+
   def test_select
     begin
       db = Sequel::Model.db = Sequel.connect(ENV['DATABASE_URL'])
@@ -18,7 +18,7 @@ def connect
       false
     end
   end
-  
+
   30.times do
     break if test_select
     $stderr.write '.'
@@ -26,7 +26,7 @@ def connect
   end
 
   raise "Database is still unavailable. Aborting!" unless test_select
-  
+
   true
 end
 
@@ -35,7 +35,7 @@ command :server do |c|
   c.desc 'Account to initialize'
   c.arg_name 'account'
   c.flag [ :a, :account ]
-    
+
   c.desc 'Policy file to load into the server'
   c.arg_name 'file_name'
   c.flag [ :f, :file ]
@@ -49,9 +49,9 @@ command :server do |c|
   c.default_value ENV['BIND_ADDRESS'] || '0.0.0.0'
   c.arg_name 'address'
   c.flag [ :b, :'bind-address' ]
-  
+
   c.action do |global_options,options,args|
-    exit_now! "No command arguments are allowed" unless args.empty?
+    help_now! unless args.empty?
 
     account = options[:account]
 
@@ -61,12 +61,12 @@ command :server do |c|
     if account
       system "rake account:create[#{account}]" or exit $?.exitstatus
     end
-    
+
     if file_name = options[:file]
       raise "account option is required with file option" unless account
       system "rake policy:load[#{account},#{file_name}]" or exit $?.exitstatus
     end
-    
+
     exec "rails server -p #{options[:port]} -b #{options[:'bind-address']}"
   end
 end
@@ -77,19 +77,19 @@ command :policy do |cgrp|
   cgrp.arg_name "account file_name"
   cgrp.command :load do |c|
     c.action do |global_options,options,args|
-      account = args.shift or exit_now! "Expecting account argument"
-      file_name = args.shift or exit_now! "Expecting file_name argument"
-      exit_now! "No additional command arguments are allowed" unless args.empty?
-  
+      account = args.shift or help_now! "Expecting account argument"
+      file_name = args.shift or help_now! "Expecting file_name argument"
+      help_now! unless args.empty?
+
       connect
 
       exec "rake policy:load[#{account},#{file_name}]"
     end
   end
-  
+
   cgrp.desc "Watch a file and reload the policy if it's modified"
   cgrp.long_desc <<-DESC
-To trigger a reload of the policy, replace the contents of the watched file with the path to 
+To trigger a reload of the policy, replace the contents of the watched file with the path to
 the policy. Of course, the path must be visible to the container which is running "conjurctl watch".
 This can be a separate container from the application server. Both the application server and the
 policy watcher should share the same backing database.
@@ -103,9 +103,9 @@ $ conjurctl watch /run/conjur/policy/load)"
   cgrp.arg_name 'account file_name'
   cgrp.command :watch do |c|
     c.action do |global_options,options,args|
-      account = args.shift or exit_now! "Expecting account argument"
-      file_name = args.shift or exit_now! "Expecting file_name argument"
-      exit_now! "No additional command arguments are allowed" unless args.empty?
+      account = args.shift or help_now! "Expecting account argument"
+      file_name = args.shift or help_now! "Expecting file_name argument"
+      help_now! unless args.empty?
 
       connect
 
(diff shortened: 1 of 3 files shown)


lib/tasks/role.rake
```diff
@@ -1,11 +1,12 @@
...
   desc "Retrieve the API key for the given role"
   task :"retrieve-key", [:role_id] => [:environment] do |t, args|
     begin
       creds = Credentials.first!(role_id: args[:role_id])
-      $stderr.puts creds.api_key
+      puts creds.api_key
     rescue Sequel::NoMatchingRow
-      $stderr.puts "Role '#{args[:role_id]}' not found"
+      $stderr.puts "error: #{args[:role_id]} is not a role"
+      puts
```
ryanprior: Note for reviewers: this "puts" is to make sure you get n lines of output when you request keys for n roles, no matter whether the roles exist. This is convenient for scripting.

kgilpin: It may be the case that the Role does not have any Credentials created yet. In that case, the user's expectation is that the Credentials will be created now.

I would suggest that you first try and find the role (it's an error if it doesn't exist), then use `Role#api_key`.

bin/conjur-cli.rb
```diff
@@ -131,8 +131,8 @@ def test_select
   DESC
   cgrp.command :generate do |c|
     c.action do |global_options,options,args|
-      exit_now! "No command arguments are allowed" unless args.empty?
```
izgeri: When you remove "No command arguments are allowed", when you run `conjurctl data-key generate lol wat` the output is
```
error: parse error: 
```

If, instead of your update, you just replace `exit_now` with `help_now` (and leave everything else the same), then it outputs
```
error: parse error: No command arguments are allowed
```
before the help info.

I think this is clearer. There are several commands impacted by similar updates.

bin/conjur-cli.rb
```diff
@@ -9,7 +9,7 @@
 # Attempt to connect to the database.
 def connect
   require 'sequel'
-  
+
```
dividedmind: Please try to configure your editor so that it only fixes whitespace on modified lines. This makes for a lot less noisy diffs. Thanks!

bin/conjur-cli.rb
dividedmind: I second that. BTW - since exiting because no args were expected looks like a common pattern - you might want to consider creating a utility method or a dedicated exception class.

bin/conjur-cli.rb
```diff
@@ -197,13 +197,13 @@ def test_select
   cgrp.desc "Retrieve a role's API key"
   cgrp.command :"retrieve-key" do |c|
     c.action do |global_options,options,args|
-      exit_now! "Must specify role ID" unless args.count == 1
-      
-      role_id = args.shift
-      
+      help_now! "specify at least one role id" unless args.count > 0
```
dividedmind: You can also stick 
```ruby
arguments :strict
subcommand_option_handling :normal
```
in the `App` class, declare arguments with `arg` instead of `arg_name` and have GLI check the arguments for you, eg:
```ruby
arg :role_id, :multiple
c.action { ... }
```

This will also automatically make it bail if arguments are given but not declared, so most of all these checks that are sprinkled around can come off.

lib/tasks/role.rake
dividedmind: You should put this kind of note in a comment instead of here.

(BTW, I agree with what @kgilpin said. Plus I don't like the wording change: `foo is not a role` suggests to me that `foo` exists although it's not a role, but something else.)

bin/conjur-cli.rb
```diff
@@ -18,40 +19,39 @@ def test_select
...
-  
+
   true
 end
 
 desc 'Run the application server'
 command :server do |c|
   c.desc 'Account to initialize'
-  c.arg_name 'account'
+  c.arg :account
```
dividedmind: This actually doesn't do anything at all for flags -- it's for command arguments -- you should remove them. You can put `arg_name` back in where it makes sense (it doesn't when you just repeat the flag; I'd do `arg_name :name` for account, `arg_name :path` for file, maybe `arg_name :ip` for address. Port should be pretty self-explanatory. (Do watch how the help text changes as you set these.)

bin/conjur-cli.rb
```diff
@@ -61,35 +61,36 @@ def test_select
...
   end
 end
 
 desc "Manage the policy"
 command :policy do |cgrp|
-  cgrp.desc "Load the policy from a file"
-  cgrp.arg_name "account file_name"
+  cgrp.desc "Load MAML policy from file(s)"
+  cgrp.arg :account
+  cgrp.arg_name "filename"
```
dividedmind: You should remove this `arg_name` -- not only it doesn't make sense here, but it also disables strict argument handling.

bin/conjur-cli.rb
```diff
@@ -100,13 +101,13 @@ def test_select
 
 $ conjurctl watch /run/conjur/policy/load)"
   DESC
-  cgrp.arg_name 'account file_name'
+
+  cgrp.arg :account
+  cgrp.arg_name "filename"
```
dividedmind: Ditto.

bin/conjur-cli.rb
```diff
@@ -195,15 +189,15 @@ def test_select
...
-      
-      role_id = args.shift
-      
+
       connect
-      
-      exec "rake role:retrieve-key[#{role_id}]"
+
+      args.each do |id|
+        print `rake role:retrieve-key[#{id}]`
```
dividedmind: Why the change to `print`?

ryanprior: `exec "rake"` replaces the Ruby process with the rake process, while ``print `rake` `` launches a subprocess. In order to allow the loop to finish, I made the change.

bin/conjur-cli.rb
```diff
@@ -195,9 +193,9 @@ def test_select
 
       connect
 
-      args.each do |id|
+      args.map { |id|
         print `rake role:retrieve-key[#{id}]`
-      end
+      }.all?
```
dividedmind: I think you forgot to change to `system` here. ```map{print `foo`}.all?``` will always be false.

bin/conjur-cli.rb
```diff
@@ -75,16 +75,15 @@ def test_select
...
     c.action do |global_options,options,args|
       account, *file_names = args
       connect
 
-      file_names.each do |file_name|
-        exec "rake policy:load[#{account},#{file_name}]"
-      end
+      file_names.map { |file_name|
+        system "rake policy:load[#{account},#{file_name}]"
+      }.all?
```
dividedmind: You still need to use `exit_now!` (or just `fail`) -- it appears that GLI doesn't look at the return value of an action.

bin/conjur-cli.rb
```diff
@@ -152,25 +149,21 @@ def test_select
...
 
   cgrp.desc "Delete an organization account"
-  cgrp.arg_name 'account'
+  cgrp.arg :account
   cgrp.command :delete do |c|
     c.action do |global_options,options,args|
-      account = args.shift or exit_now! "Expecting account argument"
-      exit_now! "No additional command arguments are allowed" unless args.empty?
-      
+      account=args.first
```
izgeri: the spacing is different here than elsewhere

izgeri:
I "approved changes" because the issues that I brought up have been resolved - I am not merging because it's not clear to me which issues that other devs brought up have not yet been resolved, so I will leave that to the others who have not yet approved the changes.

PR #481, opened 2017-11-14
Increases maximum POST request size to 10MiB

### LINKED ISSUE #480: Loading a large policy (~4k roles) results in server error
Certain large valid policies which load cleanly in Conjur Enterprise or as a default root policy (using `conjurctl server -f`) will result in a server error when loaded using the Conjur CLI tool.

## Steps to reproduce
1. Set up Conjur server, Postgres, and Conjur client as per https://github.com/cyberark/conjur-policy-generator/blob/loadtest/loadtest/test.v5.sh
2. Run:
   ```
   docker-compose -f docker-compose.v5.yml exec client \
     conjur policy load root /var/policies/2k.2k.0.yml
   ```

### Expected results
Policy loads cleanly, creates ~4k roles
### Actual results
```
error: 500 Internal Server Error
```
Server log: [error-log.txt](https://github.com/cyberark/conjur/files/1469523/error-log.txt)

## Other notes

* Loading the same policy (`2k.2k.0.yml`) as the default policy works fine. You can test this by changing `docker-compose.v5.yml` to make the Conjur server's `command` be:

   `command: server -a loadtest -f /var/policies/2k.2k.0.yml`

* Loading the same policy using Conjur Enterprise works, which you can verify using docker-compose.v4.yml and the Conjur Enterprise CLI.

* Loading a smaller policy also generated by the the same software works fine. You can test this by running:
   ```
   docker-compose -f docker-compose.v5.yml exec client \
     conjur policy load root /var/policies/500.500.0.yml
   ```

ryanprior:
Closes #480 

It configures Rails to increase the maximum POST request size to 10MiB. This allows loading of large policies. Previous (Rails default) max size was 4KiB.

During policy load testing, we found that a policy with ~4k roles would cause the server to give a 500 error. This change addresses that issue except in the case of vast policies (100k roles is still under 5MB.)

https://github.com/cyberark/conjur/blob/allow-large-policy-loads%2F480/config/initializers/limits.rb#L8

1. Run through steps to set up the Conjur dev containers as per README.md
1. checkout the `cyberark/conjur-policy-generator` repo (currently private, sorry non-CyberArgonauts)
1. load the `conjur-policy-generator/loadtest/policies/20k.20k.0.yml` policy

https://jenkins.conjur.net/job/cyberark--conjur/job/allow-large-policy-loads%252F480/

No, the current docs don't mention max policy size.

Yes, the API docs should be updated to indicate that 10 MiB is the max.


config/initializers/limits.rb
```diff
@@ -2,3 +2,7 @@
 def secrets_version_limit
   ( ENV['SECRETS_VERSION_LIMIT'] || 20 ).to_i
 end
+
+# Max size of policies that can be loaded (or data sent to POST endpoints in
+# general)
+Rack::Utils.key_space_limit = 1048576 * 10 # 10 MiB
```
apotterri: It wasn't clear to me why setting an attribute named `key_space_limit` would change the maximum policy size. A little more explanation here, describing how they're related, would be good.

apotterri:
How long does it take to generate the big policy and load it? It would really be good to have a test written for it.

ryanprior:
It takes about a minute to generate & load the policy. We can make a test for that.

jvanderhoof:
Looks good from my perspective.  Nice work Ryan!

config/initializers/limits.rb
```diff
@@ -2,3 +2,11 @@
...
+
+# This sets the default size of a Rack::Utils::KeySpaceConstrainedParam, which
```
apotterri: This comment should get combined with the one above it. Having them separate, and having a comment after the setting it's documenting, makes it harder to understand.

PR #488, opened 2017-11-20
Nginx proxy tutorial

ryanprior:
Closes CON-4461 (Example exists for using an nginx container to serve Conjur securely with TLS)

Provides an example of using Conjur with nginx to provide TLS for Conjur traffic.

Background context is provided in `docs/tutorials/integrations/nginx.md`, reproduced in part here:

>  it is crucial to set up TLS correctly when you deploy Conjur...

> Do not leave this to chance: even a small flaw can totally compromise your
> secrets. Setting up Conjur and TLS without the appropriate expertise is like
> packing your own parachute and jumping out of a plane.

The tutorial flow is in `docs/tutorials/integrations/nginx.md`, the supporting files are in `docs/integrations/tutorials/nginx/`

Run through the tutorial flow and check the container logs to verify that TLS is being used.

### DIFF
--- a/docs/tutorials/integrations/nginx.md
+++ b/docs/tutorials/integrations/nginx.md
@@ -135,11 +135,11 @@ network and proxying the traffic through to Conjur.
 ## Breaking Down the Tutorial
 
 These files show how the proxy setup works. Note that, while this tutorial uses
-Docker and NGINX, there's nothing technically important about using those
-particular technologies or even about using containers at all. You can replicate
-the same strategy using a different endpoint such as [HAProxy][haproxy-tls] and
-services that are run in VMs or on bare metal. For the purpose of a tutorial
-that can conveniently run on a laptop, we provide this setup:
+Docker containers and NGINX, there's nothing magic about those technologies. You
+can replicate the same strategy using a different endpoint such as
+[HAProxy][haproxy-tls] and services that run in virtual machines or on bare
+metal. To create a tutorial that can run as conveniently on a laptop as in the
+cloud, we provide this setup.
 
 [haproxy-tls]: https://www.haproxy.com/documentation/aloha/7-0/haproxy/tls/
 
@@ -192,7 +192,7 @@ not to the Internet or to the Local Area Network (LAN).
 proxy:
   image: nginx:1.13.6-alpine
   ports:
-    - "443:443"
+    - "8443:443"
   volumes:
     - ./default.conf:/etc/nginx/conf.d/default.conf:ro
     - ./tls/nginx.key:/etc/nginx/nginx.key:ro
@@ -211,7 +211,7 @@ certificate, and a private key related to the certificate. Explanation of those
 files follows below. The files are made accessible from the local file system
 for read-only access by the container.
 
-#### Production tip
+#### Production tips
 
 For the convenience of a tutorial, we automatically generate a self-signed
 certificate and provide it to the proxy service. For reasons that are described
@@ -224,6 +224,11 @@ authentic Conjur server. Your security team can provide certificates for your
 organization, or you can create a certificate for any domain or sub-domain you
 control with [certbot][certbot], which uses [Let's Encrypt][lets-encrypt] to
 provide certificates for no cost.
+
+To avoid conflicting with other services that might be running on the tutorial
+user's port 443, we remap the port to 8443. On the production machine, the port
+mapping should be changed to "443:443" instead of "8443:443".
+
 
 [certbot]: https://certbot.eff.org/
 [lets-encrypt]: https://letsencrypt.org/
@@ -243,8 +248,8 @@ tinkering. It is connected to the proxy service, allowing it to access Conjur
 via TLS.
 
 The "sleep" and "infinity" bits ensure that this container stays up for the
-duration of the demo. Without these options, the `conjurinc/cli5` images gives
-you an ephemeral stateless Conjur container that performs a single command and
+duration of the demo. Without these options, the `conjurinc/cli5` image gives
+you an ephemeral stateless client container that performs a single command and
 exits, a desirable behavior for common ops use cases.
 
 [postgres-image]: https://hub.docker.com/_/postgres/
@@ -272,7 +277,7 @@ ssl_certificate     /etc/nginx/nginx.crt;
 ssl_certificate_key /etc/nginx/nginx.key;
 ```
 
-This block gives NGINX its directions on how to perform TLS. ("ssl" is a name
+This block gives NGINX its directions on how to perform TLS ("ssl" is a name
 for an older standard for TLS and is still often used interchangeably.)
 
 The `certificate` is a public key, and the `certificate_key` is the
@@ -300,14 +305,18 @@ self-signed certificate. This allows you to use TLS in testing and staging, but
 it does not allow clients to automatically authenticate the identity of the
 Conjur server.
 
-In production, don't use a self-signed certificate. It's better than nothing,
-but it's not a sustainable security practice because you're going to have to
+#### Production tip
+
+Don't use a self-signed certificate in production. It's better than nothing, but
+it's not a sustainable security practice because you're going to have to
 manually verify that you're not talking to a man in the middle.
 
 Instead, ask your security team to provide a certificate signed by a trusted
-root and modify `default.conf` to use that certificate instead.
+root and use that instead.
 
-Let's look at what parts you might modify:
+#### Modifying tls.conf for development use
+
+These are blocks that you might want to change:
 
 ```ini
 [ dn ]
@@ -352,3 +361,14 @@ does to accomplish each step.
 * Configure the Conjur client and log in as admin
 
 [rotate-api-key]: https://www.conjur.org/api.html#authentication-rotate-an-api-key
+
+## Up and running
+
+Now that we've got a TLS endpoint for our Conjur server, you can check it with
+your web browser.
+
+The status page is available at https://localhost:8443 but your browser will
+warn you about the self-signed certificate. To override the warning and see the
+page, you'll have to instruct your browser to trust the certificate. Once you
+switch to using your own certificate, the browser warning will go away
+automatically.
--- a/docs/tutorials/integrations/nginx.md
+++ b/docs/tutorials/integrations/nginx.md
@@ -69,10 +69,14 @@ Conjur successfully.
 tutorial will show you how to use Docker to install Conjur and NGINX and
 configure them to provide Conjur with TLS.
 
+[NGINX]: https://www.nginx.com
+
 ### Prerequisites
 
 This tutorial requires Docker and a terminal application. Prepare by following
 the prerequisite instructions found on [Install Conjur][prerequisites].
+
+[prerequisites]: /get-started/install-conjur.html#prerequisites
 
 Additionally, you will need the tutorial files from the Conjur source code
 repository. Here's how you get them:
@@ -120,7 +124,14 @@ $ docker-compose exec client bash
 
 ## Breaking Down the Tutorial
 
-These files show how the proxy setup works:
+These files show how the proxy setup works. Note that, while this tutorial uses
+Docker and NGINX, there's nothing technically important about using those
+particular technologies or even about using containers at all. You can replicate
+the same strategy using a different endpoint such as [HAProxy][haproxy-tls] and
+services that are run in VMs or on bare metal. For the purpose of a tutorial
+that can conveniently run on a laptop, we provide this setup:
+
+[haproxy-tls]: https://www.haproxy.com/documentation/aloha/7-0/haproxy/tls/
 
 ### docker-compose.yml
 
@@ -198,14 +209,49 @@ exits, a desirable behavior for common ops use cases.
 
 ### default.conf
 
-TODO describe more
+This configuration file tells NGINX how to use TLS and behave as a proxy for
+Conjur.
 
-* `default.conf` configures NGINX to use a self-signed certificate for TLS and
-  sets up a proxy to Conjur
-* `tls/tls.conf` sets the parameters of the self-signed certificate
-* `start.sh` shows the installation flow
+```nginx
+listen              443 ssl;
+server_name         proxy;
+access_log          /var/log/nginx/access.log;
+```
 
-### About that self-signed certificate
+This block sets up a few basic properties of the NGINX server. Its hostname is
+`proxy`, it listens on the standard port for HTTPS (port 443) and it has a
+location for its access logs, useful for monitoring traffic.
+
+```nginx
+ssl_certificate     /etc/nginx/nginx.crt;
+ssl_certificate_key /etc/nginx/nginx.key;
+```
+
+This block gives NGINX its directions on how to perform TLS. ("ssl" is a name
+for an older standard for TLS and is still often used synomymously.)
+
+The `certificate` is a public key, and the `certificate_key` is the
+corresponding private key. NGINX maintains [documentation][nginx-https] about
+how to configure the server for HTTPS, including production optmization
+guidelines and the values of many default settings, so that is worth reading.
+
+[nginx-https]: https://nginx.org/en/docs/http/configuring_https_servers.html
+
+```nginx
+location / {
+  proxy_pass http://conjur;
+}
+```
+
+This part instructs NGINX to proxy incoming traffic (secured by TLS) through to
+the Conjur server.
+
+### tls/tls.conf
+
+This file describes to `openssl` what options to use when generating a
+self-signed certificate. This allows you to use TLS in testing and staging, but
+it does not allow clients to automatically authenticate the identity of the
+Conjur server.
 
 In production, don't use a self-signed certificate. It's better than nothing,
 but it's not a sustainable security practice because you're going to have to
@@ -214,5 +260,46 @@ manually verify that you're not talking to a man in the middle.
 Instead, ask your security team to provide a certificate signed by a trusted
 root and modify `default.conf` to use that certificate instead.
 
-[NGINX]: https://www.nginx.com
-[prerequisites]: /get-started/install-conjur.html#prerequisites
+Let's look at what parts you might modify:
+
+```ini
+[ dn ]
+C=US
+ST=Wisconsin
+L=Madison
+O=CyberArk
+OU=Onyx
+CN=proxy
+```
+
+This block describes the distinguished name of the certificate using the
+(C)ountry, (ST)ate, (L)ocation, (O)rganization, (O)rganizational (U)nit, and
+(C)ommon (N)ame. You'll want to change all these to suit your own organization.
+
+```ini
+[ alt_names ]
+DNS.1 = localhost
+DNS.2 = proxy
+IP.1 = 127.0.0.1
+```
+
+This block describes the names by which the server will be known, including its
+hostnames and IP addresses. You'll want to modify it to match the hostnames and
+IP addresses you use.
+
+### start.sh
+
+Here's the outline of the tutorial flow. Read through the file to see what it
+does to accomplish each step.
+
+* Pull required containers from Docker Hub
+* Remove containers, certs and keys created in earlier tutorial runs (if any)
+* Create a self-signed certificate and key for TLS
+* Generate a data key for Conjur encryption of data at rest
+  - Move this key to a safe place before deploying in production!
+* Start services and wait a little while for them to become responsive
+* Create a new account in Conjur and fetch its API key
+  - [Rotate the admin's API key][rotate-api-key] regularly!
+* Configure the Conjur client and log in as admin
+
+[rotate-api-key]: https://www.conjur.org/api.html#authentication-rotate-an-api-key
--- a/docs/tutorials/integrations/nginx.md
+++ b/docs/tutorials/integrations/nginx.md
@@ -1,0 +1,9 @@
+---
+title: Tutorial - Nginx Proxy
+layout: page
+section: tutorials
+description: Conjur Tutorial - Nginx
+---
+
+If you're deploying Conjur in production, you need to set up Transport Layer
+Security (TLS) so that requests made to the server are encrypted in transit.
--- a/docs/tutorials/integrations/nginx.md
+++ b/docs/tutorials/integrations/nginx.md
@@ -1,8 +1,8 @@
 ---
-title: Tutorial - Nginx Proxy
+title: Tutorial - NGINX Proxy
 layout: page
 section: tutorials
-description: Conjur Tutorial - Nginx
+description: Conjur Tutorial - NGINX
 ---
 
 If you're deploying Conjur in production, you need to set up Transport Layer
@@ -32,7 +32,7 @@ is called a **man in the middle attack**.
 
 Even if I'm not able to impersonate the Conjur server, I could still learn your
 secrets by joining your network and listening for traffic coming and going from
-the Conjur server. This is called **passive survillance**.
+the Conjur server. This is called **passive surveillance**.
 
 ### TLS defeats passive and active attacks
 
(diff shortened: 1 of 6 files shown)


jvanderhoof:
Ryan, 
This is a great start.  I'd love to see you take the documentation down a level to the technical aspects of setting up the proxy. What steps would someone need to take to add an internally generated cert (organization with their own CA provider)?  Let's expose end users to the painful details of the `start.sh` file.  We're speaking to developers with ops experience, or ops people.  They want the details.

As an outline, maybe something like:
1. Generate certificate (allow user to create their own cert or use a provide cert)
2. Show Nginx config and highlight the custom parts (cert & key, proxy pass to an internal URL or IP).
3. Start Conjur & Proxy as separate processes (as opposed to using docker-compose) to mimic real world deployment.

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,135 @@
...
+$ # start the Conjur+NGINX tutorial servers
+$ cd conjur/docs/tutorials/integrations/nginx
+$ ./start.sh
+```
+
+Take a look at the logs for the Conjur or NGINX servers:
+
+```sh-session
+$ docker-compose logs conjur
+$ docker-compose logs proxy
```
izgeri: It could be clearer that the user should be doing this from the same `nginx` dir as above

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,135 @@
...
+
+```sh-session
+$ docker-compose exec client bash
+# conjur authn whoami
+# conjur list
+```
+
+These files show how the proxy setup works:
+
+* `docker-compose.yml` declares services and the connections between them
```
izgeri: Maybe talk a little bit here about how this file sets up the links / how the services are linked together?

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,135 @@
...
+* `default.conf` configures NGINX to use a self-signed certificate for TLS and
+  sets up a proxy to Conjur
+* `tls/tls.conf` sets the parameters of the self-signed certificate
+* `start.sh` shows the installation flow
```
izgeri: Worth diving into in more detail - the file is well commented, but if setting this up for production what things would they need to change? Could they use this file? Why are the config steps at the end needed?

docs/tutorials/integrations/nginx.md
izgeri: Also, something is weird here. After running through the tutorial and coming back to this command, here is what I see:
```
$ docker-compose logs conjur
Attaching to nginx_conjur_1
conjur_1    | => Booting Puma
conjur_1    | => Rails 4.2.8 application starting in production on http://0.0.0.0:80
conjur_1    | => Run `rails server -h` for more startup options
conjur_1    | => Ctrl-C to shutdown server
conjur_1    | [1] Puma starting in cluster mode...
conjur_1    | [1] * Version 3.8.2 (ruby 2.2.8-p477), codename: Sassy Salamander
conjur_1    | [1] * Min threads: 5, max threads: 5
conjur_1    | [1] * Environment: production
conjur_1    | [1] * Process workers: 2
conjur_1    | [1] * Preloading application
conjur_1    | [1] * Listening on tcp://0.0.0.0:80
```

izgeri: Note in particular that it doesn't seem to be working through port 443 or over https - can you be clearer about what they should be looking for in the logs to verify TLS? I'm not seeing it in the logs

izgeri: I did get a 200 response (and the "Your server is running!" page) when I ran `curl -i https://localhost:443 --cacert tls/nginx.crt`, however - and a "failure to connect" message when trying localhost:80.

docs/tutorials/integrations/nginx.md
izgeri: Also, the version of postgres that you use leads to a warning in the docker logs for the conjur server:

```
conjur_1    | .......SEQUEL DEPRECATION WARNING: Calling a dataset filtering method with multiple arguments or an array where the first argument/element is a string is deprecated and will be removed in Sequel 5.  Use Sequel.lit("account(resource_id) = ?", "test") to create an SQL fragment expression and pass that to the dataset filtering method, or use the auto_literal_strings extension.
conjur_1    | /var/lib/gems/2.2.0/gems/sequel-4.46.0/lib/sequel/dataset/query.rb:1293:in `filter_expr'
conjur_1    | /var/lib/gems/2.2.0/gems/sequel-4.46.0/lib/sequel/dataset/query.rb:1249:in `add_filter'
conjur_1    | /var/lib/gems/2.2.0/gems/sequel-4.46.0/lib/sequel/dataset/query.rb:1034:in `where'
conjur_1    | /opt/conjur-server/app/models/resource.rb:70:in `search'
conjur_1    | /var/lib/gems/2.2.0/gems/sequel-4.46.0/lib/sequel/model/base.rb:1058:in `search'
conjur_1    | /opt/conjur-server/app/controllers/resources_controller.rb:14:in `index'
conjur_1    | /var/lib/gems/2.2.0/gems/actionpack-4.2.8/lib/abstract_controller/base.rb:198:in `process_action'
conjur_1    | /var/lib/gems/2.2.0/gems/actionpack-4.2.8/lib/action_controller/metal/rendering.rb:10:in `process_action'
conjur_1    | /var/lib/gems/2.2.0/gems/actionpack-4.2.8/lib/abstract_controller/callbacks.rb:20:in `block in process_action'
conjur_1    | /var/lib/gems/2.2.0/gems/activesupport-4.2.8/lib/active_support/callbacks.rb:117:in `call'
```

maybe it'd be worth using a different version just to avoid this bit of ugliness (if possible)?

docs/tutorials/integrations/nginx.md
ryanprior: So what's happening here is the Conjur server is listening on port 80, but it isn't exposing that port except on the local Docker network. Meanwhile, the nginx container is exposing its port 443 and proxying through to Conjur.

I will add this explanation to the tutorial!

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,218 @@
...
+
+This file declares services to be used in the tutorial. Let's break down each declaration:
+
+```yaml
+database:
+  image: postgres:9.3
+```
+
+Conjur requires a Postgres database to store encrypted secrets and other data.
+This service uses the [official Postgres image][postgres-image] from DockerHub.
```
dividedmind: Since we're doing best practices here, we should also show how to configure postgres in a way that prevents outside access. This can be accomplished by using TLS authentication, or, in this scenario, more simply by setting up the network so that only the conjur container can connect. You note below that the containers can only connect over private Docker network; even if that's true (not sure if that network isn't shared between all the containers on the host), there's still the matter of nginx container being able to connect to postgres, which is potentially dangerous (if you own nginx you can backdoor into pg).

docs/tutorials/integrations/nginx/tls/tls.conf
```diff
@@ -0,0 +1,39 @@
...
+x509_extensions = v3_ca # The extentions to add to the self signed cert
+req_extensions  = v3_req
+x509_extensions = usr_cert
+
+[ dn ]
+C=US
+ST=Wisconsin
+L=Madison
+O=Onyx
+OU=CyberArk
```
dividedmind: Shouldn't these be swapped? O is organization, OU is organizational unit. Just saying... ;)

docs/tutorials/integrations/nginx/tls/tls.conf
```diff
@@ -0,0 +1,39 @@
...
+CN=proxy
+
+[ usr_cert ]
+basicConstraints=CA:FALSE
+nsCertType                      = client, server, email
```
dividedmind: I don't think email is necessary here. Neither is client, unless we're using that to authenticate (eg. to pg).

docs/tutorials/integrations/nginx/tls/tls.conf
```diff
@@ -0,0 +1,39 @@
...
+keyUsage = nonRepudiation, digitalSignature, keyEncipherment
+extendedKeyUsage = serverAuth, clientAuth, codeSigning, emailProtection
```
dividedmind: codeSigning is definitely not needed here; cf. above for client and email.

docs/tutorials/integrations/nginx/tls/tls.conf
```diff
@@ -0,0 +1,39 @@
+[req]
+default_bits = 2048
+prompt = no
+default_md = sha256
+req_extensions = req_ext
+distinguished_name = dn
+x509_extensions = v3_ca # The extentions to add to the self signed cert
+req_extensions  = v3_req
+x509_extensions = usr_cert
```
dividedmind: I'm confused about these extensions. If you want the same extensions for req and x509, you should point them to the same section instead of duplicating it. Additionally you reference `req_ext` which does not exist.

docs/tutorials/integrations/nginx.md
ryanprior: This is interesting @dividedmind I will look into it. To start me out: what if anything do we do for hardening PG within the Conjur Enterprise appliance? Is it doing TLS in there, or does it have a special sub-network or something? Like you say this is a good opportunity to set the standard for the most secure practice we can recommend.

jvanderhoof: Great point @dividedmind.  Why don't we add this to the documentation queue.  I feel like Postgres connection security is beyond the scope of this story.

docs/tutorials/integrations/nginx/tls/tls.conf
ryanprior: I gave some effort to this and other recommendations for the TLS conf, but was only able to make the DN changes & keep the cert working. Let's take a second pass at this later and optimize/simplify the tls.conf.

docs/tutorials/integrations/nginx/tls/tls.conf
ryanprior: I tried to remove this & ran into certificate errors. See comment below.

docs/tutorials/integrations/nginx/tls/tls.conf
ryanprior: I tried removing this duplication but ran into certificate errors. See comment below.

jvanderhoof:
Looks like a solid first pass.

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+their API key to the Conjur server. The Conjur server validates that the
+identity is authentic, then checks that the provided ID is authorized to
+[retrieve the secret][secret-get]. Assuming this check passes, the Conjur server
+returns the secret value to the client.
+
+However, this flow would be vulnerable in two ways. Suppose I were to
+impersonate the Conjur server and listen with my own illegitimate server on
+`http://conjur.local`. Then when your client goes to fetch a secret, I can take
+the API key you send and impersonate you to the real Conjur server. Now I
+control your identity and can learn your secrets without you finding out. This
```
izgeri: worth mentioning something about the audit functionality in EE Conjur?

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+secrets by joining your network and listening for traffic coming and going from
+the Conjur server. This is called **passive surveillance**.
+
+### TLS defeats passive and active attacks
+
+Transport Layer Security allows your client to verify that it's talking to the
+real Conjur server, and it uses standard secure technology to encrypt your
+secrets in transit. This means:
+
+* Your Conjur server will be `https:` instead of `http:`, just like a secure
```
izgeri: Maybe "Your Conjur URL will be prefixed by `https` instead of `http`" instead?

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+  website
+* Because the client knows it's talking to the real server, the "man in the
```
izgeri: "the client knows" --> "the client can verify"?

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+For these reasons, it is crucial to set up TLS correctly when you deploy Conjur.
+
+### Protect Your Secrets
+
+Do not leave this to chance: even a small flaw can totally compromise your
+secrets. Setting up Conjur and TLS without the appropriate expertise is like
+packing your own parachute and jumping out of a plane.
+
+That doesn't mean you should close the tab and walk away. It means you should get
+in touch with us and your own security team so we can ensure that you can employ
```
izgeri: "employ" --> "deploy"? I feel like you're really talking here about how they set it up, not so much how they use it

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+Conjur successfully.
+
+[authn]: https://www.conjur.org/api.html#authentication-authenticate-post
+[secret-get]: https://www.conjur.org/api.html#secrets-retrieve-a-secret-get
+
+## Using NGINX To Proxy Traffic With TLS To Conjur
+
+[NGINX][NGINX] is a web proxy that's Free Software and easy to configure. This
```
izgeri: I'd un-capitalize "Free Software" and maybe write "free and open source" instead

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+tutorial will show you how to use Docker to install Conjur and NGINX and
+configure them to provide Conjur with TLS.
```
izgeri: "provide Conjur with" --> "use"

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+
+This tutorial requires Docker and a terminal application. Prepare by following
+the prerequisite instructions found on [Install Conjur][prerequisites].
+
+[prerequisites]: /get-started/install-conjur.html#prerequisites
+
+Additionally, you will need the tutorial files from the Conjur source code
+repository. Here's how you get them:
+
+1. Install Git: https://git-scm.com/book/en/v2/Getting-Started-Installing-Git
```
izgeri: Can you make "Install Git" a hyperlink instead of including the URL in the doc?

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+1. Clone the Conjur repository
+   
+   In your terminal application, run:
+   ```sh-session
+   $ git clone https://github.com/cyberark/conjur.git
+   ```
+   This will create a folder called `conjur` in the current directory.
```
izgeri: "the current directory" --> "your working directory"?

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+
+### The Good Part
+
+To start out our experience on a high note, let's get the full Conjur+TLS stack
+up and running so we can inspect it.
+
+The tutorial script will install Conjur and NGINX, configure them to work
+together, and connect a Conjur client to the running server. This is a full
```
izgeri: What's "the running server"? We "install Conjur and NGINX", but it's not clear what "the running server" is that we're connecting the "Conjur client" to

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+```sh-session
+$ # start the Conjur+NGINX tutorial servers
+$ cd conjur/docs/tutorials/integrations/nginx
+$ ./start.sh
+```
+
+Take a look at the logs for the Conjur or NGINX servers:
+
+```sh-session
+$ # in the 'nginx' directory we navigated to previously:
```
izgeri: the way you've rewritten this, I think this line isn't needed anymore

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+$ docker-compose logs conjur
+$ docker-compose logs proxy
+```
+
+Run some commands in the Conjur client:
+
+```sh-session
+$ # in the 'nginx' directory we navigated to previously:
```
izgeri: the way you've rewritten this, I think this line isn't needed anymore

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+Conjur requires a Postgres database to store encrypted secrets and other data.
+This service uses the [official Postgres image][postgres-image] from DockerHub.
+
+#### Production tip
+
+Just as we use a proxy in this tutorial to encrypt & authenticate traffic to the
+Conjur server using TLS, you will also want to use TLS for your production
+Postgres database. If you're using [Amazon RDS][rds], it already has TLS support
+built-in. If you're hosting your own database, you'll want to use a technique
+similar to the one described here in order to secure its traffic.
```
izgeri: This should be reworded to be more clear that this tutorial _does not_ instruct on how to secure the DB

Maybe change it to say "you will also want to use TLS for your production Postgres database - but instructions for doing so are beyond the scope of this tutorial."

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+[HAProxy][haproxy-tls] and services that run in virtual machines or on bare
+metal. To create a tutorial that can run as conveniently on a laptop as in the
+cloud, we provide this setup.
+
+[haproxy-tls]: https://www.haproxy.com/documentation/aloha/7-0/haproxy/tls/
+
+### docker-compose.yml
+
+This file declares services to be used in the tutorial. Let's break down each declaration:
+
```
izgeri: Can you add a line
`---`
here too? It will make it clearer that the "Production tip" is relating to the DB

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+can replicate the same strategy using a different endpoint such as
+[HAProxy][haproxy-tls] and services that run in virtual machines or on bare
+metal. To create a tutorial that can run as conveniently on a laptop as in the
+cloud, we provide this setup.
+
+[haproxy-tls]: https://www.haproxy.com/documentation/aloha/7-0/haproxy/tls/
+
+### docker-compose.yml
+
+This file declares services to be used in the tutorial. Let's break down each declaration:
```
izgeri: This may be beyond the scope, but it would be cool if we included a simple architectural diagram (get those visual learners to quickly see what's going on here)

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+ssl_certificate_key /etc/nginx/nginx.key;
+```
+
+This block gives NGINX its directions on how to perform TLS ("ssl" is a name
+for an older standard for TLS and is still often used interchangeably.)
+
+The `certificate` is a public key, and the `certificate_key` is the
+corresponding private key. NGINX maintains [documentation][nginx-https] about
+how to configure the server for HTTPS, including production optimization
+guidelines and the values of many default settings, so that is worth reading.
```
izgeri: I'd kill ", so that is worth reading" and maybe just add "useful" before the documentation link

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+This block describes the names by which the server will be known, including its
+hostnames and IP addresses. You'll want to modify it to match the hostnames and
+IP addresses you use.
+
+### start.sh
+
+Here's the outline of the tutorial flow. Read through the file to see what it
+does to accomplish each step.
+
+* Pull required containers from Docker Hub
```
izgeri: "containers" --> "container images"

docs/tutorials/integrations/nginx/start.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+
+# Pull required containers from Docker Hub
```
izgeri: "containers" --> "container images"

docs/tutorials/integrations/nginx.md
dividedmind: I'd vote to leave it as-is.

docs/tutorials/integrations/nginx.md
dividedmind: PG in the EE runs locally and uses Linux 'unix' authentication for local user over the domain socket. TCP connections are only allowed for replication (& for audit push) and they require TLS client authentication.

docs/tutorials/integrations/nginx.md
ryanprior: I would like to do this, too. That's on the list for the next time we revise this.

docs/tutorials/integrations/nginx.md
```diff
@@ -0,0 +1,374 @@
...
+layout: page
+section: tutorials
+description: Conjur Tutorial - NGINX
+---
+
+If you're deploying Conjur in production, you need to set up Transport Layer
+Security so that requests made to the server are encrypted in transit.
+
+*Note: if you use Conjur Enterprise, we handle this for you using an audited
+implementation that works similar to the technique described here.*
```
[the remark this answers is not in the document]
ryanprior: changed

docs/tutorials/integrations/nginx.md
ryanprior: mentioned

docs/tutorials/integrations/nginx.md
ryanprior: changed

docs/tutorials/integrations/nginx.md
ryanprior: changed

docs/tutorials/integrations/nginx.md
ryanprior: changed, good catch

docs/tutorials/integrations/nginx.md
ryanprior: I like Free Software better, but I changed it to "a high performance free open source web server" since that's how NGINX describes itself.

docs/tutorials/integrations/nginx.md
ryanprior: changed

docs/tutorials/integrations/nginx.md
ryanprior: yep done

docs/tutorials/integrations/nginx.md
ryanprior: clarified

docs/tutorials/integrations/nginx.md
ryanprior: ok

docs/tutorials/integrations/nginx.md
ryanprior: ok

docs/tutorials/integrations/nginx.md
ryanprior: added

docs/tutorials/integrations/nginx.md
ryanprior: clarified & added link to Postgres recommendations on this topic

docs/tutorials/integrations/nginx.md
ryanprior: changed

docs/tutorials/integrations/nginx.md
ryanprior: changed

docs/tutorials/integrations/nginx/start.sh
ryanprior: changed

izgeri:
LGTM at current

docs/tutorials/integrations/nginx/tls/tls.conf
dividedmind: Ok, can you create an issue?

izgeri:
@jvanderhoof any concerns about merging?

izgeri:
merging after confirming it's good with @jvanderhoof in slack.

PR #490, opened 2017-11-28
Fix issue with multifile policies returning 500 error

### LINKED ISSUE #489: Policy replace can fail when user is deleted and removed from group
If you have a multi-file root policy, and you call `conjur policy load --replace root file1.yml` where file1 references groups not defined in file1, you will get a 500 error.

Fixing this bug will include sending a more useful error response and updating documentation to be more clear about handling multi-file policie.

izgeri:
Closes #489 

## Code changes
If I load a policy with groups in it and then reload a policy using the `--replace` flag, referencing the groups but no longer defining them, I will get an unhandled `Sequel:ForeignKeyConstraintViolation` exception. The message for the exception will look like:
```
PG::ForeignKeyViolation: ERROR:  insert or update on table \"role_memberships\" violates foreign key constraint \"role_memberships_member_id_fkey\"\nDETAIL:  Key (member_id)=(cucumber:group:security-admin) is not present in table \"roles\".\n
```
In this update, we handle the `Sequel:ForeignKeyConstraintViolation` exception for a violation to the `role_memberships_member_id_fkey` or the `role_memberships_role_id_fkey` foreign key constraint. If the thrown `Sequel:ForeignKeyConstraintViolation` is for a violation of a different foreign key constraint, the returned JSON will pass on the info of the original `Sequel:ForeignKeyConstraintViolation` exception. If it is indeed the `role_memberships_member_id_fkey` or `role_memberships_role_id_fkey` foreign key constraint that has been violated, it will be parsed to raise a `not_found` exception.

## New tests
For valuable context, please review the new `replace.feature` suite of cucumber tests.

## Updates to cuke suite
In addition to the cucumber tests and updated code to handle the special cases of the `Sequel:ForeignKeyConstraintViolation` exception, some of the cucumber supporting functionality has been updated to allow more general use. In particular,
- Testing for `error` key in response has been moved to its own step
- Rather than throwing an exception in `invoke` whenever the HTTP code doesn't match the `status` value (even if no `status` value is passed to the function), we added an `expect` statement that is only parsed if an integer `status` value is actually passed to the function and set the request response in the `@result` variable directly in the `invoke` function, mirroring what is happening when the request is successful
- There is a new `json_result` method that will take the API request response and parse it as JSON, so whether or not the request is successful we can retrieve a JSON-parsed response

## Documentation changes
The documentation has been updated so that the [Loading a Policy](http://conjur.org/reference/policy.html#loading-a-policy) section will now include a `Multi-file Policy` subsection

## Jenkins build
Jenkins build [here](https://jenkins.conjur.net/job/cyberark--conjur/job/489-policy-replace-multifile-failure/)


cucumber/policy/features/replace.feature
```diff
@@ -0,0 +1,33 @@
...
+A policy can be reloaded using the --replace flag
+
+  Scenario: A multifile policy with one modified file fails on reload
+
+    Given I replace the policy by loading the policy file "multifile1.yml"
+    And I load a "root" policy file "multifile2.yml"
+    When I replace the policy by loading the policy file "multifile2_with_deletes.yml"
+    Then there is an error
+    And the error code is "not_found"
+    And the error message is "ID cucumber:group:developers is not a valid role"
```
kgilpin: This message makes it sound a little bit like `cucumber:group:developers` is not a valid role id. I think it would be more clear to say that the role does not exist.

cucumber/policy/features/replace.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Replacing a policy
+
+A policy can be reloaded using the --replace flag
+
+  Scenario: A multifile policy with one modified file fails on reload
+
+    Given I replace the policy by loading the policy file "multifile1.yml"
```
kgilpin: To me, moving the policies to files makes it harder to understand what the cucumber test is doing. Can we discuss putting the policies in the cucumber code itself, like the other tests?

cucumber/policy/features/support/policy/multifile2.yml
```diff
@@ -0,0 +1,35 @@
+- !user
```
kgilpin: Overall I think we could probably reproduce the problem with smaller policies. That way they could be embedded directly in the Cucumber test code.

cucumber/policy/features/support/world.rb
```diff
@@ -26,8 +26,8 @@ def invoke status: :ok, &block
         puts result if @echo
       end
     rescue RestClient::Exception => e
-      @exception = e
-      raise e unless status == e.http_code
+      expect(e.http_code).to eq(status) if status.is_a? Integer
```
kgilpin: `status` should always be the same data type, so I recommend we remove the `if` test.

docs/reference/policy.md
```diff
@@ -56,6 +56,42 @@ $ conjur policy load root conjur.yml
 Policy loaded
 {% endhighlight %}
 
+#### Multi-file Policies
```
kgilpin: Great addition!

app/controllers/application_controller.rb
```diff
@@ -57,11 +58,42 @@ def no_matching_row e
...
         code: "not_found",
         target: target,
-        message: e.message,
+        message: e.message
       }.compact
     }, status: :not_found
   end
 
+  def foreign_key_constraint_violation e
+    logger.debug "#{e}\n#{e.backtrace.join "\n"}"
```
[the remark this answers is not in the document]
izgeri: In my testing, the backtrace didn't propagate to the record_not_found method (it was nil)

app/controllers/application_controller.rb
```diff
@@ -57,11 +58,42 @@ def no_matching_row e
...
+
+    # check if this is a violation of role_memberships_member_id_fkey
+    if e.message.index(/role_memberships_member_id_fkey/) ||
```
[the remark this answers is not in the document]
izgeri: Done

cucumber/policy/features/replace.feature
izgeri: So updated

cucumber/policy/features/support/world.rb
izgeri: Actually, status is a symbol by default - so this update will only check the http_code against the status if a numeric value has been passed to the `invoke` function (which appears to be how it's used from reviewing the other cukes)

cucumber/policy/features/support/policy/multifile2.yml
izgeri: This is probably true - I created these tests with the policies I found the error with, but though simple they are more complex than what is needed. I can trim them down if you like, but I'd prefer not to unless there's a real benefit to doing so.

cucumber/policy/features/replace.feature
izgeri: I agree that putting the policy in the test would be more consistent with the other tests in this suite. However, the policies I was working with when I got the tests up were long enough that it seemed less neat to put them directly in the test file. Creating a policy directory is consistent with what's been done on other projects for longer policy files, and I tried to name them in a way that it would be clear which files you were looking for if you hopped into that directory to explore.

I'm open to revising this if you think it's important, but I can imagine future tests that we might add where I'd also want to put the policy in a separate file.

cucumber/policy/features/support/world.rb
izgeri: The alternative would be converting `status` within `invoke` by calling something like `status = Rack::Utils::SYMBOL_TO_STATUS_CODE[status] if !status.is_a? Integer` to make sure the comparison always works as expected.

app/controllers/application_controller.rb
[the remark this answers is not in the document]
kgilpin: Oh, because you just created the exception with `new`, you didn't raise it.

cucumber/policy/features/support/world.rb
kgilpin: In my view, numerical status codes are a better choice because that's what's specified in the HTTP spec. The symbols is more of a Rails-ism, so somebody using the cucumber tests as a reference who's not a Ruby programmer will be familiar with the numerical codes but not the symbols.

app/controllers/application_controller.rb
```diff
@@ -57,11 +58,45 @@ def no_matching_row e
...
+          break 
+        end 
+      end
+
+      # the member ID is inside the second set of parentheses of the key_string
+      key_index = key_string.index(/\(/, 1) + 1
+      key = key_string[ key_index, key_string.length - key_index - 1 ]
+
+      exc = Exceptions::RecordNotFound.new key, message: "Role #{key} does not exist"
+      record_not_found exc
```
kgilpin: The reason why `record_not_found` doesn't see a backtrace is that you are calling `record_not_found` with the exception, but the exception is never raised.

kgilpin: One way to handle this would be to have a method `render_record_not_found` that is called from here and from `record_not_found`. That removes the `if` test on the log statement in `record_not_found`, which is good because a future developer isn't going to understand why it's there.

cucumber/policy/features/support/policy/multifile2.yml
kgilpin: So, there is a real benefit I had in mind: There is a form of documentation generated from cucumber tests. You can see examples of it here : https://relishapp.com/conjur/possum-policy/docs/users-and-groups-can-be-added-to-groups

If the policies are embedded in the tests, then the generated cucumber documentation is a self-contained spec of the project's behavior. If the tests reference external files, then the generated documentation will not contain the policies, and so the user will have to guess, infer, or track down the Git repo of the project to really understand what's happening.

kgilpin: I think this problem could be reproduced with policies that are just a few lines long. Given:

```
# groups.yml
- !group developers
```
```
# users.yml
- !user alice
- !grant
  role: !group developers
  members:
  - !user alice
```

If I load "groups" and "users" together (concatenated), and then "users" alone, then the exception you are testing for should be raised.

cucumber/policy/features/support/world.rb
kgilpin: I would like to see the tests use numerical status codes, because codes are given by the HTTP spec, and also to be consistent with other tests such as https://relishapp.com/conjur/possum-api/docs/rbac-privileges-control-whether-a-role-can-update-and-or-fetch-a-secret

app/controllers/application_controller.rb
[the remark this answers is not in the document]
izgeri: Right - but I can't reraise and have it hit the `not_found` exception handler ([see here](https://simonecarletti.com/blog/2009/11/re-raise-a-ruby-exception-in-a-rails-rescue_from-statement/) for an example). I'm not sure why a newly instantiated Exception object wouldn't have its backtrace set - that only gets set on `raise`??

cucumber/policy/features/support/world.rb
izgeri: I am implementing this, but it is not as pretty as I'd like.

I want the status codes to remain optional, because I am calling invoke from some generic step definitions that don't assume what the output is going to be (and it could be successful or not, so I don't want to set the default to `success`).

This is leading invoke to look like this:
```
def invoke status: nil, &block
    begin
      @result = yield
      raise "Expected invocation to be denied" if status && status != 200
      @result.tap do |result|
        puts result if @echo
      end
    rescue RestClient::Exception => e
      expect(e.http_code).to eq(status) if status
      @result = e.response.body
    end
  end
```

At this point I just really want the status code check to live someplace else. We already have a "the error code is [code]" step definition - why not just update the cukes that are passing invoke a status code to use this step definition?

app/controllers/application_controller.rb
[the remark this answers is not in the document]
izgeri: Anywho, I used your suggestion to have a `render_record_not_found` function that gets called from here and `record_not_found`, so that should resolve this issue.

PR #494, opened 2017-11-30
Be specific about which Conjur image targeting during build/test/push [CON-4493]

ghost:
Closes CON-4493.

This PR resolves image congestion when multiple CI pipelines are run at once to build/test/push the Conjur image. With this merge it is now impossible to accidentally pick up the wrong image.

Previously we were naively taking the `conjur:latest` image and retagging it for push. This worked fine most of the time, but a race condition exists where another `conjur:latest` image is created before the retag and push to repositories. That newer image would be picked up, even if it didn't pass tests or was on an incompatible branch.

View the jenkins job, see how it is correctly passing the TAG through build/test/push stages.

Not required. You can run build.sh and test.sh locally if you like.

N/A

https://jenkins.conjur.net/job/cyberark--conjur/job/prevent-image-collision/

N/A

N/A

N/A

N/A

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,4 +1,4 @@
-FROM conjur
+ARG VERSION=latest
+FROM conjur:${VERSION}
 
 RUN bundle --no-deployment --without ''
-RUN ln -sf /src/conjur-server/bin/conjurctl /usr/local/bin/conjurctl
--- a/test.sh
+++ b/test.sh
@@ -1,12 +1,14 @@
 #!/bin/bash -ex
 
+TAG="$(< VERSION)-$(git rev-parse --short HEAD)"
+
 function finish {
 	docker rm -f $pg_cid
 	docker rm -f $server_cid
 }
 trap finish EXIT
 
-export CONJUR_DATA_KEY="$(docker run --rm conjur data-key generate)"
+export CONJUR_DATA_KEY="$(docker run --rm conjur:$TAG data-key generate)"
 
 pg_cid=$(docker run -d postgres:9.3)
 
@@ -18,7 +20,7 @@ server_cid=$(docker run -d \
 	--link $pg_cid:pg \
 	-e DATABASE_URL=postgres://postgres@pg/postgres \
 	-e RAILS_ENV=test \
-	conjur server)
+	conjur:$TAG server)
 
 cat << "TEST" | docker run \
 	-i \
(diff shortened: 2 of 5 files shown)


Dockerfile.test
```diff
@@ -1,4 +1,4 @@
-FROM conjur
+ARG VERSION=latest
+FROM conjur:${VERSION}
```
apotterri: Oh, hey, can docker on our executor's handle this now? That'd be very exciting....

test.sh
```diff
@@ -47,5 +49,5 @@ rm -rf spec/reports/*
 rm -rf cucumber/api/features/reports/*
 rm -rf cucumber/policy/features/reports/*
 
-bundle exec rake jenkins || true
+bundle exec rake jenkins
```
apotterri: I know you hate `|| true`. But, the way this script and the pipeline are currently written, no test results will get collected when the tests fail.

apotterri (changes requested):

Dockerfile.test
dustinmm80: Yeah, I had been wondering this about it since we upgraded Docker on the executors, and it works! I think we can simplify some of our bash/sed-fu that does the same thing in some of our projects.

Right now only ARG can come before FROM

https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

test.sh
dustinmm80: Good eye. I don't hate it, just wish this wasn't the way it worked :) I'll update this PR so it's working again.

dustinmm80: I think I've found a better way to do this, without requiring us to hijack exit codes. Since this is a Jenkins-specific workaround I think it's better to handle this in Jenkins logic, not in our application code.

Since each stage can also have a `post` block, I think it's better to use 

`post { always { junit ... } }` to capture the reports.

This pattern is mentioned here: https://jenkins.io/doc/book/pipeline/jenkinsfile/#handling-failures. See the updated Jenkinsfile in this PR.

What do you think Alan?

dustinmm80:
I also added coverage report collection to this PR: https://jenkins.conjur.net/job/cyberark--conjur/job/prevent-image-collision/Coverage_Report/

test.sh
apotterri: In general, this seems like a good approach to me. The only thing that seems a little weird is using `always` in the `post` block. But, I guess, if this is doc'ed as the current best practice, it's a good way to go.

apotterri (approved):

PR #504, opened 2018-01-09
adding wait method to conjurctl

### LINKED ISSUE #503: conjurctl has a command to wait for Conjur to be ready
Add a `conjurctl` command that will check for when the Conjur server is ready. This command can be used when deploying applications that rely on Conjur, so that the applications will wait for Conjur to be ready before proceeding.

izgeri:
Closes #503 

This PR adds a `wait` command to `conjurctl` so that applications that depend on Conjur have a method to wait for it to be running before proceeding. It assumes that the `CONJUR_APPLIANCE_URL` environment variable is present on the container running the command.

To test this out, you can stand up a Conjur instance using the standard `./build.sh` and `dev/start.sh` commands, call `conjurctl wait` from within the Conjur container, open another Conjur container with `docker exec -it $(docker-compose ps -q conjur) bash;`, and run `conjurctl server`. In the waiting container you will see:
```
Waiting for Conjur to be ready...
.
.
.
```
until you run `conjurctl server` to start Conjur; once the server starts you will see "Done."

### DIFF
--- a/bin/conjur-cli.rb
+++ b/bin/conjur-cli.rb
@@ -225,25 +225,31 @@ command :wait do |c|
 
   c.action do |global_options,options,args|
     puts "Waiting for Conjur to be ready..."
-    retry_num = 0
+
     retries = options[:retries].to_i
-    Struct.new('WaitError', :code)
-    while retry_num < retries
-      uri = URI.parse("http://localhost:#{options[:port]}")
+    port = options[:port]
+
+    conjur_ready = lambda do
+      uri = URI.parse("http://localhost:#{port}")
       begin
         response = Net::HTTP.get_response(uri)
-      rescue StandardError
-        response = Struct::WaitError.new('000')
+        response.code.to_i < 300
+      rescue
+        false
       end
-
-      break if response.code == "200"
-      puts "."
-      sleep 1
-      retry_num += 1
     end
 
-    exit_now! "No response" unless retry_num < retries
-    puts "Done."
+    retries.times do
+      break if conjur_ready.call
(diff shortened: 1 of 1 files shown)


bin/conjur-cli.rb
```diff
@@ -209,4 +209,16 @@ def test_select
   end
 end
 
+desc "Wait for the Conjur server to be ready"
+command :wait_for_conjur do |c|
+  c.action do |global_options,options,args|
+    puts "Waiting for Conjur to be ready..."
+    while `curl -o /dev/null -s -w '%{http_code}' $CONJUR_APPLIANCE_URL` != "200"
```
kgilpin: How is `CONJUR_APPLIANCE_URL` going to be set? We know that the service is going to be on localhost. I think we may need the port number as an argument (default: 80).

kgilpin: Since this is a Ruby function, I would rather see this in Ruby rather than shelling out to `curl`. Example along the lines of https://docs.ruby-lang.org/en/2.0.0/Net/HTTP.html

```ruby
res = Net::HTTP.get_response(URI("http://localhost:#{port}"))
break if res.is_a?(Net::HTTPSuccess)
STDOUT.write(".")
sleep 1
```

kgilpin: `conjurctl wait_for_conjur` seems a bit redundant. Maybe just `conjurctl wait`?

Also a timeout would be another good option to have. "wait forever" is a Jenkins killer.

izgeri:
LGTM, too

PR #505, opened 2018-01-10
Update changelog to add missing changelog notes

izgeri:
There were items missing or uncharacterized from the 0.2.0 release, and items missing from Unreleased. This file should be correct in its current state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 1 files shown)


kgilpin:
Looks good although I think we should label this 0.3.0 and release it

CHANGELOG.md
```diff
@@ -6,10 +6,23 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
```
kgilpin: It seems like below ## Unrelaesed it should say "Nothing"?

See

https://github.com/cyberark/conjur/blob/271f85b3d472ace1e61a7bd115df711afab267fa/CHANGELOG.md

izgeri: Based on the example [here](http://keepachangelog.com/en/1.0.0/) it should be blank.

PR #510, opened 2018-02-09
Adds hint to URL encode identifiers, with examples table

### LINKED ISSUE #508: API docs should say that {login} needs to be URL-encoded
We should explicitly state in the API documentations that logins with `/`s in them need to be URL encoded.

https://www.conjur.org/api.html

For example, the correct call for authenticating host `host/myapp-01` is:

```
localhost:8080/authn/quick-start/host%2Fmyapp-01/authenticate`
```

not 

```
localhost:8080/authn/quick-start/host/myapp-01/authenticate
```

The `{login}` in `/authn/{account}/{login}/authenticate` should be URL encoded.

This came up in Conjur Slack.

ryanprior:
Closes #508 

It adds additional hints to the API docs so that people know to URL-encode identifiers when they are part of an API URL. Additionally, it provides a table of examples to clarify common use cases.

People making requests to the Conjur API have been stumped by unexpected responses, only to eventually discover that the problem was a result of failure to URL encode their identifiers. This pull request is intended, then, to help make the Conjur API more approachable.

Read the new partial: `apidocs/src/partials/url_encoding.md`

Notice that it's been included in the body of any API route that includes a partial.

Generate the API docs as per https://github.com/cyberark/conjur/tree/master/apidocs#working-on-the-api-docs and visit localhost:3000

screen shot 2018-02-09 at 4 19 17 pm

https://jenkins.conjur.net/job/cyberark--conjur/job/apidocs-url-encoding-hint/

No

Nope

### DIFF
--- a/apidocs/src/partials/url_encoding.md
+++ b/apidocs/src/partials/url_encoding.md
@@ -1,0 +1,15 @@
+#### Note: entity IDs must be url-encoded
+
+Any identifier included in the URL must be [percent encoded][percent-encoding]
+to be recognized by the Conjur API.
+
+Examples:
+| Identifier             | Encoded                    |
+|------------------------|----------------------------|
+| `myapp-01`             | `myapp-01` _(no change)_   |
+| `alice@devops`         | `alice%40devops`           |
+| `prod/aws/db-password` | `prod%2Faws%2Fdb-password` |
+| `research+development` | `research%2Bdevelopment`   |
+| `sales&marketing`      | `sales%26marketing`        |
+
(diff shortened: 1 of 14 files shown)


apidocs/src/partials/url_encoding.md
```diff
@@ -0,0 +1,15 @@
+#### Note: entity IDs must be url-encoded
+
+Any identifier included in the URL must be [percent encoded][percent-encoding]
```
dustinmm80: Can you change the text here to be URL-encoded instead of percent encoded? We should use the same terminology throughout the docs. Everywhere else we say URL-encoded.

PR #516, opened 2018-04-03
Authn Core moved into a gem

jvanderhoof:
Moves to `authn_core` functionality into a gem. This will make it easier to share with new authenticators.


kgilpin:
But new Authenticators will (for now) be part of the same repo. So why do this?

jvanderhoof:
To create some modularity within the repo. Jonah's `authn-ldap` work is in an engine, which is also in the project, but located in the lib folder and added via Bundler.

PR #521, opened 2018-04-10
Policy log

dividedmind:
- Add `commited_at` column in `policy_versions` table that is automatically set to wallclock time on commit of a transaction or on creating a new version, whichever comes first.
- Add `current_policy_version` SQL function, looking up the policy version that hasn't been commited yet.
- Create a log of changes that a policy load causes by the way of database triggers. Modifying data outside of policy load currently emits a warning; it needs to be better understood when and if this should be an error.
- After loading a policy, emit log messages with the changes made.
- Add a `audit_socket` configuration option; setting it causes the log messages to be formatted with RFC5424 structured data and emitted to that socket. A syslog daemon can then listen on that socket and process the messages further, eg. by forwarding it to a system of permanent record, reformatting as eg. CEE (JSON), etc.

This is the first and most important part of the audit support in v5. It's based on open standards and allows leveraging COTS and free software tooling to process the records further based on specific requirements.

Probably the most interesting part will be the database migrations. RFC 5424 formatter is a bit rudimentary at this point and need refactoring and adding tests.

Simply running a server and loading a policy should result in policy log messages getting printed to the log. By setting `audit_socket` configuration option structured data can be obtained; note that something needs to be listening on the other end or the server won't start.


app/models/audit.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    def logger
+      @logger ||= Rails.logger
+    end
+
+    attr_writer :logger
+  end
+
+  module SDID
```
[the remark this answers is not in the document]
apotterri: I went back and looked at the spec, which helped me understand this better. I still think a comment here would be appropriate.

config/environments/appliance.rb
```diff
@@ -0,0 +1,67 @@
...
+  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
+  # the I18n.default_locale when a translation cannot be found).
+  config.i18n.fallbacks = true
+
+  # Send deprecation notices to registered listeners.
+  config.active_support.deprecation = :notify
+
+  config.log_level = :info
+  config.middleware.use Audit::RememberUuid
+  config.audit_socket = '/run/conjur/audit.socket'
```
apotterri: These three lines look the only changes to a default Rails environment file.  Are there other changes here I missed?

app/models/audit.rb
```diff
@@ -0,0 +1,99 @@
+require 'syslog'
+require 'logger'
+require 'time'
+
+module Audit
+  class << self
+    def notice msg, msgid, data
+      logger.info LogMessage.new msg, msgid, data, Syslog::LOG_NOTICE
```
kgilpin: Where does the severity of `LOG_NOTICE` lie in terms of significance?

app/models/audit.rb
```diff
@@ -0,0 +1,99 @@
...
+      @severity = severity
+    end
+
+    attr_reader :msgid, :structured_data, :severity
+  end
+
+  # Middleware to store request ID in a thread variable
+  class RememberUuid < Struct.new :app
+    def call env
+      Thread.current[:request_id] = env["action_dispatch.request_id"]
```
kgilpin: Maybe `conjur_request_id` for safety

app/models/audit.rb
```diff
@@ -0,0 +1,99 @@
...
+    SEVERITY_MAP = {
+      Logger::Severity::DEBUG => Syslog::LOG_DEBUG,
+      Logger::Severity::ERROR => Syslog::LOG_ERR,
+      Logger::Severity::FATAL => Syslog::LOG_CRIT,
+      Logger::Severity::INFO => Syslog::LOG_INFO,
+      Logger::Severity::WARN => Syslog::LOG_WARNING
+    }
+
+    # TODO
+    FACILITY = 32
```
kgilpin: Is this actually TODO? Please replace with a comment regarding this constant.

app/models/audit.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    def call severity, time, progname, msg
+      severity = if msg.respond_to? :severity
+        msg.severity
+      else
+        SEVERITY_MAP[Logger::Severity.const_get severity] rescue severity
+      end
+      sd = msg.structured_data if msg.respond_to? :structured_data
```
kgilpin: When would `msg` have `structured_data` and when wouldn't it?

app/models/audit.rb
```diff
@@ -0,0 +1,99 @@
...
+      progname ||= 'conjur'
+
+      # Use the request UUID when called with REST. Otherwise use PID.
+      pid = Thread.current[:request_id] || Process.pid
+
+      msgid = msg.msgid if msg.respond_to? :msgid
+      sd = format_sd sd
+
+      fields = [timestamp, hostname, progname, pid, msgid, sd, msg]
+      ["<#{severity + FACILITY}>1", *fields.map {|x| x || '-'}].join(" ") + "\n"
```
kgilpin: Is there any possibility of a newline in `x`?

app/models/loader/orchestrate.rb
```diff
@@ -118,6 +118,67 @@ def load
       store_passwords
 
       store_public_keys
+
+      emit_audit
```
kgilpin: We need to emit these after the transaction is committed, and keep track of whether each one has been sent.

app/models/loader/orchestrate.rb
```diff
@@ -118,6 +118,67 @@ def load
...
+    end
+
+    SDID = Audit::SDID
```
kgilpin: Some of these names are really cryptic. As a casual observer of the code I don't want to have to read the RFC to understand what this means. I would like to see some comments indicating what the variable and constant names mean and how they are used.

app/models/loader/orchestrate.rb
```diff
@@ -118,6 +118,67 @@ def load
...
+
+    OPS_MAP = {
+      "INSERT" => "add",
```
kgilpin: This seems to be humanizing the names? It's important to have a comment which explains where this is used; and why it's called `OPS_MAP`?

app/models/loader/orchestrate.rb
```diff
@@ -118,6 +118,67 @@ def load
...
+      "DELETE" => "remove",
+      "UPDATE" => "change"
+    }
+
+    SUBJECT_MAP = {
```
kgilpin: Indicate what this is for? It doesn't appear to be used in this file.

app/models/loader/orchestrate.rb
```diff
@@ -118,6 +118,67 @@ def load
...
+      :privilege => 'privilege'.freeze
+    }
+
+    def emit_audit
+      auth = policy_version.role.id
+      data = {
+        SDID::POLICY => { id: policy_id, version: policy_version.version },
+        SDID::AUTH => { user: auth }
+      }
+      format = "#{auth} %sed %s".freeze
```
kgilpin: What is the significance of this particular format string?

app/models/loader/orchestrate.rb
```diff
@@ -118,6 +118,67 @@ def load
...
+      :resource_id => 'resource'.freeze,
+      :role_id => 'role'.freeze,
+      :member_id => 'member'.freeze,
+      :name => 'annotation'.freeze,
+      :privilege => 'privilege'.freeze
+    }
+
+    def emit_audit
+      auth = policy_version.role.id
+      data = {
```
kgilpin: Why is this just called `data`? I would like it to be easier to understand what's happening. Does this become the `subject`? If so, or if not, what's the connection between `data` and `subject`?

app/models/loader/orchestrate.rb
```diff
@@ -118,6 +118,67 @@ def load
...
+        SDID::POLICY => { id: policy_id, version: policy_version.version },
+        SDID::AUTH => { user: auth }
+      }
+      format = "#{auth} %sed %s".freeze
+
+      def humanize_subject table, pk
```
kgilpin: Where will the string built by this function appear?

app/models/loader/orchestrate.rb
```diff
@@ -118,6 +118,67 @@ def load
...
+        when 'resources'
+          "resource #{pk[:resource_id]}"
+        when 'permissions'
+          "permission of #{pk[:role_id]} to #{pk[:privilege]} on #{pk[:resource_id]}"
+        when 'role_memberships'
+          "#{pk[:ownership] == 't' ? 'owner' : 'member'}ship of #{pk[:member_id]} in #{pk[:role_id]}"
+        end
+      end
+
+      def sd_subject pk
```
kgilpin: What is an `sd_subject`?

app/models/loader/orchestrate.rb
```diff
@@ -118,6 +118,67 @@ def load
...
+            pk[:ownership] == 't' ? 'owner' : 'member'
+          else
+            k.to_s.chomp "_id"
+          end, v]
+        end.compact]
+      end
+
+      db[:policy_log].where(policy_id: policy_id, version: policy_version.version).each do |log|
+        op = OPS_MAP[log[:operation]]
+        msg = format % [op.chomp('e'), humanize_subject(log[:kind], log[:subject])]
```
kgilpin: Singularize [:kind]?

app/models/policy_version.rb
```diff
@@ -33,6 +33,12 @@ class PolicyVersion < Sequel::Model(:policy_versions)
   alias policy resource
   alias policy= resource=
 
+  dataset_module do
+    def current
```
kgilpin: This appears to be used in the test cases only?

config/environments/appliance.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
+  # the I18n.default_locale when a translation cannot be found).
+  config.i18n.fallbacks = true
+
+  # Send deprecation notices to registered listeners.
+  config.active_support.deprecation = :notify
+
+  config.log_level = :info
+  config.middleware.use Audit::RememberUuid
```
kgilpin: Does this need to be inserted before or after any other specific middleware?

config/environments/appliance.rb
kgilpin: How would I enable audit using in `RAILS_ENV=production`? For authn-local, the audit socket location is basically hard-coded and the service just starts if the socket exists. Any thoughts on the benefits of one approach vs the other?

config/initializers/authenticator.rb
```diff
@@ -6,11 +6,6 @@
 
 Sequel.extension :migration
 
-# Token authentication is optional for authn routes, and it's not applied at all to authentication.
-Possum::Application.config.middleware.use Conjur::Rack::Authenticator,
```
kgilpin: Just curious why we moved this?

db/migrate/20180410071554_current_policy.rb
```diff
@@ -0,0 +1,58 @@
+Sequel.migration do
+  up do
+    alter_table :policy_versions do
+      set_column_type :created_at, :timestamptz
```
kgilpin: For my edification: What is `timestamptz` and what makes it a better choice?

db/migrate/20180410071554_current_policy.rb
```diff
@@ -0,0 +1,58 @@
...
+      LANGUAGE plpgsql AS $$
+        BEGIN
+          UPDATE policy_versions pv
+            SET commited_at = clock_timestamp()
+            WHERE commited_at IS NULL;
+          RETURN new;
+        END;
+      $$;
+
+      -- deferred constraint trigger will run on transaction commit
```
kgilpin: But we are calling `emit_audit` before the transaction is committed, so `policy_versions_commit` will not have happened. Yes?

db/migrate/20180410071554_current_policy.rb
```diff
@@ -0,0 +1,58 @@
...
+      CREATE CONSTRAINT TRIGGER commit_current
+        AFTER INSERT ON policy_versions
+        INITIALLY DEFERRED
+        FOR EACH ROW
+        WHEN (NEW.commited_at IS NULL)
+        EXECUTE PROCEDURE policy_versions_commit();
+
+      -- if any version is current while creating new one, commit it
```
kgilpin: Can you explain what this means in practice?

db/migrate/20180410071554_current_policy.rb
```diff
@@ -0,0 +1,58 @@
...
+      CREATE TRIGGER only_one_current
```
kgilpin: When would we try and load a policy in a transaction when the previous one is still open? Should that be an error instead? I would rather forbid things that are weird and should never happen rather than try and handle them in a way that's hard to understand.

db/migrate/20180410071554_current_policy.rb
```diff
@@ -0,0 +1,58 @@
...
+        BEFORE INSERT ON policy_versions
+        FOR EACH ROW
+        EXECUTE PROCEDURE policy_versions_commit();
+
+      CREATE FUNCTION current_policy_version()
+        RETURNS SETOF policy_versions
```
kgilpin: Could there really be more than one? What does it mean to return `SETOF`? When this function is called elsewhere, it seems like the caller assumes there will only be one result.

db/migrate/20180410092453_policy_log.rb
```diff
@@ -0,0 +1,87 @@
...
+    key = %i(policy_id version)
+    create_table :policy_log do
+      String :policy_id, null: false
+      Integer :version, null: false
+      foreign_key key, :policy_versions, on_delete: :cascade
+      index key
+
+      column :operation, :policy_log_op, null: false
+      column :kind, :policy_log_kind, null: false
+      column :subject, :hstore, null: false
```
kgilpin: Why is the data in an audit record called `subject`? That's really confusing to me.

db/migrate/20180410092453_policy_log.rb
```diff
@@ -0,0 +1,87 @@
...
+      column :at, :timestamptz, null: false, default: Sequel.function(:clock_timestamp)
```
kgilpin: We will add `roles` and `resources` for indexing and query purposes?

db/migrate/20180410092453_policy_log.rb
```diff
@@ -0,0 +1,87 @@
...
+            END IF;
+            current = current_policy_version();
+            IF current.resource_id = subject.policy_id THEN
+              INSERT INTO policy_log(
+                policy_id, version,
+                operation, kind,
+                subject)
+              SELECT
+                current.resource_id, current.version,
+                TG_OP::policy_log_op, '#{table}'::policy_log_kind,
```
kgilpin: Singularize `kind`.
`role_memberships` as a `kind` of operation sounds too SQL-y?
The `subject` of the insert includes all the insert data? That's not intuitive from the field name.
Subject data as `resource_id` seems SQL-y. Would prefer simply `resource`.
Role and/or privilege grant should include the `owner` field.
We need to know the role which initiated each action.
We need to be sure that we can index the roles and resources as in v4.
`operation` as all-caps is odd compared to the other fields.
We should take the opportunity to use cucumbers as the definitive specification of audit data formats.
We should compare the audit data formats to CyberArk audit, to incorporate the learnings and best practices.
IP address is not being captured.

What's the syntax `TG_OP::policy_log_op`?
How does a table know its `policy_log_kind`?

For my edification: What does slice on an hstore and `literal primary_key` do? What does `literal` do?

db/migrate/20180410092453_policy_log.rb
```diff
@@ -0,0 +1,87 @@
...
+                slice(hstore(subject), #{literal primary_key})
+              ;
+            ELSE
+              RAISE WARNING 'modifying data outside of policy load: %', subject.policy_id;
```
kgilpin: Is this a warning or an error? Is it here so that mucking with the tables isn't blocked by auditing? Do we want to enforce that table modification can only be performed if there is an open policy_version?

If we were sure that every change was audited, then the current state of the audited tables could be built from scratch by replaying the audit. If we allow unaudited access, then we can't make that guarantee. 

I could imagine a switch that allows unaudited writes in dev and test; without that switch in place, this line would be an error and abort the transaction.

db/migrate/20180410092453_policy_log.rb
```diff
@@ -0,0 +1,87 @@
...
+            END IF;
+            RETURN subject;
+          END;
+        $$ LANGUAGE plpgsql
+        SET search_path FROM CURRENT;
+
+        CREATE TRIGGER policy_log
+          AFTER INSERT OR UPDATE ON #{table}
+          FOR EACH ROW
+          WHEN (NEW.policy_id IS NOT NULL)
```
kgilpin: When is `policy_id` NULL?

spec/models/policy_version_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    it 'returns the currently loading policy version' do
+      pv = PolicyVersion.create \
+        policy: policy('current'),
+        role: owner,
+        policy_text: '[]'
+      expect(PolicyVersion.current).to eq pv
+    end
+  end
+
+  it "commits previously active policy version" do
```
kgilpin: What does `commits` mean here, since it's not really committing the transaction?

app/models/audit.rb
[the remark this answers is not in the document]
dividedmind: Fair. I wanted to refactor this whole part anyway, so I'll make sure it's clearer.

app/models/loader/orchestrate.rb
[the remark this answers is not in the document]
dividedmind: Good catches, thanks. Not freezing is an omission, `SUBJECT_MAP` is a leftover from a previous iteration.

db/migrate/20180410071554_current_policy.rb
```diff
@@ -0,0 +1,58 @@
+Sequel.migration do
+  up do
+    alter_table :policy_versions do
+      set_column_type :created_at, :timestamptz
+      add_column :commited_at, :timestamptz,
```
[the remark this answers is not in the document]
dividedmind: Touche! Thanks :)

config/environments/appliance.rb
dividedmind: No, there are none.

app/models/audit.rb
dividedmind: Thing is, syslog and Ruby severity levels are different. Notice is between warn and info, eg. "Normal but significant conditions." Cf. https://en.wikipedia.org/wiki/Syslog#Severity_level

app/models/audit.rb
dividedmind: I kept going back and forth on this. However there's nothing Conjur-specific about this request ID.

app/models/audit.rb
dividedmind: Yes, this should be configurable once this code is factored out.

app/models/audit.rb
dividedmind: If it's a normal string it won't. This formatter is meant to be a compatible drop-in for normal Ruby logger formatter.

app/models/audit.rb
dividedmind: Depends on the input. Formatters don't usually guard against it, so I chose not to here, too.

app/models/loader/orchestrate.rb
dividedmind: I was thinking about it and I really don't think it's necessary. If audit can't be sent in full, the transaction should just be rolled back.

app/models/loader/orchestrate.rb
dividedmind: Fair points.

app/models/loader/orchestrate.rb
dividedmind: It's mapping SQL operations to humanized names for formatting a log message.

app/models/loader/orchestrate.rb
dividedmind: Yup, leftover, thanks. I changed it to the `case` below.

app/models/loader/orchestrate.rb
dividedmind: This is just for the human-readable message.

app/models/loader/orchestrate.rb
dividedmind: `data` is the structured data to be output with the message. `subject` is the subject of this operation. `subject` field in the input gets translated to `subject@43868` SD-ID of the output.

config/environments/appliance.rb
dividedmind: With `RAILS_ENV=production` as with the other environments the audit messages go to the normal log sink (formatted with that sink's formatter). 

This is 'hardcoded' here specifically for `appliance` environment which is what the environment is set to in the appliance, which seems appropriate; this is a configuration file after all. Perhaps this configuration file should go in the appliance instead of this repository, but I thought it'd make it harder to manage (and also that's how we had it in v4).

config/initializers/authenticator.rb
dividedmind: When I was figuring out how to add the middleware I noticed this; I figured it makes more sense to have it in a configuration file than in an initializer which ought to be more generic. I don't feel strongly on this subject however :)

db/migrate/20180410071554_current_policy.rb
dividedmind: `timestamptz` is short for `TIMESTAMP WITH TIME ZONE` which is exactly the same as `TIMESTAMP WITHOUT TIME ZONE` except if you for some reason use local time instead of UTC, Postgres translates it accordingly on I/O and always stores UTC in the database.

We ought to use UTC everywhere anyway, but this assumption has proven to be a source of a couple bugs and mistifying test failures.

db/migrate/20180410071554_current_policy.rb
dividedmind: Agreed. However, this will require refactoring some existing tests, and I wasn't sure we aren't doing it somewhere outside of tests, too.

db/migrate/20180410071554_current_policy.rb
dividedmind: A function returning `SETOF` returns a table of records of given type. This particular function only ever returns a table that has at most one row, but it simplifies its usage significantly that it returns a set (eg. you can use it in `FROM` clause).

db/migrate/20180410092453_policy_log.rb
dividedmind: No, not here. This is just for pushing data out. A separate service will do querying.

db/migrate/20180410092453_policy_log.rb
dividedmind: I'm confused as to what's confusing. It's the PK of the subject of this policy log record.

db/migrate/20180410092453_policy_log.rb
dividedmind: This is meant to be SQL-y. It's just the log of DB operations. It's going to be pushed out to a separate storage and querying backend. Which role initiated each action is already stored in `policy_versions`.

`TG_OP::policy_log_op` is a cast to `policy_log_op` type. Table doesn't know its `policy_log_kind`, it's embedded in the text of the trigger function. `literal primary_key` transforms the `primary_key` to SQL literal, eg. it makes `%w(foo bar)` into `('foo', 'bar')`.

db/migrate/20180410092453_policy_log.rb
dividedmind: I left it as a warning for now since it happens in tests a bit and I wasn't sure if it isn't happening elsewhere.

db/migrate/20180410092453_policy_log.rb
dividedmind: I've seen it happen in tests.

spec/models/policy_version_spec.rb
dividedmind: Finalization of the policy version load. I'll push a commit shortly which renames it so it'll hopefully become clearer.

app/models/audit.rb
kgilpin: Wouldn't it be great to have a comment in the code to this effect?

app/models/loader/orchestrate.rb
kgilpin: But then you have audits of some things happening that never happened?

app/models/loader/orchestrate.rb
kgilpin: I guess I need to be more explicit. With these questions I am asking for a comment to be added to the code so that the person who comes along will be able to see the explanation. If the response is only in a PR comment then it will be essentially lost once the PR is merged.

app/models/loader/orchestrate.rb
kgilpin: Please comment accordingly.

app/models/loader/orchestrate.rb
kgilpin: Please comment accordingly.

app/models/loader/orchestrate.rb
kgilpin: Please comment accordingly.

config/environments/appliance.rb
kgilpin: But, if I am using the open source edition, and I want to get the audit messages, I want to be able to add a setting to enable the audit without having to modify a source file. And I want to run with `RAILS_ENV=production`

db/migrate/20180410092453_policy_log.rb
kgilpin: But it will want to see indexed columns on the table, right?

kgilpin:
Thanks for all the responses. The only thing I ask at this point is that some of the answers be added to the code as comments so they are not lost when the PR closes.

kgilpin:
@dividedmind I am approving the PR but I would like to see some comments added as I've written above. I don't want to be blocking the merge though so once the comments are in, we are good.

app/models/loader/orchestrate.rb
dividedmind: Only in tight disaster scenarios, which I doubt will ever happen in practice. Even if it does happen, the entries be easily identified as spurious by virtue of duplicate policy version (or referring to a nonexistent one).

dividedmind:
@kgilpin I did add some comments in f1e277d, let me know if you still feel something is missing.

db/migrate/20180410092453_policy_log.rb
dividedmind: It won't use this table at all :) I suppose you could think of that service as being an audit receiver, to (ab)use v4 parlance.

app/models/loader/orchestrate.rb
apotterri: What happens if the process reading data from the `audit_socket` crashes in the middle of reading events from the socket? Will that cause the transaction to abort? If so, and given that we don't have control over that process, I can definitely imagine that happening in practice.

OTOH, we're storing the audit records in the database, right? So, I guess we could say that the db table is the system of record, and the audit stream might have spurious events in it.

dividedmind: It will. But we do have control over that process, at least on the appliance; namely, it's the local syslog daemon. It's highly unlikely it crashes in the middle of reading events from the socket.

dividedmind:
@apotterri are there any other changes yo

app/models/loader/orchestrate.rb
dividedmind: We won't have control over other processes, but this is the responsibility of the user. Conjur core does not provide a system of record for the audit data, it just produces an event stream.

dividedmind:
Agreed. Thanks.

PR #522, opened 2018-04-12
Authn ldap - initial end to end without tests

izgeri:
Creates working end to end authn-ldap, ready for initial PR

Also: Fixes the "Role table is hit twice" problem by creating a class MemoizedRole that delegates to Role, but memoizes results of lookups by id.

This has been tested manually, but will need quick followup projects for:

- Cucumber tests
- Unit tests for AuthenticationSecurity (+ refactoring)
- Pulling AuthenticateController logic into an object
- Unit testing that object
- Updating documentation


app/controllers/authenticate_controller.rb
```diff
@@ -1,30 +1,78 @@
...
+
+    validate_role_exists!(role_id)
+    
+    case authenticator
+    when 'authn-ldap'
+      validate_security!(authenticator, account, service_id, username)
+      raise Unauthorized unless ldap_authenticator.valid?(username, password)
+      role = Role[role_id]
+    # default conjur: authn
     else
```
izgeri: @jonahx it should check specifically for `when 'authn'` and return 401 if they do not return an authenticator that is valid

config/application.rb
```diff
@@ -13,6 +13,7 @@
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
+Bundler.require(:local)
```
jvanderhoof: We shouldn't need this anymore (it was to load the locally referenced gems).

app/controllers/authenticate_controller.rb
```diff
@@ -1,30 +1,89 @@
...
+    logger.debug("Unexpected AuthenticatorSecurity Error: #{e.message}")
+    raise Unauthorized
   end
 
-  def credentials_lookup
-    roleid = Role.roleid_from_username(account, params[:id])
-    @credentials = Credentials[roleid]
-    unless @credentials
-      logger.debug "Credentials for #{roleid} not found"
+  def validate_role_exists!(role_id)
```
izgeri: this is just checking that the `role_id` isn't falsey, but to me `validate_role_exists` implies that we are checking in the DB that the role referenced by the role_id exists

lib/authenticators/ldap/authn.rb
```diff
@@ -0,0 +1,40 @@
+require 'active_support'
```
jvanderhoof: This file looks like a duplicate of a class in `lib/authenticators/ldap/authn_ldap.rb`.  Do we need them both?

app/controllers/authenticate_controller.rb
```diff
@@ -1,30 +1,89 @@
...
+    service_id     = params[:service_id]
+    account        = params[:account]
+    username       = params[:id]
+    password       = request.body.read
+    role_id        = MemoizedRole.roleid_from_username(account, username)
+
+    validate_role_exists!(role_id)
+    
+    case authenticator
+    when 'authn-ldap'
```
izgeri: all custom authn are going to have to  `validate_security!`, so it would be nice if this was written so that everyone adding a custom authn didn't have to remember to do that

app/controllers/authenticate_controller.rb
```diff
@@ -1,30 +1,89 @@
...
+      validate_security!(authenticator, account, service_id, username)
+      raise Unauthorized unless ldap_authenticator.valid?(username, password)
+      role = MemoizedRole[role_id]
+    # default conjur: authn
     else
```
izgeri: we need to be careful that we send a 401 when the `authenticator` is something besides a custom authn that is handled explicitly or exactly `authn`. that is, we shouldn't have a blanket `else` here - it should be `when 'authn'` (do the usual standard authn stuff) and `else` 401.

app/models/authn_local.rb
```diff
@@ -49,7 +49,7 @@ def run
             conn.puts issue_token(claims)
```
izgeri: these changes could be removed from this PR. they live on in a separate branch should we choose to pursue them.

dev/sample-ldap-webservice-entitlement.yml
```diff
@@ -0,0 +1,3 @@
+- !grant
```
izgeri: this can be merged into dev/sample-ldap-policy.yml

app/models/authn_local.rb
jvanderhoof: Agreed.  Alan my run into UI issues with this change.

PR #528, opened 2018-04-20
Improve Development flow

jvanderhoof:
This PR adds a `dev/start` script that brings up a development environment.  A user can now pass in a `--authn-ldap` flag to start OpenLDAP and load the required policy.

This effort is the beginning of simplifying our development & test experience on a per project basis.  There will be additional work to bring the Dockerfile used for development & production closer together. We'll also be improving the experience of running automated tests.

Most of the action happens in the `dev/start` shell file.

Pull this PR down and follow the directions for setting up a development environment with Authn-LDAP.  After starting the server, issue a Curl request to authenticate as `alice`:

```sh
$ curl -v -X POST -d "alice" http://localhost:3000/authn-ldap/test/cucumber/alice/authenticate
```
The result should be a Conjur token.

### DIFF
--- a/dev/stop
+++ b/dev/stop
@@ -1,0 +1,5 @@
+#!/bin/bash -ex
+
+docker-compose down -v
+
+rm -f data_key


dev/start
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash -ex
+
+# Determine which extra services should be loaded when working with authenticators
```
apotterri: If we're going to use this as an example of how to write a `start` script,  we should use `getopt` for argument parsing: http://www.bahmanm.com/blogs/command-line-options-how-to-parse-in-bash-using-getopt .

README.md
```diff
@@ -86,25 +86,36 @@ environment with a database container (`pg`, short for *postgres*), and a
...
    ```sh-session
-   $ ./build.sh
+   $ cd dev
+   $ ./start
+   ...
+   root@f39015718062:/src/conjur#
    ```
-3. start the container:
+
+   Once the start.sh script finishes, you're in a Bash shell in the Conjur
```
apotterri: This should be `start`, not `start.sh`.

dev/docker-compose.yml
```diff
@@ -51,8 +54,6 @@ services:
 
   client:
     image: conjurinc/cli5
-    # entrypoint: sleep
-    # command: infinity
```
apotterri: I'd be inclined to leave these in, and maybe even uncomment them. Because we no longer install the CLI in the conjur container, you almost always want to bring up a client.

jonahx:
It won't let me comment on lines you didn't change, so I think we should note:

- That can rspec can be run from the host.  (It works for me, anyway.  If that's not true in general, ignore this comment)
- I do think a single command, run from the host, that does exactly what will ci will do is important (ie, run all the tests the same).  For the dev process, the current description is fine.

dev/start
jonahx: @apotterri (partial) counterpoint:
https://mywiki.wooledge.org/BashFAQ/035#Manual_loop

dev/docker-compose.yml
jvanderhoof: I'll leave them in. I also added a CLI dev tool to handle loading policy: https://github.com/cyberark/conjur/pull/529

jvanderhoof:
@jonahx - I'm working on a separate testing set of changes.

dev/start
apotterri: That appears to be comparing it with `getopts` (with an `s`), which is older and less functional than `getopt`.  I believe all the criticisms they offer of `getopts` have now been addressed in the new command.

I think we use the standard thing, rather than rolling our own in each project.

PR #534, opened 2018-04-26
Resource visibility restrictions

dividedmind:
This pull request makes it so resources are only visible to users which hold a permission on them or own them. Since we don't need the account visibility cutoff anymore, I also added a `/resources` route which searches all resources the user might see, regardless of the account.


config/routes.rb
```diff
@@ -40,6 +40,8 @@ def matches?(request)
 
       get "/resources/:account" => "resources#index"
 
+      get "/resources" => "resources#index"
+
```
apotterri: Seems like `/resource/:kind` could be useful here, too.

dividedmind: I assume you mean `/resources/:kind`; then it would clash with the account route. I suppose if it was `/resource/:kind` (singular) then it wouldn't clash, but it would be a bit unexpected.

It's a good idea to add a route qualified only with a kind, but I'm not sure how would that look to make sense. `/resource/:kind` is weird because it's not just one resource; I almost wish we had `/resources/:kind` in the first place, and `/accounts/:account/resources/...` would then qualify by account. Of course we could merge the routes and have `/resources/:account_or_kind`, but the meaning would depend on the state of the system and that'd _really_ be confusing.

Hmm... what do you think about a wildcard placeholder, eg. `/resources/_/:kind`?

kgilpin: I think `/resources?kind=foo` suffices

config/routes.rb
```diff
@@ -40,6 +40,8 @@ def matches?(request)
 
       get "/resources/:account" => "resources#index"
```
kgilpin: I feel like we should drop this? Or at least deprecate and remove it from the doc.

cucumber/api/features/resource_visibility.feature
```diff
@@ -1,7 +1,109 @@
...
+  Scenario: Resources from a foreign account can be visible
+    Given I create a new resource in a foreign account
+    And I create a new user "alice"
+    And I permit user "alice" to "try" it
+
+    When I login as "alice"
+    And I successfully GET "/resources"
+    Then the resource list should include the newest resource
+
+    When I successfully GET "/resources/cucumber"
```
kgilpin: I think at this point it's better to document this as `GET /resources?account=cucumber`

cucumber/api/features/resource_visibility.feature
```diff
@@ -1,7 +1,109 @@
...
+    Then the resource list should not include the newest resource
+
+  Scenario: Resources without permissions or ownership are not visible
+    Given I create a new resource called "probe"
+    And I create a new user "alice"
+
+    When I login as "alice"
     And I successfully GET "/resources/cucumber"
```
kgilpin: Again, account as parameter rather than path component.

cucumber/api/features/resource_visibility.feature
```diff
@@ -1,7 +1,109 @@
...
+    Given I create a new resource called "probe"
+    And I successfully POST "/secrets/cucumber/test-resource/probe" with body:
+    """
+    v-1
+    """
+    And I create a new user "alice"
+
+    When I login as "alice"
+    And I GET "/secrets/cucumber/test-resource/probe"
+    Then the HTTP response status code is 404
```
kgilpin: Wouldn't this already be covered by the user not having `execute` on the variable?

db/migrate/20180422043957_resource_visibility.rb
```diff
@@ -0,0 +1,53 @@
...
+      RETURNS boolean
+      LANGUAGE sql STABLE STRICT AS $$
+        WITH RECURSIVE search(role_id) AS (
+          -- We expand transitively back from the set of roles that the
+          -- resource is visible to instead of relying on all_roles().
+          -- This has the advantage of not being sensitive to the size of the
+          -- role graph of the argument and hence offers stable performance
+          -- even when a powerful role is tested, at the expense of slightly
+          -- worse performance of a failed check for a locked-down role.
+          -- This way all checks take ~ 1 ms regardless of the role.
```
kgilpin: Just for my sanity, this is the same optimization that we made in `authz`, right?

config/routes.rb
apotterri: Yeah, I meant `/resources/:kind`.

@kgilpin's suggestion that we use a query string works for me.

config/routes.rb
apotterri: This feels like it should be a separate story. What do you think?

cucumber/api/features/resource_visibility.feature
apotterri: I think this is different. SecretsController starts by looking for the secret's resource, which will fail if the user has no privilege on the secret.

It seems like we should have scenarios analogous to this one for the other controllers that do this kind of checking (PoliciesController and HostFactoryTokensController). I'll add those.

db/migrate/20180422043957_resource_visibility.rb
apotterri: I took a look at the code in `authz`, the functionality here looks the same to me.

config/routes.rb
kgilpin: Yes, it does.

cucumber/api/features/secrets_authorization.feature
```diff
@@ -39,3 +39,13 @@ Feature: RBAC privileges control whether a role can update and/or fetch a secret
     """
     v-1
     """
+
+  Scenario: Fetching a secret without any permission on it
+    If the user doesn't have any permission or ownership of a secret, fetching
+    it should return 404 (not 403) even if it exists.
```
kgilpin: Yep, I like this.

PR #535, opened 2018-04-26
Authn ldap

jonahx:

# What does this pull request do?

- Adds framework for pluggable authenticators (details below)
- Refactors AuthenticateController logic into an `Authentication::Strategy` object

# What background context can you provide?

### To add a new authenticator:

1. Create a new directory under `/app/domain/authentication`.  For example:
```
/app/domain/authentication/my_authenticator
```
2. That directory must contain a file named `authenticator.rb`, with the
   following structure:
```ruby
module Authentication
  module MyAuthenticator
    
    class Authenticator
      def initialize(env:)
        # initialization code based on ENV config
      end

      def valid?(input)
        # input has 5 attributes:
        #
        #     input.authenticator_name
        #     input.service_id
        #     input.account
        #     input.username
        #     input.password
        #
        # return true for valid credentials, false otherwise
      end
    end

  end
end
```

That's it.  However, before using the authenticator...  

### security requirements

1. Enable it by adding it to the following `ENV` variable:
```
CONJUR_AUTHENTICATORS=my-authenticator/optional-service-id,another-authenticator/service-id
```
2. Add the authenticator as a webservice in your conjur policy:
```yaml
- !policy
  id: conjur/my-authenticator/optional-service-id
```
3. Add any users that need to access it to your policy, and grant them the
   `authenticate` privilege.

### Using the authenticator

1. To authenticate, `POST` to:
```
/my-authenticator/optional-service-id/my-account/username/authenticate
```
with the password as text in the request body

### Other Notes

1. Your authenticator directory can contain other supporting files used by your
   authenticator.
2. Conjur will instantiate your autheticator at bootup.  By default, when your
   authenticator is instantiated by conjur, it will be passed the `ENV` through
   the kwarg `env`.  If you don't need any configuration from the environment,
   you can opt out like so:
```ruby
module Authentication
  module MyAuthenticator
    
    class Authenticator
      def self.requires_env_arg?
        false
      end

      def initialize
        # you could also omit this altogether
      end

      def valid?(input)
        # same as before
      end
    end

  end
end
```

## Technical Notes

- The architecture is objects nested like Russian dolls.  All dependencies are
  passed explicitly through the constructor.
- It also uses `Dry::Struct` quite a bit.  You can think of this like an
  `ostruct` with built-in type checking, which cleans up what would otherwise
  be verbose validation and initialization code.

# Where should the reviewer start?

- Start in the `AuthenticateController`, which delegates to the
  `Authentication::Strategy` object, which delegates to `Security` and the
  individual authenticators.

# How should this be manually tested?

- Run the cukes.  See the `cyberark/conjur` README for details.

# Link to build in Jenkins (if appropriate)

# Questions:
> Does this have automated Cucumber tests?

Yes

Perhaps.

The PR above will be converted to full documentation in a followup project.


app/controllers/authenticate_controller.rb
```diff
@@ -1,90 +1,30 @@
...
-  def validate_security!(authenticator, account, service_id, user_id)
-    security = Authenticators::Security.new(authn_type: authenticator,
-      account: account, role_class: MemoizedRole)
-    security.validate(service_id, user_id)
-  rescue AuthenticatorNotEnabled, ServiceNotDefined, NotAuthorizedInConjur => e
-    logger.debug(e.message)
-    raise Unauthorized
   rescue => e
-    logger.debug("Unexpected Authenticators::Security Error: #{e.message}")
+    puts '*******************************'
```
jvanderhoof: What do you think about using the application `logger` for output?

ci/test-ldap
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash -ex
```
jvanderhoof: This file should be removed. It's now handled by `ci/test --cucumber-authenticators`

config/application.rb
```diff
@@ -28,7 +28,13 @@ class Application < Rails::Application
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
 
-    config.autoload_paths << Rails.root.join('lib')
+    # config.autoload_paths << Rails.root.join('lib')
```
jvanderhoof: With the bulk of the work in `app/domains/...`, do we still need these changes?

jonahx: Don't think it's needed for my code anymore.  Tomorrow I'll try pulling it and rerun the cukes.

app/controllers/authenticate_controller.rb
jonahx: Yeah probably best.  I'll make that change.  I think eventually we should return json error response bodies, but that's a different update.

jonahx:
@jvanderhoof Just pushed fixes for your request above

jonahx:
@kgilpin Responding to your comments in one block:
- Yes, the TODOs / comments will all be resolved.  We just wanted to get things rolling.
- As Jason said,`eager_load` is needed for the introspection that allows new authenticators to be auto-detected.  This could probably be rewritten by introspecting the file system rather than the ruby classes, but would make things messier.  Unless there is a good reason we should avoid eager loading (which is possible), I think we should keep it.
- Currently all errors will be logged by the rescue block in the authenticate controller.  Which I think is fine for now, but lmk if you disagree.  In a later story, we can convert the messages (only those that wouldn't leak security information indirectly) to the MS json format we use.

spec/lib/authenticators/security_spec.rb
```diff
@@ -0,0 +1,75 @@
+# require 'authenticators/security'
```
apotterri: If this isn't ready to go yet, maybe it should get deleted from the PR?

PR #541, opened 2018-05-04
Authn ldap strategy

jonahx:

https://ca-il-jira.il.cyber-ark.com:8443/browse/CONJ-5033

Adds unit tests for the Strategy objects and it's sub-object.

`strategy_spec.rb` file.

Run the rspec tests, `./test --rspec`.

https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--conjur/detail/authn-ldap-strategy/4/pipeline/

#### Questions:

The doc tasks are already in JIRA

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,12 +96,17 @@ To use it:
    root@f39015718062:/src/conjur#
    ```
 
-   Once the `start` script finishes, you're in a Bash shell in the Conjur
+   Once the `start` script finishes, you're in a Bash shell inside the Conjur
    server container.
 
-   **LDAP Authentication**
+   After staring Conjur, your instance will be configured with the following:
+   * Account: `cucumber`
+   * User: `admin`
+   * Password: Run `conjurctl role retrieve-key cucumber:user:admin` inside the container shell to retrieve the admin user API key (which is also the password)
 
-   To use LDAP to authentication, run `start` with the `--authn-ldap` flag:
+  #### LDAP Authentication
+
+   To enable a user to log into Conjur using LDAP credentials, run `start` with the `--authn-ldap` flag:
 
    ```sh-session
    $ cd dev
@@ -153,7 +158,7 @@ The `./cli` script, located in the `dev` folder is intended to streamline these 
 $ ./cli --help
 
 NAME
-    cli - Developement tool to simplify working with a Conjur container.
+    cli - Development tool to simplify working with a Conjur container.
 
 SYNOPSIS
     cli [global options] command [command options] [arguments...]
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -158,7 +153,7 @@ The `./cli` script, located in the `dev` folder is intended to streamline these
 $ ./cli --help
 
 NAME
-    cli - Development tool to simplify working with a Conjur container.
+    cli - Developement tool to simplify working with a Conjur container.
```
izgeri: this typo crept back in

izgeri: actually, I'm not sure why a lot of the changes to this file were dropped

migrate.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 
-heroku run conjurctl db migrate
+heroku run possum db migrate
```
izgeri: does this undo a recent change? I feel like this might be connected to a recent issue

app/domain/authentication/strategy.rb
```diff
@@ -8,15 +8,31 @@ class Strategy < ::Dry::Struct
...
 
+
     class Input < ::Dry::Struct
       attribute :authenticator_name, Types::NonEmptyString
       attribute :service_id,         Types::NonEmptyString.optional
       attribute :account,            Types::NonEmptyString
       attribute :username,           Types::NonEmptyString
       attribute :password,           Types::NonEmptyString
+
+      def to_access_request(env)
```
apotterri: Maybe I'm just missing some context, but I don't understand this method name. Looking at the places it's used didn't really help, either. How is it meant to be read?

ci/test
```diff
@@ -42,7 +42,7 @@ function run_cucumber_tests() {
   profile="$1"
 
   # Generate a data key
-  export CONJUR_DATA_KEY="$(openssl rand -base64 32)"
+  CONJUR_DATA_KEY="$(openssl rand -base64 32)"
```
apotterri: If you don't `export` this, it does nothing, because it's not used anywhere else in this file. Better to just remove it altogether.

migrate.sh
apotterri: This is wrong -- it's what broke the eval.conjur.org deploy last week. The command should be `conjurctl`.

app/domain/authentication/strategy.rb
apotterri: Maybe part of the problem is that I don't really understand what an `Input` is.

ci/test
apotterri: So, yes, this change would have fixed the problem of having each cucumber run get a different `CONJUR_DATA_KEY`. *None* of them would be using the one set here. :)

Unless I'm missing something, it should go.

PR #547, opened 2018-05-12
WIP Credential factory

kgilpin:

A Credential Factory is a Conjur service which creates a new set of credentials each time it is called. A credential factory is a secure and scalable way to expose a credential-issuing backend, such as AWS Security Token Service, to Conjur clients.

See dependency PR - https://github.com/conjurinc/conjur-policy-parser/pull/9

AWS Security Token Service (STS) is an AWS service which issues temporary credentials with a client-specified IAM policy. 

Conjur Credential Factory is a way to expose the capabilities of STS-like services without over-privileging clients or requiring clients to directly handle permanent AWS credentials. A client can use the familiar "Conjur secrets fetch" interface to obtain a temporary set of AWS credentials with a set of privileges that are pre-defined by the security administrator. Many different credential factories can be defined on a single set of permanent AWS credentials. Access to credential factories is managed exactly the same way as access to variables, providing a familiar workflow and granular control.

Read and understand the new cucumber tests and the CredentialFactory.md.

CredentialFactory.md shows how to try this from the Rails console. You can also invoke `GET /secrets/cucumber/webservice/myapp/aws-credentials` via cURL (with a valid access token, of course).

https://jenkins.conjur.net/job/cyberark--conjur/job/feature%252Fcredential-factory/

#### Questions:

> Does this have automated Cucumber tests?

Yes

Yes

Not completely.

Yes

### DIFF
--- a/CredentialFactory.md
+++ b/CredentialFactory.md
@@ -1,8 +1,8 @@
 # Credential Factory
 
-Generates credentials dynamically from a backend which supports it, such as AWS Security Token Service.
+A credential factory is a Conjur service which creates a new set of credentials each time it is called. A credential factory is a secure and scalable way to expose a credential-issuing backend, such as AWS Security Token Service, to Conjur clients.
 
-# Policy configuration
+# Example : AWS STS credential factory
 
 Load a policy file, such as [aws_credential_factory.yml](./run/aws_credential_factory.yml). In one policy, such as `aws-dev`, define variables for AWS access key id, secret, and region. For example:
 
@@ -17,13 +17,13 @@ Create a `secrets-users` group with access to these credentials.
 
 Load valid credentials into the variables and set the `region`.
 
-In another policy, such as `myapp`, define a `!credential-factory` object and add an entitlement which grants it the `secrets-users` group (for example, `aws-dev/secrets-users`. The Credential Factory will access these credentials in order to call the AWS Security Token Service [GetFederationToken](https://docs.aws.amazon.com/STS/latest/APIReference/API_GetFederationToken.html) function. You must provide an annotation `credential-factory/policy`, which is the IAM policy which will be applied to the generated credentials.
+In another policy, such as `myapp`, define a `!webservice` object and add an entitlement which grants it the `secrets-users` group (for example, `aws-dev/secrets-users`. This web service will access these credentials in order to call the AWS Security Token Service [GetFederationToken](https://docs.aws.amazon.com/STS/latest/APIReference/API_GetFederationToken.html) function. You must provide an annotation `credential-factory/policy`, which is the IAM policy which will be applied to the generated credentials.
 
 You can also define optional annotations `credential-factory/duration-seconds` and `credential-factory/user-name`. 
 
-Then, for any role which you want to use the Credential Factory, give it the `execute` privilege on the `!credential-factory`. The client role **does not** need any privileges on the long-lived AWS credentials.
+Then, for any role which you want to use the Credential Factory, give it the `execute` privilege on the `!webservice`. The client role **does not** need any privileges on the long-lived AWS credentials.
 
-The client role can now fetch a secret using the Conjur API using the object id `<acct>:credential_factory:<the-factory-id>`, just like fetching a secret from a variable using `<acct>:variable:<the-variable-id>`. 
+The client role can now fetch a secret using the Conjur API using the object id `<acct>:webservice:<the-factory-id>`, just like fetching a secret from a variable using `<acct>:variable:<the-variable-id>`. 
 
 The data will be returned in JSON format with the following fields:
 
@@ -36,7 +36,7 @@ The data will be returned in JSON format with the following fields:
 
 The MIME type is `application/json`.
 
-# Example
+# Example : Obtaining credentials in `rails console`
 
 1. Start up `./dev/start.sh`
 2. In the `conjur` container, run the server using the AWS credential factory example policy:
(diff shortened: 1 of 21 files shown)


CredentialFactory.md
```diff
@@ -0,0 +1,79 @@
...
+  "session_token": "FQoDYXdzEH4aD...7bz7Nw+NL0mKNr8ztYF",
+  "expiration": "2018-04-15 22:04:26 UTC",
+  "federated_user_id": "<acct-id>:myapp",
+  "federated_user_arn": "arn:aws:sts::<acct-id>:federated-user/myapp"
+}
+```
+
+# TODO
+
+* Should the `expiration` result be set as a header instead?
```
dividedmind: Seems to me that the standard `Expires` header fits the bill perfectly.

Gemfile
```diff
@@ -16,6 +16,8 @@ gem 'sequel-rails'
 gem 'pg'
 gem 'sequel-postgres-schemata', require: false
 
+gem 'aws-sdk-core'
```
dividedmind: Not a fan of adding this to the Gemfile at the base level. Can we try to load the gems dynamically and disable the feature if they're not available? We can keep them in Gemfile in a separate group so that it's easier to test and deploy, while still possible to disable.

app/controllers/application_controller.rb
```diff
@@ -157,7 +157,16 @@ def record_exists e
 
   def forbidden e
     logger.debug "#{e}\n#{e.backtrace.join "\n"}"
-    head :forbidden
+    if e.message != "ApplicationController::Forbidden"
+      render json: {
```
dividedmind: I think this should render text unless JSON is explicitly requested in an `Accept` header.

app/models/credential_factory.rb
```diff
@@ -0,0 +1,49 @@
+module CredentialFactory
+  class << self
+    def values factory_resource
+      provider_annotation = factory_resource.credential_factory_provider
+      raise ArgumentError, %Q(Annotation "credential-factory/provider" not found on #{factory_resource.id.inspect}) unless provider_annotation
+      require "credential_factory/#{provider_annotation.value.underscore}"
```
dividedmind: This will fail with 500 if the plugin is not found. I'm not sure what the correct code would be -- it's not exactly a client error, so suggests 5xx. Maybe 404, with an informative message? An informative message would be nice regardless of the code.

app/models/credential_factory.rb
```diff
@@ -0,0 +1,49 @@
...
+      factory_class = self.const_get(provider_annotation.value.downcase.camelize)
+
+      annotations = factory_resource.annotations.select do |annotation|
```
dividedmind: It's not clear to me how that'll work with the database, or if at all. What do you think about using `factory_resource.annotations.where(name: /^credential-factory/)` to make the database do the filtering? Nb, I think it should be pretty harmless to pass `provider` annotation to the provider as well, and -- who knows -- maybe even useful.

app/models/credential_factory.rb
```diff
@@ -0,0 +1,49 @@
...
+            raise Exceptions::RecordNotFound, id
+          end
+        end
+      end
+      dependent_secrets = dependent_variables.inject({}) do |memo, variable|
+        # Don't reveal that a variable exists if the credential factory doesn't have permission to execute it.
+        unless factory_resource.role.allowed_to?('execute', variable)
+          msg = "#{factory_resource.resource_id.inspect} does not have 'execute' privilege on #{variable.resource_id.inspect}"
+          Rails.logger.info msg
+          raise Exceptions::Forbidden, "#{factory_resource.resource_id.inspect} does not have 'execute' privilege on #{variable.resource_id.inspect}"
```
dividedmind: I wonder -- should these messages be exposed to the client?

app/models/credential_factory.rb
```diff
@@ -0,0 +1,49 @@
...
+        end
+
+        Resource[id].tap do |variable|
+          unless variable
+            Rails.logger.info "Variable #{id.inspect} required by #{factory_resource.resource_id.inspect} does not exist"
+            raise Exceptions::RecordNotFound, id
+          end
+        end
+      end
+      dependent_secrets = dependent_variables.inject({}) do |memo, variable|
```
dividedmind: Protip: instead of
```ruby
xh = xs.inject({}) do |memo, x|
  memo[k] = f(x)
  memo
end
```
you can do
```ruby
xh = xs.map do |x|
  [k, f(x)]
end.to_h
```

app/models/credential_factory/base.rb
```diff
@@ -0,0 +1,36 @@
+module CredentialFactory
+  module Base
+    def self.included cls
+      cls.extend ClassMethods
+    end
+
+    def annotations
+      resource.annotations.inject({}) do |memo, entry|
```
dividedmind: There's gotta be a better way. :/

app/models/credential_factory/base.rb
```diff
@@ -0,0 +1,36 @@
...
+    module ClassMethods
+      # Require annotation +name+ to be in the hash +annotations+, or a configuration error is raised.
+      def require_annotation annotations, name
+        annotations[name] or raise ArgumentError, "Annotation #{name.inspect} is required"
+      end
+
+      # From an example annotation name, build a set of annotation names which are related by 
+      # the prefix on the example. If the example is 'aws/ci/secret_access_key' and the +names+ are
+      # 'foo' and 'bar', this method returns 'aws/ci/foo' and 'aws/ci/bar'. 
+      def build_variable_ids example, names
```
dividedmind: Nitpick: this method produces a lot of garbage. How about `names.map &example[/.*\//].method(:+)`?

run/aws_credential_factory.yml
```diff
@@ -0,0 +1,55 @@
...
+          "Version" : "2012-10-17",   
+          "Statement" : [
+            {
+              "Action" : [           
+                "sts:GetCallerIdentity"
+              ],           
+              "Effect" : "Allow",           
+              "Resource" : "*"       
+            }
+          ]
```
dividedmind: Gotta love YAML. 
However, in the JSON heredoc you missed the closing brace.

jvanderhoof:
Closing this as it's been stale for quite some time.

Issue #544, opened 2018-05-07
Variable Expiration and Rotation


jvanderhoof:
**Feature Overview**
Variables in Conjur can be configured to exist for a period of time before they disappear.  This feature becomes most powerful when it's paired with rotation.

**Workflow**
Let's look at a Conjur variable:
```yml
- !variable
      id: database/username
      annotations:
        expiration: 2018-07-28T23:14:04Z
```

dividedmind:
Couple notes:
- Expiration is an attribute of a particular value, not variable. As such it probably should not be an annotation and definitely does not belong in a policy document.
- The expiration itself could be handled in core simply by storing validity interval on each value and returning the valid (n)one. A check constraint and triggers or rules and could ensure that the intervals are disjoint.

kgilpin:
I agree with Rafał

jonahx:
# Let's Discuss Rotators

A basic demo to get the conversation started is available on the `rotator-test` branch.  It has a random rotator hardcoded into a "watcher" process for automatically expiring and rotating any resources with a `ttl` annotation, and a variable set.

Below I'll explain:

1. How to run the demo
2. How it works
3. How we can build rotators on this base (This is the part to discuss)

## Running the demo

  1. Checkout the `rotator-test` branch.  
  2. Per normal, `cd dev && ./start`.  This will drop you into a dev container.
  3. `conjurctl server`.  This will start the "watcher" process along with the api.
  4. In a 2nd terminal tab, bash into your dev container again.  Then in this new tab:
  5. `conjur policy load -root sample_policy.yaml` Which will create some `variable` resources with `ttl` annotations.
  6. To kick off the rotation, set the first secret value on any of the variables defined in the sample policy.  Eg:
    ```
    conjur variable values add rotate-15s password123
    ```
  Do this for a couple variables.
  7. Back in the original tab, you'll see the variables expiring and being rotated at the intervals specified by their `ttl`.

## How it works

  1. We've added an `expires_at` column to the `secrets` table.
  2. We've added a new query to `model/Secret` called `freshly_expired` which returns any expired secrets.
  3. When the server starts up, we fork off a new "watcher" process:
    ```ruby
    Process.fork do
      exec "rake expiration:watch"
    end
    ```

The current dummy implementation just loops and checks for expired variables:

```ruby
require 'iso_8601_duration'

namespace :expiration do
  desc "Watch for expired variables and rotate"
  task :watch => :environment do

[...]

PR #549, opened 2018-05-14
WIP Rotator design

kgilpin:
Closes [relevant GitHub issues, eg #544 ]

This is currently a WIP PR, primarily for discussion purposes.


db/migrate/20180508164825_add_expiration.rb
```diff
@@ -0,0 +1,7 @@
+Sequel.migration do
+  change do
+    alter_table :secrets do
+      add_column :expires_at, DateTime
```
kgilpin: Can you create a GitHub issue to revisit the use of DateTime here?

app/models/secret.rb
```diff
@@ -20,7 +20,37 @@ def latest_public_keys account, kind, id
...
+
+# Album.group_and_count(:artist_id).having{count.function.* >= 10}
+# # SELECT artist_id, count(*) AS count FROM albums
+# # GROUP BY artist_id HAVING (count(*) >= 10)
+
+    def freshly_expired
+      Sequel::Model.db[<<-EOS
+        SELECT resource_id, expires_at, value AS ttl
+        FROM annotations
+        NATURAL JOIN (
```
kgilpin: I wonder if this would be more efficient as a UNION

lib/tasks/expiration.rake
```diff
@@ -0,0 +1,21 @@
+require 'iso_8601_duration'
+
+namespace :expiration do
+  desc "Watch for expired variables and rotate"
+  task :watch => :environment do
+
+    #TODO: perhaps use clockwork / similar rather than a sleep loop
+    while true
+      Secret.freshly_expired.each do |secret|
+        puts "Secret #{secret[:resource_id]} expired!  Resetting it..."
```
kgilpin: When this expired secret is found and used to create a delayed job (or similar), an update should be made to the secret to prevent another delayed job from being created in the next loop cycle.

jonahx: so actually, on reflection, one counterpoint to using delayedjob idea, is that the current implementation will already handle retries.  if a variable does not get rotated, it will get picked up again as needing rotation on the next loop iteration (after 1s).  as long we handle errors properly so that the loop process itself doesn't die, that might be enough?  are there any additional features we need out of delayedjob.  it does require a DB back so the complexity cost isn't trivial....

lib/tasks/expiration.rake
```diff
@@ -0,0 +1,21 @@
+require 'iso_8601_duration'
+
+namespace :expiration do
+  desc "Watch for expired variables and rotate"
+  task :watch => :environment do
+
```
kgilpin: We should verify that the database is writable before allowing this loop to run.

jonahx: Meaning because we don't want this process running on a follower -- only on a master?   What's the best way to verify this?

lib/tasks/expiration.rake
kgilpin: Maybe `conjurctl` should be responsible for terminating the parent process if any of the sub-processes dies. Because we can't let any of them stop running (the main server process, authn-local or rotation)

kgilpin: Also, see the above comment about ensuring that the database is writeable before this loop is allowed to run. That will ensure that standbys / followers don't try and run the rotator loop, which could cause multiple processes to be rotating the same backend password at the same time (which would be bad).

lib/tasks/expiration.rake
kgilpin: You can use `SELECT pg_is_in_recovery();` which will return TRUE if the db is in standby and FALSE if not.

app/models/secret.rb
dividedmind: I can't imagine how a `UNION` could even be useful here. Anyway, let's leave out the performance for later -- there isn't even an index on `expires_at` yet.

lib/iso_8601_duration.rb
```diff
@@ -0,0 +1,136 @@
+# Manually copied, renamed, and slightly altered version of:
```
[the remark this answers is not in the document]
dividedmind: Alternatively: can we use eg. https://rubygems.org/gems/iso8601 here instead?

lib/tasks/expiration.rake
jonahx: > Maybe conjurctl should be responsible for terminating the parent process if any of the sub-processes dies. Because we can't let any of them stop running (the main server process, authn-local or rotation)

Can you elaboarate on this one?  How were you thinking the parent process would monitor the sub-process?

kgilpin: Well, instead of `Process.waitall` (https://github.com/cyberark/conjur/blob/master/bin/conjur-cli.rb#L81), when any process exits start terminating the other ones as well.

This feels like a separate story.

app/models/secret.rb
```diff
@@ -20,7 +20,37 @@ def latest_public_keys account, kind, id
...
+          SELECT max(expires_at) AS expires_at, resource_id 
+          FROM secrets
+          GROUP BY resource_id
+          HAVING (
+            max(expires_at) IS NULL OR max(expires_at) < NOW()
```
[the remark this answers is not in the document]
kgilpin: `IS NULL` is finding variables that have a `ttl` annotation but no `secrets` values yet.

app/models/secret.rb
kgilpin: Because there are two separate cases: expired secrets, and variables with a `ttl` but no secrets.

lib/tasks/expiration.rake
kgilpin: It looks like @dividedmind is making a similar suggestion with something called "rufus scheduler"

app/models/secret.rb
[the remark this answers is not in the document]
dividedmind: Yes, but this is unlikely to happen in real use -- if there is no secret then there's nothing to rotate.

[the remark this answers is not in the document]
jonahx: I like this.  Would have used it if I had known about it / found it when searching.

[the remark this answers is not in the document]
dividedmind: Actually, no, it's not true -- the NULL clause is looking for manually added values (eg. without an expiration date). I posit that for simplicity we should instead set expiration when manually adding values -- can be `-infinity` -- for ttl-equipped variables, either automatically or manually.

app/models/secret.rb
dividedmind: This query does not handle the second case and I believe it's unlikely to be useful anyway.

app/models/secret.rb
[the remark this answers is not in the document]
dividedmind: (Alternatively do `MAX(COALESCE(expires_at, '-infinity')) < NOW()`.)

app/models/secret.rb
```diff
@@ -20,7 +20,37 @@ def latest_public_keys account, kind, id
...
+# Album.group_and_count(:artist_id).having{count.function.* >= 10}
+# # SELECT artist_id, count(*) AS count FROM albums
+# # GROUP BY artist_id HAVING (count(*) >= 10)
+
+    def freshly_expired
+      Sequel::Model.db[<<-EOS
+        SELECT resource_id, expires_at, value AS ttl
+        FROM annotations
+        NATURAL JOIN (
+          SELECT max(expires_at) AS expires_at, resource_id
```
dividedmind: I don't think we're actually using the expiration time here?

app/models/secret.rb
```diff
@@ -20,7 +20,37 @@ def latest_public_keys account, kind, id
...
     end
+
+
+   # WITH expired_secrets AS (SELECT resource_id FROM secrets GROUP BY resource_id HAVING max(expires_at) < NOW()) SELECT resource_id, value AS ttl FROM annotations NATURAL JOIN expired_secrets WHERE name = 'ttl'
+
+# Album.group_and_count(:artist_id).having{count.function.* >= 10}
+# # SELECT artist_id, count(*) AS count FROM albums
+# # GROUP BY artist_id HAVING (count(*) >= 10)
+
+    def freshly_expired
```
dividedmind: Nitpick: this should be called `pending_expirations`, otherwise it looks like it's a filter on secrets, which it isn't.

dividedmind:
After some discussion, I'm okay with a polling process for now. 

Here's a list of changes I'd require:
- wrap every rotation in `db.transaction { rotate rescue nil }` for reliability and transactionality,
- also rotate resources that don't have secret values attached yet (eg. using something like `SELECT resource_id, value AS ttl FROM annotations a WHERE name = 'ttl' AND COALESCE((SELECT MAX(expires_at) FROM secrets s WHERE a.resource_id = s.resource_id), '-infinity'::timestamptz) < NOW()`).

Ideas for design changes for your consideration:
- Move all logic out of the rake task -- it should be a thin facade calling some method in a loop; this will make testing and development easier.
- Since we're also 'rotating' resources without any secret values on them -- `pending_expiration` probably belongs in a different module. Maybe at this point it makes sense to have a dedicated rotation/expiration module, which would include that method and the `rotate_pending_secrets` (or whatever the name makes sens) from the point above.

app/models/secret.rb
jonahx: naming "nitpicks" are always welcomed.  they're important.

app/domain/rotation/rotator_class.rb
```diff
@@ -0,0 +1,74 @@
...
+    def initialize(cls)
+      Validation.new(cls).validate!
+      @cls = cls
+    end
+
+    def needs_additional_variables?
+      @cls.respond_to?(:required_variables)
+    end
+
+    # For example: "aws/some-variable"
```
jvanderhoof: Would the SomeVariable class would look like this:
```
module Rotation
  module Rotators
    module Aws
      class SomeVariableRotator
        def initialize
          ...
```

app/domain/rotation/master_rotator.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    def resource_id(variable_name)
+      "#{@account}:variable:#{variable_name}"
+    end
+
+  end
+end
+__END__
+Recall each rotator has a `required_variables` method by which it can specify the hash of `{variable_id => value }` that will be passed to its `rotate` method.  the variable ids are resource ids of the form `<acct>:<kind>:<id>`.  but where does `<acct>` come from?  clearly we don't want to hard code it into the rotator.  but since the rotator is just run when the server boots up, there isn't (i don't think?) a logged in acct to asssociate with.  so i'm confused how this work.
```
jvanderhoof: Can we simplify this by assuming a variable uses its own account?

app/domain/rotation/master_rotator.rb
```diff
@@ -0,0 +1,93 @@
...
+        raise RotatorNotFound unless rotator
+
+        new_values = rotator.new_values(new_values_input(rotator))
+        update_all_secrets(new_values, secret['ttl'])
+      end
+    end
+
+    private
+
+    def new_values_input(rotator)
```
jvanderhoof: Would this method return a hash of current values that will cumulatively need to be rotated?  Sticking with the AWS example, would it be something like:
```ruby
{
  'cucumber/aws/region': 'us-east-1',
  'cucumber/aws/access_key_id': 'aksdflsdjflkadjflkdjfla1',
  'cucumber/aws/secret_access_key': 'dBQn6phsnbjfZFRC377KShCS29qClGWcEjj6icrojX8=',
}
```

app/domain/rotation/master_rotator.rb
```diff
@@ -0,0 +1,93 @@
...
+        sleep(seconds)
+      end
+    end
+
+    def rotate_all
+      rotations = rotation_model.required_rotations
+      Parallel.each(rotations, in_threads: 10) do |secret|
+
+        rotator = @rotators[secret['rotator']]
+        raise RotatorNotFound unless rotator
```
jvanderhoof: Down the line it might make sense to validate rotators when a policy is imported instead of having a user find a rotator mis-spelling when they try to rotate.

app/domain/rotation/rotators/example/example_rotator.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Rotators
+    module Example
+
+      class ExampleRotator
+
+        def initialize
+          @cnt = 0
+        end
+
+        def required_variables(annotated_variable)
```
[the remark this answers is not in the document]
jvanderhoof: Actually, looks more like it would return:
```yml
%w(region access_key_id secret_access_key secret_key_proxy)
```

app/domain/rotation/master_rotator.rb
jvanderhoof: Is this still going to a problem? An alternative to passing an `account` argument could be to inspect the expired value's variable to get the account.

lib/tasks/expiration.rake
```diff
@@ -0,0 +1,13 @@
+require 'iso_8601_duration'
+require 'master_rotator'
+require 'installed_rotators'
+
+namespace :expiration do
+  desc "Watch for expired variables and rotate"
+  task :watch, [:account] => [:environment] do |t, args|
+
+    #TODO how to do rake args
+    MasterRotator.new(rotators: InstalledRotators.new, account: args.account)
```
jvanderhoof: Where we'll need to rotate variables across all accounts, would it make more sense to get a variable's account by introspecting the value's variable?

app/domain/rotation/master_rotator.rb
```diff
@@ -0,0 +1,93 @@
...
+    def variable_name(resource_id)
+      resource_id.split(':', 3).last
+    end
+
+    def resource_ids(variable_names)
+      variable_names.map { |name| resource_id(name) }
+    end
+
+    def resource_id(variable_name)
+      "#{@account}:variable:#{variable_name}"
```
dividedmind: Jason is correct that rotators should work across accounts. Also I don't think we should hardcode 'variable' kind here -- there's no reason it shouldn't work regardless of the resource kind. In general, I suggest we treat the whole resource ID as an opaque whole instead of trying to dig into it.

app/domain/rotation/master_rotator.rb
```diff
@@ -0,0 +1,93 @@
...
+    def rotate_every(seconds)
+      loop do
+        rotate_all
+        sleep(seconds)
+      end
+    end
+
+    def rotate_all
+      rotations = rotation_model.required_rotations
+      Parallel.each(rotations, in_threads: 10) do |secret|
```
dividedmind: What will happen if one of the threads fails?

bin/conjur-cli.rb
```diff
@@ -78,6 +78,13 @@ def test_select
       exec "rake authn_local:run"
     end
 
+    # Start the rotation "watcher" in a separate thread
+    rotations_thread = Thread.new do
+      exec "rake expiration:watch[#{account}]"
```
dividedmind: That looks weird and it's not clear (to me) how it behaves re database connections for instance.
Would it work to just call `MasterRotator.new(rotators: InstalledRotators.new).rotate_every(1.second)` here instead?

app/domain/rotation/rotator_class.rb
jonahx: @jvanderhoof yes

app/domain/rotation/rotators/example/example_rotator.rb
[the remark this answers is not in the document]
jonahx:
```
%w(region access_key_id secret_access_key secret_key_proxy)
```
is correct.  

It's a way for the rotator to say what information it needs to perform it's calculation.  Similarly, the `new_values` method allows the rotator to specify that multiple variables need to be updated.  This will all be spelled out clearly in the docs.

app/domain/rotation/master_rotator.rb
jonahx: Close, but it's only the variable name, not the full resource_id, so it wouldn't include "cucumber."  Since rotators are not account specific, the idea is to clearly separate the concept of "variable names" (what rotators use) and the fully qualified resource_ids (which include account info).  These distinct concepts should probably be reified as objects.

app/domain/rotation/master_rotator.rb
jonahx: 👍

app/domain/rotation/master_rotator.rb
jonahx: I think I have a solution now that solves this.  Will be implementing today.

jonahx:
@dividedmind 

>  I'd probably be less rigid about the rotator interface. Matter of taste I suppose :)

app/domain/rotation/master_rotator.rb
jonahx: >  In general, I suggest we treat the whole resource ID as an opaque whole instead of trying to dig into it.

app/domain/rotation/master_rotator.rb
jonahx: I'll check, but the more operative question is: What _should_ happen?  It's actually unclear to me.  We could either have a retry strategy or just fail fast.  Retry discussion aside, I _think_ that once a rotation truly fails, that's an irrecoverable exception and all of Conjur stops.  Lmk if this is not correct.  @kgilpin you thoughts on this too are appreciated.

bin/conjur-cli.rb
jonahx: @dividedmind So changing from process to threads I _thought_ was your idea.  It may have been Kevin's.  I believe the new threads can share db connections and that was your reason for preferring them, but that might be incorrect.  

You are right to notice that using rake is sort of superfluous here... but I had to use to get the autoloading to work.  rake integrates with rails, and if I try to call `MasterRotator` directly in `conjur-cli.rb` I get a bunch of class not found errors, and that necessitates putting explicit require statements everywhere, which ironically is my preferred way of doing things, but it's very un-railsy and would be departing from the conventions used everywhere else in our codebase.  So rake is being used here for the sole purpose of hooking into the rails autoload system.  If you have a better solution I'm all ears.

app/domain/rotation/master_rotator.rb
kgilpin: Use a thread pool, which will automatically re create the lost threads?

app/domain/rotation/master_rotator.rb
```diff
@@ -0,0 +1,144 @@
...
+      #
+      def new_values
+        resource_id_keys(rotator.new_values(name_keys(current_values)))
+      end
+
+      private
+
+      def current_values
+        @secret_model.latest_resource_values(required_resources)
+        p 'current_values', current_values
```
kgilpin: Stray log message. There's at least one more elsewhere too, I don't want to litter the PR pointing them out though :-) That's been indicated as a bad PR practice.

bin/conjur-cli.rb
```diff
@@ -78,6 +78,14 @@ def test_select
       exec "rake authn_local:run"
     end
 
+    # Start the rotation "watcher" in a separate thread
+    rotations_thread = Thread.new do
+      # exec "rake expiration:watch[#{account}]"
+      exec "rake expiration:watch"
```
kgilpin: This is going to replace the current process, which is not what we want.

I suggestion going with the `fork / exec` model used for the other sub-tasks.

jonahx:
I just added a major update that's ready for a new review.

## How to Run it

1. `cd dev && ./start` will drop you into a container.  `conjurctl server` once inside.
2. In 2nd tab: `docker-compose run --no-deps --rm --entrypoint bash client`
3. Once inside there, `conjur policy load --replace root sample_policy.yaml`
4. `conjur variable values add db-reports/url http://localhost/db`
5. `conjur variable values add db-reports/username db-user`
6. `conjur variable values add db-reports/password secret`
7. In the original tab, ctrl-c stop conjur server and `conjurctl server` to restart it.

You should see debug output showing the password rotating in conjur every 10s.

## How to Review It

1. "bin/conjur-cli.rb:84" kicks it off
2. calls "expiration.rake"
3. calls MasterRotator, which calls the postgres rotator

There are a number of minor TODO comments... all of them will be done soon.

app/domain/rotation/rotator_class.rb
```diff
@@ -0,0 +1,77 @@
+# Represents the contract of a class that implements an rotator.
+#
+module Rotation
+  class RotatorClass
+
+    # Represents the rules any rotator class must conform to
+    #
+    class Validation
+
+      DoesntEndWithRotator = ::Util::ErrorClass.new(
```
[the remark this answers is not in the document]
jonahx: This is already gone.  I'll be pushing the change later :)

app/domain/rotation/rotators/postgres/postgres_rotator.rb
```diff
@@ -0,0 +1,125 @@
+require 'pg'
+
+module Rotation
+  module Rotators
+    module Postgres
+
+      class PasswordRotator
```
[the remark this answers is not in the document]
jonahx: This too :)

jonahx:
https://jenkins.conjur.net/job/cyberark--conjur/job/rotator-test/25/

Removed all the dead code and TODOs except the one mentioned in #v5-dev.  Going to merge this as a single squash it's all greenfield code essentially unconnected to master and makes sense as a unit.

Issue #542, opened 2018-05-07
AWS IAM Authenticator


kgilpin:
**Feature Overview**
The Conjur IAM Authenticator allows an AWS resource to use its AWS IAM role to authenticate with Conjur. This approach enables EC2 instances and Lambda functions to access credentials stored in Conjur without a pre-configured Conjur identity.

**Setup**
To enable an IAM Authenticator called `prod`, we'll set the following environment variable when we start Conjur:
```
CONJUR_AUTHENTICATORS=authn-iam/prod
```

In this context, `prod` is the called the `service ID`. Multiple authenticators can be configured with different service IDs. Each one implements a separate "zone" of authentication.

After Conjur is started, we'll create a policy to enable our `prod` IAM Authenticator:

```yml
# policy id needs to match the convention `conjur/authn-iam/<service ID>`
- !policy
  id: conjur/authn-iam/prod
  body:
  - !webservice

  - !group clients
 
  - !permit
    role: !group clients
    privilege: [ read, authenticate ]
    resource: !webservice
```

Next, let's create an application policy to provide a database username and password to an AWS services.  The IAM role in this example is `011915987442:MyApp`:

```yml
- !policy 
  id: myapp
  body:
  - &variables
    - !variable database/username
    - !variable database/password

  # Create a group that will have permission to retrieve variables
  - !group secrets-users

  # Give the `secrets-users` group permission to retrieve variables
  - !permit
    role: !group secrets-users
    privilege: [ read, execute ]
    resource: *variables
  
  # Create a layer to hold this application's hosts
  - !layer

  # The host ID needs to match the AWS ARN of the role we wish to authenticate.
  - !host 011915987442/MyApp

  # Add our host into our layer
  - !grant
    role: !layer
    member: 011915987442/MyApp

  # Give the host in our layer permission to retrieve variables
  - !grant
    member: !layer
    role: !group secrets-users
```

**Important:** Note above host has an ID composed of a prefix/namespace followed by the AWS account ID followed by the name of the AWS IAM role. The AWS Account ID and name of the role is extracted from the getCallerIdentity by the authenticator

Finally, let's give our `myapp` host permission to authenticate using the IAM Authenticator:

```yml
- !grant
  role: !group conjur/authn-iam/prod/clients
  member: !host myapp/011915987442/MyApp
```

**Workflow**

Now that the IAM Authenticator has been configure and we've permitted an IAM role to authenticate, let's look at the authentication flow of an EC2 instance or Lambda function.

[...]

PR #553, opened 2018-05-21
AWS IAM Authenticator - WIP

cyberark-bizdev:

See https://github.com/cyberark/conjur/issues/542

Not yet

https://github.com/cyberark/conjur/issues/542


app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,84 @@
+require 'active_support'
+require 'active_support/core_ext'
+require 'net/ldap'
```
jvanderhoof: Why do we need `net/ldap`?

dev/start
```diff
@@ -52,6 +55,10 @@ if [[ $ENABLE_AUTHN_LDAP = true ]]; then
   docker-compose exec conjur conjurctl policy load cucumber /src/conjur-server/dev/files/authn-ldap/policy.yml
 fi
 
+if [[ $ENABLE_AUTHN_LDAP = true ]]; then
```
jvanderhoof: Should this be `ENABLE_AUTHN_IAM`?

dev/start
```diff
@@ -52,6 +55,10 @@ if [[ $ENABLE_AUTHN_LDAP = true ]]; then
...
+  env_args="$env_args -e CONJUR_AUTHENTICATORS=authn-iam/prod"
```
jvanderhoof: For this to work, we'll need to add a IAM Authenticator policy.  For the Authn-LDAP authenticator, this policy file is located in `dev/files/authn-ldap`.  How about adding an `authn-iam` folder with a corresponding policy file to the `dev/files` directory?

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,84 @@
+require 'active_support'
+require 'active_support/core_ext'
```
jvanderhoof: Do we need these two requires?

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+require 'json'
+require 'aws-sdk-iam'
+require 'aws-sdk-core'
+
+
+module Authentication
+  module AuthnIam
+
+    class Authenticator
+
```
jvanderhoof: Looks like there might be an extra line here.

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+      
+      def initialize(env:)
+        @env = env
+      end
+
+      def valid?(input)
+        @authn_name, @account, @login, password, @service_id = input.authenticator_name, input.account, input.username, input.password, input.service_id
+
+        # JSON holding the AWS signed headers 
+        @signed_aws_headers = JSON.parse input.password
```
jvanderhoof: Instead of using class variables for passing variables into a method, how about we pass it into the `is_trusted_by_aws?` method?  Doing this will make it easier to test the `is_trusted_by_aws?` in isolation. Maybe something like:
```ruby
signed_aws_headers = JSON.parse(input.password)
is_trusted_by_aws?(signed_aws_headers) && ...
```

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+      def is_trusted_by_aws?
+
+        url = 'https://sts.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15'
+
+        # Executes the AWS Signed Request
+        uri = URI.parse(url)
+        https = Net::HTTP.new(uri.host,uri.port)
+        https.use_ssl = true
+        https.set_debug_output($stdout)
+        request = Net::HTTP::Get.new(url)
```
jvanderhoof: There's a lot of boilerplate network code here.  What do you think about pulling this out into a method.  A separate method would allow you mock the request option to validate the correct fields were added to the request object.  For example, we had something like:
```ruby
def base_aws_request
  url = 'https://sts.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15'

  # Executes the AWS Signed Request
  uri = URI.parse(url)
  https = Net::HTTP.new(uri.host,uri.port)
  https.use_ssl = true
  https.set_debug_output($stdout)
  Net::HTTP::Get.new(url)
end

def retrieve_iam_identity(aws_headers)
  aws_request = base_aws_request
  aws_headers.each do |key, value|
    aws_request.add_field(key, value)
  end

  https.request(aws_request)
end
```
This would allow us to stub the `base_aws_request` to verify the expected header attributes were added.  It would also let you mock the AWS call to test this code made the call to AWS.

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+          request.add_field(key, value)
+        end
+
+        res = https.request(request)
+
+        Rails.logger.info("request => #{request}")
+        Rails.logger.info("****> #{res.code} #{res.message}")
+        Rails.logger.info("**** Body -> #{res.body} ")
+
+        @aws_response_hash = Hash.from_xml(res.body) if res.code.eql? "200"
```
jvanderhoof: Instead of setting a class variable, can we return the hash?  It'll allow us to unit test `iam_role_matches? `.  Something like:
```ruby
  ...
  if res.code.eql?("200")
    Hash.from_xml(res.body)
  else
    Rails.logger.error("****> #{res.code} #{res.message}")
    false
  end
end

File: app/domain/authentication/authn_iam/authenticator.rb [hunk 9]
```diff
@@ -0,0 +1,84 @@
...
+
+        res.code.eql? "200"
+
+      end
+    
+      def iam_role_matches? resource
```
jvanderhoof: If this method takes both a resource and response hash, it'll be easy to write a unit test to verify it works as expected.

jvanderhoof:
The big thing this feature is missing is tests.  We usually write unit tests in RSpec and integration tests in Cucumber.  The Authn-LDAP authenticator has great examples.

The challenge in this project will be the integration testing.  What do you think about writing a simple Lambda function to authenticate?  We could use Cloudformation (or Terraform) to setup the IAM role and Lambda function. If we wrap this all up with Bash, we can standup the environment prior to running the authenticator integration tests.

cyberark-bizdev:
I will work on this during the week/weekend... Bear with me while attend Okta conference as well. I started working on the feedback from Jason and I will commit on that by this Friday, and start working on the unit tests by that day.

jvanderhoof:
@cyberark-bizdev: This looks like it's shaping up.  Can you please add some tests? Even a few RSpec tests would give me some confidence that this code continues to works as advertised as we work to expand V5.

jvanderhoof:
@cyberark-bizdev: We should be returning a response error 403 with a more helpful error message, maybe something like: `This role does not have permission to access this credential`.

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,95 @@
...
+      def valid?(input)
+        @account, @login, password = input.account, input.username, input.password
+
+        # JSON holding the AWS signed headers 
+        signed_aws_headers = JSON.parse password
+
+        response_hash = identity_response_hash(signed_aws_headers)
+        trusted = response_hash != false
+
+        (trusted) && (iam_role_matches? host_role, response_hash)
```
kgilpin: Consider: `trusted && iam_role_matches?(host_role, response_hash)`

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+
+      def identity_response_hash(signed_aws_headers)
+
+        res = retrieve_iam_identity(signed_aws_headers)
+
+        Rails.logger.info("****> #{res.code} #{res.message}")
+        Rails.logger.info("**** Body -> #{res.body} ")
+
+        if res.code.eql?("200")
```
kgilpin: It's not obvious that `code` will be a string, so consider comparing `code.to_i == 200`, or perhaps `code.to_i < 300` (since HTTP statuses such as 204 also indicate success).

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,95 @@
...
+        split_assumed_role = response_hash["GetCallerIdentityResponse"]["GetCallerIdentityResult"]["Arn"].split(":")
+
+        # removes the last 2 parts of login to be substituted by the info from getCallerIdentity
+        host_prefix = (@login.split("/")[0..-3]).join("/")
+        aws_account_id = split_assumed_role[4]
+        aws_role_name = split_assumed_role[5].split("/")[1]
+
+        host_to_match = "#{host_prefix}/#{aws_account_id}/#{aws_role_name}"
+
+        Rails.logger.info("host to match = #{host_to_match}")
```
kgilpin: This a `debug` level message. Debug info is also elsewhere in this PR as `info`, so I'd like to see a cleanup of the logging.

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,95 @@
...
+
+        (trusted) && (iam_role_matches? host_role, response_hash)
+
+      end
+
+      def identity_response_hash(signed_aws_headers)
+
+        res = retrieve_iam_identity(signed_aws_headers)
+
+        Rails.logger.info("****> #{res.code} #{res.message}")
```
kgilpin: It doesn't seem that it will be too clear in the log what this is referring to. So, if you want to provide diagnostic information to the server administrator, consider a single log message that contains all the info that someone might need. For example:

```
Rails.logger.info("Verification of IAM identity failed with HTTP code #{res.code.inspect}: #{res.message}")
```

app/domain/authentication/authn_iam/authenticator.rb
cyberark-bizdev: Removed!

app/domain/authentication/authn_iam/authenticator.rb
cyberark-bizdev: Removed!

app/domain/authentication/authn_iam/authenticator.rb
cyberark-bizdev: Removed the extra line!

app/domain/authentication/authn_iam/authenticator.rb
cyberark-bizdev: refactored and cleaned up

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      def is_trusted_by_aws?
+
+        url = 'https://sts.amazonaws.com/?Action=GetCallerIdentity&Version=2011-06-15'
+
+        # Executes the AWS Signed Request
+        uri = URI.parse(url)
+        https = Net::HTTP.new(uri.host,uri.port)
+        https.use_ssl = true
+        https.set_debug_output($stdout)
```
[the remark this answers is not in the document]
cyberark-bizdev: Incorporated, and refactored.

app/domain/authentication/authn_iam/authenticator.rb
cyberark-bizdev: Incorporated!

app/domain/authentication/authn_iam/authenticator.rb
cyberark-bizdev: Fixed.

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+
+        @signed_aws_headers.each do |key, value|
+          request.add_field(key, value)
+        end
+
+        res = https.request(request)
+
+        Rails.logger.info("request => #{request}")
+        Rails.logger.info("****> #{res.code} #{res.message}")
+        Rails.logger.info("**** Body -> #{res.body} ")
```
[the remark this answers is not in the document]
cyberark-bizdev: Fixed!

app/domain/authentication/authn_iam/authenticator.rb
cyberark-bizdev: Incorporated!

app/domain/authentication/authn_iam/authenticator.rb
cyberark-bizdev: Incorporated, and refactored

app/domain/authentication/authn_iam/authenticator.rb
cyberark-bizdev: Cleaned up and messages still present were sent to debug

dev/start
cyberark-bizdev: Fixed.

dev/start
cyberark-bizdev: Added the root policy file with authn-iam in place

jvanderhoof:
I pulled out key commits and merged this functionality into PR #586.

PR #554, opened 2018-05-21
Audit querying

dividedmind:
This is a simple webservice to query audit data. Please note it requires a separate process to fill in the audit database and as such is completely separate from the rest of this app. Let's discuss whether and how and where we want to split this out.

### DIFF
--- a/config/environments/appliance.rb
+++ b/config/environments/appliance.rb
@@ -4,4 +4,5 @@ Rails.application.configure do
   config.log_level = :info
   config.middleware.use Audit::RememberUuid
   config.audit_socket = '/run/conjur/audit.socket'
+  config.audit_database ||= 'postgres://:5433/audit'
 end
--- a/config/environments/appliance.rb
+++ b/config/environments/appliance.rb
@@ -1,67 +1,7 @@
+load File.expand_path '../production.rb', __FILE__
+
 Rails.application.configure do
-  # Settings specified here will take precedence over those in config/application.rb.
-
-  # Code is not reloaded between requests.
-  config.cache_classes = true
-
-  # Eager load code on boot. This eager loads most of Rails and
-  # your application in memory, allowing both threaded web servers
-  # and those relying on copy on write to perform better.
-  # Rake tasks automatically ignore this option for performance.
-  config.eager_load = true
-
-  # Full error reports are disabled and caching is turned on.
-  config.consider_all_requests_local       = false
-  config.action_controller.perform_caching = true
-
-  # Enable Rack::Cache to put a simple HTTP cache in front of your application
-  # Add `rack-cache` to your Gemfile before enabling this.
-  # For large-scale production use, consider using a caching reverse proxy like
-  # NGINX, varnish or squid.
-  # config.action_dispatch.rack_cache = true
-
-  # Disable serving static files from the `/public` folder by default since
-  # Apache or NGINX already handles this.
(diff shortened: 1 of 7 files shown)


config/environments/appliance.rb
```diff
@@ -1,67 +1,8 @@
...
-  # Set this to true and configure the email server for immediate delivery to raise delivery errors.
-  # config.action_mailer.raise_delivery_errors = false
-
-  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
-  # the I18n.default_locale when a translation cannot be found).
-  config.i18n.fallbacks = true
-
-  # Send deprecation notices to registered listeners.
-  config.active_support.deprecation = :notify
+load File.expand_path '../production.rb', __FILE__
```
apotterri: 👍

lib/conjur/audit/webservice.rb
```diff
@@ -0,0 +1,71 @@
...
+        return to_path_hash [*keys, value] if value
+        return keys.first if keys.length == 1
+        { keys.first => to_path_hash(keys.drop 1) }
+      end
+
+      # filters dataset based on given query
+      # keys not in _fields_ are assumed to be in structured data
+      # eg. subject@43868/role=foo looks for [subject@43868 role="foo"]
+      def filter dataset, query
+        fields = %w(facility severity hostname appname procid msgid)
```
apotterri: It seems like we're going to want a higher level of abstraction here. Clients of this webservice are going to want to query in terms of Conjur objects, e.g. show me the events related to this resource. Using the database fields for filtering feels like it's leaking implementation detail.

dividedmind: I both agree and not. Let me explain :)

It's not using 'database' fields (though indeed it translates the query directly into them, that fact alone being a -- not exposed -- implementation detail), but standard syslog fields defined by RFC, plus any structured data fields the entries may contain. As for the latter, RFC stipulates an authority responsible for specifying them (as registered with IANA), and so for Conjur-specific fields, we will need to publish a spec once they're settled (not only for using this service but also for external consumption eg. via syslog relay). In my view this makes it all a public interface, not an implementation detail.

That said, of course we'll want to add querying in terms of Conjur objects, but I feel we should add it on as a layer of smart sugar on top of this basic and generic functionality, which is meant to generically query structured syslog data, whether it's originally produced by Conjur or not (regardless of whether we will ever use it to query other events).

apotterri: OK, that description makes sense to me.

Where would you see the sugar layer being added, though? Anywhere outside this service seems like it would imply some amount of duplicated code (e.g. in each different client library).

dividedmind: I'm not 100% sure and I'm open to suggestions. One idea I had is for Conjur-specific entries would be handled by `/audit/resource/*id` and `/audit/role/*id` or even `/audit/*id` and redirect or forward to this service with the specific query.

Although now that I think of it these couldn't be in fact implemented as simple wrappers for ONE query with this service, because it only supports conjunction of conditions, and a given role or resource can appear in many places in an audit entry. OTOH I wonder how important or useful it is to provide such general catch-all queries; yes, we had them in v4, and the fact that we ONLY had them is why the UI had to do so much filtering and transformation of the data, and that would probably be better-served by using more specific queries. Thoughts?

apotterri: Maybe you'll learn more by prototyping use of this service in the UI. We can go with this as-is for now, and update based on what you find.

PR #555, opened 2018-05-25
Allow listing of resources for a role other than the currently authenticated user

apotterri:
Closes https://ca-il-jira.il.cyber-ark.com:8443/browse/CONJ-5065

Adds support for a `role` parameter to the `/resources` route. Allows the call to list resources for a different role, provided that role is held by the caller.

The UI uses this when rendering the "Privileges Held" sections of the details pages for various resources.

The new cucumber scenario is a good place to start: https://github.com/cyberark/conjur/pull/555/files#diff-fb6cd84532d315ce061240eac6e7791c

No need for manual testing.

https://jenkins.conjur.net/job/cyberark--conjur/job/resources-for-role_20180525/

Yes.

No

API doc needs updating.

No

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -1,8 +1,9 @@
 class ResourcesController < RestController
   include FindResource
-
+  include AssumedRole
+  
   before_filter :find_resource, only: [ :show, :permitted_roles, :check_permission ]
-
+  
   def index
     options = params.slice(:account, :kind, :limit, :offset, :search).symbolize_keys
     
@@ -11,7 +12,7 @@ class ResourcesController < RestController
       options[:owner] = Role[ownerid] or raise Exceptions::RecordNotFound, ownerid
     end
     
-    scope = Resource.visible_to(query_role).search options
+    scope = Resource.visible_to(assumed_role).search options
 
     result =
       if params[:count] == 'true'
@@ -45,26 +46,10 @@ class ResourcesController < RestController
     privilege = params[:privilege]
     raise ArgumentError, "privilege" unless privilege
 
-    if query_role.allowed_to?(privilege, @resource)
+    if assumed_role.allowed_to?(privilege, @resource)
       head :no_content
     else
       head :not_found
     end
   end
-
-  # Return the role to use when finding resources: either the role
-  # specified by the request, or the currently authenticated user.
-  def query_role
-    @query_role ||= if role_param
-                      Role[role_param] or raise RecordNotFound
-                    else
-                      current_user
-                    end
-  end
-
-  def role_param
-    # support :acting_as for backwards compatibility
-    params[:role].presence || params[:acting_as].presence
-  end
-  
 end
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -1,8 +1,8 @@
 class ResourcesController < RestController
   include FindResource
-  
+
   before_filter :find_resource, only: [ :show, :permitted_roles, :check_permission ]
-    
+
   def index
     options = params.slice(:account, :kind, :limit, :offset, :search).symbolize_keys
     
@@ -11,7 +11,7 @@ class ResourcesController < RestController
       options[:owner] = Role[ownerid] or raise Exceptions::RecordNotFound, ownerid
     end
     
-    scope = Resource.visible_to(current_user).search options
+    scope = Resource.visible_to(query_role).search options
 
     result =
       if params[:count] == 'true'
@@ -38,21 +38,33 @@ class ResourcesController < RestController
     render json: Role.that_can(privilege, @resource).map {|r| r.id}
   end
 
-  # Implements the use case "check MY permission on some resource", where "me" is defined as the +current_user+.
+  # Implements the use case "check permission on some resource",
+  # either for the role specified by the query string, or for the
+  # +current_user+.
   def check_permission
     privilege = params[:privilege]
     raise ArgumentError, "privilege" unless privilege
 
-    role = if role_id = params[:role]
-      Role[role_id] or raise Exceptions::RecordNotFound, role_id
-    else
-      current_user
-    end
-
-    if role.allowed_to?(privilege, @resource)
+    if query_role.allowed_to?(privilege, @resource)
       head :no_content
     else
       head :not_found
     end
   end
+
+  # Return the role to use when finding resources: either the role
+  # specified by the request, or the currently authenticated user.
+  def query_role
+    @query_role ||= if role_param
+                      Role[role_param] or raise RecordNotFound
+                    else
+                      current_user
+                    end
+  end
+
+  def role_param
+    # support :acting_as for backwards compatibility
+    params[:role].presence || params[:acting_as].presence
+  end
+  
 end
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -108,4 +108,8 @@ class Role < Sequel::Model
   def all_roles
     Role.from(Sequel.function(:all_roles, id))
   end
+
+  def ancestor_of? role
+    Role.from(Sequel.function(:is_role_ancestor_of, id, role.id)).first[:is_role_ancestor_of]
+  end
 end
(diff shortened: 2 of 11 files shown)


dividedmind:
I'm not sure this is a good approach. This is likely not the only API where it's useful to assume another role. In v4 we had `acting_as` parameter, which was universal (even though it's not documented as such, which led to some confusion, cf. https://github.com/cyberark/conjur-api-dotnet/pull/11). Using a query parameter for that is a bit cumbersome, but what do you think of using a header instead, eg. `Conjur-Assume-Role: foo:user:bob`?

apotterri:
In the two cases I changed in ResourcesController, the user isn't really asking to assume another role. He's making a query about another role:  listing the resources the role can see, or a checking a permission. A query parameter seems appropriate to me.

I don't know whether we'll want the capability to assume another role somewhere. If we find we do, we can decide then about how to implement it.

dividedmind:
Ok, actually I do have an issue with this beyond the design one. What is to prevent one from giving `role=acct:user:admin` thus bypassing visibility restrictions? Not sure if the right solution is to implement some sort of visibility intersection or to simply require the target role to be held.

apotterri:
@dividedmind this is ready for another look

app/controllers/concerns/current_role.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def current_role_from_param(param_name)
+    find_current_role(param_name)
+  end
+  
+  private
+
+  def find_current_role(param_name)
+    acting_as = request.headers['Conjur-Assume-Role'].presence || params[param_name].presence
```
dividedmind: TIL. ActiveSupport is an endless source of surprises.

app/controllers/resources_controller.rb
```diff
@@ -38,33 +39,15 @@ def permitted_roles
     render json: Role.that_can(privilege, @resource).map {|r| r.id}
   end
 
-  # Implements the use case "check permission on some resource",
-  # either for the role specified by the query string, or for the
-  # +current_user+.
+  # Implements the use case "check MY permission on some resource", where "me" is defined as the +current_user+.
```
dividedmind: Is this accurate?

app/controllers/resources_controller.rb
```diff
@@ -38,33 +39,15 @@ def permitted_roles
...
   def check_permission
     privilege = params[:privilege]
     raise ArgumentError, "privilege" unless privilege
 
-    if query_role.allowed_to?(privilege, @resource)
+    if current_role_from_param(:role).allowed_to?(privilege, @resource)
```
dividedmind: As a user of this API I'd find it surprising that it doesn't respect `acting_as` like most other APIs.

apotterri:
Yeah, I like `assumed_role` better than `current_role`. I'll make that change.

I used `is_role_ancestor_of`, because that's what's called in `authz`.  Maybe `has_role?` would be better (with a comment indicating that transitive grants are taken into account)?

I don't think we should change `current_user` to `assumed_role`. The distinction exists in v4, and seems like a useful one to me. It will allow us to keep track of who the authenticated user is, and whether he has assumed another role.

app/models/role.rb
```diff
@@ -108,4 +108,8 @@ def allowed_to? privilege, resource
   def all_roles
     Role.from(Sequel.function(:all_roles, id))
   end
+
+  def is_ancestor_of? role
```
dividedmind: BTW, `is` seems redundant here with the `?`.

dividedmind:
See, `has_role` is the reverse; it's what got me confused :) I know this is what it was called in authz and I even remember I called it that; seemed to make sense at the time ;) A couple ideas:  `has_member_role_transitively`; `includes_role(_transitively)`. Or perhaps indeed reverse the arguments and call it `has_role(_transitively)` or `is_role_(transitively_)member_of`. (I'm actually okay with leaving this as is even, just thought maybe there's a better way to put it, so I'll leave this up to you.)

Re `current_user` I'm not proposing to remove it. Just make sure `assumed_role` is used in places where `current_user` currently is, such as https://github.com/cyberark/conjur/blob/fc61f506c1e998c4169db29ef1d344a3f5aa3ed2/app/controllers/concerns/authorize_resource.rb#L9 or https://github.com/cyberark/conjur/blob/master/app/controllers/accounts_controller.rb#L15 (and most other places as indicated by `git grep current_user`) so that the assumption of another role is consistent.

apotterri:
Alright, I'll probably just leave is_role_ancestor_of alone for now.

There aren't any places besides ResourcesController that allow a user to assume another role -- it's the only controller that includes `CurrentRole` (soon to be `AssumeRole`.

I'd really rather leave the other references to `current_user` alone, to avoid expanding the scope of this work further.

app/controllers/resources_controller.rb
apotterri: I still think this is more like a query than like the assumption of a role. Also, changing this to expect `acting_as` will require an API change, which feels beyond the scope of this work.

app/controllers/concerns/current_role.rb
```diff
@@ -0,0 +1,43 @@
...
+    return current_user unless acting_as
+
+    role = Role[acting_as]
+    raise ApplicationController::Forbidden unless role && role.is_ancestor_of?(current_user)
```
kgilpin: Raising `Forbidden` if the requested role doesn't exist seems a little weird. Maybe both cases should be 404, indicating that the role you requested doesn't exist, or maybe it does exist but you don't have access to it.

db/migrate/20180530162704_is_role_ancestor_of.rb
```diff
@@ -0,0 +1,24 @@
...
+         LANGUAGE sql
+         STABLE STRICT
+      AS $$
+        SELECT COUNT(*) > 0 FROM (
+          WITH RECURSIVE m(id) AS (
+            SELECT $2
+            UNION ALL
+            SELECT role_id FROM role_memberships rm, m WHERE member_id = id
+          )
+          SELECT true FROM m WHERE id = $1 LIMIT 1
```
kgilpin: @dividedmind this is the efficient one, right?

kgilpin:
I agree with Alan that making this feature universally applicable is too much of a scope creep for now. 

I personally feel that using a header is awkward for clients. Therefore I would have a query parameter `assume_role` or `assumed_role` trigger this behavior and not support the header option. 

So in summary, if it were totally up to me (which it's not!), I would get rid of the header option (it's just more doc to write...), rename the query parameter to `assumed_role`, test this one scenario (which seems to be already done), write an Issue to make assumed roles universal, and merge.

kgilpin:
Regarding the universality, if we don't document this feature as working universally, then we don't have to worry about whether it does or not.

dividedmind:
Ok, thanks Kevin for breaking the tie :)

If it isn't a universal feature then I agree a header makes less sense. I just assumed it would be universal when proposing to use a header; I apologize for the misunderstanding. I still see it as a bit of a missed opportunity to not implement it as a generic feature now, but I'm inclined to swallow it for the sake of moving forward ;)

kgilpin:
Someday there will be a Conjur 6

app/controllers/concerns/current_role.rb
dividedmind: Roles aren't currently visibility filtered (only resources are), so returning 404 on non-held role would be a bit weird.

For a nonexistent role 404 would be a wrong response for a header -- because 404 semantically says the the target resource (in REST sense) does not exist, which wouldn't be true. For a query parameter, I suppose returning 404 does make some sense in case the role doesn't exist.

db/migrate/20180530162704_is_role_ancestor_of.rb
dividedmind: Yes.

dividedmind:
(However I'll note it was universal in v4 even if not documented as such.)

apotterri:
I took out support for the header.

I changed the controller concern to `AssumedRole`.

I went back to using `role` as the parameter name in `#index`, so it would match the parameter in `#check_permission`.

I also changed `Role#is_ancestor_of?` to `Role#ancestor_of?`.

What do you think, @dividedmind?

dividedmind:
Way ahead of you! Already approved :)

PR #557, opened 2018-05-30
Adding `RolesController#members` search API endpoint 

micahlee:
For CONJ-5074

This PR adds the `?members` endpoint back to the Roles API. This endpoint allows searching, filtering, and paging of the Role members.

This is to support searching and paging members in the Conjur UI.

The new cucumber tests are probably the best place to start:
<https://github.com/cyberark/conjur/pull/557/files#diff-c7d2b396124fd17f55303b265705a6c7>

The cucumber API tests will test the new and updated features.

https://jenkins.conjur.net/view/cyberark/job/cyberark--conjur/job/search_role_members/

**Yes**

**Yes**

#### What's left to do:
- [x] Add `kind` filter to members
- [x] Update API documentation
- [x] Update tests
- [x] Update changelog

### DIFF
--- a/app/controllers/roles_controller.rb
+++ b/app/controllers/roles_controller.rb
@@ -1,8 +1,7 @@
 class RolesController < RestController
-  before_filter :find_role, only: [ :show, :all_memberships, :direct_memberships, :members ]
 
   def show
-    render json: @role.as_json.merge(members: @role.memberships)
+    render json: role.as_json.merge(members: role.memberships)
   end
 
   # Find all role memberships, expanded recursively. If no parameters
@@ -10,20 +9,13 @@ class RolesController < RestController
   #
   # If +params[:filter]+ is given, return (as role ids), the subset of
   # memberships that matches the filter
-  
+  #
   # If +params[:count]+ is given, return the count of memberships,
   # rather than the memberships themselves
   #
   def all_memberships
-    options = membership_options
-    
-    memberships = with_filter(options) do
-      @role.all_roles
-    end
-
-    json = render_count(memberships, options) || memberships.map(&:role_id)
-    
-    render json: json
+    memberships = filtered_roles(role.all_roles, membership_filter)
+    render_dataset(memberships) { |dataset| dataset.map(&:role_id) }
   end
 
   # Find all direct memberships, i.e don't recursively expand member
@@ -36,14 +28,8 @@ class RolesController < RestController
   # +params[:filter]+ and +params[:count]+ are handled as for +#all_memberships+
   #
   def direct_memberships
-    options = membership_options
-    
-    memberships = with_filter(options) do
-      @role.memberships_as_member_dataset
-    end
-
-    json = render_count(memberships, options) || memberships
-    render json: json
+    memberships = filtered_roles(role.memberships_as_member_dataset, membership_filter)
+    render_dataset(memberships)
   end
   
   # Find all members of this role.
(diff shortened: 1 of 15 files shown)


app/controllers/roles_controller.rb
```diff
@@ -46,8 +46,33 @@ def direct_memberships
...
+  # For each member, return the full details of the grant as a
+  # JSON object
+  #
+  # +params[:count] returns only the number of members
+  # +params[:limit] and [:offset] control the paging
+  # +params[:search] returns only the members that match the search string
+  def members
+    filter_options, render_options = members_options
+
+    render_options[:order_by] ||= :role_id
```
kgilpin: Not member_id?

micahlee: Ah, yes. You are correct.

app/controllers/roles_controller.rb
```diff
@@ -46,8 +46,33 @@ def direct_memberships
...
+  # Find all members of this role.
+  #
+  # For each member, return the full details of the grant as a
+  # JSON object
+  #
+  # +params[:count] returns only the number of members
+  # +params[:limit] and [:offset] control the paging
+  # +params[:search] returns only the members that match the search string
+  def members
+    filter_options, render_options = members_options
```
jonahx: I'd just make each of this its own method, rather than using array destructuring + the array returning `members_options` method

app/controllers/roles_controller.rb
```diff
@@ -46,8 +46,33 @@ def direct_memberships
...
+
+    render_options[:order_by] ||= :member_id
```
jonahx: And then just make this part of the new `render_options` method

app/controllers/roles_controller.rb
```diff
@@ -46,8 +46,33 @@ def direct_memberships
...
+
+    members = @role.memberships_dataset
```
jonahx: Make a method for this too.... it's really nice if the controllers are just a sequence of high-level instructions, without implementation detail

app/controllers/roles_controller.rb
```diff
@@ -46,8 +46,33 @@ def direct_memberships
...
+                   .search(filter_options)
+                   .select(:role_memberships.*)
+
+    json = render_count_or_results(members, render_options)
```
jonahx: I try to avoid any method name with an "or" or "and" in it.  It's almost always a symptom that it's mixing concerns.  In this particular case, the "error" (as I see it) was in the API design as I mentioned before, but that's just part of the requirement.  That said, we can at least make it as clear as possible that the controller action has two distinct paths.  I also try to avoid `if... else` where I can, but in this case I'd recommend exactly that.  One case for when we're returning a count, another for the normal case.  I really want to draw attention to that bifurcation because it's something I'd consider unexpected and potentially confusing.

app/controllers/roles_controller.rb
```diff
@@ -72,8 +104,18 @@ def with_filter(options, &block)
     filter ? roles.member_of(filter) : roles
   end
     
-  def render_count(roles, options)
-    { count: roles.count } if options.has_key?(:count)
+  def render_count(dataset, options)
+    { count: dataset.count } if options.has_key?(:count)
   end
   
+  def render_results(dataset, order_by: nil, offset: nil, limit: nil)
```
jonahx: I like this code, but I'd _consider_ putting it in the model, which is where I'd expect the db interactions to go.  That said, I don't feel too strongly since it's pretty simple and likely won't be reused anywhere else.

cucumber/api/features/member_list.feature
```diff
@@ -0,0 +1,204 @@
...
+            "admin_option": false,
+            "member": "cucumber:user:alice",
+            "ownership": false,
+            "policy": "cucumber:policy:root",
+            "role": "cucumber:group:employees"
+        }
+    ]
+    """
+
+    When I successfully GET "/roles/cucumber/group/employees?members&kind[]=group"
```
kgilpin: Will this also work with a plain `kind` parameter?

micahlee: Yes, and I added a test case for this. Thanks!

app/models/membership_search.rb
```diff
@@ -0,0 +1,48 @@
...
+    # Sounds just about right. A are name and id, B is rest of annotations, C is kind.
+    rank = Sequel.function(:ts_rank_cd, :textsearch, query)
+
+    left_join(:resources_textsearch, resource_id: :member_id)
+      .where("? @@ textsearch", query)
+      .order(Sequel.desc(rank))
+  end
+
+  # result_set renders a dataset to a result set using the
+  # provided order and paging parameters
```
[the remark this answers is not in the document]
micahlee: Is this really mutating it's own state? I thought each step returned a new Sequel dataset:
- `scope = self` captures the initial dataset (the association table).
- `scope = scope.where` sets scope to a new dataset with the additional where clause.
- `scope = scope.textsearch` sets the `scope` to a new dataset again adding a textsearch to the chain.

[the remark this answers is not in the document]
dividedmind: Micah is correct. Datasets are immutable by design.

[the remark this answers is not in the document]
jonahx: @dividedmind @micahlee it's still mutation even though the datasets are immutable.  You are mutating the value of self / scope.

```
a = 1
a = a + 9
```
that's mutation.  i probably shouldn't have mentioned it as i don't plan to suggest a fix in this case, but i suppose i just wanted it called out.

PR #571, opened 2018-06-16
Authenticate audit events

dividedmind:
This adds 'authenticate' audit events to the audit stream. 

Note the volume of these messages is substantial so we should consider whether they should be enabled by default, or somehow filtered. (These messages weren't present in v4.)

Comments welcome especially on the message schema (see docs/audit.md).

### DIFF
--- a/app/models/audit.rb
+++ b/app/models/audit.rb
@@ -8,6 +8,14 @@ module Audit
       logger.info LogMessage.new msg, msgid, data, Syslog::LOG_NOTICE
     end
 
+    def info msg, msgid, data
+      logger.info LogMessage.new msg, msgid, data, Syslog::LOG_INFO
+    end
+
+    def warn msg, msgid, data
+      logger.warn LogMessage.new msg, msgid, data, Syslog::LOG_WARNING
+    end
+
     def logger
       @logger ||= Rails.logger
     end
@@ -21,7 +29,7 @@ module Audit
     # cf. https://pen.iana.org
     CONJUR_PEN = 43868
     def self.conjur_sdid label
-      [label, CONJUR_PEN].join('@').freeze
+      [label, CONJUR_PEN].join('@').intern
     end
 
     POLICY = conjur_sdid 'policy'
--- a/app/models/audit.rb
+++ b/app/models/audit.rb
@@ -4,16 +4,15 @@ require 'time'
 
 module Audit
   class << self
-    def notice msg, msgid, data
-      logger.info LogMessage.new msg, msgid, data, Syslog::LOG_NOTICE
-    end
-
-    def info msg, msgid, data
-      logger.info LogMessage.new msg, msgid, data, Syslog::LOG_INFO
-    end
-
-    def warn msg, msgid, data
-      logger.warn LogMessage.new msg, msgid, data, Syslog::LOG_WARNING
+    {
+      notice: [:info, Syslog::LOG_NOTICE],
+      info: [:info, Syslog::LOG_INFO],
+      warn: [:warn, Syslog::LOG_WARNING]
+    }.each do |meth, details|
+      logger_method, syslog_severity = details
+      define_method meth do |msg, msgid, facility: nil, **data|
+        logger.send logger_method, LogMessage.new(msg, msgid, data, syslog_severity, facility)
+      end
     end
 
     def logger
@@ -39,14 +38,15 @@ module Audit
   end
 
   class LogMessage < String
-    def initialize msg, msgid, structured_data = nil, severity = nil
+    def initialize msg, msgid, structured_data = nil, severity = nil, facility = nil
       super msg
       @msgid = msgid
       @structured_data = structured_data
       @severity = severity
+      @facility = facility
     end
 
-    attr_reader :msgid, :structured_data, :severity
+    attr_reader :msgid, :structured_data, :severity, :facility
   end
 
   # Middleware to store request ID in a thread variable
@@ -65,9 +65,6 @@ module Audit
       Logger::Severity::INFO => Syslog::LOG_INFO,
       Logger::Severity::WARN => Syslog::LOG_WARNING
     }
-
-    # TODO
-    FACILITY = 32
 
     def call severity, time, progname, msg
       severity = if msg.respond_to? :severity
@@ -90,8 +87,10 @@ module Audit
       msgid = msg.msgid if msg.respond_to? :msgid
       sd = format_sd sd
 
+      facility = msg.try(:facility) || 4
+
       fields = [timestamp, hostname, progname, pid, msgid, sd, msg]
-      ["<#{severity + FACILITY}>1", *fields.map {|x| x || '-'}].join(" ") + "\n"
+      ["<#{severity + (facility << 3)}>1", *fields.map {|x| x || '-'}].join(" ") + "\n"
     end
 
   private
--- a/app/models/audit.rb
+++ b/app/models/audit.rb
@@ -37,6 +37,9 @@ module Audit
     ACTION = conjur_sdid 'action'
   end
 
+  # This inherits from String in order to preserve compatibility with Ruby's
+  # built-in formatters, which check the class of the parameter (probably as
+  # an optimization).
   class LogMessage < String
     def initialize msg, msgid, structured_data = nil, severity = nil, facility = nil
       super msg
@@ -90,7 +93,8 @@ module Audit
       facility = msg.try(:facility) || 4
 
       fields = [timestamp, hostname, progname, pid, msgid, sd, msg]
-      ["<#{severity + (facility << 3)}>1", *fields.map {|x| x || '-'}].join(" ") + "\n"
+      # per RFC5424 priority = severity + facility * 8
+      ["<#{severity + facility * 8}>1", *fields.map {|x| x || '-'}].join(" ") + "\n"
     end
 
   private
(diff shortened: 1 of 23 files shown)


app/models/audit/event/authn.rb
```diff
@@ -0,0 +1,53 @@
+require 'ostruct'
+
+class Audit::Event::Authn < OpenStruct
```
jonahx: Currently the interface for this class is only implicit: you have to read through every method to figure out what attribute assignments must be made before it can be used.  It must be made explicit one way or the other.  My preference would be:
1. Write a normal intializer which accepts the member variables.  This has the advantage of not only being explicit, but also making the client code interface clear and consistent.  There's really no reason anyone should be using attr setters with this object.  Also, you can enforce immutability.
2. If you don't want to do 1. right now,  please at least document the intended interface in comments.

app/models/audit/event/authn.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  SUCCESS_TEMPLATE = "%s successfully authenticated with authenticator %s service %s".freeze
+  FAILURE_TEMPLATE = "%s failed to authenticate with authenticator %s service %s: %s".freeze
+
+  def role_id
+    role.id
+  end
+
+  def service_id
+    service.try :id
```
jonahx: I'd prefer not to rely on Rails methods in our domain code.  I'd like to see a require active_support if you want to keep it (just to make that dep clear) or else `service.id rescue nil`

app/models/audit/event/authn.rb
```diff
@@ -0,0 +1,53 @@
+require 'ostruct'
+
+class Audit::Event::Authn < OpenStruct
+  def emit_success
```
jonahx: You don't need to fix this, but there's something confusing to me about the naming.  we have an "event" emitting a "success" (a "success event").  so an event is emitting an event, which reads oddly.  

There's also a kind of strange inversion in the design, in that this object delegates to the thing that actually writes the log.  Ie, we have an `Event` with an `Audit` (the actually log writer) inside of it, and the event calls its own `emit_success` which then calls Audit, passing the events own structured data to `Audit.info`.  

I think a simpler, clearer design would be to just have pass events to Audit's writer methods, and allow client code to use `Audit`.  But that's a larger (though not very large) refactor -- just food for thought.

app/models/resource.rb
```diff
@@ -52,6 +52,12 @@ def find_if_visible role, *a
       res = find *a
       res if res.try :visible_to?, role
     end
+
+    # Specialization to allow lookup by composite ids,
+    # eg. Resource[account, kind, id]
```
jonahx: Just a note for later (not specific to your code), but just to make you aware: we have similar logic to this in multiple places in our code base (and others I believe) including my rotator code.  At some point we should DRY it all up.

app/models/audit.rb
```diff
@@ -82,8 +87,10 @@ def call severity, time, progname, msg
       msgid = msg.msgid if msg.respond_to? :msgid
       sd = format_sd sd
 
+      facility = msg.try(:facility) || 4
```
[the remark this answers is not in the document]
dividedmind: Yeah, I'll probably do that eventually.

app/models/audit.rb
```diff
@@ -82,8 +87,10 @@ def call severity, time, progname, msg
...
+
       fields = [timestamp, hostname, progname, pid, msgid, sd, msg]
-      ["<#{severity + FACILITY}>1", *fields.map {|x| x || '-'}].join(" ") + "\n"
+      ["<#{severity + (facility << 3)}>1", *fields.map {|x| x || '-'}].join(" ") + "\n"
```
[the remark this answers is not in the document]
dividedmind: Yeah, this is an RFC thing. I'll change it to `* 8` to make it clearer.

app/models/audit/event/authn.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def message
+    if success?
+      SUCCESS_TEMPLATE % [role_id, authenticator_name, service_id]
+    else
+      FAILURE_TEMPLATE % [role_id, authenticator_name, service_id, error_message]
+    end
+  end
+
+  SUCCESS_TEMPLATE = "%s successfully authenticated with authenticator %s service %s".freeze
```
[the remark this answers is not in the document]
dividedmind: If they'd been meant for external use, I would. But since these are in a way only local, I like to keep them near the code that uses them.

app/models/audit.rb
```diff
@@ -31,14 +38,15 @@ def self.conjur_sdid label
   end
 
   class LogMessage < String
```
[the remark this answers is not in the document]
dividedmind: I'll add it, thanks.

app/models/audit.rb
```diff
@@ -31,14 +38,15 @@ def self.conjur_sdid label
...
       @structured_data = structured_data
       @severity = severity
+      @facility = facility
     end
 
-    attr_reader :msgid, :structured_data, :severity
+    attr_reader :msgid, :structured_data, :severity, :facility
   end
 
   # Middleware to store request ID in a thread variable
```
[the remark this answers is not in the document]
dividedmind: There is no specific reason. It'll be moved out at some point.

app/models/audit/event/authn.rb
dividedmind: Good point, I'll make it clearer.

app/models/audit/event/authn.rb
dividedmind: I understand where you're coming from, but I like the expressiveness of some of ActiveSupport utilities and this code is unlikely to be moved out.

app/models/audit/event/authn.rb
dividedmind: Spent some time thinking about it actually. Maybe it'll get refactored at some point.

app/models/resource.rb
dividedmind: Agreed. That's why I added this method in lieu of copying the same functionality again, even though I didn't go the one step further and found all the places where it could be used.

kgilpin:
Some of your answers to the questions in the PR really seem like they would
be very helpful as comments in the code. For example, why you inherit from
String.

dividedmind:
Listen @jonahx , I spent all night trying to get to a working solution that would make us both happy, but I spent way too much time on it already and am over any reasonable timebox. Let's leave this as is and please file a tech debt issue if you want.

Not that the time has been wasted, as I added a lot of test code and noticed and fixed a couple bugs on the way, which is why this PR grown a little bit; you might want to give it another look-over. I tried to address some of your concerns (however some are outside the scope of this change).

The commits are clean so you can go through them one-by-one, and also if you're going to merge do either rebase merge or straight merge.

For the record here are WIPs of two approached I tried:
- https://github.com/cyberark/conjur/commit/027b5986d743c2ab3a7a4b5038c76b04372adb24
- https://github.com/cyberark/conjur/commit/f1387b6e6079f4bbae39ec81d68345f69adf0b8f

jvanderhoof:
@dividedmind: I like the approach you and Jonah have taken on this one.  I'd ask that you please file issues for work that we reasonably plan to address in the future. As the implementor and visionary of this work, you're in the best place to understand the work that is being pushed out to a future date.

app/models/audit.rb
```diff
@@ -21,7 +28,7 @@ module SDID
     # cf. https://pen.iana.org
     CONJUR_PEN = 43868
     def self.conjur_sdid label
-      [label, CONJUR_PEN].join('@').freeze
+      [label, CONJUR_PEN].join('@').intern
```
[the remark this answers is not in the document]
jonahx: After your experience with my code, I now believe that any time something in code needs an explanation, having it documented elsewhere is not enough.  Either repeat the explanation directly in the code comments, or include a link to the explanation in the code comments.

app/models/resource.rb
jonahx: I don't understand you what you mean, this does copy the same functionality again.  It can't be reused beause here it exists in a very specific context (inside an audit event object), whereas in the codebase in general it's used in many different contexts.  I'm not saying this because I want you to change anything atm, just want to clarify my initial point.

app/models/audit.rb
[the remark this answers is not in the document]
dividedmind: Fair :) Could you say more about what do you feel needs explanation?

app/models/resource.rb
dividedmind: What prevents this method from being used in all the other different contexts?

security-scan.sh
```diff
@@ -18,7 +18,7 @@ done
 if [[ $RUN_BRAKEMAN = true ]]; then
   mkdir -p brakeman/reports
   chmod -R 0777 brakeman/reports
-  docker run -v "$(pwd):/tmp/" --rm -w /tmp/ codeclimate/codeclimate-brakeman brakeman -o brakeman/reports/brakeman-output.html
+  docker run -v "$(pwd):/tmp/" --rm -w /tmp/ codeclimate/codeclimate-brakeman:b804 brakeman -o brakeman/reports/brakeman-output.html
```
dividedmind: Thank you! I was just about to add this :)

jvanderhoof: No worries. I think we need to start building our own utility containers for stuff like this....

dividedmind: I don't think it's worth the effort for one-tool containers like that. I do think, however, that we ought to start pinning known good versions. Even better, actually review the images to make sure there are no backdoors or surprises of some sort, and then pin by specific SHA that we reviewed.

dividedmind:
I need to take a look at why are those new cukes failing in Jenkins. Stay tuned.

dividedmind:
(I'll go ahead and merge after I fix that, unless I hear a veto.)

PR #575, opened 2018-06-19
Adding paging params to audit engine and additional test

micahlee:
Fixes https://github.com/conjurinc/appliance/issues/216

This PR adds `:offset` and `:limit` parameters to the audit search endpoints to support paging the results. 

This PR also includes and update to Sprockets to resolve [CVE-2018-3760] Path Traversal in Sprockets. This was blocking the build on the vulnerability scan.

See the new test 
(<https://github.com/cyberark/conjur/compare/add_paging_to_audit?expand=1#diff-5d5386e82cd58fedccb05c222859234e>) 
for the expected behavior, and the updated controller 
(https://github.com/cyberark/conjur/compare/add_paging_to_audit?expand=1#diff-88088d05d0d394a58ae4da4e198d1280)
for the implementation changes

The new RSpec test can be run for the audit engine.

https://jenkins.conjur.net/job/cyberark--conjur/job/add_paging_to_audit/

It includes a new RSpec test

No

No

No


micahlee:
Hey @dividedmind, so we need `limit`, since that is what the UI is expecting and causing this bug. And `limit` & `offset` are paging patterns we have everywhere else already (and you may potentially need them within since/until as well). So how about I open a ticket for adding those later, since they don't directly relate to this bug?

dividedmind:
`limit` is a no-brainer, but `offset` doesn't work very well generally for interactive speeds and leads to a very drunk experience with fast-changing data such as audit. 

I can see @jvanderhoof already merged it, which is IMO a bad decision. Also, doesn't it require corresponding changes in the UI now? I'm pretty sure audit in v4 didn't support `offset` (specifically because of these issues) and so the UI code uses `till` and `since`.

micahlee:
@dividedmind, I agree this could have used a little more consensus before merging. I would have been ok with removing `offset` and only implementing `limit` in this PR, since that's what the UI was already passing as a parameter, and just needed to be implemented.

jvanderhoof:
I pulled this in part because I believe there are unfortunate experiences in both approaches. By standardizing a return size, we insure pagination works as expected (showing 5 more events doesn't return 8 more events). This comes at the cost of potentially seeing duplicate events as a user pages through results.

I'm not sure which is worse, but made the call to go with limit/offset.  We can circle back around in the future as we get more feedback.

PR #581, opened 2018-06-21
Rotator cukes

jonahx:

Adds general rotation framework and the postgres rotator, along with cukes.

Entrypoint is `bin/conjur_cli.rb:82`.  But the true entrypoint for rotator logic is in master_rotator.

I'll add more detail later, but it's important to know that rotators communicate with conjur *only* via the `ConjurFacade`.  

`cd dev && ./start --rotators` will drop you into the cuke container, and startup a testdb container whose rotation we test against.  You just need to open the conjur container, do a `conjurctl server`, then tab back to the cuke container and do `bundle exec cucumber -p rotators`

https://jenkins.conjur.net/job/cyberark--conjur/job/rotator-cukes/17/


app/domain/rotation/master_rotator.rb
```diff
@@ -60,6 +60,7 @@ def initialize(avail_rotators:, facade:, secret_model:)
       def run
         rotator.rotate(@facade)
       rescue => e
+        raise e
```
apotterri: Reraising `e` doesn't look right here. It'll bypass the error handling below.

apotterri: Reraising `e` doesn't look right here. It'll bypass the error handling below.

cucumber/rotators/features/password.feature
```diff
@@ -0,0 +1,31 @@
...
+               rotation/postgresql/password/length: 32
+    """
+    And I create a db user "test" with password "secret"
+    And I watch for changes in "db-reports/password" and db user "test"
+    And I add the value "testdb" to variable "db-reports/url"
+    And I add the value "test" to variable "db-reports/username"
+    And I add the value "secret" to variable "db-reports/password"
+
+  Scenario: Values are rotated according to the policy
+    Given I wait for 3 seconds
```
apotterri: Like we talked about before, I think you're asking for trouble by making your tests depend on something happening within a particular amount of time. Once Jenkins starts running several jobs on the same executor, all bets are off.

.gitignore
```diff
@@ -39,3 +39,4 @@ api.html
 
 brakeman-output.*
 .idea
+TEMP_NOTES.txt
```
micahlee: Do you mean to include this in the final outcome?

app/domain/rotation/conjur_facade.rb
```diff
@@ -38,15 +44,24 @@ def current_values(variables)
     #
     def update_variables(new_values, &rotator_code)
       @db.transaction do
+
+        # lock_secrets_for_resource_ids(new_values.keys)
```
apotterri: I think I'd rather see this removed. Given that it's commented out, it's obviously not affecting the success of the cukes. Until we have agreement about whether we need to do locking, it doesn't seem like it belongs here.

app/domain/rotation/conjur_facade.rb
```diff
@@ -38,15 +44,24 @@ def current_values(variables)
...
+
         new_values.each do |resource_id, value|
           update_secret(resource_id, value)
         end
         rotator_code.call if rotator_code
       end
+    rescue => e
+      puts "Rotator update failed: #{e}"
```
apotterri: How can we ever wind up here? If there's no way, this `rescue` should get removed. If there's some way, it needs to do more error handling than just a `puts`.

dev/start
```diff
@@ -52,8 +58,19 @@ if [[ $ENABLE_AUTHN_LDAP = true ]]; then
...
+  services="$services testdb cucumber"
+fi
+
 docker-compose up -d --no-deps $services
 
 api_key=$(docker-compose exec -T conjur conjurctl \
 	role retrieve-key cucumber:user:admin | tr -d '\r')
-docker exec -e CONJUR_AUTHN_API_KEY=$api_key $env_args -it --detach-keys 'ctrl-\' $(docker-compose ps -q conjur) bash
+
+if [[ $ENABLE_ROTATORS = true ]]; then
```
apotterri: These `docker exec`s are complicated enough that they probably deserve to be DRYed up. Maybe do something like

```
if [[ $ENABLE_ROTATORS = true ]]; then
  container_name=cucumber
else
  container_name=conjur
fi

docker exec -e CONJUR_AUTHN_API_KEY=$api_key $env_args \
    -it --detach-keys 'ctrl-\' "$(docker-compose ps -q $container_name)" bash
```

app/domain/rotation/rotated_variable.rb
```diff
@@ -39,7 +39,6 @@ def name
     # and returns everything up to th first slash
     #
     def prefix
-      puts '@resource_id', @resource_id
       @resource_id.match(%r{(.*)/.*})[1]
```
micahlee: Wouldn't these Id utilities be more appropriate on the `Resource` class or even a new `Id` helper class? Would you mind creating a ticket for this if you don't want to address it now?

app/domain/rotation/rotators/postgresql/password.rb
```diff
@@ -103,6 +96,29 @@ def username_id
           end
         end
 
+        class RotatedDb
```
micahlee: I'm not loving this class name because it doesn't clearly communicate what it is responsible for. 

`DbPasswordUpdater`, `DbUpdater`, or if you wanted to keep it more generic, even just `DbConnection` or `Database`. 

Any of those better communicate, at least to me, what I expect an instance of the class to represent.

app/models/secret.rb
```diff
@@ -54,6 +54,10 @@ def update_expiration(resource_id, expires_at)
       Secret.where(resource_id: resource_id).update(expires_at: expires_at)
     end
 
+    def current_values_for_update(resource_ids)
+      Secret.for_update.select(resource_id: resource_ids).all
```
micahlee: It looks like this is only used in the commented out method `conjur_facade.rb#lock_secrets_for_resource_ids`, but...

Can you elaborate on the purpose of of this method? I ask because I don't believe this use of `for_update` is doing what you expect. When you end the chain with `.all`, this is going to execute the command with the update lock and return the results as an Array, with the lock released.

If you want downstream operations to use the lock, you should be returning the dataset (no `.all`) rather than the result array.

cucumber/rotators/features/password.feature
micahlee: This is where the idea of polling would fit in. That the test would read something like:
```
Given I wait until the values have rotated
```
and the step definition would perform a sleep poll checking for the updated variable versions. If you set your poll interval to the TTL, then you're probably only waiting once or twice, but you are now more resilient to variance in the runtime environment.

With the polling, you probably do want to check that the actual rotate -> rotate duration (From the model timestamps) is actually within an acceptable margin of the TTL.

cucumber/rotators/features/support/env.rb
```diff
@@ -0,0 +1,11 @@
+# TODO: find better way to share common World
```
micahlee: Can you create a ticket for this, rather than a TODO comment?

.gitignore
jonahx: yeah, it's just a file for my personal notes that i want ignored.

app/domain/rotation/master_rotator.rb
jonahx: good catch, i put that there during testing and forgot to remove.

cucumber/rotators/features/password.feature
jonahx: @alan can you elaborate on the Jenkins problem.  I don't believe there is an issue here if you look at the logic.  The test is testing what we actually want it to test -- ie, we actually do need to test that the ttl works.

@micahlee i couldn't tell what you were arguing for here.  but fwiw i'm currently using polling under the hood.  and since i poll every 1/3 of a second, it's guaranteed that i'll see 3 distinct keys if i wait 3 seconds.  if that doesn't happen, we *want* an error reported.

app/domain/rotation/conjur_facade.rb
jonahx: @apotterri ok i'll remove this and add it as an issue that we can discuss later.

app/domain/rotation/conjur_facade.rb
jonahx: We can definitely wind up here.  When Rafal integrates the audit code, he'll audit this event.

dev/start
jonahx: Agree, that was me being lazy since this script wasn't part of the main code, but I'll make this change.

app/domain/rotation/conjur_facade.rb
```diff
@@ -38,15 +44,24 @@ def current_values(variables)
...
     end
 
     private
 
+    def lock_secrets_for_resource_ids(resource_ids)
+      @secret_model.current_values_for_update(resource_ids)
+    end
```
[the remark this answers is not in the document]
jonahx: Will remove.

app/domain/rotation/rotators/postgresql/password.rb
```diff
@@ -6,22 +6,31 @@ module Postgresql
 
       class Password
 
-        def initialize(password_factory: ::Rotation::Base58Password, pg: ::PG)
+        PASSWORD_ANNOTATION = 'rotation/postgresql/password/length'
```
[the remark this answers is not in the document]
jonahx: Love, love this attention to naming detail.  Will change.

app/domain/rotation/rotated_variable.rb
jonahx: You are 100% right, in that this logic isn't specific to a rotated variable, and in fact different versions of this code exist all over the code base.  But because I didn't want to open that can of worms, and I still don't, I left it like this.  I will make a note to open an issue on this, and the related bits of code.  This is a larger issue than it seems because it opens up questions about how to share not only across features but even across gh projects which our team has never answered in a standardized way.

app/domain/rotation/rotators/postgresql/password.rb
jonahx: Fair enough, but `DbPasswordUpdater` and `DbUpdater` are [out of the question](http://www.carlopescio.com/2011/04/your-coding-conventions-are-hurting-you.html).  (I have about 5 more links on this subect if your interested :) ).

I'll go ahead and change it to `Database` though.

app/domain/rotation/conjur_facade.rb
apotterri: It needs to do more than audit, though, doesn't it? If rotation fails, it needs to reset the expiration.

jonahx: Yep, you're correct.  I forgot I added that during debugging.  I've just removed it and repushed, so now the errors will be caught by this part in MasterRotator, as intended:
```ruby
      def run
        rotator.rotate(@facade)
      rescue => e
        p "Catching error", e
        set_retry_expiration
        log_error(e)
      end
```

PR #588, opened 2018-06-26
Audit authorize events

dividedmind:
Closes conjurinc/appliance#203

Add authorization events.

There is still an open question whether implicit authorization checks should generate these events. Comments welcome.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -49,10 +49,16 @@ class ResourcesController < RestController
     privilege = params[:privilege]
     raise ArgumentError, "privilege" unless privilege
 
-    if assumed_role.allowed_to?(privilege, @resource)
-      head :no_content
-    else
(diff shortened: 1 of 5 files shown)


app/controllers/resources_controller.rb
```diff
@@ -49,10 +49,16 @@ def check_permission
...
+
+    Audit::Event::Check.new(
+      user: current_user,
+      resource: @resource,
+      privilege: privilege,
+      role: assumed_role,
+      success: result
+    ).log_to Audit.logger
+
+    head result ? :no_content : :not_found
```
apotterri: Maybe it's just me, but I find statements like this really hard to parse. It always take me an extra look, and a moment of wondering, "do the precedence rules mean that this is going to behave correctly?" I'd have a much easier time with:

```
head(result ? : no_content : :not_found)
```

PR #592, opened 2018-06-27
Endpoint for expiring rotating variables early

### LINKED ISSUE #568: Expiration service has been ported to Conjur core
GIVEN a running a running Conjur CE container, a variable with the annotation 'expiration/timestamp', and a user with permission to retrieve that variable
WHEN a variable has an annotation with a time in the future
THEN the variable value can be retrieved
AND WHEN a variable has an annotation with a time in the past
THEN the variable value request returns a 404 response

References: CONJ-4799

jonahx:
Closes #568 

• Fixes a bug in `Secret.scheduled_rotations` query
• Adds an API endpoint to instantly expire a rotating variable
• Adds cucumber tests to test that endpoint

1. `cd dev && ./start --rotators`
2. Start conjur container in another tab `conjurctl server`
3. `bundle exec cucumber -p rotators`

The `expiration.feature` is the new cuke relevant to this PR.

https://jenkins.conjur.net/job/cyberark--conjur/job/expire-api/5/

Yes

### DIFF
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -85,6 +85,23 @@ class SecretsController < RestController
     ).log_to Audit.logger
   end
 
+  # NOTE: We're following REST/http semantics here by representing this as 
+  #       an "expirations" that you POST to you.  This may seem strange given
+  #       that what we're doing is simply updating an attribute on a secret.
+  #       But keep in mind this purely an implementation detail -- we could 
+  #       have implemented expirations in many ways.  We want to expose the
+  #       concept of an "expiration" to the user.  And per standard rest, 
+  #       we do that with a resource, "expirations."  Expiring a variable
+  #       is then a matter of POSTing to create a new "expiration" resource.
+  #       
+  #       It is irrelevant that the server happens to implement this request
+  #       by assigning nil to `expires_at`.
+  #
+  #       Unfortuneatly, to be consistent with our other routes, we're abusing
+  #       query strings to represent what is in fact a new resource.  Ideally,
+  #       we'd use a slash instead, but decided that consistency trumps 
+  #       correctness in this case.
+  #
   def expire
     authorize :update
     Secret.update_expiration(@resource.id, nil)
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -86,7 +86,7 @@ class SecretsController < RestController
   end
 
   def expire
-    raise 'WE MADE IT'
+    authorize :update
     Secret.update_expiration(@resource.id, nil)
     head :created
   end
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -84,4 +84,10 @@ class SecretsController < RestController
       user: current_user
     ).log_to Audit.logger
   end
+
+  def expire
+    raise 'WE MADE IT'
+    Secret.update_expiration(@resource.id, nil)
+    head :created
+  end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -42,10 +42,14 @@ Rails.application.routes.draw do
       get     "/resources/:account"                   => "resources#index"
       get     "/resources"                            => "resources#index"
 
+      # NOTE: the order of these routes matters: we need the expire 
+      #       route to come first.
+      constraints account: /[^\/]*/, kind: /[^\/]+/, identifier: /.+/ do
+        post    "/secrets/:account/:kind/:identifier/expirations" => "secrets#expire"
+      end
       get     "/secrets/:account/:kind/*identifier" => 'secrets#show'
       post    "/secrets/:account/:kind/*identifier" => 'secrets#create'
       get     "/secrets"                            => 'secrets#batch'
-      post    "/secrets/:account/:kind/*identifier/expirations" => "secrets#expire"
 
       put     "/policies/:account/:kind/*identifier" => 'policies#put'
       patch   "/policies/:account/:kind/*identifier" => 'policies#patch'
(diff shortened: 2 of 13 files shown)


config/routes.rb
```diff
@@ -42,6 +42,11 @@ def matches?(request)
       get     "/resources/:account"                   => "resources#index"
       get     "/resources"                            => "resources#index"
 
+      # NOTE: the order of these routes matters: we need the expire 
+      #       route to come first.
+      constraints account: /[^\/]*/, kind: /[^\/]+/, identifier: /.+/ do
+        post    "/secrets/:account/:kind/:identifier/expirations" => "secrets#expire"
```
apotterri: When we originally talked about this route, I guess I was expecting that `POST`ing to `/secrets/:account/:kind/:identifier/expirations` was going to take an expiration from the client and apply it to the secret.

Since that's not what's happening, I wonder if this should be something else (though I'm not sure what)?

cucumber.yml
```diff
@@ -13,11 +13,20 @@ authenticators: >
   -r cucumber/authenticators/features/support
   -r cucumber/authenticators/features/step_definitions cucumber/authenticators
 
+# NOTE: We have to require the needed files from "api" individually, because
+#       if you mass require the folder it includes "api"s env.rb, which screws
+#       things up because (I think) it sets ENV['CONJUR_ACCOUNT'].  Cucumber
+#       profiles need to be thought through better and refactored most likely.
```
micahlee: This should be filed as an issue in the repo so we don't lose track of it.

app/controllers/secrets_controller.rb
```diff
@@ -68,4 +68,10 @@ def batch
 
     render json: result
   end
+
+  def expire
+    authorize :update
+    Secret.update_expiration(@resource.id, nil)
+    head :created
```
micahlee: Without documented conventions it's hard to have any ground to stand on but...

I don't know that I agree with `:created` as the right response here. `204 No Content` is more appropriate, in my opinion. :)

cucumber/api/features/secrets.feature
```diff
@@ -1,29 +1,32 @@
 @logged-in
 Feature: Adding and fetching secrets
 
-  Each resource in Possum has an associated list of "secrets", each of which is an arbitrary
-  piece of encrypted data. Access to secrets is governed by privileges:
+  Each resource in Possum has an associated list of "secrets", each of which is
```
micahlee: Change "Possum" to "Conjur"

config/routes.rb
micahlee: A couple of things on this:

- First, if just calling the endpoint simply expires the given secret. Then I think this should just be `/expire`.

- Second, if this were to accept a parameter to specify the new expiration (which is what I think @apotterri was asking about), then I think this should be change to the singular `/expiration`

- Third and finally, while I like how you've made the path work here. I think we need to either change the other routes to match (namely the role membership and member routes), or this one should match the others and use a query parameter constraint instead.

jonahx: @apotterri that was discussed, this was the feature we ended up going with per @jvanderhoof 's note that this was what was needed and trying to small batch it.  adding general expiration later should be close to trivial.

cucumber.yml
jonahx: @micahlee oh, it will :)

app/controllers/secrets_controller.rb
apotterri: I think if we switch to supplying an expiration, then `created` will be the right response (since you're creating an expiration for the secret).

jonahx: @micahlee I disagree here.  An expiration is being conceptualized as an http resource.  You "create" a new one to achive expiration.  So this follow REST semantics.

config/routes.rb
apotterri: I though about having the endpoint be `expire`, components of the path are supposed to be nouns, not verbs. The method you're using is supposed to be the verb.

Yes, I was suggesting accepting a parameter to specify the new expiration. if we did that, changing to `expiration` makes sense to me.

app/controllers/secrets_controller.rb
micahlee: Ok, now that I understand the intention better, I'm fine with the `created` response.

jvanderhoof: The V4 docs on expiration provide a mechanism for setting the variable to expire in the future: https://conjur.docs.apiary.io/#reference/variable/set-expiration.  Should we add similar support, thereby allowing `create` to be a more logical action?

jvanderhoof: This could be a future update as well if it falls out of scope of this feature.

config/routes.rb
micahlee: To summarize a slack discussion. I'm dropping my request for changing the semantics of the endpoint, but only the syntax (`?expirations` instead of `/expirations`) so that the endpoint is consistent with how sub-resources are handled with Roles.

jvanderhoof: If possible, can we keep this change inline with the current v4 functionality?  It'll make documentation & education a lot easier.

app/controllers/secrets_controller.rb
```diff
@@ -84,4 +84,27 @@ def audit_fetch resource, version: nil
...
+  #       But keep in mind this purely an implementation detail -- we could 
+  #       have implemented expirations in many ways.  We want to expose the
+  #       concept of an "expiration" to the user.  And per standard rest, 
+  #       we do that with a resource, "expirations."  Expiring a variable
+  #       is then a matter of POSTing to create a new "expiration" resource.
+  #       
+  #       It is irrelevant that the server happens to implement this request
+  #       by assigning nil to `expires_at`.
+  #
+  #       Unfortuneatly, to be consistent with our other routes, we're abusing
```
micahlee: "Unfortuneatly" => "Unfortunately"

PR #593, opened 2018-06-29
Audit failed variable access

dividedmind:
This is the backend part that partly closes conjurinc/appliance#201

It implements audit messages for failed variable fetches and updates.

### DIFF
--- a/app/models/audit/event/can_fail.rb
+++ b/app/models/audit/event/can_fail.rb
@@ -1,0 +1,49 @@
+require 'English'
+
+module Audit
+  class Event
+    module CanFail
+      def message
+        if success
+          success_message
+        else
+          [failure_message, error_message].compact.join(': ')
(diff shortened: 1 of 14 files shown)


app/models/audit/event/can_fail.rb
```diff
@@ -0,0 +1,49 @@
...
+          field success: true, error_message: nil
+          abstract_field :success_message, :failure_message
+          extend CanFail::ClassMethods
+        end
+      end
+
+      module ClassMethods
+        # Automatically initialize an instance with data from currently handled
+        # exception (if any).
+        def with_exception **parameters
```
apotterri: I think invocations of this method would read better if it was called `new_with_exception`.

dividedmind: Thanks, fixed.

PR #595, opened 2018-06-29
Available authenticators endpoint

micahlee:
For

This PR adds an endpoint in Conjur, `/authenticators`, that returns the installed and configured authenticators.

The route is setup between:
- `config/application.rb`
- `config/routes.rb`

Because this is going to be access by the Info service on the appliance, which has no identify, this endpoint is not authenticated. I welcome opinions on that, or ideas on how to not have it be so.

The data for the endpoint itself is assembled in:
`app/controllers/authenticate_controller.rb`

There is a test for the expected behavior in:
`cucumber/authenticators/features/available_authn.feature`

1) Start a development server, this alone should show the installed authenticator strategies and the default `authn` option at the `/authenticators` endpoint.

2) Load authenticator policies (per https://github.com/cyberark/conjur/blob/master/AUTHENTICATORS.md) and update the `CONJUR_AUTHENTICATORS` environment variable to change the information returned by the endpoint.

https://jenkins.conjur.net/job/cyberark--conjur/job/available_authenticators/

Yes

Maybe?

No

I don't know

### DIFF
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -11,10 +11,27 @@ end
 
 class AuthenticateController < ApplicationController
 
+  AUTHN_RESOURCE_PREFIX = "conjur/authn-".freeze
+
+  def index 
+    authenticators = {
+      # Installed authenticator plugins
+      installed: installed_authenticators.keys.sort,
+    
+      # Authenticator webservices created in policy
+      configured: configured_authenticators.sort,
+
+      # Authenticators white-listed in CONJUR_AUTHENTICATORS
+      enabled: enabled_authenticators.sort
+    }
+
+    render json: authenticators
+  end
+
   def authenticate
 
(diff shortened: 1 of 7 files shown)


app/controllers/authenticate_controller.rb
```diff
@@ -11,10 +11,28 @@
 
 class AuthenticateController < ApplicationController
 
+  AUTHN_RESOURCE_PREFIX = "conjur/authn-".freeze
+
+  def index 
+
```
dividedmind: Stray empty line.

config/routes.rb
```diff
@@ -10,7 +10,9 @@ def matches?(request)
 
 Rails.application.routes.draw do
   scope format: false do
+
```
dividedmind: Here too.

dividedmind: Here too.

PR #604, opened 2018-07-04
Some optimization in audit engine

dividedmind:
Fixes conjurinc/appliance#240.

- Add a gen_log.rb script which quickly generates large volume of audit data in mostly realistic shape (at least as seen by the RDBMS).
- Add missing indexes on the audit.messages table (note this won't be picked up by the appliance build yet).
- Only create audit.messages table in migrations if it doesn't exist (so appliance can use them).
- Use more efficient JSON serialization in the audit engine.

### DIFF
--- a/engines/conjur_audit/app/controllers/conjur_audit/messages_controller.rb
+++ b/engines/conjur_audit/app/controllers/conjur_audit/messages_controller.rb
@@ -3,7 +3,7 @@ require_dependency "conjur_audit/application_controller"
 module ConjurAudit
   class MessagesController < ApplicationController
     def index
-      render json: messages, status: messages.any? ? :ok : :not_found
+      render json: messages_json, status: messages.any? ? :ok : :not_found
     end
 
     private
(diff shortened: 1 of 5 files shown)


engines/conjur_audit/contrib/gen_log.rb
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env ruby
```
micahlee: Would this file not be more appropriate in the `engines/conjur_audit/bin` directory?

engines/conjur_audit/app/controllers/conjur_audit/messages_controller.rb
```diff
@@ -65,7 +65,16 @@ def messages
         dataset = dataset.limit(limit)
       end
 
-      @messages ||= dataset.all
+      @messages ||= dataset
+    end
+
+    # Convert messages to JSON.
+    # Note: For performance, we use JSON.dump directly.
```
micahlee: Thanks for adding these remarks on why this improves the performance. TIL!

engines/conjur_audit/contrib/gen_log.rb
dividedmind: That's an excellent question. The answer is that putting it in `bin` would place the onus on us to maintain it, whereas as it is, it's just a tool script that I wrote that might be useful to somebody in the future (most probably myself...). The usual caveat of `contrib` directory that YMMV and it's not exactly maintained. 

(The name comes from the fact that often those scripts are contributed, but deemed useful and valuable enough to keep.)

engines/conjur_audit/app/controllers/conjur_audit/messages_controller.rb
dividedmind: FWIW, I found it by profiling. There is another optimization opportunity that should give x20 improvement relating to timestamp handling, but I feel it's good enough for now that fetching 1e5 records takes ~10 seconds.

PR #605, opened 2018-07-04
Style fixes

dividedmind:
Fixes for some style issues. It's best to go through the commits one by one.


jvanderhoof:
There are a LOT of exceptions from the standard Ruby Style Guide here. It feels like this is more a reflection of our currently practices, not alignment with the broader audience.

Particularly for an open source project, can we try to move closer to the Ruby standards?

dividedmind:
I'm really sure why is this discussion here, not at #597, but regarding 'exceptions' from 'the' 'standard' 'Ruby Style Guide': there is no standard style guide. If there was a single style allowed in a language, other styles would be disallowed (or would cause deprecation warnings). 'Particularly for open source project': every project has its own style, and when you contribute to open source projects you're used to have to change the style depending on the specific project you contribute to. It's like when you go somewhere people talk differently, and it's to be expected and it's okay.

IMO the important thing is to be consistent, because so much of that is a matter of taste. And if rubocop has an 'enforced style' option of something we use, we can enforce that consistency. You'll note that I only selectively disabled a handful of cops where I felt any imposed style is too rigid (because I believe the choice of the clearest style depends on the context, and enforcing one in particular is a detriment to readability); other than that, I had rubocop enforce a style which is prevalent in our codebase, even if it's not the default.

That said, I'm okay with any style changes as long as it doesn't force me to use monospace fonts (or tabs!). Otherwise I'll have to hire @jonahx to reformat my every commit ;)

Issue #601, opened 2018-07-03
Host and User resources can be CIDR restricted


jvanderhoof:
As a Conjur operator, I want to be able to restrict the IP range a user or host can authenticate from.

GIVEN a host or a user resource with a `cidr` annotation
WHEN I authenticate from an IP address within the CIDR block
THEN I can authenticate
AND WHEN I authenticate from an IP address outside the CIDR block
THEN my authentication attempt fails.

*Notes*
Valid CIDR options:
cidr: "192.168.101.1"
cidr: "192.168.0.0/16" 
cidr: "192.168.0.1,192.168.1.10" 

Old CLI specific docs: https://developer.conjur.net/reference/services/authentication/

kgilpin:
Where did the discussion happen about what to name the database column, etc?

micahlee:
Hey @kgilpin, The discussion in slack begins here: 
https://conjurhq.slack.com/archives/CA8AT283D/p1530814551000475

kgilpin:
We should probably use github issues for discussion otherwise all the
context gets hidden

kgilpin:
Or at least summarize it here when we are done.

PR #616, opened 2018-07-09
WIP CIDR Authentication Restriction for Users and Hosts

micahlee:
PR 2 of 2 for https://github.com/cyberark/conjur/issues/601

* Adds storage of CIDR restrictions for User and Host authentication credentials
* Adds origin verification to authentication if network restrictions are set

- [ ] Add review instructions

- [ ] Add manual test instructions

#### Link to build in Jenkins
https://jenkins.conjur.net/job/cyberark--conjur/job/restricted_to_cidr/

Yes

Possibly

No

Yes

#### Remaining Tasks:
- [ ] Add origin verification
- [ ] Add tests for origin verification
- [ ] Add tickets for documentation updates


cucumber/api/features/step_definitions/request_steps.rb
```diff
@@ -83,6 +83,14 @@
   end
 end
 
+When(/^I authenticate as "([^"]*)" with account "([^"]*)"/) do |login, account|
+  user = lookup_user(login, account)
+  user.reload
+  steps %Q{
```
jonahx: This is one of things I'll be putting in my best practices docs, but `steps` is considered an anti-pattern.  From the cucumber docs:

> While it is possible to call steps from step definitions in Cucumber-Ruby,
> this is not recommended.  The best way to achieve composition and reuse, is to
> use the features of your programming language. If you want to combine several
> actions into one step, extract individual (helper) methods and call these
> methods from your step definition.

lib/util/cidr.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Util
+
+  # provides helper methods for interacting with CIDR network addresses 
+  module CIDR
+    class << self
+
+      # returns the formatted CIDR formatted string for a network
+      def format_cidr cidr
```
jonahx: I've been encouraging people to follow the rule "if you method returns something, name it after the thing being returned," so this would be `formatted_cidr`.  Then verb names are reserved for method with side effects.  Simple rule that imo really improves readability.

lib/util/cidr.rb
```diff
@@ -0,0 +1,23 @@
...
+        cidr.is_a?(IPAddr) ? "#{cidr}/#{cidr_mask cidr}" : cidr.to_s
+      end
+
+      # returns the length of the netmask in bits
+      def cidr_mask cidr
+        mask = cidr.instance_variable_get(:@mask_addr).to_s(2)[/\A(1*)0*\z/, 1]
```
jonahx: Is Cidr an object you don't own?  Seems like this should just be a simple accessor call?

app/models/loader/handlers/password.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Loader
+  module Handlers
+    # extends the policy loader to store role credentials (passwords). This occurs
+    # after the initial policy load, since variable values are not part of the policy load
+    module Password
+  
+      # When a public key is encountered in a policy, it is saved here. It can't be written directly into
```
apotterri: This comment and the corresponding one in `PublicKey` both seem to have some problems: this comment mentions "public key", but the module is for loading passwords.

I know you just copied-and-pasted them from the original, but can you fix them, please?

app/models/loader/handlers/public_key.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Loader
+  module Handlers
+    # extends the policy loader to store public keys for users as variables. This happens
+    # after the initial policy load, since variable values are not part of the policy load
+    module PublicKey
+  
+      # When a public key is encountered in a policy, it is saved here. It can't be written directly into
```
apotterri: This module is for loading a pubic key, but it mentions the `credentials` table, which AFAIK, doesn't have anything to do with public keys.

app/models/loader/handlers/restricted_to.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Loader
+  module Handlers
+    # extends the policy loader to store the authentication network restriction
+    # on a role's credentials.
+    module RestrictedTo
+  
+      # When a cidr restriction is encountered in a policy, it is saved here. It can't be written directly
```
apotterri: Pretty sure this comment is right, though. ;)

cucumber/api/features/authn_restricted_to.feature
```diff
@@ -0,0 +1,22 @@
+Feature: User and Host authentication can be network restricted
+
+  Background:
+    Given I am the super-user
+    And I successfully PUT "/policies/cucumber/policy/root" with body:
```
jonahx: Add comments about alice IP is bogus, and the bob ones are local and CI ones

app/models/role.rb
```diff
@@ -67,11 +67,29 @@ def by_login login, account:
...
+
+  def valid_origin?(ip_addr)
+    ip = IPAddr.new(ip_addr)
+    restricted_to.blank? || restricted_to.any? do |cidr|
+      cidr.include?(ip)
+    end
+  end
+
+  def restricted_to
+    credentials = self.credentials ||= Credentials.new(role: self)
```
apotterri: I feel like the extra local variable here and in `#modify_credentials` takes away from the readability of the code. Maybe just use `self.credentials` instead?

cucumber/api/features/authn_restricted_to.feature
```diff
@@ -0,0 +1,22 @@
...
+    """
+    - !user
+      id: alice
+      restricted_to: 176.0.0.1
+
+    - !user
+      id: bob
+      restricted_to: [ "127.0.0.1", "172.0.0.0/8" ]
```
apotterri: How did you choose `172.0.0.0/8` here?

lib/util/cidr.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Util
+
+  # provides helper methods for interacting with CIDR network addresses 
+  module CIDR
```
jonahx: Something about this wasn't sitting right with me -- utility classes usually don't -- and I was kind of mulling it over in the back of my mind and trying to figure out why, and then it popped out at me... you have an object here begging to be born.  It's an upgraded version of ruby's `IPAddr` class -- with these additional methods.  So rather than working with plain Strings and sending them to this utility methods to do what you want, make it so you can `cidr = Cidr.new(ip_addr)` and now your client code can `cidr.to_s` (or `cidr.formatted_string`) and `cidr.mask`.

kgilpin:
Are we going to add CIDR for hosts created via host factory?

cucumber/api/features/step_definitions/request_steps.rb
micahlee: That makes sense, and I will refactor this.

lib/util/cidr.rb
micahlee: Sure, no problem!

lib/util/cidr.rb
micahlee: It is not, it is the `IPAddr` class, but I did some digging and I have some ideas for how to clean this up a little.

app/models/loader/handlers/public_key.rb
micahlee: Sure, thanks for catching that. I will clean up the comments.

cucumber/api/features/authn_restricted_to.feature
micahlee: I will be added a comment for this per Jonah's request. To summarize:
- `127.0.0.1` is to support running this test locally
- `172.0.0.0/8` is for running this test within docker-compose containers. The default docker bridge subnet is `172.17.0.0/16`. When docker-compose creates new networks, it changes the second octet, so `172.0.0.0/8` allows me to permit any connections in docker-compose environments for CI.

I had attempted to try and specify a more specific subnet in the docker-compose config for a more narrow restriction, but I was running into overlapping network issues and didn't have enough visibility into docker on the CI agents to dig into it further.

lib/util/cidr.rb
micahlee: That's fair. I'll dig into that some more. Thanks!

micahlee:
@kgilpin, that would be a natural next step. 

@jvanderhoof, do you have any thoughts? Did you intend for this to be included in this story? It seems like it would stand alone well enough to be its own story, but I'm willing to go either way.

To clarify, the hosts created by host factories are being given a restriction in the database with this changeset, it's just empty. The change or additional story would be to modify the host factory API to allow specifying a CIDR restriction for the created host.

app/models/host_factory_token.rb
```diff
@@ -44,6 +44,10 @@ def valid?
     !expired?
   end
 
+  def cidr
```
jonahx: should this be cidrs?  since it's a list of them?

lib/util/cidr.rb
```diff
@@ -1,23 +1,30 @@
...
-        mask.length
-      end
-      
+  class CIDR
+    extend Forwardable
+
+    def_delegators :@cidr, :include?
+
+    def initialize(ip_addr)
+      @cidr = ip_addr
```
jonahx: When decorating an object, which is what we're doing here, it's better to keep the terminology clear about the base/wrapped object and the wrapper object.  So I'd name this `@ip_addr` since the object we're initializing is what is, in fact the "cidr"

PR #618, opened 2018-07-10
Audit secret fetches failed due to lack of visibility

### LINKED ISSUE #606: Emit events for fetches failed due to lack of visibility
GIVEN a running conjur instance
WHEN you try to fetch a secret you can't see
THEN you get 404
BUT the audit event about it still gets recorded.

dividedmind:
Closes #606 

Note there's some refactoring in this pull request to support the final commit with the change. It's probably best to review the commits one-by-one.

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -3,8 +3,6 @@
 class AccountsController < ApplicationController
   include AuthorizeResource
 
-  before_action :find_or_create_accounts_resource
-
   def index
     authorize :read
     
@@ -29,8 +27,8 @@ class AccountsController < ApplicationController
 
   protected
 
-  def find_or_create_accounts_resource
-    @resource = Account.find_or_create_accounts_resource
+  def resource
+    @resource ||= Account.find_or_create_accounts_resource
   end
 
   def account_name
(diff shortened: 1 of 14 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -3,8 +3,6 @@
 class AccountsController < ApplicationController
   include AuthorizeResource
 
-  before_action :find_or_create_accounts_resource
```
micahlee: 💯

app/controllers/concerns/find_resource.rb
```diff
@@ -9,8 +9,19 @@ def resource_id
     [ params[:account], params[:kind], params[:identifier] ].join(":")
   end
 
-  def find_resource
-    @resource ||= Resource.find_if_visible current_user, resource_id: resource_id
-    raise Exceptions::RecordNotFound, resource_id unless @resource
+  def resource!
```
micahlee: What's the reason for the bang on this method name?

apotterri: Probably because it doesn't do any visibility checking.

dividedmind: Yup, I deliberated for a good couple minutes on the right name for the method, until I remembered the Ruby convention of having bang methods being the counterpart of non-bang, only more dangerous. Seemed fitting.

PR #622, opened 2018-07-11
Moving Linting/Static Analysis/Vulnerability Scanning to Code Climate

jvanderhoof:
Adds Rubocop scanning using our specific settings.

Moving to Code Climate provides better visibility for style, static analysis, and vulnerability scanning to GitHub PRs.

Code Climate has an unfortunate issue where it can't load files into `.rubocop.yml` using `inherit_from`.  To work around this, I've loaded the contents of our Rubocop Todo & our current conventions.


dividedmind:
Didn't you end up removing our settings?

dividedmind:
Also, does codeclimate just ignore includes, or does it break? Because if it just ignores them, maybe it'd make more sense to move settings to rubocop.yml and leave todos in another file. It's better if it does not respect the todos anyway, since it only comments on changed lines IIUC.

PR #637, opened 2018-07-21
WIP: Adds rotator for Gogs

ryanprior:

This pull request adds a rotator for Gogs user account passwords.

In [cdemo](https://github.com/conjurdemos/cdemo) we use Gogs to show off secrets flows driven by source control changes. ("gitops")

To better demonstrate best practices around user accounts, we want a rotator to allow Conjur to manage user account credentials automatically.

The new rotation code is in `app/domain/rotation/rotators/gogs/`.

1. start up a [Gogs server](https://hub.docker.com/r/gogs/gogs/)
2. create a user account called "test"
3. load a policy in Conjur like so:
   ```
    - !policy
     id: gogs-test
     body:
       - !variable
         id: password
         annotations:
           - gogs/url: http://localhost:8080/
           - gogs/user_name: test
           - rotation/rotator: gogs
           - rotation/ttl: PT1M
   ```
   *Substitute the actual URL of your local gogs server if different.*
4. load the initial password into Conjur: `conjur variable values add gogs-test/password secret`
*Substitute the actual password if not "secret"*
5. Observe that the password should rotate immediately, and thereafter each minute.
6. Fetch the value of the variable and check that you can actually use it to log in (`conjur variable value gogs-test/password`)

#### Further development

This rotator should be very nearly the same as would be needed to rotate passwords in [Gitea](https://gitea.io/en-us/), a popular fork.

In addition, this has code that's useful for any rotator where the pattern is basically "log in and use a web form with CSRF to change a password" as opposed to cases where a specific API is called for. Abstracting that out and documenting it to make it trivial to write this kind of authenticator would be nice.

No but it probably does need tests & documentation. Until it has those I'll consider it WIP.

Yeah, that would be really cool actually.

Not yet, but it should be.

### DIFF
--- a/app/domain/rotation/rotators/gogs/user_account.rb
+++ b/app/domain/rotation/rotators/gogs/user_account.rb
@@ -1,0 +1,72 @@
+return if defined? Rotation::Rotators::Gogs::UserAccount
+
+require 'rest-client'
+require 'securerandom'
+
+module Rotation
+  module Rotators
+    module Gogs
+      class UserAccount
+        def initialize()
+        end
+
+        def rotate(facade)
+          variable_id = facade.rotated_variable.resource_id
+          user_name = facade.annotations['gogs/user_name']
+          base_url = facade.annotations['gogs/url']
+          old_password = facade.current_values([variable_id])[variable_id]
+          new_password = new_password(base_url, user_name, old_password)
+          facade.update_variables({variable_id => new_password})
+        end
+
+        private
+
+        def csrf_token(url, cookies={})
+          response = RestClient.get(url, {cookies: cookies})
+          csrf_rx = /<input.+?name="_csrf".+?value="(.+?)"/
+          csrf_rx.match(response.body).captures.first
+        end
+
+        def auth_cookies(url, user_name, password, remember=false)
+          token = csrf_token url
+          begin
+            RestClient.post(url,
+                            {
+                              _csrf: token,
+                              user_name: user_name,
+                              password: password,
+                              remember: remember
+                            })
+          rescue RestClient::ExceptionWithResponse => e
+            e.response.cookies
+          end
+        end
+
+        def new_password(base_url,
+                         user_name,
+                         old_password,
+                         new_password=SecureRandom.base64(32))
+          cookies = auth_cookies("#{base_url}/user/login", user_name, old_password)
+          token = csrf_token "#{base_url}/user/settings/password", cookies
+          begin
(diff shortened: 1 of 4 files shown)


app/domain/rotation/rotators/gogs/user_account.rb
```diff
@@ -0,0 +1,72 @@
...
+                              password: new_password,
+                              retype: new_password
+                            },
+                            {cookies: cookies})
+          rescue RestClient::ExceptionWithResponse => e
+            case e.response.code
+            when 302
+              new_password
+            else
+              raise e
```
jonahx: i'd add a comment for clarity about what's happening here.  afaict, you are doing the post _assuming_ that it will 302 redirect you when successful?  and therefore that _all_ requests are assumed to raise an error of _some_ sort, but only "successful" ones will be 302?

Needs clarification either way.  Also, assuming that interpretation is correct, then it seems that you are not handling the "error" condition... ie, the case where the post unexpectedly succeeds!

also just a style preference but since case is verbose i'd prefer:
```
raise e unless password_update_succeeded?(e)
new_password
```
and then you define `password_update_succeeded?` to check for 302

app/domain/rotation/rotators/gogs/user_account.rb
```diff
@@ -0,0 +1,72 @@
...
+        def initialize()
+        end
+
+        def rotate(facade)
+          variable_id = facade.rotated_variable.resource_id
+          user_name = facade.annotations['gogs/user_name']
+          base_url = facade.annotations['gogs/url']
+          old_password = facade.current_values([variable_id])[variable_id]
+          new_password = new_password(base_url, user_name, old_password)
+          facade.update_variables({variable_id => new_password})
```
jonahx: My one major comment: You'll want to do something similar to this:

https://github.com/cyberark/conjur/blob/master/app/domain/rotation/rotators/postgresql/password.rb#L38

to ensure that the two updates (to conjur, and to gogs) occur as a unit.  Theoretically the way it currently is gogs could update and the conjur `update_variables` could fail and you'd be out of sync.

jvanderhoof:
Looks like this PR has been inactive for about 4 months.  I'm going to close this for now.  Please re-open if this is a priority.

ryanprior:
I never came up with a plan for how to test this, and it didn't seem appropriate to merge it without tests. At some point it'll become a priority and I will raise the question of testing so we can work something out.

PR #674, opened 2018-08-10
Improve large policy initial load performance

### LINKED ISSUE #669: Large policies load requests exceed the desired request response time
#### Description

Large policy load requests (e.g. 18,000 variables) take a long to complete (3-6 minutes). This exceeds the default timeout for puma request workers (1 min) and in general is a symptom of a another performance or architecture issue that should be improved. 

An example policy that causes this issue is available here: https://github.com/conjurinc/appliance/issues/263

#### Relevant Source
https://github.com/cyberark/conjur/blob/261cce419214fad4679d16be74520e8f654c7327/config/puma.rb#L7-L12

micahlee:
Closes #669 

This PR makes two changes to policy loading to make it more efficient:
1. Annotations on a resource are inserted in bulk when loading the policy file 
    (`app/models/loader/types.rb`)

2. INSERT policy log records are added in bulk, rather than using the trigger
    (`app/models/loader/orchestrate.rb`)

When loading large policies, the load time is largely dominated by the trigger processing for the policy change log. This was discovered when loading policies generated by the Vault Synchronizer.

To disable the triggers during the bulk load, I tried a couple of methods before landing on the current:

1. `ALTER TABLE table ENABLE/DISABLE TRIGGER {trigger name}`

    This didn't work because `ALTER TABLE` obtains an `ACCESS EXCLUSIVE LOCK`, which breaks concurrent policy loading.

2. `SET session_replication_role = replica;`
    
    This didn't work because it disables *ALL* triggers for the current session, and there are a number of other triggers that are necessary, and would become cumbersome to re-implement as bulk ops.

3. Configuration setting
    The method I settled on was to use a configuration setting, scoped to the session, that the trigger function is aware of. When we set this setting to bypass the trigger to `true`, then the trigger will observe the setting value and not create its own policy log.

Each of the  efficiency changes is in its own commit.

An example large policy can be found on the original issue. In my testing, it originally loaded in ~200 seconds, with these updates it loads in ~50 seconds.

https://jenkins.conjur.net/job/cyberark--conjur/job/policy_load_performance/

No


jvanderhoof:
Looks promising @micahlee!  There's a couple RSpec tests failing on parallel tests, which we'll need to figure out if we can find a non-blocking way to pull this off....

kgilpin:
Do you think we need a test to ensure that the audit records are identical?

app/models/loader/orchestrate.rb
```diff
@@ -255,13 +255,67 @@ def insert_new
...
+    end
+
+    def enable_policy_log_trigger
+      db.run 'SET myvars.skip_insert_policy_log_trigger = false'
+    end
+
+    def insert_policy_log_records(table)
+      primary_key_columns = Array(Sequel::Model(table).primary_key).map(&:to_s).pg_array
+      db.run <<-POLICY_LOG
+          INSERT INTO policy_log(
```
apotterri: I think it's important that we ensure that policy log entries generated here match those generated when `myvars.skip_insert_policy_log_trigger` is `true`.

Seems like there are two options: 

* DRY up the code, and have both paths call a function to do the `INSERT INTO ... SELECT`.
* Add an automated test that shows that entries generated both ways are identical.

I think the first option is preferable, but it's not immediately obvious to me that it's possible. @micahlee, can you investigate and see what you can come up with?

app/models/loader/orchestrate.rb
```diff
@@ -255,13 +255,67 @@ def insert_new
...
+      #     This didn't work because it disables *ALL* triggers for the
+      #     current session, and there are a number of other triggers that 
+      #     are necessary, and would become cumbersome to re-implement as bulk ops.
+      #
+      # 3. Configuration setting
+      #     The method I settled on was to use a configuration setting, scoped
+      #     to the session, that the trigger function is aware of. When we set
+      #     this setting to bypass the trigger to `true`, then the trigger will
+      #     observe the setting value and not create its own policy log.
+      db.run 'SET myvars.skip_insert_policy_log_trigger = true'
```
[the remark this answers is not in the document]
micahlee: Yes, `SET LOCAL` does work equivalently for this case. I will make the change.

app/models/loader/orchestrate.rb
```diff
@@ -255,13 +255,63 @@ def insert_new
...
+            r.kind,
+            r.subject
+          FROM
+            #{schema_name}.#{table},
+            policy_log_#{table}_record(
+              /* We need to convert the temp schema table type
+                 to the publish schema type for the same table.
+                 We have to use string serialization to accomplish
+                 this. */
+              #{table}::text::#{table},
```
micahlee: @apotterri, do you have any ideas on a better way to accomplish this?

The two types have the same column definitions, but I couldn't find a way to directly cast the temporary schema type to the public schema without marshaling through `text`.

This doesn't appear to impact the performance in a meaningful way, it just feels unpolished to me.

app/models/loader/orchestrate.rb
```diff
@@ -255,13 +255,63 @@ def insert_new
...
+            r.kind,
+            r.subject
+          FROM
+            #{schema_name}.#{table},
+            policy_log_#{table}_record(
+              /* We need to convert the temp schema table type
+                 to the public schema type for the same table.
+                 We have to use string serialization to accomplish
+                 this. */
+              #{table}::text::#{table},
```
micahlee: @apotterri, do you have any ideas on a better way to accomplish this?

The two types have the same column definitions, but I couldn't find a way to directly cast the temporary schema type to the public schema without marshaling through `text`.

This doesn't appear to impact the performance in a meaningful way, it just feels unpolished to me.

app/models/loader/orchestrate.rb
```diff
@@ -306,13 +292,23 @@ def insert_policy_log_records(table)
...
-          '#{table}'::policy_log_kind,
-          slice(hstore(#{table}), #{db.literal(primary_key_columns)})
+            r.policy_id,
+            r.version,
+            r.operation,
+            r.kind,
+            r.subject
           FROM
-          #{schema_name}.#{table}
+            #{schema_name}.#{table},
```
apotterri: This doesn't look right. Why are we joining the table with the rows returned from the function?

db/migrate/201808131137612_policy_log_trigger_bypass.rb
```diff
@@ -2,13 +2,48 @@
...
       # find the primary key of the table
       primary_key = schema(table).select{|x,s|s[:primary_key]}.map(&:first).map(&:to_s).pg_array
-      execute """
+
+      # Each table has a function format the policy record for that type.
+      # This function is used both by the policy_log trigger for each
+      # table, as well as for the bulk insert policy log in the policy
+      # load orchestrator.
+      execute <<-SQL
+        CREATE OR REPLACE FUNCTION policy_log_#{table}_record(
```
apotterri: Maybe, instead of creating a function for each table, we can have a single function that takes the table name as an argument? This SO answer seems like it describes a reasonable technique: https://stackoverflow.com/a/10711349 . (I did read the objection in the subsequent answer, and I'm ok with disregarding the advice in it.)

I don't really understand the other problem you asked about, but maybe this approach will get around it. It not, can you describe it in more detail (e.g. are the errors you get when don't marshal through `text`)?

db/migrate/201808131137612_policy_log_trigger_bypass.rb
```diff
@@ -0,0 +1,147 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  tables = %i(roles role_memberships resources permissions annotations)
+  
+  up do
+    # `public_log_record` is a custom type to record the column
```
kgilpin: public_log_record -> policy_log_record

apotterri:
It might be nice to clean up the commits here before merging, or I guess you could do a squash merge.

micahlee:
@apotterri, thanks! Yes, at this point I will plan on a squash merge.

Thanks @kgilpin, I've fixed that comment. Let me know if there is anything else you think I should address.

PR #686, opened 2018-08-16
tagging docker images with minor releases

### LINKED ISSUE #279: Conjur CE Images tagged to ensure an easy pull for minor releases
As a consumer of the Conjur CE image, I want an easy way to control which image pull from a registry.

For example, being able to pull cyberark/conjur:latest to get the latest code is great. But, sometimes I'd like to stay at a minor release, and only pull patches. Having a tag like `1.1-latest` applied would enable me to pick up any patches applied to version `1.1`.

Given a version `1.1.1` (from `VERSION` file), in the Jenkins pipeline, let's create and push three different Conjur tags: 
- `cyberark/conjur:1`
- `cyberark/conjur:1.1`
- `cyberark/conjur:1.1.1`

This will allow users to choose the desired version level.

orenbm:
Closes #279 

instead of having only one docker image per version (i.e 1.1.1-refid) we have one for each minor release (i.e 1-refid, 1.1-refid, 1.1.1-refid)

version-utils.sh -> build.sh -> ci/test -> push-image.sh

run build.sh & test.sh

https://jenkins.conjur.net/job/cyberark--conjur/job/279-minor-releases-tags/

### DIFF
--- a/push-image.sh
+++ b/push-image.sh
@@ -8,47 +8,49 @@
 # shellcheck disable=SC1091
 . version_utils.sh
 
-TAG="${1:-$(version_tag)}"
+TAGS="${1:-$(version_tags)}"
+IFS=',' read -ra TAGS_ARR <<< "$TAGS"
 
-SOURCE_IMAGE="conjur:$TAG"
 INTERNAL_IMAGE='registry.tld/conjur'
 INTERNAL_IMAGE_NEW='registry.tld/cyberark/conjur'  # We'll transition to this
 DOCKERHUB_IMAGE='cyberark/conjur'
 QUAY_IMAGE='quay.io/cyberark/conjur'
 
 function main() {
-  echo "TAG = $TAG"
+  for tag in "${TAGS_ARR[@]}"; do
+    echo "TAG = $tag"
 
-  tag_and_push $INTERNAL_IMAGE $TAG
-  tag_and_push $INTERNAL_IMAGE_NEW $TAG
+    tag_and_push $INTERNAL_IMAGE $tag
+    tag_and_push $INTERNAL_IMAGE_NEW $tag
 
-  if [ "$BRANCH_NAME" = "master" ]; then
-    local latest_tag='latest'
-    local stable_tag="$(< VERSION)-stable"
+    if [ "$BRANCH_NAME" = "master" ]; then
+      local latest_tag='latest'
+      local stable_tag="$(< VERSION)-stable"
 
-    echo "TAG = $stable_tag, stable image"
+      echo "tag = $stable_tag, stable image"
 
-    tag_and_push $INTERNAL_IMAGE $latest_tag
-    tag_and_push $INTERNAL_IMAGE $stable_tag
+      tag_and_push $INTERNAL_IMAGE $latest_tag
+      tag_and_push $INTERNAL_IMAGE $stable_tag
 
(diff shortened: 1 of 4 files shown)


jvanderhoof:
@orenbm: this is great! I think the only place we really need to tag images is in the `push-image.sh` file. I have a PR (#687) that only publishes a Docker image with a version label when the master branch is tagged with the version tag. This means tagging the master branch with `v1.1.1` causes Jenkins to publish the image `cyberark/conjur:1.1.1` to DockerHub.

Can we tie into that release architecture?  That would augment the above to additionally release:
- `cyberark/conjur:1.1`
- `cyberark/conjur:1`

push-image.sh
```diff
@@ -8,47 +8,49 @@
...
+      tag_and_push $DOCKERHUB_IMAGE $latest_tag
+      tag_and_push $DOCKERHUB_IMAGE $stable_tag
 
-    tag_and_push $QUAY_IMAGE $TAG
-    tag_and_push $QUAY_IMAGE $latest_tag
-    tag_and_push $QUAY_IMAGE $stable_tag
-  fi
+      tag_and_push $QUAY_IMAGE $tag
+      tag_and_push $QUAY_IMAGE $latest_tag
+      tag_and_push $QUAY_IMAGE $stable_tag
```
jvanderhoof: What about just limiting the image tagging to here?  We only want images published once tests are passed.  What do you think about something like:
```sh
...
VERSION=$(< VERSION)
minor_regex="^([0-9]+\.[0-9]+)"
if [[ "$VERSION" =~ $minor_regex ]]; then
  minor_version="${BASH_REMATCH[1]}"
  tag_and_push $INTERNAL_IMAGE $minor_version
  tag_and_push $INTERNAL_IMAGE_NEW $minor_version
fi

major_regex="^([0-9]+)"
if [[ "$VERSION" =~ $major_regex ]]; then
  major_version="${BASH_REMATCH[1]}"
  tag_and_push $INTERNAL_IMAGE $major_version
  tag_and_push $INTERNAL_IMAGE_NEW $major_version
fi
...
```

dividedmind:
I like this approach more than https://github.com/cyberark/conjur/pull/687 -- not only because the latter breaks builds ;) but also because this tags all the major/minor/patch version combinations.

The bash seems hard to follow however, how about something like this instead?
```bash
gen_versions()
{
  local $version=$1
  while [[ $version == *.* ]]; do
    echo $version
    version=${version%.*}
  done
}

versions="`gen_versions $VERSION`"

for v in $versions; do tag_and_push_version $source $target $v; done
```

orenbm:
Yes, as i understand the US we need to tag 3 images per build and in @jvanderhoof 's code we tag only one. 

@dividedmind - does the code you wrote replace my changes in `version_utils.sh`? we can add that as a function there. I must admit that i find my code more readable but I'm not a bash expert.

jvanderhoof:
I like the conciseness of @dividedmind's suggestion.  The function returns the combination:
```sh
gen_versions()
{
  local version=$1
  while true; do
    echo $version
    [[ $version = *.* ]] || return
    version=${version%.*}
  done
}

VERSION='5.3.2'
versions="`gen_versions ${VERSION%-*}`"
for v in $versions; do
  echo $v
done
```
prints: 
```
5.3.2
5.3
5
```

jvanderhoof:
I'm going to close this PR until we get a chance to circle back around and address it.

Issue #651, opened 2018-07-26
Alpha - Conjur signs certificates using a configured CA


jvanderhoof:
## Status
- **Completion ETA**: Alpha Version Complete (August 31, 2018)
- **Confidence**: N/A

## Context

PKI, or Public Key Infrastructure is a mechanism to enable organizations to authenticate and secure communication between nodes in a network. Conjur will operate as a Certificate Authority, signing and managing certificates for hosts managed by Conjur.

Most PKI systems do not lend themselves to automation. The primary objective of this initiative is to enable organizations to leverage PKI practices at scale, in high automation environments.

## Value
What is the benefit of Conjur acting like a CA? 
- Allows an organization to leverage mutual TLS to provide application to application authentication using tools like Nginx and Envoy. 
- Eliminates the need for engineers to create application-to-application authentication systems, allowing them to focus on shipping features. 
- Provides application to application connection transparency (through mutual TLS), which allows a security team to track how applications connect to each other.

## Feature Demo

To demonstrate Conjur acting as a certificate authority, we'll demonstrate the following scenarios:
- Issue a request (using Curl) to a server protected with mutual TLS.
- Connect to a simple website using a web browser configured with mTLS (potentially)

## Setup

A best practice is to utilize an intermediate certificate to performing signing.  This allows the root certificate to be kept securely, preferably in an air gapped environment. We will need to include the intermediate private key

Policy:
```yml
- !policy 
  id: conjur/<service-id>/ca
  body:
    # Signed certificates will be valid for up to a year
    - !webservice
        annotations:
          ca/private-key: ops/ca/private-key
          ca/private-key-password: ops/ca/private-key-password # If the import PEM key is encrypted
          ca/certificate-chain: ops/ca/private-key-chain
          ca/max_ttl: P1M

    - !group clients

    # Allow hosts in the `clients` group to be signed 
    - !permit
      role: !group clients
      privilege: [ sign ]
      resource: !webservice
```

Multiple certificate authorities can be configured by creating multiple CA services. This approach further simplifies the process of replacing intermediate certificate (due to breach or expiration). A new CA can be created, then remote nodes refresh their certificates using the new CA. 

Multiple CAs can also be utilized to separate departments or divisions within an organization. Each certificate chain needs to include the appropriate certificates. We leave this as an exercise for the user.

## Enable Host Signing
With our CA setup, let's look at a policy to enable application servers to create signed certificates:
```yml
- !policy 
  id: my-app
  body:
    - !layer
    
[...]

PR #689, opened 2018-08-16
Conjur CA Services

### LINKED ISSUE #684: PKI host certificates are issued through the Conjur API
I, as a Conjur user, want to sign my host certificates using the Conjur API, so that I can automate the certificate lifecycle for my hosts.

#### Task Breakdown

- [x] Add routes for CA service endpoint
- [x] Add controller for CA services
- [x] Add CI for testing CSR signing for hosts

- [x] Factor CA logic out of controller
- [x] Allow serve to return JSON or PEM based on `Accept` header
- [x] Use POST instead of PUT

- [x] Update response codes to match Feature Experience Document
- [x] Add developer readme
- [x] Address remaining code climate issues

micahlee:
Closes #684 
Blocked by #708

This PR add support for certificate authority (CA) services to be configured using Conjur policy that are capable of signing host CSRs using a provided CA certificate.

This is in support of the epic #651.

The best place to start to understand how the feature works is with the
cucumber tests in `cucumber/api/features/certificate_authority.feature`.

After that, the `CertificateAuthorityController` is a good entrypoint into the code changes.

This can be manually tested using cucumber:
* `cd dev && ./start` and `conjurctl server` in the container
* In a new bash session, `cd dev && ./cli exec` and in the container:
    ```sh-session
    bundle exec cucumber --profile api cucumber/api/features/certificate_authority.feature
    ```

Also, a full demo of this capability is available here:
https://github.com/conjurdemos/misc-util/tree/master/demos/certificate-authority/mutual-tls

https://jenkins.conjur.net/job/cyberark--conjur/job/ca_signing_service/

Yes

Probably

Not yet. This is being tracked in #688

This PR contains a developer README. API and end-user docs will be separate stories.


config/routes.rb
```diff
@@ -60,6 +60,8 @@ def matches?(request)
       post    "/policies/:account/:kind/*identifier" => 'policies#post'
 
       get     "/public_keys/:account/:kind/*identifier" => 'public_keys#show'
+
+      put     "/ca/:account/:service_id/hosts/*identifier" => 'certificate_authority#sign_host'
```
kgilpin: PUT? I would expect a POST.

app/controllers/certificate_authority_controller.rb
```diff
@@ -0,0 +1,143 @@
...
+    raise RecordNotFound.new("No Host #{host_id}") unless host
+
+    raise Forbidden.new unless current_user.allowed_to?('sign', ca_resource)
+    raise Forbidden.new('Requestor is not CSR host') unless requestor_is_host?
+    raise Forbidden.new('CSR cannot be verified') unless csr.verify(csr.public_key)
+    raise Forbidden.new('CSR CN does not match host') unless host_name_matches?(csr)
+    
+    certificate = sign_csr
+
+    render json: {
```
kgilpin: Why are we returning JSON instead of just the certificate?

jvanderhoof:
My initial thinking around the JSON response was to send back relevant information, but looking that information, it doesn't make much sense:
- `created_at` - not important as expiration date is what determines validitity
- `id` - required by the host to generate the signing request, so duplicate
- `expires_at` - this can be gleaned from the resulting CRT
- `crt` - actually required :)

So in summary, let's update the epic to include: 
- `PUT` to `POST`
- CRT is returned for a successful signing request

app/controllers/certificate_authority_controller.rb
```diff
@@ -0,0 +1,143 @@
+# frozen_string_literal: true
+
+class CertificateAuthorityController < RestController
+
+  def sign_host
```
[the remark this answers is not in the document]
jvanderhoof: Let's update Rubocop to allow a bit more complexity in methods (unless the change in request from JSON to CRT negates this).

kgilpin:
It’s also worth comparing to
https://www.vaultproject.io/api/secret/pki/index.html#sign-certificate

app/domain/ca/certificate_authority.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+
+    protected
+
+    def private_key
+      @private_key ||= OpenSSL::PKey::RSA.new secret(private_key_var)
+    end
+
+    def private_key_password
+      @private_key_password ||= secret(private_key_password_var)
```
micahlee: This PR includes this method to read a potentially supplied `private_key_password`, unlocking the key is a follow-on story. This is tracked on the Feature Experience Document.

jonahx:
While I didn't go over it in full detail, this code looks very good to me.

Unfortunately, as I feared, we duplicated a lot of each other's work.  I even wrote mine as library code, put it in a util folder, and was planning to make it into a gem so we could easily re-use it across projects.  If you want to see the relevant files, they're here:

https://github.com/cyberark/conjur/tree/k8s-refactor-642/app/domain/util/open_ssl

(some of them have unit tests as well)

Probably makes sense to add a refactor, if we want, as a followup project, since it would be a largish change and you're already ready to go as is.

My only comment for the future is "beware of putting logic in rails controllers."  Especially library code of any kind.  Controllers lock up the code and make it difficult to re-use and unit test.  The controller's job should ideally be "border" related duties.  Unpacking parameters, converting domain errors to http status codes and responses, etc.  Out of the box rails really encourages mixing of concerns in controllers, which isn't a problem when the logic is simple and unlikely to be re-used, but is a problem for generic code you do want to re-use.

This problem of re-use through libraries, and more importantly discovery (how could you have known relevant code was already available?  how can we make things like this searchable?  the unlucky case like this where it wasn't even merged yet is less interesting, but the general challenge is important) is something I've been thinking about a lot, and have discussed a bit with @jvanderhoof.  

I think we should start making it a priority to solve.

PR #726, opened 2018-09-04
Implement role membership search

micahlee:
#### What does this PR do?
This PR implements the keyword search of role memberships in the Conjur API. 

The Conjur UI supports searching the membership for a given role by keyword. This was not currently implemented in the open source / V5 API.

Closes https://github.com/conjurinc/conjur-ui/issues/200

The implementation is added in `app/controllers/roles_controller.rb`. The updates to the `Role` Sequel model are in `app/models/role.rb` with a refactor of the `MembershipSearch` module to support reuse between `membership` and `member` search.

The most simple way to test this is through the new cucumber test in the API profile for searching memberships.

```
bundle exec cucumber --profile api cucumber/api/features/memberships.feature:85
```

https://jenkins.conjur.net/job/cyberark--conjur/job/membership_search/

Yes

Yes

No

N/A

No

### DIFF
--- a/app/models/membership_search.rb
+++ b/app/models/membership_search.rb
@@ -3,68 +3,59 @@
 # MembershipSearch provides the `search`` extension method
 # for Role membership datasets to allow text searching
 # and paging of Role members
-class MembershipSearch < Module
+module MembershipSearch 
 
-  def initialize(id_column)
-    super() do 
-      define_search
-      define_filter_kind(id_column)
-      define_textsearch(id_column)
-      define_result_set(id_column)
-    end
+  # @param search [String] - a search term in the resource id
+  def search(search: nil, kind: nil)
+    filter_kind(kind).textsearch(search)
   end
 
-  def define_search
-    # @param search [String] - a search term in the resource id
-    define_method :search do |search: nil, kind: nil|
-      filter_kind(kind).textsearch(search)
-    end
+  def filter_kind(kind)
+    return self unless kind
+
+    kind_function = Sequel.function(:kind, search_key)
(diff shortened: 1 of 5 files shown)


app/models/membership_search.rb
```diff
@@ -1,51 +1,70 @@
...
 # and paging of Role members
-module MembershipSearch
-  # @param search [String] - a search term in the resource id
-  def search(search: nil, kind: nil)
-    filter_kind(kind).textsearch(search)
-  end
-
-  def filter_kind(kind)
-    return self unless kind
+class MembershipSearch < Module
```
apotterri: `class` already inherits from `Module`, so it's unnecessary (and kind of confusing) to say it again.

app/models/membership_search.rb
```diff
@@ -1,51 +1,70 @@
...
 
-    where(Sequel.lit("kind(member_id) in ?", Array(kind))) 
+  def initialize(id_column)
```
apotterri: I don't think you need to jump through the hoops of defining new methods here. The methods in the Module you pass to `extend:` have access to information about the association, as described here: https://guides.rubyonrails.org/v4.2/association_basics.html#association-extensions .

Also, the method has been renamed `extending` in Rails 4 and later, so you should probably update the associations to use the new name.

micahlee:
Hi @apotterri, thanks again for helping me come up with a cleaner solution for this. I really like the way the outcome turned out. 

This is ready for you to take another look when you are ready!

apotterri:
Looks good now, thanks for making these changes.

PR #731, opened 2018-09-10
729  refactor kubectl exec class

### LINKED ISSUE #729: Refactor kubectl exec class
The kubectl_exec class was refactored as part of the k8s refactor effort but it has been exceedingly difficult to write unit tests for it. We have decided to continue allowing to be tested as a side effect of the cukes.

To make sure it doesn't hinder our efforts to fix the cukes when merging the larger refactor branch, I think it's worth testing and merging the kubectl exec changes separately.

jtuttle:
#### What does this PR do?

Refactors the kubectl_exec class in isolation before the full refactor merge.

Closes #729 

Cukes should suffice

https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--conjur/activity?branch=729--refactor-kubectl-exec-class

Not yet.

### DIFF
--- a/app/domain/authentication/authn_k8s/kubectl_exec.rb
+++ b/app/domain/authentication/authn_k8s/kubectl_exec.rb
@@ -265,7 +265,7 @@ module Authentication
 
       def server_url(cmds, stdin)
         api_uri = @kubeclient.api_endpoint
-        base_url = "ws://#{api_uri.host}:#{api_uri.port}"
+        base_url = "wss://#{api_uri.host}:#{api_uri.port}"
         path = "/api/v1/namespaces/#{@pod_namespace}/pods/#{@pod_name}/exec"
         query = query_string(cmds, stdin)
         "#{base_url}#{path}?#{query}"
--- a/app/domain/authentication/authn_k8s/kubectl_exec.rb
+++ b/app/domain/authentication/authn_k8s/kubectl_exec.rb
@@ -116,10 +116,10 @@ module Authentication
       def add_websocket_event_handlers(ws, body, stdin)
         this_obj = self
         
-        ws.on(:open, ->() { this_obj.on_open(ws, body, stdin) })
-        ws.on(:message, ->() { this_obj.on_message })
-        ws.on(:close, ->() { this_obj.on_close })
-        ws.on(:error, ->() { this_obj.on_error })
+        ws.on(:open,    &->() { this_obj.on_open(ws, body, stdin) })
+        ws.on(:message, &->() { this_obj.on_message })
+        ws.on(:close,   &->() { this_obj.on_close })
+        ws.on(:error,   &->() { this_obj.on_error })
       end
       
       def on_open(ws, body, stdin)
--- a/app/domain/authentication/authn_k8s/kubectl_exec.rb
+++ b/app/domain/authentication/authn_k8s/kubectl_exec.rb
@@ -117,7 +117,7 @@ module Authentication
         else
           @logger.debug("Pod #{@pod_name} : channel open")
 
-          if @stdin
+          if stdin
             data = WebSocketMessage.channel_byte('stdin') + body
             ws.send(data)
             ws.send(nil, type: :close)
--- a/app/domain/authentication/authn_k8s/kubectl_exec.rb
+++ b/app/domain/authentication/authn_k8s/kubectl_exec.rb
@@ -39,12 +39,15 @@ module Authentication
         @messages = Hash.new { |hash,key| hash[key] = [] }
       end
 
-      # This is taking either a string or message DATA
-      def save_message(msg, stream: nil)
+      def save_message(msg)
         puts "* SAVING MESSAGE: #{msg}"
         strm ||= stream_name(msg)
         raise "Unexpected channel: #{channel(msg)}" unless strm
         @messages[strm.to_sym] << msg
+      end
+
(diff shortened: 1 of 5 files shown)


app/domain/authentication/authn_k8s/kubectl_exec.rb
```diff
@@ -1,133 +1,214 @@
...
+
+    class Messages
+      attr_reader :messages
+      
+      def initialize
+        @messages = Hash.new { |hash,key| hash[key] = [] }
+      end
+
+      def save_message(msg)
+        strm ||= stream_name(msg)
```
jonahx: Coming back to this code I wrote, I found the "channel" and "stream" names confusing and not immediately clear.  I looked at the docs around websockets and event machine, and didn't find any mention of "stream" in the sense i've used it here.  I think we should change to "channel_name" and "channel_number" terminology, unless there is something even more correct.  but those at names at least make clear the essential distinction in the context of this code.

app/domain/authentication/authn_k8s/kubectl_exec.rb
```diff
@@ -1,133 +1,214 @@
...
+        raise "Unexpected channel: #{channel(msg)}" unless strm
+        @messages[strm.to_sym] << msg_data(msg)
+      end
 
-      attr_reader :pod, :container, :timeout, :ws
+      def save_string(str, stream: nil)
```
jonahx: Should nil be the default here?  It doesn't seem to make any sense to save messages to the nil channel....

Also, this should be named clearly, perhaps (using the channel terminology suggested above):  `save_into_channel(msg)`

app/domain/authentication/authn_k8s/kubectl_exec.rb
```diff
@@ -1,133 +1,214 @@
...
+          body: generate_file_tar_string(path, content, mode)
+        )
+      end
 
-          def message_close msg
-            Rails.logger.debug "Pod #{@pod.metadata.name.inspect} message :#{close} : #{msg.data[1..-1]}"
+      private
+
+      def add_websocket_event_handlers(ws, body, stdin)
+        # These callbacks have access to local variables, but we can't use the
```
jonahx: Ok I had an epiphany about this.  We can get rid of these local methods, _and_ write each callback as its own KubectlExec method, to avoid this extra long method.  POC here:

https://eval.in/1056138

app/domain/authentication/authn_k8s/kubectl_exec.rb
```diff
@@ -1,133 +1,214 @@
...
+          stream_state.close
+          logger.debug("Pod #{pod_name} error : #{e.inspect}")
+          
+          messages.save_string(e.inspect, stream: :error)
         end
+      end
 
-        ws.send "\n"
-
+      def wait_for_close_message(ws)
```
jonahx: Can remove `ws` arg

app/domain/authentication/authn_k8s/kubectl_exec.rb
```diff
@@ -1,133 +1,214 @@
...
+      def server_url(cmds, stdin)
+        api_uri = @kubeclient.api_endpoint
+        base_url = "wss://#{api_uri.host}:#{api_uri.port}"
+        path = "/api/v1/namespaces/#{@pod_namespace}/pods/#{@pod_name}/exec"
+        query = query_string(cmds, stdin)
+        "#{base_url}#{path}?#{query}"
+      end
 
-      def generate_file_tar_string path, content, mode
+      def generate_file_tar_string(path, content, mode)
```
jonahx: We should fix this name, it's a carry over from the old code and doesn't really make sense...

app/domain/authentication/authn_k8s/kubectl_exec.rb
```diff
@@ -1,133 +1,214 @@
...
 require 'rubygems/package'
 
+require 'active_support/time'
+require 'websocket-client-simple'
+
+module WebSocket
+  module Client
+    module Simple
+      class Client
+        alias_method :send_msg, :send
```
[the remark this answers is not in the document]
jtuttle: Ah that's true. I still hate that they overrode the `:send` method though.

app/domain/authentication/authn_k8s/kubectl_exec.rb
jtuttle: Yeah this was super confusing when I was trying to find that bug. Sounds good to me.

app/domain/authentication/authn_k8s/kubectl_exec.rb
jtuttle: Since we're only using this for errors right now, I'm going to make it `save_error_string` and just have it use the error channel.

app/domain/authentication/authn_k8s/kubectl_exec.rb
jtuttle: Looks great, will work it in.

app/domain/authentication/authn_k8s/kubectl_exec.rb
jtuttle: afaik this is still doing exactly what it says. What would you call it?

jtuttle: Changing to `tar_file_as_string` from Slack convo.

PR #732, opened 2018-09-11
Use Authenticators to implement both /login and /authenticate

micahlee:
#### What does this PR do?

This PR updates authentication to implement both `/login` and `/authenticate` with authenticators. It also updates the LDAP authenticator to support `/authenticate` with LDAP password (the original, now deprecated, behavior) or with the Conjur API key return by LDAP authentication `/login`

There are guiding principals for these Conjur authentication changes:
1) People users should `/login` with a password and services should `/authenticate` with a key or token.
2) `/login` should return with a token that is accepted by `/authenticate` for the same authenticator
3) Authenticators generally MUST implement `/authenticate`, but may implement `/login`
4) To be used by the Conjur UI, an authenticator MUST implement `/login`

Closes https://github.com/conjurinc/conjur-ui/issues/204

Reading the updated `AUTHENTICATORS.md` readme then walking through the commits is probably the best way to understand the intention behind this PR and the steps I took to implement it. The commits have been cleaned up into as discreet of steps as I could make them.

These changes have automated tests in the CI pipeline.

If you wanted to manually test this, the easiest way might be to run the Conjur UI from this branch https://github.com/conjurinc/conjur-ui/compare/mlee-authn-ldap. This branch expects to be able to login using LDAP, and the user `alice` with password `alice` should work.

https://jenkins.conjur.net/job/cyberark--conjur/job/authenticator_backed_login/

Yes

Yes

I'm not sure.

The developer README is updated

#### TODO:

**Coding**
- [x] Refine design for authenticators to implement `/login` vs `/authenticate` (e.g. method names, fallback mechanisms, etc.)
- [x] Move primary LDAP logic to `/login` rather than `/authenticate`. `/authenticate` should attempt LDAP but fall back to the default authenticate (with API key).
- [x] Resolve relevant code climate issues
- [x] Clean up history to make it easier to review
- [x] Address PR review comments

**Testing**
- [x] LDAP unit tests should use login rather than authenticate to test password and admin behavior
- [x] Add integration tests for LDAP login and authenticating with returned token

**Documentation**
- [x] Update authenticators documentation
- [x] Update CHANGELOG
- [x] Consider documentation update needs for https://docs.conjur.org/Latest/en/Content/Operations/Services/Authentication-new.htm


app/controllers/concerns/basic_authenticator.rb
```diff
@@ -2,19 +2,52 @@
...
-        authentication.authenticated_role = credentials.role
+      authenticator_login(username, password).tap do |response|
+        authentication.authenticated_role = ::Role[response.role_id]
         authentication.basic_user = true
       end
+    rescue ::Authentication::Strategy::InvalidCredentials, 
+           ::Authentication::Strategy::InvalidOrigin,
+           ::Authentication::Security::NotAuthorizedInConjur
+      raise ApplicationController::Unauthorized
     end if request.authorization =~ /^Basic / # we need to check the auth method.
```
jtuttle: Instead of sticking this conditional at the end of the block where it's kind of hidden, could we treat it as an early return above the block? i.e.:

```
return if request.authorization !~ /^Basic /
```

app/domain/authentication/login_result.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'dry-struct'
+require 'types'
+
+module Authentication  
+  # Response from an authenticator login method
+  class LoginResponse < ::Dry::Struct
```
jtuttle: There's a mismatch between the class and file names here. `login_result` vs `LoginResponse`.

app/domain/authentication/strategy.rb
```diff
@@ -77,6 +79,24 @@ def conjur_token(input)
...
+    def login(input)
+      authenticator = authenticators[input.authenticator_name]
+
+      validate_authenticator_exists(input, authenticator)
+      validate_security(input)
+
+      key = authenticator.login(input)      
+      raise InvalidCredentials unless key
+
+      # TODO: Should be audit_login_success
```
jtuttle: What's this `TODO` for? Something worth fixing in this branch?

app/domain/authentication/strategy.rb
```diff
@@ -77,6 +79,24 @@ def conjur_token(input)
...
+      audit_success(input)
+      new_login(input, key)
+    rescue => err
+      # TODO: Should be audit_login_failure
```
jtuttle: Another `TODO` worth fixing in this branch?

docs/AUTHENTICATORS.md
```diff
@@ -93,6 +127,26 @@ module Authentication
         # initialization code based on ENV config
       end
 
+      # OPTIONAL
```
jtuttle: The `valid?` method had this commented info inside the method definition. Might be nice to match that (or change `valid?` to have the comments above the methods definition) for consistency.

spec/controllers/authenticate_controller_spec.rb
```diff
@@ -6,6 +6,45 @@
...
   let(:login) { "u-#{random_hex}" }
   let(:account) { "rspec" }
+  let(:authenticator) { "authn" }
+
+  context "#login" do
+    shared_examples_for "successful authentication" do
+      it "succeeds" do
+        post :login, params
+        expect(response).to be_ok
+        expect(response.body.length).to be >= 44
```
jtuttle: Could we possible seed the response values with some clever use of `allow` and then check the exactly response value rather than a length here? You can pass an expected value into the shared example I believe.

spec/controllers/authenticate_controller_spec.rb
```diff
@@ -6,6 +6,45 @@
   let(:password) { "password" }
   let(:login) { "u-#{random_hex}" }
   let(:account) { "rspec" }
+  let(:authenticator) { "authn" }
+
+  context "#login" do
+    shared_examples_for "successful authentication" do
```
jtuttle: Maybe call this `successful login`?

spec/controllers/authenticate_controller_spec.rb
```diff
@@ -6,6 +6,45 @@
...
+      it "succeeds" do
+        post :login, params
+        expect(response).to be_ok
+        expect(response.body.length).to be >= 44
+      end
+    end
+    
+    shared_examples_for "authentication denied" do
```
jtuttle: Maybe call this `login denied`?

spec/controllers/authenticate_controller_spec.rb
```diff
@@ -6,6 +6,45 @@
...
+      it "is unauthorized" do
+        post :login, account: account, authenticator: authenticator
+        expect(response.code).to eq("401")
+      end
+    end
+    
+    context "without auth" do
+      it_should_behave_like "authentication denied"
+    end
+    context "when user doesn't exist" do
```
jtuttle: Mind sticking some empty lines into this block of text? At least above each `context`?

config/routes.rb
```diff
@@ -20,13 +20,12 @@ def matches?(request)
     end
 
     constraints account: /[^\/\?]+/ do
-      get  '/authn/:account/login' => 'credentials#login'
```
jtuttle: Does this mean we're going to have to update our Conjur test files across various project to define the `authn` authenticator in the `CONJUR_AUTHENTICATORS` env var if we want to continue using the CLI to authenticate when another authenticator is present?

For example, in the k8s deploy scripts, we define `authn-k8s` but currently that doesn't prevent authentication using the Conjur CLI from working. If I'm understanding this correctly, we would now need to define `authn-k8s` AND `authn` to allow authentication both through the k8s authenticator and CLI?

jtuttle: Also, what was the impetus for this change? Is there a reason we've decided to move things around?

app/controllers/concerns/basic_authenticator.rb
micahlee: Will do!

app/domain/authentication/login_result.rb
micahlee: Thanks, I will fix this.

app/domain/authentication/strategy.rb
micahlee: No, but I will clean up the todo comments.

spec/controllers/authenticate_controller_spec.rb
micahlee: Good idea, I'll take a look into that.

config/routes.rb
micahlee: The immediate cause for these changes is to support allowing users to login to the UI using LDAP credentials. We explored a few different options for this, but most of them had one design smell or another. In the end we settled on a principal that resonates well with the existing design: Humans login, services authenticate.

So what this intends to lay out is a way in which authenticators may support either use case, where the overall flow (`User` -> `Login` -> `Authenticate`, or `Host` -> `Authenticate`) doesn't change, and token/key between login and authenticate is consistent enough that the UI and other end-user tools can support it without a special case for each authenticator.

You are correct that this PR would currently require you to include `authn` in the whitelisted authenticators if you also use another. I would be on board with including it behind the scenes with the enabled login authenticators to make this more fully backwards compatible.

app/controllers/concerns/basic_authenticator.rb
```diff
@@ -2,19 +2,52 @@
...
-      credentials = Credentials[Role.roleid_from_username(account, username)]
-      if credentials && credentials.authenticate(password)
-        authentication.authenticated_role = credentials.role
+      authenticator_login(username, password).tap do |response|
+        authentication.authenticated_role = ::Role[response.role_id]
         authentication.basic_user = true
       end
+    rescue ::Authentication::Strategy::InvalidCredentials, 
+           ::Authentication::Strategy::InvalidOrigin,
+           ::Authentication::Security::NotAuthorizedInConjur
```
micahlee: Per Jonah: `InvalidOrigin` and `NotAuthorized` could return a more useful message with the `Unathorized` response

config/routes.rb
jtuttle: Okay, I think it's probably better to require users to explicitly include `authn` to keep the functionality because otherwise we would have to treat disabling `authn` as a special case, which sounds annoying.

The only trouble is that i's going to become important to communicate in docs that you need to include `authn` in `CONJUR_AUTHENTICATORS` in order to continue using the CLI w/ Conjur. Otherwise people might get a bit confused when they try to enable the k8s authenticator and can no longer use the CLI to load policy or something.

Any thoughts on this @jonahx ?

config/routes.rb
```diff
@@ -20,13 +20,12 @@ def matches?(request)
...
-      put  '/authn/:account/password' => 'credentials#update_password'
-      put  '/authn/:account/api_key'  => 'credentials#rotate_api_key'
```
micahlee: From John in slack:
> why are we moving `update_password` and `rotate_api_key`? is it because you can perform those actions through the UI and we also want to be able to use LDAP creds?

https://conjurhq.slack.com/archives/CA8AT283D/p1537368039000100

micahlee: >Those only require the login "level" of authentication to perform. But with this, login isn't its own special case anymore, so they need to be authenticator aware. These also don't make sense outside of built-in authenticator though, so I do need to revisit that again.

app/controllers/concerns/basic_authenticator.rb
```diff
@@ -2,19 +2,52 @@
 
 module BasicAuthenticator
```
micahlee: Per Jonah: Consider moving this into the authenticator controller directly

config/routes.rb
micahlee: To summarize our discussion in slack here: https://conjurhq.slack.com/archives/CA8AT283D/p1537367685000100

This would already be an issue when logging in with the CLI when a non-default authenticator is configured. For the built-in Conjur login flow to work, `authn` would need to be present in the enabled authenticators (`CONJUR_AUTHENTICATORS` env variable).

app/controllers/concerns/basic_authenticator.rb
micahlee: @jonahx, this is currently used by both the `CredentialsController` and the `AuthenticationController`, so I'm inclined to leave it as a concern. Let me know if you have a better design idea though!

micahlee:
@jtuttle and @jonahx, I believe I have now addressed all of the issues we've discussed for this PR. Let me know when you get a chance to take another look, and if you have any suggestions to further improve it!

I tried to make the updates in discreet commits to make it easier to review what has changed from before, beginning with commit 8199094.

Thanks!

app/domain/authentication/authn_ldap/authenticator.rb
```diff
@@ -7,10 +7,14 @@
 module Authentication
   module AuthnLdap
 
-    class Authenticator
+    class Authenticator < ::Dry::Struct
+      attr_reader :role_cls
+      attr_reader :credentials_cls
```
jtuttle: What's using these `attr_reader`s? Or is this just something you have to specify for `::Dry::Struct`? If not, I would omit them and refer to the variables using `@` notation within the class.

cucumber/authenticators/features/support/authenticator_helpers.rb
```diff
@@ -8,12 +8,18 @@ module AuthenticatorHelpers
...
+  attr_reader :response_body, :http_status, :rest_client_error, :ldap_auth_key
 
-  def authenticate_with_ldap(service_id:, account:, username:, password:)
+  def login_with_ldap(service_id:, account:, username:, password:)
+    path = "#{conjur_hostname}/authn-ldap/#{service_id}/#{account}/login"
+    get(path, user: username, password: password)
+    @ldap_auth_key=response_body
+  end
+
+  def authenticate_with_ldap(service_id:, account:, username:, key:)
```
jtuttle: Can we rename `key` to `api_key` or is that too specific?

app/domain/authentication/authn_ldap/authenticator.rb
jonahx: @jtuttle You need them to make the object constructor.  As I explained to Micah (and may have forgotten to explain to you), my use of `Dry::Struct` for these command objects was a clear error and abuse of `Dry::Struct` in retrospect.  `Dry::Struct` should be reserved for value objects only.  The correct solution here is to rewrite these old objects with `CommandClass` or even vanilla ruby.

app/controllers/authenticate_controller.rb
```diff
@@ -19,6 +18,12 @@ def index
...
   end
 
+  def login
+    result = perform_basic_authn
+    raise Unauthorized, "Client not authenticated" unless authentication.authenticated?
+    render text: result.authentication_key
+  end
+
   def authenticate
     authentication_token = ::Authentication::Strategy.new(
```
jonahx: The creation of `Authentication::Strategy` can be moved to a private method an memoized.  It could even be put in an initializer but I'm a bit hazy on the best practice around doing that.

app/controllers/concerns/basic_authenticator.rb
```diff
@@ -8,13 +8,54 @@ module BasicAuthenticator
...
+    end
+  end
+
+  private
+
+  def authenticator_login(username, password)
+    authentication_strategy.login(login_input(username, password))
+  end
+
+  def authentication_strategy
```
jonahx: can be memoized

app/domain/authentication/authn_ldap/authenticator.rb
micahlee: As mentioned in slack, the inheritance from `::Dry::Struct` should be removed.

jonahx:
Fix of my old ctor, and removal of `included` block, per our slack convo:

https://github.com/cyberark/conjur/tree/wip/jonah-authenticator_backed_login

app/domain/authentication/login_response.rb
```diff
@@ -0,0 +1,16 @@
...
+  # Response from an authenticator login method
+  class LoginResponse < ::Dry::Struct
+    # If successful, the role id of the logged in Role
+    attribute :role_id,             ::Types::NonEmptyString.optional
+
+    # If successful, the key that should be return to the client (e.g.
+    # the Conjur API key for standard authentication)
+    attribute :authentication_key,  ::Types::NonEmptyString.optional
+  end
+end
```
jonahx: perfect use for Dry::Struct

config/routes.rb
jonahx: Per slack convo, this has been the case for a while and was the intentional design.

app/controllers/authenticate_controller.rb
```diff
@@ -60,23 +48,38 @@ def k8s_inject_client_cert
...
+  end
+
+  def authentication_input
+    ::Authentication::Strategy::Input.new(
+      authenticator_name: params[:authenticator],
+      service_id:         params[:service_id],
+      account:            params[:account],
+      username:           params[:id],
+      password:           request.body.read,
+      origin:             request.ip,
```
jonahx: Occurs to me now the `origin` is a bit vague, especially when you see it in a context like

> "User is not authorized to login from the current origin"

Does that mean a particular IP address?  A country?  A network?  Looks like you just lifted the name from existing code on the `Role` class, but I'd say no reason to let a bad name spread.  I'd suggest `ip_address` or `request_ip` instead...

spec/controllers/authenticate_controller_spec.rb
```diff
@@ -6,6 +6,47 @@
...
+        let(:basic_password) { api_key }
+        include_context "authenticate Basic"
+        it_should_behave_like "successful login"
+      end
+      context "with Token auth" do
+        include_context "authenticate Token"
+        it_should_behave_like "login denied"
+      end
+    end
+  end
```
jonahx: One of the main reason for creating `Strategy` as an object decoupled is so that we can test this logic in pure ruby, without going through Rails.  The same tests, written as unit tests, will execute instantly, as well as be more focused (ie, you're just testing input / output of a method).

app/domain/authentication/strategy.rb
```diff
@@ -77,6 +79,22 @@ def conjur_token(input)
...
+      key = authenticator.login(input)      
+      raise InvalidCredentials unless key
+
+      audit_success(input)
+      new_login(input, key)
+    rescue => err
+      audit_failure(input, err)
+      raise err
+    end
+
```
jonahx: Having this extra method on `Strategy` pushes me just to the edge (maybe over) of how far I like to stretch SRP.  If I rewriting this today I'd use two command objects, perhaps `ExchangeCredentialsForToken` (previously `Strategy`)  and `ExchangeCredentialsForApiKey`.

All of the shared logic would be pulled out into objects, which both of them would delegate to: eg, `ValidateSecurity`, `ValidateAuthenticatorExists`, etc....

jonahx:
Looks like it needs a rebase before it can be merged.

I'm going to approve and let you decide how to proceed with the comments... would be worthwhile fixing some of the CC issues as well.  

I think this is mergeable as is, but we should try to capture the remaining work in followon tasks if we decide to table it for now.

app/domain/authentication/strategy.rb
micahlee: Tracking here: https://github.com/cyberark/conjur/issues/747

micahlee:
Hey @jtuttle, did you have any additional comments or improvements for this PR? Thanks!

PR #733, opened 2018-09-12
Prevent anonymous LDAP authentication

### LINKED ISSUE #672: Users can authenticate with empty LDAP password
LDAP servers can include passwordless authentication.  This allows a user to authenticate to LDAP using only their username.

I've validated the authenticator attempts to reach out to LDAP when the password is empty.  We need to update this so any authentication request with an empty body returns an error (potentially `400`).

micahlee:
#### What does this PR do?
This PR ports over the check from V4 LDAP authentication that specifically forbids blank passwords. This is to ensure that anonymous LDAP binding is not allowed in Conjur, even if the LDAP server enables it.

Closes #672 

It is a simple change in `app/domain/authentication/authn_ldap/authenticator.rb` with additional tests added to verify the behavior.

This is tested automatically in CI.

https://jenkins.conjur.net/job/cyberark--conjur/job/no_blank_ldap_password/

Yes

Yes

No

No

I don't think so.

### DIFF
--- a/spec/app/domain/authentication/authn_ldap/authenticator_spec.rb
+++ b/spec/app/domain/authentication/authn_ldap/authenticator_spec.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Authentication::AuthnLdap::Authenticator do
+  let(:authenticator_instance) do
+    Authentication::AuthnLdap::Authenticator.new(env: {})
+  end
+  
+  let(:input) do
+    ::Authentication::Strategy::Input.new(
+      authenticator_name: 'ldap',
+      service_id: 'test',
+      account: 'test',
+      username: username,
+      password: password,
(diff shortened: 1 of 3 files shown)


spec/app/domain/authentication/authn_ldap/authenticator_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      password: password,
+      origin: '127.0.0.1',
+      request: nil
+    )
+  end
+
+  before do
+    allow_any_instance_of(Net::LDAP)
+      .to receive(:bind_as)
+      .and_return(:valid_login?)
```
jonahx: should this be `and_return(true)` or `and_return(<some obj>)`?

micahlee: Good catch! I had intended this to make it easier to control the mocked LDAP response, but I ended up only needing `true` for these cases. Thanks!

PR #736, opened 2018-09-17
Authn OIDC - Draft #1

moticless:
#### What does this PR do?
Review of new OpenID authenticator with one API

- Familiar with OpenID protocol
- Design 
[Design.1.-.1.API.with.input.of.authorization.token.and.output.of.access.token (2).pdf](https://github.com/cyberark/conjur/files/2389394/Design.1.-.1.API.with.input.of.authorization.token.and.output.of.access.token.2.pdf)
- Relevant only as a milestone for the UI development

https://github.com/conjurinc/appliance/issues/277

- start review from the entry point of the API
- Run the `dev/start` script with argument:`--authn-oidc`  
- by setting OKTA environment and run REST request against oidc authenticator 
wip


app/controllers/authenticate_controller.rb
```diff
@@ -40,25 +33,54 @@ def authenticate
...
+  def handle_authentication_error(e)
     logger.debug("Authentication Error: #{e.message}")
     e.backtrace.each do |line|
       logger.debug(line)
     end
     raise Unauthorized
   end
 
-  private
+  def new_authentication_strategy
```
jonahx: Pulling this out was a good move, and it can even memoized with the standard memoization pattern:
```
@new_authentication_strategy ||= ::Authentication::Strategy... 
```
It could even be created in a rails initializer though that's something I'd like to think through a bit before doing...

app/domain/authentication/authn_oidc/authentication_service.rb
```diff
@@ -0,0 +1,95 @@
...
+
+        return id_token, user_info
+      rescue => e
+        raise OIDCAuthenticationError.new(e)
+      end
+
+      private
+
+      def validate_variable_exists (variableName)
+        resource = Resource["#{conjur_account}:variable:#{service_id}/#{variableName}"]
```
jonahx: `Resource` should be injected as a dependency so that:
1. The classes dependencies are all clearly visible at the top
2. So that the class can be easily tested.

I create a gem for doing this without any boilerplate, which I think really cleaned up the code for similar objects in our k8s refactor:

https://github.com/jonahx/command_class

It will help to take a look at the K8s classes which use it:

- https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/authenticator.rb
- https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/inject_client_cert.rb
- https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/validate_pod_request.rb

Let me know if you have any questions about it.

app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+
+        username = user_info.preferred_username
+        input.instance_variable_set(:@username, username)
+
+        true
+      end
+
+      private
+
+      def authenticator_name
```
jtuttle: Is there a reason you need these four getters? I would suggest removing them and just referring to the instance variable directly w/ `@`.

app/domain/authentication/authn_oidc/authentication_service.rb
```diff
@@ -0,0 +1,95 @@
+require 'uri'
+require 'openid_connect'
+
+module Authentication
+  module AuthnOidc
+    class AuthenticationService
```
jtuttle: I know you were told to look at the K8s authenticator as a model but we just finished doing a huge refactor that might serve as a better model. I would suggest taking a look (it's in master now) and restructuring to be more in line with that.

yonikr: Thanks @jtuttle ! What for example are you referring to?

app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+
+        oidc_authn_service = AuthenticationService.new(service.identifier, conjur_account)
+        id_token, user_info = oidc_authn_service.get_user_details(request_body)
+
+        # TODO: validate id_token - if not raise error
+
+        validate_user_info(user_info, id_token.sub)
+
+        username = user_info.preferred_username
+        input.instance_variable_set(:@username, username)
```
jonahx: `input` is intentionally immutable, and breaking the immutability with this backchannel method seriously undermines the design.

I'll discuss alternative solutions to the larger problem this is being used to solved in other comments or in my review at the end.

app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      def valid?(input)
+        @authenticator_name = input.authenticator_name
+        @service_id = input.service_id
+        @conjur_account = input.account
+        @request_body = input.request.body.read
+
+        verify_service_enabled
+
+        oidc_authn_service = AuthenticationService.new(service.identifier, conjur_account)
```
jonahx: `AuthenticationService` should be injected.  See my other command where I discuss `CommandClass`

app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      def initialize(env:)
+        @env = env
+      end
+
+      def valid?(input)
+        @authenticator_name = input.authenticator_name
+        @service_id = input.service_id
+        @conjur_account = input.account
+        @request_body = input.request.body.read
```
jonahx: The `valid?` method should be pure: It shouldn't be mutating state on the object.

This is problematic for multiple reasons:
1. It makes the code flow more complex to reason about.
2. (see my other comment on `CommandClass`) These authenticator objects are ultimately intended to be created once, at boot time.  When we actually do this (which will probably be soon), code like the code above will actually not be threadsafe.

The `command_class` gem offers a simple alternative that will make this kind of code unnecessary.

In the case of `request_body` you'd simply create a memoized private method (`@input` will be automatically available):
```
def request_body
  @request_body ||= @input.request.body.read
end
```

For the other 3, you could either do the same, or use the `Forwardable` class since this is just a case of delegation to `@input` .  For an example see:

https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/validate_pod_request.rb#L17

app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+        @conjur_account
+      end
+
+      def service_id
+        @service_id
+      end
+
+      def request_body
+        @request_body
+      end
```
jonahx: The above 4 methods should be changed to reflect the suggestions in the comment at line 18.

app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      def service
+        @service ||= Resource["#{conjur_account}:webservice:conjur/#{authenticator_name}/#{service_id}"]
+      end
```
jonahx: `Resource` should be an injected dependency.

app/domain/authentication/strategy.rb
```diff
@@ -77,6 +77,24 @@ def conjur_token(input)
...
+      validate_security(input)
+      validate_origin(input)
+
+      audit_success(input)
+      new_token(input)
+    rescue => e
+      audit_failure(input, e)
+      raise e
+    end
+
```
jonahx: This method is problematic in itself as well as for the ripple effects it causes in the `Authenticator` itself (discussed in other comments), but I suspect it will go away naturally when we implement the new design, in which retrieving the token will be its own API route.  That is, when the `authenticate` route becomes simply exchanging the session cookie containing the encrypted id token for a conjur token, then it will fit within the original `Strategy` object, without modification.

I think the additional route of the new design may naturally clean up some of the issues I mentioned in other comments as well.

spec/app/domain/authentication/authn_oidc/authenticator_spec.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Authentication::AuthnOidc::Authenticator do
+  def non_equal_id_token_subject
```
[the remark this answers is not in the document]
jonahx: This is correct.  The rspec idiom is to use `let` when setting up the doubles, spies, or other objects that will be used in the tests themselves (the `it` statements)

spec/app/domain/authentication/authn_oidc/authenticator_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  let (:authenticator_instance) do
+    Authentication::AuthnOidc::Authenticator.new(env:[])
+  end
+
+  it "validates user_info with valid value" do
+    subject = authenticator_instance
+    expect { subject.send(:validate_user_info, valid_user_info, valid_user_info_subject) }.to_not raise_error
+  end
```
jonahx: We should never unit test private methods.  Only test the public behavior of the object.

spec/routing/authn_routing_spec.rb
```diff
@@ -12,7 +12,24 @@
...
+  end
+
+  it "routes POST /authn-oidc/the-account/authenticate to authenticate#authenticate_oidc" do
+    expect(post: '/authn-oidc/the-account/authenticate').to route_to(
+      controller: 'authenticate',
+      action: 'authenticate_oidc',
+      account: 'the-account'
+    )
+  end
+
```
jonahx: I don't think there's much benefit to these tests.  It's almost testing the rails routing framework itself, and besides, what they're resting will be implicitly tested by our end to end tests.

I would remove them.

dev/files/authn-oidc/phantomjs/keycloak_login.js
```diff
@@ -0,0 +1,93 @@
+var steps=[];
```
jonahx: I had a lot of suggestions for cleaning up this js file, so it made more sense to just refactor it and show by doing.  Here is the cleaned up version:

https://gist.github.com/jonahx/8236760f0b498002877e16e8906561b5

A few notes about the things I did:

- Replaced the spread out logic involving `setInterval`, `loadInProgress`, and `executeRequestsStepByStep` and the `onLoadStarted` / `onLoadFinished` handlers with the object `PhantomSteps`.
- Move all `require`s to the top
- Pull out repeated code into utility methods.
- Use a utility method and an object, rather than string manipulation, to create the login url with query string.

dev/files/authn-oidc/phantomjs/keycloak_login_tmp.js
```diff
@@ -0,0 +1,94 @@
+var steps=[];
```
jonahx: This file can be cleaned up similar to the other one.

jonahx:
Some general comments after looking over the JS.  These are things the _probably_ don't make sense for this first PR, but which I think we should do at some point before the project officially finishes:

1. The testing infrastructure (and I know you guys were just building off what was already there, so no blame is intended here...) is too complex imo... significantly more complex than any of the actual code.  We are using the keycloak service, a number of bash files, docker, phantomjs, etc... 

Ideally, we could find a way to simplify that.  But if that turns out not to be feasible, I think we need to add extensive documentation about how the tests work and how all the pieces mentioned above fit together.  A new developer coming to this code should be able to ramp up quickly and have 0 struggle understanding our testing infrastructure.

2. PhantomJS is a now abandoned/frozen project.  It does not support modern JS syntax, and its API is heavier weight than modern alternatives like [Nightmare.js](http://www.nightmarejs.org/).  

3. Another alternative we could consider is avoiding a JS web driver altogether.  After all, we don't actually test an JS here.  We're simply testing http requests and responses.  https://github.com/rack-test/rack-test or similar should be sufficient and would avoid a lot of this infrastructure.

app/domain/authentication/authn_oidc/authentication_service.rb
```diff
@@ -0,0 +1,101 @@
...
+        Resource["#{conjur_account}:variable:#{service_id}/provider-uri"].secret.value
+      end
+
+      def redirect_uri
+        @redirect_uri
+      end
+
+      def discover
+        # TODO: should not run in production
+        disable_ssl_verification
```
jonahx: I'm confused here.  Is this test only code?

app/domain/authentication/authn_oidc/authentication_service.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      def get_client
+        @client ||= OpenIDConnect::Client.new(
+            identifier: client_id,
+            secret: client_secret,
+            redirect_uri: redirect_uri,
+            token_endpoint: discover.token_endpoint,
+            userinfo_endpoint: discover.userinfo_endpoint
+        )
+      end
```
jonahx: Rename above to `open_id_client` per my other note about "name after the returned thing" convention.

app/domain/authentication/authn_oidc/authentication_service.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      def decode_id_token(id_token)
+        # TBD: catpture exception: JSON::JWK::Set::KidNotFound: JSON::JWK::Set::KidNotFound and try refresh signing keys
+        OpenIDConnect::ResponseObject::IdToken.decode id_token, discover.jwks
+      end
```
jonahx: Rename to `decoded_id_token`

app/domain/authentication/strategy.rb
```diff
@@ -77,6 +77,24 @@ def conjur_token(input)
       raise e
     end
 
+    def conjur_token_oidc(input)
+      authenticator = authenticators[input.authenticator_name]
+
+      validate_authenticator_exists(input, authenticator)
+
+      # this method also injects the username to input
```
jonahx: Our new plan involves releasing with this version without separate endpoints.

However, I realized we can (and should) still split up the "get user details" and "check if valid" steps.

I would probably be inclined to create two command objects for these steps, although it could also be achieved by adding a new method to the authenticator.

This will allow us to avoid the hack above that mutates input, which needs to remain immutable.

As a separate issue which I should probably take care of myself, since it affects code unrelated to OIDC itself, but which is worth pointing out, we should eventually have an OIDC specific "Strategy" object rather than editing the existing one.  This is related to this existing issue: https://github.com/cyberark/conjur/issues/747

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: This was done so that we can use the already written code such as `validate_security` etc. which gets the username from the input. If this isn't good we need to find a way to change the design a bit in other authenticators too.

app/controllers/authenticate_controller.rb
orenbm: done

app/domain/authentication/authn_oidc/authentication_service.rb
orenbm: I added a ticket for this, we'll get to it after the split

spec/app/domain/authentication/authn_oidc/authenticator_spec.rb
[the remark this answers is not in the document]
orenbm: done

dev/start
```diff
@@ -63,6 +67,52 @@ if [[ $ENABLE_AUTHN_LDAP = true ]]; then
...
+  echo "Starting Conjur server"
+  api_key=$(docker-compose exec -T conjur conjurctl \
+  	role retrieve-key cucumber:user:admin | tr -d '\r')
+  docker-compose exec $env_args -d conjur conjurctl server
+
+  echo "Sleep 30 sec for Conjur server to start"
+  sleep 30
+
+  echo "Setting keycloak variables values in conjur"
+  keycloak_CLIENT_ID=$(docker-compose exec oidc-keycloak printenv CLIENT_ID | tr -d '\r')
```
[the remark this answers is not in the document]
orenbm: done

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: done

app/domain/authentication/authn_oidc/authentication_service.rb
orenbm: I added a ticket for this, we'll get to it after the split

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: I added a ticket for this, we'll get to it after the split

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: I added a ticket for this, we'll get to it after the split

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: I removed these getters as suggested by @jtuttle

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: Do you mean this - https://github.com/cyberark/conjur/blob/97f1e3a3b4edb9e7bc1f3da973f651ea9006b009/app/domain/authentication/webservice.rb#L25?

In the method below (`resource_id`) you d othe exact string building we're doing so I'm a bit confused.

Regarding the injected dependency - this will be done after the split.

app/domain/authentication/strategy.rb
orenbm: let's get back to this after the new design is implemented and see if it's still a problem

spec/app/domain/authentication/authn_oidc/authenticator_spec.rb
orenbm: I agree. At this point this is the only thing we have a UT for so I removed the file

spec/routing/authn_routing_spec.rb
orenbm: These tests aren't part of our PR (we just added some tests) so I'd rather not get this decision as part of the review

app/domain/authentication/authn_oidc/authentication_service.rb
```diff
@@ -0,0 +1,101 @@
...
+      # Constructs AuthenticationService from the <service-id>, which is typically something like
+      # conjur/authn-oidc/<service-id>.
+      def initialize(service_id, conjur_account)
+        @service_id = service_id
+        @conjur_account = conjur_account
+
+        validate_variable_exists("client-id")
+        validate_variable_exists("client-secret")
+        validate_variable_exists("provider-uri")
+      end
```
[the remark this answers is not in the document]
orenbm: Checking if `client_id` (for instance) is nil isn't enough as `client_id` is the secret itself and if the varaible doesn't exist we'll get a null pointer exception on the variable itself. 

How about this?

```
def client_id
        resource = Resource["#{conjur_account}:variable:#{service_id}/client-id"]
        if resource.nil? || resource.secret.nil?
          raise OIDCConfigurationError, "Variable [#{service_id}/#{variableName}] not found in Conjur"
        end
        
        resource.secret.value
      end
``` 

without the validation calls.

Regarding the calls to the database vs. restarting Conjur, i guess the main question is how expensive is it to restart Conjur. The customer will need to update the `client-secret` but in long intervals so we can guide the user to restart conjur once he updated a value (@InbalZilberman wdyt?).

in that case we can have this method:
```
      def client_id
        if @client_id.nil?
          resource = Resource["#{conjur_account}:variable:#{service_id}/client-id"]
          if resource.nil? || resource.secret.nil?
            raise OIDCConfigurationError, "Variable [#{service_id}/client-id] not found in Conjur"
          end

          @client_id = resource.secret.value
        end

        @client_id
      end
```
which will have only one call to the database.

tell me what you think

app/domain/authentication/authn_oidc/authentication_service.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        client = get_client
+        client.authorization_code = authorization_code
+        access_token = client.access_token!
+
+        client.host = URI.parse(provider_uri).host
+        id_token = decode_id_token(access_token.id_token)
+        user_info = access_token.userinfo!
+
+        return id_token, user_info
```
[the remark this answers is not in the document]
orenbm: done

app/domain/authentication/authn_oidc/authentication_service.rb
orenbm: no. we'll add the ssl verification later

app/domain/authentication/authn_oidc/authentication_service.rb
orenbm: done

app/domain/authentication/authn_oidc/authentication_service.rb
orenbm: done

dev/files/authn-oidc/phantomjs/keycloak_login.js
nessiLahav: First of all thanks a lot of creating a better js code (I do not familiar with js, I just searched for quick POC to create an infrastructure for testing environment),
I tried to run your new js script, but as you mentioned phantomjs does not support ES6 and most of the syntax is relevant to ES6 (according to http://jshint.com/),
According to your next comment which I completely agree with you, we will going to replace phantoms with something else,
So I think we will delayed this fix until we will decide on which tool will replace phantomjs, anyway this will happen after the split of the APIs (I created a card in our board),
What do you think?

app/domain/authentication/authn_oidc/authentication_service.rb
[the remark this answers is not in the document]
orenbm: Also, is this secure? I'm not sure how this works in ruby but in the approach where we save the secret in memory it may be insecure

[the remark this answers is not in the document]
InbalZilberman: from user perspective there are 2 aspects here:
1. client secret is a secret that can actually be rotated and changed hence should be dynamic. same goes to client id. provider uri isnt - maybe we can use annotation? @jonahx , @orenbm WDYT?
2. I think that once we defined something as a variable the end user expectations are to have its value being filled in via var value at any time. Telling our end users for X vars you now need to do a restart kind of miss the point of Conjur abilities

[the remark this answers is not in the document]
orenbm: I agree that we can't restart for some variable changes and not restart for others s it's a confusing UX. 

Regarding using the provider-uri as an annotation (or env var?) and not as a conjur variable, I guess we should discuss this in a different thread, regardless of this PR. 

I'll wait to see what Jonah says as we already posted some questions for him in this thread.

app/domain/authentication/authn_oidc/authentication_service.rb
nessiLahav: We have an issue with our OpenID 3rd party lib, which doesn’t support self sign certificate, and prevent us to communicate with keycloak (the OpenID provider in our dev&test environment ),
We aware to this issue and we planned to handle this in the new design of the 3 APIs ( after this PR).

app/domain/authentication/authn_oidc/authenticator.rb
jonahx: > In the method below (resource_id) you d othe exact string building we're doing so I'm a bit confused.

Correct.  I'm saying (to avoid duplicating logic, and also for clarity) simply use the existing object to accomplish the same thing you are accomplishing.

> Regarding the injected dependency - this will be done after the split.

Sounds good.

app/domain/authentication/authn_oidc/authenticator.rb
jonahx: Let's discuss this on Tuesday.  I may have done some refactoring of this by then.

app/domain/authentication/authn_oidc/authenticator.rb
jonahx: This one really needs to be fixed before we integrate this code into master.  If there's a time issue I can take this particular refactor off your hands since I'll be able to implement it quickly I believe.

dev/files/authn-oidc/phantomjs/keycloak_login.js
jonahx: I think that's a good decision.

I would strongly consider looking into to replacing it with something that doesn't use js or a headless browser at all, as I suggested.

app/domain/authentication/authn_oidc/authentication_service.rb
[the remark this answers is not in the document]
jonahx: Agree 100% with @InbalZilberman's comments.  So it sounds like we are going to do the validation and fetching of the secret values on every request.

Given that, I thought about this quite a bit... it seems like a small thing at first, but on further reflection I think that "verifying that a secret exists and getting its current value" is really its own thing and not specific to OIDC.  I am inclined to do this with a dedicated command object in the main method (currently `user_details`) and then to pass the secrets returned by the command object `FetchRequiredSecrets` as arguments down to other methods that need them.

I went ahead and created the command object for this, just to validate that my idea would work.  You can copy it from here, drop it in into `app/domain/conjur` in your branch, and then integrate it as described above:

https://gist.github.com/jonahx/4c9b050e87a38a0ca3fae0324129f510

I will try to integrate this into master before tuesday, so check for that, and if it's there then you only have to do the integration of it into your OIDC code.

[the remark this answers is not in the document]
orenbm: thanks!

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: I think this is a great thing to do together in Boston. it will be very educating to do this together.

app/domain/authentication/authn_oidc/authentication_service.rb
yonikr: @jtuttle @orenbm @jonahx I understand this can be done after merge to master, right?

app/domain/authentication/authn_oidc/authenticator.rb
yonikr: @orenbm @jonahx Let's shoot for working on UI and not authenticators together while we are in Boston

orenbm: NP, I added a ticket for this. we'll address it after the merge

PR #753, opened 2018-10-04
Handle host builder scenario when host role already exists

### LINKED ISSUE #744: Creating a new host using hf token can cause an exception, depending on the name
Creating a host called `brand-new-host` using an hf token causes an exception, while naming it `brand_new_host` works fine.

#### Steps to reproduce
1. login to Conjur as admin
1. load this policy as root:
   ```yaml
   - !policy
     id: database
     body:
       - !layer users
       - !host-factory
         id: users
         layers: [ !layer users ]
   ```
1. create a host factory token: `conjur hostfactory tokens create database/users`
1. use the token to create a host: `conjur hostfactory hosts create $token brand-new-host`

Expected result: new host created, id and API key returned
Actual result: 500 error

#### Workaround
If you instead create the host with id `brand_new_host`, you get the expected result.

#### Side note

With the request to create a new host, the Conjur CLI is sending some unfamiliar data. Examples:

```
__parent__[%23%3CGLI%3A%3ACommand%3A%3AParentKey%3A0x000000018bf6b8%3E]&id=brand-new-host
```

```
__parent__[%23%3CGLI%3A%3ACommand%3A%3AParentKey%3A0x00000002bd76b0%3E]&id=brand_new_host
```

All I'd expect to see is the last part, `id=whatever`. Why is there a ParentKey with some kind of pointer address or something in there?

#### Session log

```sh-session
$ conjur hostfactory tokens create database/users
[
  {
    "token": "[redacted]",
    "expiration": "2018-09-21T20:53:24+00:00",
    "cidr": [

    ]
  }
]
$ conjur hostfactory hosts create [redacted] "brand-new-host"
RestClient.post "https://eval.conjur.org/host_factories/hosts", "__parent__[%23%3CGLI%3A%3ACommand%3A%3AParentKey%3A0x000000018bf6b8%3E]&id=brand-new-host", "Accept"=>"*/*", "Accept-Encoding"=>"gzip, deflate", "Authorization"=>"Token token=\"1eydjya19ht5xd1tyyf5k8dm7me3gq3g5420nbxhdk84nyr1j5t1r\"", "Content-Length"=>"89", "Content-Type"=>"application/x-www-form-urlencoded", "User-Agent"=>"rest-client/2.0.2 (linux-gnu x86_64) ruby/2.4.1p111"
# => 500 InternalServerError | text/html 0 bytes

error: 500 Internal Server Error


$ conjur hostfactory hosts create [redacted] "brand_new_host"
RestClient.post "https://eval.conjur.org/host_factories/hosts", "__parent__[%23%3CGLI%3A%3ACommand%3A%3AParentKey%3A0x00000002bd76b0%3E]&id=brand_new_host", "Accept"=>"*/*", "Accept-Encoding"=>"gzip, deflate", "Authorization"=>"Token token=\"1eydjya19ht5xd1tyyf5k8dm7me3gq3g5420nbxhdk84nyr1j5t1r\"", "Content-Length"=>"89", "Content-Type"=>"application/x-www-form-urlencoded", "User-Agent"=>"rest-client/2.0.2 (linux-gnu x86_64) ruby/2.4.1p111"
# => 201 Created | application/json 261 bytes
{
  "created_at": "2018-09-21T19:53:56.525+00:00",
  "id": "ryan.prior@cyberark.com:host:brand_new_host",
  "owner": "ryan.prior@cyberark.com:host_factory:database/users",
  "permissions": [

  ],
  "annotations": [

  ],
  "api_key": "[redacted]"
}
```

micahlee:
#### What does this PR do?

This PR updates the host builder for host factories to check for the existence of a Host role and its resource with the requested ID prior to checking permissions and attempting to rotate the api key.

If a host role already exists, but there is no corresponding resource to check, the host builder will now return `403 forbidden` because we cannot verify ownership.

If a host was previously created and then removed by either a host factory or policy, a `Role` record would exist for the host, but there would not exist a `Resource` record. A bug in the host builder assumed that if a `Role` was present, then the `Resource` would also be present. This caused the permission check to return a `500` server error for the missing record. 

Closes #744 

The PR consists of 2 supporting commits to update the test infrastructure, and 1 last commit that adds automated tests for this scenario and updates the host builder to include the resource existence check.

This is tested with the automated CI tests, but can be run manually with
```
bundle exec cucumber -p api cucumber/api/features/host_factory_rotate_host.feature
```

https://jenkins.conjur.net/job/cyberark--conjur/job/host-factory-744/

Yes

I don't think so

No

No

### DIFF
--- a/cucumber/api/features/host_factory_rotate_host.feature
+++ b/cucumber/api/features/host_factory_rotate_host.feature
@@ -1,0 +1,75 @@
+Feature: Rotate a host api key using the host factory.
+
+  Background:
+    Given I am the super-user
+    And I successfully PUT "/policies/cucumber/policy/root" with body:
+    """
+    - !policy
+      id: database
+      body:
+        - !layer users
+        - !host-factory
+          id: users
+          layers: [ !layer users ]
+    """
+    And a host factory token for "database/users"
+    And I authorize the request with the host factory token
+    And I successfully POST "/host_factories/hosts?id=brand-new-host"
+    And I log out
+    Given I am the super-user
+    And I successfully PUT "/policies/cucumber/policy/root" with body:
+    """
+    []
+    """
+
+  Scenario: The host factory can rotate the host api key
+
+    Given I am the super-user
+    And I successfully PUT "/policies/cucumber/policy/root" with body:
+    """
+    - !policy
+      id: database
+      body:
+        - !layer users
+        - !host-factory
+          id: users
+          layers: [ !layer users ]
+    """
+    And a host factory token for "database/users"
+    And I authorize the request with the host factory token
+    And I POST "/host_factories/hosts?id=brand-new-host"
+    Then the HTTP response status code is 403
+
+  Scenario: The host factory can rotate the host api key
+
+    Given I am the super-user
+    And I successfully PUT "/policies/cucumber/policy/root" with body:
+    """
+    - !policy
+      id: database
+      body:
+        - !layer users
+        - !host-factory
+          id: users
+          layers: [ !layer users ]
+
+    - !host 
+      id: brand-new-host
+      owner: !host-factory database/users
+    """
+    And a host factory token for "database/users"
+    And I authorize the request with the host factory token
+    And I successfully POST "/host_factories/hosts?id=brand-new-host"
+    Then the HTTP response status code is 201
+    And the JSON should be:
+    """
+    {
+      "annotations" : [],
+      "id": "cucumber:host:brand-new-host",
+      "owner": "cucumber:host_factory:database/users",
+      "policy": "cucumber:policy:root",
+      "api_key": "@response_api_key@",
+      "permissions": [],
+      "restricted_to": []
+    }
+    """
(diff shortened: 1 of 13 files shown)


cucumber/api/features/host_factory_rotate_host.feature
```diff
@@ -0,0 +1,75 @@
...
+    """
+    - !policy
+      id: database
+      body:
+        - !layer users
+        - !host-factory
+          id: users
+          layers: [ !layer users ]
+    """
+    And a host factory token for "database/users"
```
jonahx: This line makes it sounds like you also "successfully PUT a toke".  I think the step def matcher should be change from 
```
Given(/^a host factory token(?: for "([^"]*)")?$/) do |host_factory_id|
```
to
```
Given(/^I have a host factory token(?: for "([^"]*)")?$/) do |host_factory_id|
```

micahlee:
Hey @jonahx, I updated the cucumber test steps like you recommended so that they use the active voice instead of passive. Let me know if you see other improvements I can make! Thanks!

PR #756, opened 2018-10-09
Oidc error handling added

orenbm:
#### What does this PR do?
Adds error handling cases for OIDC authentication

Authentication Strategy

https://jenkins.conjur.net/job/cyberark--conjur/job/oidc-error-handling/

no

### DIFF
--- a/app/domain/authentication/authn_oidc/authenticator.rb
+++ b/app/domain/authentication/authn_oidc/authenticator.rb
@@ -29,7 +29,7 @@ module Authentication
 
       def validate_user_info
         raise OIDCAuthenticationError, subject_err_msg unless valid_subject?
-        raise OIDCAuthenticationError, no_profile_err_msg unless preferred_username
+        raise OIDCAuthenticationError, no_username_err_msg unless preferred_username
       end
 
       def user_info
@@ -76,8 +76,8 @@ module Authentication
         @service ||= Resource["#{conjur_account}:webservice:conjur/#{authenticator_name}/#{service_id}"]
       end
 
-      def no_profile_err_msg
-        "[profile] is not included in scope of authorization code request"
+      def no_username_err_msg
+        "username not found in OIDC access token"
       end
 
       def subject_err_msg
--- a/app/domain/authentication/security.rb
+++ b/app/domain/authentication/security.rb
@@ -14,6 +14,8 @@ module Authentication
       "Webservice '{0}' is not defined in the Conjur policy")
     NotAuthorizedInConjur = ::Util::ErrorClass.new(
       "User '{0}' is not authorized in the Conjur policy")
+    NotDefinedInConjur = ::Util::ErrorClass.new(
+      "User '{0}' is not defined in Conjur")
 
     class AccessRequest < ::Dry::Struct
       attribute :webservice, ::Types.Instance(::Authentication::Webservice)
@@ -54,7 +56,7 @@ module Authentication
       account      = req.webservice.account
       user_role_id = role_class.roleid_from_username(account, req.user_id)
       user_role    = role_class[user_role_id]
-      raise NotAuthorizedInConjur, req.user_id unless user_role
+      raise NotDefinedInConjur, req.user_id unless user_role
 
       # Ensure user has access to the service
       has_access = user_role.allowed_to?('authenticate', webservice_resource)
--- a/app/domain/conjur/fetch_required_secrets.rb
+++ b/app/domain/conjur/fetch_required_secrets.rb
@@ -7,7 +7,7 @@ module Conjur
     'Missing required resource: {0}'
   )
   RequiredSecretMissing = Util::ErrorClass.new(
-    'Missing secret for resource: {0}'
+    'Missing value for resource: {0}'
   )
 
   FetchRequiredSecrets = ::CommandClass.new(
--- a/spec/app/domain/authentication/security_spec.rb
+++ b/spec/app/domain/authentication/security_spec.rb
@@ -104,7 +104,7 @@ RSpec.describe Authentication::Security do
       )
     end
   end
- 
+
   context "A whitelisted, authorized webservice and authorized user" do
     it "validates without error" do
       subject = full_access_security_double
@@ -139,13 +139,13 @@ RSpec.describe Authentication::Security do
   end
 
   context "A whitelisted, authorized webservice and non-existent user" do
-    it "raises a NotAuthorizedInConjur error" do
+    it "raises a NotDefinedInConjur error" do
       subject = Authentication::Security.new(
         role_class: nil_user_role_class,
         resource_class: full_access_resource_class
       )
       expect { subject.validate(valid_access_request) }.to(
-        raise_error(Authentication::Security::NotAuthorizedInConjur)
+        raise_error(Authentication::Security::NotDefinedInConjur)
       )
     end
   end


app/controllers/application_controller.rb
```diff
@@ -7,6 +7,9 @@ class ApplicationController < ActionController::API
   class Unauthorized < RuntimeError
   end
 
+  class NotImplemented < Exceptions::NotImplemented
+  end
+
```
jonahx: I don't think we need this shortcut name.  Just refer to it as `Exceptions::NotImplemented`

app/controllers/application_controller.rb
```diff
@@ -177,6 +181,16 @@ def unauthorized e
...
 
+  def not_implemented e
+    logger.debug "#{e}\n#{e.backtrace.join "\n"}"
+    render json: {
+      error: {
+        code: "not_implemented",
+        message: e.message
+      }
+    }, status: :not_implemented
+  end
```
jonahx: I know you guys are probably just following along with the existing code here, and you should probably leave this as for consistency (we can clean up all of later if we want), but I wanted to point out that the way we are doing this is fragmented.  A better design imo would to make the error objects responsible for their own representation as json, ie, giving the a `to_json` method.  That would clean up all the code like this on this page, enabling us to write `render json: e.to_json`.

app/controllers/authenticate_controller.rb
```diff
@@ -77,7 +77,16 @@ def handle_authentication_error(err)
...
+
+    case err
+    when Conjur::RequiredResourceMissing
+    when Conjur::RequiredSecretMissing
+    when Authentication::Security::ServiceNotDefined
+    when Authentication::Security::NotWhitelisted
+      raise NotImplemented, err.message
+    else
+      raise Unauthorized
+    end
```
jonahx: This belongs in the relevant controllers.  Didn't look into the first two, but `Authentication::Security::ServiceNotDefined` and `Authentication::Security::NotWhitelisted` belong in the `AuthenticateController` -- they're not application wide concerns.

app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -26,6 +23,9 @@ def call
       def validate_id_token_claims
         expected = { client_id: client_id, issuer: issuer } # , nonce: 'nonce'}
         @user_details.id_token.verify!(expected)
+      rescue => e
+        raise OIDCAuthenticationError, e.message if is_invalid_token_error?(e)
+        raise e
       end
```
jonahx: Rather than rescue all (which you should never do), check for something specific, and then re-raise for all others, simply do: 

```
      def validate_id_token_claims
        expected = { client_id: client_id, issuer: issuer } # , nonce: 'nonce'}
        @user_details.id_token.verify!(expected)
      rescue OpenIDConnect::ResponseObject::IdToken::InvalidToken => e
        raise OIDCAuthenticationError, e.message
      end
```
This will remove the need for the `is_invalid_token_error?` method.

As a side note for future reference, when naming a method that returns a boolean, use the `?` suffix but not the `is_` prefix, since using both is redundant.  Ie, the above method would be named `invalid_token_error?`.  Otoh, variable names cannot end with a `?`, so in that case you can use the `is_` prefix where appropriate.

app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -85,6 +85,10 @@ def subject_err_msg
         "User info subject [#{user_info.sub}] and id token subject " +
          "[#{id_token_subject}] are not equal"
       end
+
+      def is_invalid_token_error? err
+        err.class <= OpenIDConnect::ResponseObject::IdToken::InvalidToken
+      end
```
jonahx: per above comment, this can be removed.

app/domain/authentication/authn_oidc/get_user_details.rb
```diff
@@ -59,6 +59,9 @@ def user_info
 
       def access_token
         @access_token ||= oidc_client.access_token!
+      rescue => e
+        raise OIDCAuthenticationError, e.message if is_oauth_error?(e)
+        raise e
```
jonahx: See above comment - same applies here.

app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -26,6 +23,9 @@ def call
       def validate_id_token_claims
         expected = { client_id: client_id, issuer: issuer } # , nonce: 'nonce'}
         @user_details.id_token.verify!(expected)
+      rescue => e
+        raise OIDCAuthenticationError, e.message if is_invalid_token_error?(e)
+        raise e
```
jonahx: Rather than rescue all (which you should never do), check for something specific, and then re-raise for all others, simply do: 

```
      def validate_id_token_claims
        expected = { client_id: client_id, issuer: issuer } # , nonce: 'nonce'}
        @user_details.id_token.verify!(expected)
      rescue OpenIDConnect::ResponseObject::IdToken::InvalidToken => e
        raise OIDCAuthenticationError, e.message
      end
```
This will remove the need for the `is_invalid_token_error?` method.

And since we're catching only the specific error we're interested in, all other errors will be passed through automatically, so there's no need to explicitly re-raise them.

As a side note for future reference, when naming a method that returns a boolean, use the `?` suffix but not the `is_` prefix, since using both is redundant.  Ie, the above method would be named `invalid_token_error?`.  Otoh, variable names cannot end with a `?`, so in that case you can use the `is_` prefix where appropriate.

app/domain/authentication/authn_oidc/get_user_details.rb
```diff
@@ -133,6 +136,10 @@ def oidc_client
           userinfo_endpoint: discovered_resource.userinfo_endpoint
         )
       end
+
+      def is_oauth_error? err
+        err.class <= Rack::OAuth2::Client::Error
+      end
```
jonahx: See above comments.  This should be removed too.

app/models/exceptions/not_implemented.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Exceptions
+  class NotImplemented < RuntimeError
+    attr_reader :message
+
+    def initialize message
+      @message = message
+    end
```
jonahx: The body of this `class` is superfluous: it's just re-implementing the default behavior of `RuntimeError`: 
```
[1] pry(main)> e = RuntimeError.new('test')
=> #<RuntimeError: test>
[2] pry(main)> e.message
=> "test"
```

app/controllers/authenticate_controller.rb
orenbm: I don't understand. This code is in `authenticate_controller` as you say

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: i don't follow

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: thanks for the note

app/domain/authentication/authn_oidc/authenticator.rb
orenbm: I removed it as this method is redunant by rescuing the specific error. Not sure why this is relevant to the above comment

app/controllers/authenticate_controller.rb
orenbm: what I'm trying to accomplish here is that in cases where we have an error in the authn flow, and we need further user configuration then we'll raise a NotImplemented error which will cause a 501 response code

PR #758, opened 2018-10-10
Fix k8s builds 755

### LINKED ISSUE #755: CI Pipeline fails sporadically on k8s tests
Example failed builds, with different test failures/errors:

https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--conjur/detail/master/468/pipeline
```
Scenario: Authenticate as a Pod.                                # ./cucumber/kubernetes/features/authenticate.feature:3
    Given I login to authn-k8s as "pod/inventory-pod"             # cucumber/kubernetes/features/step_definitions/login_steps.rb:33
    Then I can authenticate with authn-k8s as "pod/inventory-pod" # cucumber/kubernetes/features/step_definitions/authenticate_steps.rb:23
      no implicit conversion of nil into String (TypeError)
      ./cucumber/kubernetes/features/step_definitions/authenticate_steps.rb:28:in `initialize'
      ./cucumber/kubernetes/features/step_definitions/authenticate_steps.rb:28:in `new'
      ./cucumber/kubernetes/features/step_definitions/authenticate_steps.rb:28:in `/^I( can)? authenticate with authn-k8s as "([^"]*)"( without cert and key)?$/'
      ./cucumber/kubernetes/features/authenticate.feature:5:in `Then I can authenticate with authn-k8s as "pod/inventory-pod"'

```

https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--conjur/detail/master/464/pipeline
```
  Scenario: Authenticate as a Deployment.                                       # ./cucumber/kubernetes/features/authenticate.feature:15
    Given I login to authn-k8s as "deployment/inventory-deployment"             # cucumber/kubernetes/features/step_definitions/login_steps.rb:33
    Then I can authenticate with authn-k8s as "deployment/inventory-deployment" # cucumber/kubernetes/features/step_definitions/authenticate_steps.rb:23
      no implicit conversion of nil into String (TypeError)
      ./cucumber/kubernetes/features/step_definitions/authenticate_steps.rb:28:in `initialize'
      ./cucumber/kubernetes/features/step_definitions/authenticate_steps.rb:28:in `new'
      ./cucumber/kubernetes/features/step_definitions/authenticate_steps.rb:28:in `/^I( can)? authenticate with authn-k8s as "([^"]*)"( without cert and key)?$/'
      ./cucumber/kubernetes/features/authenticate.feature:17:in `Then I can authenticate with authn-k8s as "deployment/inventory-deployment"'

```

sgnn7:
#### What does this PR do?
Should fix majority of intermittent issues with k8s tests
 
Closes #755 

Going through individual commits - the full diff is a bit messy to look otherwise

Rebuild the branch a few times. Ensure that it doesn't break on k8s tests.

https://jenkins.conjur.net/job/cyberark--conjur/job/fix-k8s-builds-755/

Probably does not need one

Yes - same tests as before, just that these pass.

Not really

Doesn't seem to have a need for it

No

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,22 +2,25 @@ FROM ubuntu:16.04
 
 ENV DEBIAN_FRONTEND=noninteractive
 
+ENV PORT 80
+
+EXPOSE 80
+
 RUN apt-get update -y && \
     apt-get install -y software-properties-common && \
-    apt-add-repository -y ppa:brightbox/ruby-ng && \
-    apt-get update -y
+    apt-add-repository -y ppa:brightbox/ruby-ng
 
-RUN apt-get install -y \
-      build-essential \
-      ruby2.5 ruby2.5-dev \
-      postgresql-client \
-      libpq-dev \
-      unattended-upgrades \
-      ldap-utils \
-      git \
-      curl \
-      update-notifier-common \
-      tzdata
+RUN apt-get update -y && \
+    apt-get install -y build-essential \
+                       curl \
+                       git \
+                       libpq-dev \
+                       ldap-utils \
+                       postgresql-client \
+                       ruby2.5 ruby2.5-dev \
+                       tzdata \
+                       unattended-upgrades \
+                       update-notifier-common
 
 RUN gem install -N -v 1.16.1 bundler
 
@@ -25,18 +28,14 @@ RUN mkdir -p /opt/conjur-server
 
 WORKDIR /opt/conjur-server
 
-ADD Gemfile      .
-ADD Gemfile.lock .
+COPY Gemfile \
+     Gemfile.lock ./
 
 RUN bundle --without test development
 
-ADD . .
+COPY . .
 
 RUN ln -sf /opt/conjur-server/bin/conjurctl /usr/local/bin/
-
-ENV PORT 80
-
-EXPOSE 80
 
 ENV RAILS_ENV production
 
--- a/app/domain/authentication/authn_k8s/inject_client_cert.rb
+++ b/app/domain/authentication/authn_k8s/inject_client_cert.rb
@@ -5,6 +5,7 @@ module Authentication
 
     InjectClientCert = CommandClass.new(
       dependencies: {
+        logger: Rails.logger,
         resource_repo: Resource,
         conjur_ca_repo: Repos::ConjurCA,
         k8s_object_lookup: K8sObjectLookup,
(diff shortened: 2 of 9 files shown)


app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -27,13 +27,17 @@ def validate
...
 
       def install_signed_cert
-        exec = @kubectl_exec.new(
-          pod_name: spiffe_id.name,
-          pod_namespace: spiffe_id.namespace,
-          logger: Rails.logger,
-          kubeclient: @k8s_object_lookup.kubectl_client
+        pod_namespace = spiffe_id.namespace
+        pod_name = spiffe_id.name
+        puts "Copying SSL cert to #{pod_namespace}/#{pod_name}"
```
jvanderhoof: What do you think about logging these instead of using `puts`?

sgnn7: Hmm... haven't used the logger previously in the way we use it in the project - do we have an example I can compare/copy-paste?

sgnn7: Ugh - ignore this - we use it in the other file I worked on lol

jvanderhoof: :) it'll at least let us turn the volume up and down.

Dockerfile
```diff
@@ -2,42 +2,41 @@ FROM ubuntu:16.04
...
 
-ADD Gemfile      .
-ADD Gemfile.lock .
+COPY Gemfile \
+     Gemfile.lock ./
 
 RUN bundle --without test development
 
-ADD . .
+COPY . .
```
jtuttle: Are all of the changes to this file just syntax? The only one that looks like it might not be is this line. Is this part of the fix?

app/domain/authentication/authn_k8s/kubectl_exec.rb
```diff
@@ -26,106 +27,68 @@ def channel_names
...
-      def copy(path, content, mode)
-        execute(
-          [ 'tar', 'xvf', '-', '-C', '/' ],
-          stdin: true,
-          body: tar_file_as_string(path, content, mode)
-        )
-      end
-      
+
+      # on_* methods are public since ws_client needs public callbacks that it will
```
jtuttle: I think you could remove the `that it will invoke` from this comment to fit it on one line. Why else would it need public callbacks unless it was planning to invoke them. :)

app/domain/authentication/authn_k8s/kubectl_exec.rb
```diff
@@ -153,23 +116,29 @@ def on_message(msg, ws_client)
...
+
+        error_info = err.inspect
+        @logger.debug("Pod #{@pod_name} error : #{error_info}")
+        @message_log.save_error_string(error_info)
       end
 
       private
 
+      def kubeclient
+        @kubeclient ||= @k8s_object_lookup.kubectl_client
```
jtuttle: What is `@k8s_object_lookup` resolving to...? I don't see it defined anywhere else in the class.

Dockerfile
sgnn7: They're not just syntax, they're a lot are reordering to reduce/reuse cache layers. This particular change though is due to the fact that `ADD` is generally used to put something in the image from a compredded file or the web vs the `COPY` which is an explicit local FS copy.

app/domain/authentication/authn_k8s/kubectl_exec.rb
sgnn7: It's in the dependencies: https://github.com/cyberark/conjur/blob/3e2b06c183da59a177f6c06c055c8b6bab8001d1/app/domain/authentication/authn_k8s/kubectl_exec.rb#L58

PR #761, opened 2018-10-13
Fix openshift issues

sgnn7:
#### What does this PR do?
- Fixes error handling when array arg is used
- Fixes throwing of error in authn_k8s when controller cannot be found

https://jenkins.conjur.net/job/cyberark--conjur/job/fix-openshift-issues/

No

### DIFF
--- a/app/domain/util/error_class.rb
+++ b/app/domain/util/error_class.rb
@@ -12,8 +12,9 @@ module Util
           @args = args
         end
         define_method(:to_s) do
-          @args.each.with_index.reduce(msg) do |m,(x,i)|
-            m.gsub(Regexp.new("\\{#{i}}"), x || 'nil')
+          @args.each_with_index.reduce(msg) do |m, (x, arg_index)|
+            x_stringified = x.nil? ? 'nil' : x.to_s
+            m.gsub(Regexp.new("\\{#{arg_index}}"), x_stringified)
           end
         end
       end
--- a/app/domain/authentication/authn_k8s/validate_pod_request.rb
+++ b/app/domain/authentication/authn_k8s/validate_pod_request.rb
@@ -58,7 +58,7 @@ module Authentication
 
       def validate_controller
         return if controller_object
-        raise ControllerNotFound.new(controller, object, namespace)
+        raise ControllerNotFound.new(k8s_host.controller, k8s_host.object, k8s_host.namespace)
       end
 
       def validate_pod_metadata


sgnn7:
@izgeri I think those CodeClimate warnings might be junk:
- Without a nil check, we can't print all falsey value-types for that method (i.e. `false` or `nil` itself). We also cannot use `empty?` or `blank?` since they will not print all value types correctly (e.g. `""`, `false`, `nil`).
- That class is one of the smallest meta-programming classes I've seen around and I don't think that the code can physically be shrunk any further without joining lines together which will hurt readability.

Sidenote: I would probably go further and say that the statement count rule needs to be relaxed as 6 statements max is a rather excessively low default even for regular code.

sgnn7:
@jtuttle / @jonahx Can you take a look at this PR and see what we can do about the codeclimate issues?

app/domain/util/error_class.rb
```diff
@@ -12,8 +12,9 @@ def initialize(*args)
           @args = args
         end
         define_method(:to_s) do
-          @args.each.with_index.reduce(msg) do |m,(x,i)|
-            m.gsub(Regexp.new("\\{#{i}}"), x || 'nil')
+          @args.each_with_index.reduce(msg) do |m, (x, arg_index)|
```
jvanderhoof: @sgnn7, I'm having trouble groking this.  Can you run through what we're trying to accomplish here?

sgnn7: There are multiple problems with the old impl of the class:
- `gsub` cannot auto-convert some types (bools, arrays, fixnums, classes) into strings automatically (error raised)
- `false` was invalidly printed as `nil` since the code does a falsy check
- `each.with_index` is functionally equivalent to `each_with_index` given how we use it

The new code ensures that we use an explicit conversion to string of second gsub param and that we only label 'nil' things that are actually nil and not just falsy.

Try running https://gist.github.com/sgnn7/46dbfbb4bb6d63f527a2b778ddb9315a to see the diff in output

izgeri:
@sgnn7 after talking with @jtuttle it sounds like the CC issues in this PR are not insurmountable. Can you take a look to see if it's easy enough to add a test for at least the `k8s_host` changes?

sgnn7:
Sure and you read my mind lol - was going to see how easy/hard was going to see to add those this morning if the PR wasn't merged yet.

izgeri:
tests will be added with #763 in a future release.

Issue #383, opened 2017-09-19
New template for Integrations pages


typaulhus:
Update landing page for each Integration with:
-Intro blurb
- Integration details
- Link to GitHub
- Next Steps: tutorials, etc.

typaulhus:
These pages are more content updates, and not a template change.

PR #770, opened 2018-10-26
Rollback login implementation for authn-ldap

micahlee:
This PR removes the '/login' endpoint support for authn-ldap to exchange for a conjur role token. Instead the authn-ldap only supports `/authenticate` directly with the ldap credentials.

This change:
- Updates the authn-ldap module to remove the login handler.
- Updates the rspec tests to test LDAP binds using authenticate rather than login.
- Updates the cuke tests to test LDAP auth with authenticate rather than login.


izgeri:
@micahlee does this resolve #383? If so can we please update the description to indicate that? Right now this PR isn't tied to any issues

PR #774, opened 2018-10-30
Housekeeping: Refactor host factories controller a bit

dividedmind:
This fixes all reek and rubocop warnings and errors on this file.

### DIFF
--- a/app/controllers/host_factories_controller.rb
+++ b/app/controllers/host_factories_controller.rb
@@ -1,36 +1,52 @@
 # frozen_string_literal: true
 
+# This controller is responsible for creating host records using
+# host factory tokens for authorization.
 class HostFactoriesController < ApplicationController
   include BodyParser
 
-  before_filter :verify_token,  only: :create_host
+  before_action :validate_token
 
   # Ask the host factory to create a host.
-  # This requires the host factory's token in the Authorization header. 
+  # This requires the host factory's token in the Authorization header.
   def create_host
-    id = params.delete(:id)
-    raise ArgumentError, "id" if id.blank?
-
-    @host_builder = HostBuilder.new(@host_factory.account, id, @host_factory.role, @host_factory.role.layers, params)
-    host, api_key = @host_builder.create_host
+    host, api_key = do_create_host
     response = host.as_json
     response['api_key'] = api_key
-    
+
     render json: response, status: :created
   end
-  
+
   protected
-  
-  def verify_token
-    token = request.headers['Authorization'] or raise Unauthorized
-    if token.to_s[/^Token token="(.*)"/]
-      token = $1
-    else
-      raise Unauthorized
-    end
-    
-    @token = HostFactoryToken.from_token token
-    raise Unauthorized unless @token && @token.valid? && @token.valid_origin?(request.ip)
-    @host_factory = @token.host_factory
+
+  def do_create_host
+    host_factory = hf_token.host_factory
+    role = host_factory.role
+    HostBuilder.new(
+      host_factory.account,
+      params.require(:id),
+      role,
+      role.layers,
+      params.except(:id)
+    ).create_host
+  end
+
+  # Note that while all three of the methods below raise plain Unauthorized on
+  # an error, these are in fact different errors (missing token, token not
+  # found, token expired or wrong origin), it's just we choose to make them
+  # indistinguishable to the client.
+  # Please resist the temptation to roll them all into one conditional.
+
+  def validate_token
+    raise Unauthorized unless hf_token.valid?(origin: request.ip)
+  end
+
+  def hf_token
+    @hf_token ||= HostFactoryToken.from_token(auth_token) or raise Unauthorized
+  end
+
+  def auth_token
+    request.headers['Authorization'].to_s[/^Token token="(.*)"/, 1] \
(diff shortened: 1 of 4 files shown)


app/controllers/host_factories_controller.rb
```diff
@@ -1,34 +1,45 @@
...
+    @hf_token ||= HostFactoryToken.from_token(auth_token).tap do |token|
+      raise Unauthorized unless token && [
+        token.valid?,
+        token.valid_origin?(request.ip)
+      ].all?
+    end
+  end
+
+  def auth_token
+    @auth_token ||= begin
```
jtuttle: Doesn't look like this needs to be memoized. `def auth_token` is only called once so I don't think you need to set `@auth_token`.

app/controllers/host_factories_controller.rb
```diff
@@ -1,34 +1,45 @@
...
+      token = request.headers['Authorization'].to_s
+      token && token[/^Token token="(.*)"/, 1] or raise Unauthorized
```
jtuttle: I don't think the first `token` in this conditional will ever return `false` and cause the conditional to fail. On the above line, `nil.to_s` would evaluate to `""`, which will be `true` for conditional purposes. I believe you can remove the `token &&` on this line without changing the behavior.

app/controllers/host_factories_controller.rb
```diff
@@ -1,34 +1,45 @@
...
+      params.except(:id)
+    ).create_host
+  end
+
+  def hf_token
+    @hf_token ||= HostFactoryToken.from_token(auth_token).tap do |token|
+      raise Unauthorized unless token && [
+        token.valid?,
+        token.valid_origin?(request.ip)
+      ].all?
```
jtuttle: This is neat syntax but might be simpler as a `token_valid?` helper method, like so:

```
def hf_token
  @hf_token ||= HostFactoryToken.from_token(auth_token).tap do |token|
     raise Unauthorized unless token && token_valid?(token)
  end
end

def token_valid?(token)
  token.valid? && token.valid_origin?(request.ip)
end
```

Same number of lines and it's a bit simpler. I could go either way here, though.

app/controllers/host_factories_controller.rb
dividedmind: You're right, just a habit I guess :) I'll remove it.

app/controllers/host_factories_controller.rb
dividedmind: Great point!

app/controllers/host_factories_controller.rb
dividedmind: Reek always complains about utility methods like that which don't touch the instance state. I'm always on the fence whether it has a point or not ;)

jonahx: @dividedmind It does have a point, but technically the solution is to create another object that accepts the param in its initializer, and while sometimes this cleans things up, sometimes the cure is worse than the disease.  My rule of thumb is having 1 or 2 "static" helper methods like that is ok... not worth creating a whole new object.

Also, there's always value in separating out validation into its own step.

All that said, your original and John's suggestion both look ok to me in this case.

dividedmind: @jonahx actually I've found a good existing place for that logic, see next commit. Let me know what you think :)

jonahx: I like where you moved it.

You can remove the `or raise...` here:

```
 def hf_token
    @hf_token ||= HostFactoryToken.from_token(auth_token) or raise Unauthorized
  end
```

because the validate part handles it, and indeed you only want the `validate` one to raise.

dividedmind: It does not actually :) It assumes the token isn't nil to validate.

jonahx: Yeah, I would move the nil check to the validate method then.  I don't like the repeated `raise Unauthorized`

dividedmind: Reek complains about the nil check and is partly right. These are different errors (authorization missing vs. invalid); just because we report it to the client in the same way doesn't change that fact.

dividedmind: I'll add a comment to that effect.

jonahx:
@dividedmind,

> These are different errors (authorization token not found vs. invalid due to expiration or origin); just because we report it to the client in the same way doesn't change that fact.

That's fine.  It's orthogonal to the point I was making.  You still don't want to mix validation with object creation or lookup.  Having the `or raise Unauthorized` in `hf_token` is simply not what you want from a method named `hf_token`.  A method with that name should return a token, not raise http errors.

> Reek complains about the nil check and is partly right.

Reek is 100% right but again it's an orthogonal problem and it's not avoided with the code you have now.  `or raise Unauthorized` is also a nil check.  

What reek is asking for is a type that properly represents "the return value of `HostFactoryToken.from_token`".  You're probably familiar with this as a Maybe or Result type, or as the Null object pattern.  This is ultimately a better approach imo, but it's also a much bigger change, and we don't do this anywhere else in our code, and it's also not "the rails way."  Hence I'm trying  to be pragmatic with recommending the safe navigation operator as a much lighter weight alternative.

PR #791, opened 2018-11-13
Add BodyParser to HostFactoryTokensController

dsbyrne:
Fixes request params not being parsed from the body of `x-www-form-urlencoded` requests when making requests to create or delete a host factory token


sgnn7:
Yay!

Issue #830, opened 2019-01-14
Pre-compile assets before packaging


sgnn7:
Assets for Conjur (possum) are not pre-compiled before packaging so if the container is just using short-timeout health/readiness checks the asset pipeline will keep restarting the process without success and the container will never be "Ready".

AC:
- [ ] Assets for Conjur/Possum are pre-compiled in the packaging

sgnn7:
CC: @micahlee

PR #831, opened 2019-01-21
Pre-compile Rails Assets on Docker Build

micahlee:
#### What does this PR do?
This PR updates the docker build to precompile the Rails assets for Conjur.

Previously, the initial connection to the Conjur container would take several seconds to complete due to asset compilation. This caused issues in systems such as Kubernetes, where the slow initial connection could surpass the health check timeout and cause the health check to fail. 

Connected to #830 

https://jenkins.conjur.net/job/cyberark--conjur/job/830-asset-precompile/

Yes

### DIFF
--- a/config/initializers/sequel.rb
+++ b/config/initializers/sequel.rb
@@ -2,7 +2,6 @@
 
 Sequel.split_symbols = true
 Sequel.extension :core_extensions, :postgres_schemata
-Sequel::Model.db.extension :pg_array, :pg_inet, :pg_hstore
 Sequel::Model.plugin :validation_helpers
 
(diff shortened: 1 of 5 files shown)


config/initializers/sequel.rb
```diff
@@ -2,7 +2,6 @@
 
 Sequel.split_symbols = true
 Sequel.extension :core_extensions, :postgres_schemata
-Sequel::Model.db.extension :pg_array, :pg_inet, :pg_hstore
```
micahlee: This is moved in to `config/application.rb` to allow us to set the extension conditionally (it requires a working database connection).

PR #881, opened 2019-02-12
Strategy refactor

### LINKED ISSUE #852: Dev - Refactoring strategy
childof #838

orenbm:
#### What does this PR do?
Refactors Strategy class and splits each method (login, authenticate, oidc_login, oidc_authenticate) into it's own CommandClass. 

This makes the authenticator flow more readable and testable.

The design of the Strategy class was good at the beginning when all authenticators worked the same and the flow was simple. Once other authentication methods were introduced, the Strategy class became hard to maintain. 

closes #852 

authenticate_controller -> authentication/authenticate.rb, authentication/login.rb, authentication/authn-oidc/oidc_authenticate.rb, authentication/authn-oidc/oidc_login.rb

Run Postman requests for OIDC login & authenticate

https://jenkins.conjur.net/job/cyberark--conjur/job/strategy-refactor/ 

> Does this work have automated integration and unit tests?
Yes

No

N/A

No


app/domain/authentication/login.rb
```diff
@@ -0,0 +1,58 @@
...
+    def login(input)
+      authenticator = @authenticators[input.authenticator_name]
+
+      @validate_authenticator_exists.(input: input, authenticator: authenticator)
+      @validate_security.(input: input, enabled_authenticators: @enabled_authenticators)
+
+      key = authenticator.login(input)
+      raise InvalidCredentials unless key
+
+      @audit_event.(
```
jonahx: This might all fit on one line.

app/domain/authentication/login.rb
```diff
@@ -0,0 +1,58 @@
...
+        input: input,
+          success: true,
+          message: nil
+      )
+
+      new_login(input, key)
+    rescue => e
+      @audit_event.(
```
jonahx: this too...

app/domain/authentication/validate_authenticator_exist.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Authentication
+
```
jonahx: Change `Exist` to `Exists` if you can do so easily in your IDE...

app/domain/authentication/validate_authenticator_exist.rb
```diff
@@ -0,0 +1,21 @@
...
+  ) do
+
+    def call
+      validate_authenticator_exists
+    end
+
+    private
+
+    def validate_authenticator_exists
+      raise AuthenticatorNotFound, @input.authenticator_name unless @authenticator
```
jonahx: As we discussed the other day, I believe the class `ValidateAuthenticatorExists` is superfluous.  Where you would write in client code:
```
@validate_authenticator_exists.(input, authenticator)
```
you can just as easily write:
```
raise AuthenticatorNotFound, input.authenticator_name unless authenticator
```
Note that in general a CommandClass with no dependencies is equivalent to a pure function whose arguments are the same arguments as the command class's `call` method.  

In this case, it's not technically pure because you can raise an error.  However, you might as well just use a function with a verb phrase name or, if the function is only a single line as is the case here, you can inline it where you need it.  But I think adding the machinery of `CommandClass` here is unnecessary and potentially confusing.

app/domain/authentication/validate_origin.rb
```diff
@@ -0,0 +1,26 @@
...
+    },
+    inputs: %i(input)
+  ) do
+
+    def call
+      validate_origin
+    end
+
+    private
+
```
jonahx: I would put the two lines of code in `validate_origin` directly inside `call` (after all, the class name `ValidateOrigin` describes what the `call` method does, so it's already described by that).

app/domain/authentication/validate_security.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Authentication
+  ValidateSecurity = CommandClass.new(
+    dependencies: {
+      security: ::Authentication::Security.new
+    },
+    inputs: %i(input enabled_authenticators)
+  ) do
+
```
jonahx: same thing here, just move line 18 directly inside call and delete the private method `validate_security`.

app/models/get_role_by_login.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+GetRoleByLogin = CommandClass.new(
+  dependencies: {
+    role_cls: ::Role
+  },
+  inputs: %i(username account)
+) do
+
```
jonahx: Same thing here, I'd move line 17 into call and delete `role`.

spec/app/domain/authentication/authenticate_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+  # (_  _)( )_( )( ___)  (_  _)( ___)/ __)(_  _)/ __)
+  #   )(   ) _ (  )__)     )(   )__) \__ \  )(  \__ \
+  #  (__) (_) (_)(____)   (__) (____)(___/ (__) (___/
+
+
+  context "An unavailable authenticator" do
+    subject do
+      input_ = input(
+        authenticator_name: 'AUTHN-MISSING'
+      )
```
jonahx: formatting note: remove line breaks if putting it all on one line is under 80 chars or so.

spec/app/domain/authentication/authenticate_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+  ####################################
+  # Security doubles
+  ####################################
+
+  let (:mocked_security_validator) {double("MockSecurityValidator")}
+  let (:mocked_origin_validator) {double("MockOriginValidator")}
+
+  before(:each) do
+    allow(Authentication::ValidateSecurity)
+      .to receive(:new)
```
jonahx: Ok, this is the first real conceptual error I've found so far.  We never want to do our mocks like this, where we are changing the behavior of a class name to act differently from how it would in production.  

In fact, in many ways the whole point of the command class dependencies is so that we can inject mock objects directly into the command class constructors in tests.  That is, the interface exists so we don't have to do this "Make the hardcoded thing act different in tests" hack.   Stuff like `Authentication::ValidateSecurity.new` in the production just gives our command classes _default_ dependencies so our production client code doesn't have to explicitly name them everywhere.  But we do expect to have to explicitly override these defaults in the tests.

spec/app/domain/authentication/authenticate_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+    it "raises AuthenticatorNotFound" do
+      expect {subject}.to raise_error(
+                            Authentication::AuthenticatorNotFound
+                          )
+    end
+  end
+
+  context "An available authenticator" do
+    context "that receives invalid credentials" do
+      subject do
```
jonahx: I think we should remove the `input` helper function, and instead explicitly create the input we want in each test.  We want the tests to be fully independent (here they are coupled via the helper function `input`) and to be locally readable (here we have to scroll back up to `input` to remind ourselves what the default values are when reading.  This is a case where duplication is actually desirable.

spec/app/domain/authentication/authenticate_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+      authenticator_name: authenticator_name,
+      service_id: service_id,
+      account: account,
+      username: username,
+      password: password,
+      origin: origin,
+      request: request
+    )
+  end
+
```
jonahx: I don't think this abstraction buys us anything.   Wherever you would currently write:
```
authenticators['authn-always-pass']
```
just write:
```
authenticator(pass: true)
```
It's equally readable and clear.

spec/app/domain/authentication/authn-oidc/oidc_spec.rb
```diff
@@ -0,0 +1,242 @@
...
+  let (:mocked_oidc_authenticator) {double("MockOidcAuthenticator")}
+  let (:mocked_security_validator) {double("MockSecurityValidator")}
+  let (:mocked_origin_validator) {double("MockOriginValidator")}
+
+  before(:each) do
+    allow(Resource).to receive(:[])
+                         .with(/#{account}:variable:conjur\/authn-oidc/)
+                         .and_return(mocked_resource)
+
+    allow(Authentication::AuthnOidc::Authenticator)
```
jonahx: See comment above about overriding hardcoded class behavior.

app/domain/authentication/login.rb
orenbm: I don't think so. `key` is needed later on as an input to `new_login` (row 37)

orenbm: @jonahx am I wrong?

orenbm: oh you were talking about the `audit_event` line. got it

app/domain/authentication/login.rb
orenbm: done

app/domain/authentication/validate_authenticator_exist.rb
orenbm: Yes, I was doubtful about this one

spec/app/domain/authentication/authenticate_spec.rb
orenbm: got it

spec/app/domain/authentication/authenticate_spec.rb
orenbm: Fun fact: it's your code :)

Anyway, this is needed as the line you mention (`authenticators['authn-always-pass']`) is not on the test code but in prod. We just pass `authenticators` as input to the `Authentication::Authenticate` CommandClass.

Let me know if I'm wrong.

spec/app/domain/authentication/authenticate_spec.rb
orenbm: cool, I didn't know it's possible to override a dependency in your CommandClass. Refactored this throughout the code

spec/app/domain/authentication/authenticate_spec.rb
jonahx: > Fun fact: it's your code :)

jonahx: Ignore the above comment.  Per our slack convo, the `authenticators['authn-always-pass']` stuff is fine as is.

PR #883, opened 2019-02-15
update conjur-policy-parser

apotterri:
#### What does this PR do?
Update to v3.0.3 of the policy parser, to support annotations on policy
objects.

The lack of this functionality appears to be due to an oversight in the original implementation.

Connected to conjurinc/conjur-policy-parser#13

Changelog, yes. I'm not sure whether the version should be updated now or not.

Yes: https://github.com/conjurinc/conjur-policy-parser/blob/possum/spec/types/policy_spec.rb#L3

no

no

no

### DIFF
--- a/ci/parse-changelog
+++ b/ci/parse-changelog
@@ -1,0 +1,11 @@
+#!/bin/bash -ex
+
+cd "$(dirname $0)"
+
+docker run --rm \
+  -v $PWD/..:/src/conjur-server \
+  -w /src/conjur-server \
+  ruby:2.5 bash -ec "
+    gem install -N parse_a_changelog
+    parse ./CHANGELOG.md
+  "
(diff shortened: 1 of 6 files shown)


apotterri:
@jtuttle If you have a minute, can you take a look at this, please? I changed the way `parse_a_changelog` gets invoked, want to make sure I'm not missing something.

ci/parse-changelog
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -ex
+
+cd "$(dirname $0)"
+
+docker run --rm \
```
jtuttle: Thanks, I did it this way in `ldap-sync` after realizing what I did in `conjur` was silly but hadn't had time to go back and fix it yet.

apotterri: No worries, thanks for taking a look.

PR #888, opened 2019-02-25
Verifying that username exists before auditing in OIDC

### LINKED ISSUE #887: audit event doesn't crash when no username is displayed

orenbm:
#### What does this PR do?
fixes a bug in which the user got an indiscriptive exception of nul pointer rather than the real problem. This happens  when we have an exception before we get the username from the OIDC provider.

closes #887 

oidc-login
oidc-authenticate

Run oidc flow with an invalid grant and verify the message says that

### DIFF
--- a/spec/app/domain/authentication/authn-oidc/oidc_spec.rb
+++ b/spec/app/domain/authentication/authn-oidc/oidc_spec.rb
@@ -272,7 +272,7 @@ RSpec.describe 'Authentication::Oidc' do
       end
     end
 
-    context "that receives authenticate request" do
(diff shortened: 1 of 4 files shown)


spec/app/domain/authentication/authn-oidc/oidc_spec.rb
```diff
@@ -224,5 +271,35 @@
                               )
       end
     end
+
+    context "that receives authenticate request" do
```
jtuttle: Could you jsut add `"and fails on oidc details retrieval"` to the end of this description to be consistent with the one above?

PR #891, opened 2019-02-27
874 oidc authn e2e

### LINKED ISSUE #874: Dev end to end flow with hardcoded email as Conjur user in root policy
childof #838

orenbm:
#### What does this PR do?
Enables e2e development in parallel on the oidc authenticate. I refactored the old authenticate flow and it is not accessed now. I also created the template for the new authn oidc flow with hardcoded username so developers can code in parallel on different parts of the feature.

closes #874 

authenticate_controller -> authn_oidc/authenticate.rb

start the server with the `--authn-oidc` flag
Run the `authenticate-oidc` request in the Postman Collection

### DIFF
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -70,7 +70,7 @@ class AuthenticateController < ApplicationController
   #
   # Returns Conjur access token
   def authenticate_oidc
-    authentication_token = ::Authentication::AuthnOidc::Authenticate.new.(
+    authentication_token = ::Authentication::AuthnOidc::AuthenticateOidcConjurToken.new.(
       authenticator_input: oidc_authenticator_input
     )
     render json: authentication_token
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -69,8 +69,17 @@ class AuthenticateController < ApplicationController
   # - Introspect ID Token
   #
   # Returns Conjur access token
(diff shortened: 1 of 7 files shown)


InbalZilberman:
@orenbm - few comments:
1. Can you make alice a user that looks as much as possible a result of an LDAP sync see here: https://developer.conjur.net/reference/services/ldap_sync use annotations and so on.
2. Can you please add the var we agreed on IDTokenUserProperty to the list of vars?
3. I would expect the full flow to have alice perform something in the system. I know the fetching of the secret is not the main course of the authn but it is the result you would of expect from it.

orenbm:
thanks @InbalZilberman for the comments!

1. This scenario isn't really with LDAP. We don't have a group, only one user which we want to see that is authorized. So the only thing I can add is the annotations but what's the point in that? It doesn't test anything.

2. I added a ticket for this. Once we implement it in the app we will add it to the tests.

3. Fetching a secret is tested in other scenarios, where a user has a conjur token and we check that he can retrieve a secret. This integration test is for oidc to see that a valid user gets a token.

InbalZilberman:
1. We wont know how annotation affect if we wont try it. As we expect this to be our main flow let's try as much as possible. 
3. I think this is very important to test that the access token that the oidc-authn provide is valid and one can use it to perform any other action in Conjur ( get secret is an example). 
If this is not part of this CL please make sure it is done in the test plan.

PR #895, opened 2019-03-03
Extracting conjur username from id token

### LINKED ISSUE #894: extract preferred username from token

orenbm:
#### What does this PR do?
Reads the id token from the input of the oidc authn request and extracts the 'preferred_username' field

At this point we assume the input is id token and that the username is in the 'preferred_username'  field.

closes #894


app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -20,6 +20,8 @@ def call
       private
 
       def access_token(input)
+        request_body = AuthnOidc::AuthenticateRequestBody.new(input.request)
```
nessiLahav: Do you think it should be a command class ? (because its only action is to parse the request )

app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -51,6 +52,14 @@ def new_token(input)
           username: input.username
         )
       end
+
+      def conjur_username(request_body)
+        id_token_username_field = "preferred_username"
+        conjur_username = request_body.id_token[id_token_username_field]
+        raise IdTokenFieldNotFound, id_token_username_field unless conjur_username
```
nessiLahav: Check if there is any possibility that the user name will be empty from some reason

app/domain/authentication/authn_oidc/authenticate.rb
orenbm: It's not really an action, it's more of an object representing the request body of the authn oidc request. You can look at `request_body.rb` (changed to `login_request_body.rb` in this PR) to see how we do it same for the login request.

In case the request will change and we will add more parameters to the input it will be easier if this will stay as a request object

authn-oidc-files/keycloak/create_user
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+keycloak/bin/kcadm.sh config credentials \
+  --server http://localhost:8080/auth \
+  --realm master \
+  --user "$KEYCLOAK_USER" \
+  --password "$KEYCLOAK_PASSWORD"
+keycloak/bin/kcadm.sh create users \
+  -s username=alice \
```
nessiLahav: Lets fetch this info from environment variable

authn-oidc-files/phantomjs/fetchAuthCode
```diff
@@ -11,8 +11,8 @@ cp -rf keycloak_login.js keycloak_login_tmp.js
 
 echo "Replacing user and password in keycloak_login_tmp.js"
 
-sed -i 's/TO_BE_REPLACED_USER/'"$KEYCLOAK_USER"'/g' keycloak_login_tmp.js
-sed -i 's/TO_BE_REPLACED_PASSSWORD/'"$KEYCLOAK_PASSWORD"'/g' keycloak_login_tmp.js
+sed -i 's/TO_BE_REPLACED_USER/'"alice"'/g' keycloak_login_tmp.js
```
nessiLahav: Lets fetch this info from environment variable

app/domain/authentication/authn_oidc/authenticate.rb
orenbm: good call. I handled this and added a UT.

btw, the expected behaviour IMO is the same as with no preferred_username at all.

authn-oidc-files/keycloak/create_user
orenbm: Done

authn-oidc-files/phantomjs/fetchAuthCode
orenbm: Done

app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -28,8 +30,7 @@ def access_token(input)
 
         # Validate ID Token is active
 
-        # Retrieve mail from response
-        conjur_username = "alice"
+        conjur_username = conjur_username(request_body)
```
jtuttle: You could pass `request_body.id_token` into this method since the method doesn't need anything else from the request body.

app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -51,6 +52,14 @@ def new_token(input)
           username: input.username
         )
       end
+
+      def conjur_username(request_body)
+        id_token_username_field = "preferred_username"
+        conjur_username = request_body.id_token[id_token_username_field]
+        raise IdTokenFieldNotFound, id_token_username_field unless conjur_username&.present?
```
jtuttle: I don't think you would ever need the safe navigation operator `&.` when using `present?`. The method is available on `nil`.

authn-oidc-files/keycloak/create_user
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
jtuttle: Are these files in `authn-oidc-files` used in the CI process? If so, could we move them into the `ci` folder? Seems awfully specific to a certain part of Conjur to be living at the top level.

app/domain/authentication/authn_oidc/authenticate.rb
orenbm: done

app/domain/authentication/authn_oidc/authenticate.rb
orenbm: correct. didn't know that.

authn-oidc-files/keycloak/create_user
orenbm: @jtuttle they are used both in ci and in dev. that's why there in a common path. Do you still think they should be in `ci` and referred to from `dev` as `../ci/authn-oidc-files/keycloak`?

orenbm: after our conversation in slack, moving to `ci` folder

PR #897, opened 2019-03-03
added variable for id token user property

### LINKED ISSUE #889: User can modify the id token property to link to Conjur username
Another variable called "IDTokenUserProperty" will be set with the name of the property in the ID Token by which we will map the user name in Conjur.

If the variable doesn’t exist, or it has no value, the app should fail as we do with other oidc vars

orenbm:
#### What does this PR do?
the user property for ID token is configurable via a variable

we want to let the user configure the property from the id token with which she'd like to map to a conjur username

closes #889


authn-oidc-files/policy.yml
```diff
@@ -14,6 +14,9 @@
   - !variable
     id: provider-uri
 
+  - !variable
```
InbalZilberman: I believe this var is defined twice in the same policy

app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -53,8 +58,7 @@ def new_token(input)
         )
       end
 
-      def conjur_username(request_body)
-        id_token_username_field = "preferred_username"
+      def conjur_username(request_body, id_token_username_field)
         conjur_username = request_body.id_token[id_token_username_field]
         raise IdTokenFieldNotFound, id_token_username_field unless conjur_username&.present?
```
InbalZilberman: As I am not really familiar with the code..  I maybe asking a dummy question
the logic of this POC is only for conjur users that are located in root (and root only) where in the code do you make this assumption?

authn-oidc-files/policy.yml
orenbm: Yes, we have 2 policies in this file - one for okta and one for keycloak

app/domain/authentication/authn_oidc/authenticate.rb
orenbm: we don't. why should we limit this in the code?

orenbm: it is a limitation we have by design

cucumber/authenticators/features/step_definitions/authn_common_steps.rb
```diff
@@ -1,3 +1,5 @@
+require_relative '../../../../cucumber/api/features/step_definitions/authz_steps'
```
jtuttle: Are you only using a single step definition from this file? If so, I would just duplicate it here and then add a TODO comment and a GitHub issue that say something about finding a good way to share step definitions between different cucumber categories.

orenbm: done

Issue #896, opened 2019-03-03
test full oidc flow - with secret retrieval


orenbm:
Test user annotations & secret retrieval

PR #901, opened 2019-03-06
Verify OIDC authenticate returns a token that is valid for secret retrieval

orenbm:
#### What does this PR do?
I added to the OIDC authenticate integration test a validation that the retrieved token is valid also for secret retrieval.

At this point the Conjur OIDC token factory inherits the regular Conjur token. In case sometime in the future this will change we want to verify the token is still a valid conjur token

Connected to #896 

authn_oidc.feature

run the authenticators cukes

### DIFF
--- a/cucumber/api/features/step_definitions/request_steps.rb
+++ b/cucumber/api/features/step_definitions/request_steps.rb
@@ -27,6 +27,12 @@ When(/^I( (?:can|successfully))? DELETE "([^"]*)"$/) do |can, path|
   end
 end
 
+When(/^I( (?:can|successfully))? GET "([^"]*)" with user "([^"]*)"$/) do |can, path, username|
+  try_request can do
+    get_json path, user: username
+  end
+end
+
 When(/^I( (?:can|successfully))? GET "([^"]*)" with parameters:$/) do |can, path, parameters|
   params = YAML.load(parameters)
   path = [ path, params.to_query ].join("?")
--- a/cucumber/api/features/step_definitions/request_steps.rb
+++ b/cucumber/api/features/step_definitions/request_steps.rb
@@ -27,9 +27,9 @@ When(/^I( (?:can|successfully))? DELETE "([^"]*)"$/) do |can, path|
   end
 end
 
-When(/^I( (?:can|successfully))? GET "([^"]*)" with user "([^"]*)"$/) do |can, path, username|
+When(/^I( (?:can|successfully))? GET "([^"]*)" with authorized user$/) do |can, path|
   try_request can do
-    get_json path, user: username
+    get_json path, token: ConjurToken.new(@response_body)
   end
 end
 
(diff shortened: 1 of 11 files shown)


cucumber/api/features/step_definitions/request_steps.rb
```diff
@@ -27,6 +27,12 @@
   end
 end
 
+When(/^I( (?:can|successfully))? GET "([^"]*)" with user "([^"]*)"$/) do |can, path, username|
+  try_request can do
+    get_json path, user: username
+  end
+end
+
```
orenbm: @jtuttle I added this step so we can run a GET request with the last retrieved token. Should I add also methods for PUT, POST & DELETE? I don't need them at the moment

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -36,19 +36,40 @@ Feature: Users can authneticate with OIDC authenticator
       member: !user alice
 
     - !user alice@conjur.net
```
orenbm: @jtuttle I'd prefer not to have this in the background policy but to add a new policy in the `A valid id token with email as id-token-user-property` scenario. This, however, doesn't work. do you know why?

You can reproduce by removing this line and adding this to the beginning of the scenario:
```
Given a policy:
    """
   - !user alice@conjur.net
    """
```

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -36,19 +36,40 @@ Feature: Users can authneticate with OIDC authenticator
...
     And I am the super-user
     And I successfully set OIDC variables
 
   Scenario: A valid id token to get Conjur access token
-    When I successfully authenticate via OIDC
+    When I successfully authenticate via OIDC with id token:
+    """
+    {"preferred_username": "alice","email": "alice@conjur.net"}
+    """
+
```
jtuttle: Can you remove this blank line plz so that the test is continuous? Same with the ones below.

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -36,19 +36,40 @@ Feature: Users can authneticate with OIDC authenticator
...
     Then "alice" is authorized
 
+    # Verify the returned access token is valid for retrieving a secret
+    Given a policy:
```
jtuttle: Usually you would put all of the "Given" blocks at the beginning of a cucumber test. Cucumber tests generally follow a `Given`-`When`-`Then` flow (with `And`s inbetween if needed). It's easier to reason about them this way.

cucumber/authenticators/features/authn_oidc.feature
orenbm: done

cucumber/authenticators/features/authn_oidc.feature
orenbm: done

PR #904, opened 2019-03-07
Rearrange primary key on permissions table

dividedmind:
The primary key used to be (privilege, resource_id, role_id) which is not very useful for how some queries are structured and this mismatch led to abysmal performance of some of them with a lot of permissions present.

Nothing seems to search primarily by privilege; OTOH eg. visibility check searches by resource_id and extracts role_id which should be possible with an index only scan after rearranging to (resource_id, role_id, privilege).

### DIFF
--- a/db/migrate/20190307154241_change_permissions_primary_key.rb
+++ b/db/migrate/20190307154241_change_permissions_primary_key.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  # The primary key used to be (privilege, resource_id, role_id) which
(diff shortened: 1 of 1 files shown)


db/migrate/20190307154241_change_permissions_primary_key.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  # The primary key used to be (privilege, resource_id, role_id) which
```
micahlee: Great explanation!

Issue #907, opened 2019-03-11
Conjur server should not be started when running in `dev/server/


orenbm:
The server should not be started when the start script is finished. currently we need to kill the process before we can run it again, this should be fixed

PR #909, opened 2019-03-12
Killing conjur server process after loading variables in dev/start

orenbm:
#### What does this PR do?
Killing the conjur server process after using it for loading variables.

In the oidc development we needed to start the server in order to load variables and values. This made the container start with the server already started which caused problems.

Connected to #907 

run `dev/start --authn-oidc` and inside the container run `ps -ef` to verify the server is down.

### DIFF
--- a/dev/files/killConjurServer
+++ b/dev/files/killConjurServer
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+kill $(ps aux | grep '[r]uby /usr/local/bin/conjurctl server' | awk '{print $2}')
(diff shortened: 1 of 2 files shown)


dev/files/killConjurServer
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+kill $(ps aux | grep '[r]uby /usr/local/bin/conjurctl server' | awk '{print $2}')
```
jtuttle: Looks like everything else in this folder is in a subdirectory corresponding to the relevant authenticator. Since this is only used in the OIDC authenticator, should we put it in an `authn-oidc` subdir?

orenbm: @jtuttle i thought about that but this isn't really oidc logic and it could be used in other scenarios that need loading variables. That's why i put this in the general folder. wdyt?

jtuttle: Yeah okay, that's fine.

Issue #910, opened 2019-03-13
Delete client-id and client-secret


nessiLahav:
childof #838

Pay attention that our previous implementation still need those variables

PR #911, opened 2019-03-13
Remove redundant oidc vars

orenbm:
#### What does this PR do?
As we don't need anymore the client-id & client-secret from OIDC I removed it from production & test code. The login flow still works as I removed the ClientConfiguration object and replaced it with a generic GetOIDCSecrets.

connected to #910


app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -23,10 +23,11 @@ def call
       def access_token(input)
         request_body = AuthnOidc::AuthenticateRequestBody.new(input.request)
 
-        oidc_client_configuration = @get_oidc_client_configuration.(
-          redirect_uri: nil, # not needed for this request
-            service_id: input.service_id,
-            conjur_account: input.account
+        required_variable_names = %w(provider-uri id-token-user-property)
+        oidc_secrets            = @fetch_oidc_secrets.(
```
jtuttle: I would remove the space between `oidc_secrets` and `=` here. This kind of indentation can be useful when you have lists of one-line variables being set but I don't think it's very useful here.

app/domain/authentication/authn_oidc/fetch_oidc_secrets.rb
```diff
@@ -0,0 +1,58 @@
+require 'command_class'
+require 'conjur/fetch_required_secrets'
+
+module Authentication
+  module AuthnOidc
+
+    # TODO: (later version) Fix CommandClass so we can add errors directly
+    # inside of it
+    #
+    # TODO: list any OIDC or other errors here.  The errors are part of the
```
jtuttle: Do we have issues for these `TODO`s so that they don't get forgotten?

app/domain/authentication/authn_oidc/fetch_oidc_secrets.rb
```diff
@@ -0,0 +1,58 @@
...
+    # Errors from FetchRequiredSecrets
+    #
+    RequiredResourceMissing = ::Conjur::RequiredResourceMissing
+    RequiredSecretMissing   = ::Conjur::RequiredSecretMissing
+
+    FetchOidcSecrets = CommandClass.new(
+      dependencies: {
+        fetch_secrets: ::Conjur::FetchRequiredSecrets.new
+      },
+      inputs:       %i(conjur_account service_id required_variable_names)
```
jtuttle: Why the space after `inputs:` here?

app/domain/authentication/authn_oidc/fetch_oidc_secrets.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      def variable_id(var_name)
+        "#{@conjur_account}:variable:conjur/authn-oidc/#{@service_id}/#{var_name}"
+      end
+
+      def secret_value(var_name)
+        required_secrets[variable_id(var_name)]
+      end
+
+      def oidc_secrets
```
jtuttle: I found this method really hard to follow because it relies on so many complex chained-together helper methods. I know it's generally good practice in Ruby to minimize method size, but I would suggest scaling it back a bit here. I find something like this (which I think does the same thing) easier to follow:

```
def required_resource_ids
  @required_variable_names.map { |var_name| variable_id(var_name) }
end

def variable_id(var_name)
  "#{@conjur_account}:variable:conjur/authn-oidc/#{@service_id}/#{var_name}"
end

def oidc_secrets
  secrets = {}

  required_secrets = @fetch_secrets.(resource_ids: required_resource_ids)
  
  @required_variable_names.each do |variable_name|
    full_var_name = variable_id(var_name)
    secrets[variable_name] = required_secrets[full_var_name]
  end

  secrets
end
```

jtuttle: Note that this is no longer memoizing `required_secrets`, but I think you were only doing that because it was being constructed inside the loop, which I don't think is necessary in this rewrite.

app/domain/authentication/authn_oidc/authenticate.rb
orenbm: done

app/domain/authentication/authn_oidc/fetch_oidc_secrets.rb
orenbm: created an issue for the first TODO and removed it.

The second one is just wrong, i removed it. we have another place to list OIDC errors

app/domain/authentication/authn_oidc/fetch_oidc_secrets.rb
orenbm: just default indentation of RubyMine. Changed it now in preferences.

app/domain/authentication/authn_oidc/fetch_oidc_secrets.rb
orenbm: i agree, this looks clearer

orenbm:
> My head isn't quite in the OIDC stuff enough to know what the reason for this change was. Could you give me some context? Was this due to the decision to validate the token locally instead of calling out to the authenticator service?

We needed the `client-id` & `client-secret` for the login flow in the previous feature. In this use-case, as you can see in @moticless 's PR, we don't need these parameters for verifying the id_token.

Issue #875, opened 2019-02-11
Dev - ID Token is validated


orenbm:
childof #838

blocked by #893

PR #914, opened 2019-03-14
Decoding and verifying ID Token

moticless:
#### What does this PR do?
Adding ID Token decoding (from JWT format), verifying signing and expiration.
none

Connected to #903 
Connected to #875 
start with authenticate.rb and then decode_and_verify_id_token.rb 
have configured provider. Retrieve from provider ID Token, and POST Authenticate with the ID Token


app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -29,15 +31,9 @@ def access_token(input)
...
         )
 
-        # Prepare ID token introspect request
-
-        # send id token to OIDC Provider
-
-        # Get JSON from OIDC Provider
-
-        # Validate ID Token is active
+        id_token_attributes = @decode_and_verify_id_token.(oidc_client_configuration.provider_uri, request_body.id_token)
```
jtuttle: Can you please split this up over a few lines so it isn't so long (same with the line below)?

app/domain/authentication/authn_oidc/decode_and_verify_id_token.rb
```diff
@@ -0,0 +1,34 @@
...
+  module AuthnOidc
+    DecodeAndVerifyIdToken = CommandClass.new(
+      dependencies: {
+      },
+      inputs:       %i(provider_uri id_token_jwt)
+    ) do
+
+      def call
+        # decode id token and Validate signing.
+        @id_token_decoded = OpenIDConnect::ResponseObject::IdToken.decode(id_token_jwt, get_cert(provider_uri))
```
jtuttle: Please split up any lines over 80 chars in this file.

app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -30,17 +31,20 @@ def access_token(input)
...
-
-        # send id token to OIDC Provider
-
-        # Get JSON from OIDC Provider
-
-        # Validate ID Token is active
+        id_token_attributes = @decode_and_verify_id_token.(
+          provider_uri: oidc_secrets["provider-uri"],
+            id_token_jwt: request_body.id_token
+        )
```
jonahx: Regarding the structure of the `call` method here: A couple points about how `CommandClass` is intended to be used.  

1. Currently `call` calls `access_token`, which has the actual high-level logic.  We probably want to move the logic that's currently in `access_token` into `call`.  Remember the name of the command class should be enough to summarize what it does.  If `Authenticate` is not clear enough, the command class itself can be named `FetchAccessToken` or `FetchAuthenticationToken`, etc.  `call` then carries out the steps implicit in that name.

2. The call method itself should read like a table of contents, or a series of sentences.  One of the huge benefits of command classes is that they reveal your high-level logic as clearly as a bulleted list of steps would, in documentation.  The philosphy is: The code itself should be as clear as you describing in English what it does.  For example, here's the `call` method in the [K8s authenticator](https://github.com/cyberark/conjur/blob/6b92244b7481bcf1f72f2ad3896dddcbcf0675f0/app/domain/authentication/authn_k8s/authenticator.rb#L18):

```
      def call
        validate_cert_exists
        validate_the_request
        validate_header_cert
      end
```

In the current implementation, things are starting to get away from us just a bit... we have the details of parameter passing, local variables, etc, obscuring our high-level logic.  The ideal is: You never have to read _code_ to figure out what the code does.  Instead, you're reading English sentences (in the form of method names) to figure it out.  *If* you are interested in more detail, you can jump to those method definitions.

app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -59,8 +63,8 @@ def new_token(input)
         )
       end
 
-      def conjur_username(id_token, id_token_username_field)
-        conjur_username = id_token[id_token_username_field]
+      def conjur_username(id_token_attributes, id_token_username_field)
```
jonahx: Note: You should never have to pass parameters around like when using a `CommandClass`.  It's explicitly designed to relieve you of this burden (I should document this specifically in the gem docs, in fact).  

All of our inputs are available as member variables.  And anything *derived* from our inputs can be made available as a (memoized, if needed) helper method.

You need not worry about concurrency or multiple requests calling `call` simultaneously.  Under the hood, every time `call` is invoked a new private `Call` object is created.  The input member variables exist on *that* object.  So it is completely thread safe.

app/domain/authentication/authn_oidc/decode_and_verify_id_token.rb
```diff
@@ -0,0 +1,38 @@
...
+
+module Authentication
+  module AuthnOidc
+    DecodeAndVerifyIdToken = CommandClass.new(
+      dependencies: {
+      },
+      inputs: %i(provider_uri id_token_jwt)
+    ) do
+
+      def call
```
jonahx: See my comments above on the other `call` method.  Those would apply here too.

Issue #879, opened 2019-02-11
Prepare OIDC code for next iteration


orenbm:
childof #838

As we had 2 iterations for the OIDC feature (login flow & authn id token flow), we have some code that will not run in production but will be re-visited in the future for further development. We want to keep the code but to keep the whole repo clean. What we will do is:

- disable the login flow feature in cucumber
- restructure oidc code so we have separation between the flows
- disable the route from the REST call to the authn-controller

PR #926, opened 2019-03-20
Restructure oidc flows

orenbm:
#### What does this PR do?
Separate OIDC files to folders and modules by API Call. I did this change for the production code and for specs.

Please take extra care to the OIDC spec separation. In order to separate the specs to 3 files, there is some duplicated code. I wasn't sure what the best practice here between the following options:

- separate to 3 files with duplications
- have 1 `oidc_spec_helper` file with the common code
  - is this possible?
  - this makes less readability. wdyt about that?
- leave the specs in one file like we had before

Please let me know what you think is the best option.

As we had 2 iterations for the OIDC feature (login flow & authn id token flow), we have some code that will not run in production but will be re-visited in the future for further development. We want to keep the code but to keep the whole repo clean. What we will do is:

- disable the login flow feature in cucumber
- restructure oidc code so we have separation between the flows
- disable the route from the REST call to the authn-controller

Connected to #879 

run rspec & cucumber


app/domain/authentication/authn_oidc/get_conjur_oidc_token/client.rb
```diff
@@ -0,0 +1,80 @@
...
+    module GetConjurOidcToken
+      class Client
+        def initialize(client_id:, client_secret:, redirect_uri:, provider_uri:)
+          @client_id = client_id
+          @client_secret = client_secret
+          @redirect_uri = redirect_uri
+          @provider_uri = provider_uri
+        end
+
+        def oidc_id_token_details!(authorization_code)
```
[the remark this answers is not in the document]
orenbm: while working on this with @jonahx i understood that a bang suggests that there is a critical section in this method, such as calling a database. that's why i added the bang.

Anyway, i read a bit and it seems that in most cases bang will indeed be used for mutation so i'll remove it in this case.

app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
```diff
@@ -0,0 +1,103 @@
...
+        def conjur_username
+          id_token_username_field = oidc_secrets["id-token-user-property"]
+
+          conjur_username = id_token_attributes[id_token_username_field]
+          raise IdTokenFieldNotFound, id_token_username_field unless conjur_username.present?
+
+          conjur_username
+        end
+
+        def validate_security
```
orenbm: The following method appear in  several command classes. is there any way to remove duplications?
@jtuttle

orenbm:
@jtuttle please review once again.

Regarding the separation to another branch - let's do this after the pr is merged

app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
jtuttle: I wouldn't worry about it too much. The method is pretty small and it's kind of nice to have them in helper methods for the sake of clarity.

Issue #872, opened 2019-02-11
create automated tests according to test plan


orenbm:
Before we release this feature we should verify that the automated tests, defined in the test plan, are tested

childof #838

orenbm:
https://ca-il-confluence.il.cyber-ark.com/display/rndp/ID+Token+test+plan

orenbm:
This is already implemented

PR #934, opened 2019-03-26
Oidc tests

orenbm:
#### What does this PR do?
Added integration tests for OIDC

Connected to #872 

authn_oidc.feature

run ci/test


ci/docker-compose.yml
```diff
@@ -84,6 +84,12 @@ services:
       - KEYCLOAK_APP_USER=alice
       - KEYCLOAK_APP_USER_PASSWORD=alice
       - KEYCLOAK_APP_USER_EMAIL=alice@conjur.net
+      - KEYCLOAK_SECOND_APP_USER=bob
```
jtuttle: Is there some reason these user credentials need to be defined in the container env? It looks like we're just using the container env as a way to temporarily store the credentials so that they can be extracted and used in the bash script that calls `create_user` later on but maybe I'm missing something.

app/controllers/application_controller.rb
```diff
@@ -174,6 +178,11 @@ def forbidden e
     head :forbidden
   end
 
+  def bad_request e
+    logger.debug "#{e}\n#{e.backtrace.join "\n"}"
+    head :bad_request
```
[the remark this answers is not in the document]
orenbm: there is actually no real standard for this issue and 400 makes more sense to me as the request is bad and shouldn't be re-run without modification

ci/docker-compose.yml
orenbm: yes, we want one place to have all the keycloak variables. these variables are used for `create_user` both in `ci/test` & `dev/start` so we put it here to have 1 source of truth.

This is relevant for `KEYCLOAK_APP_USER, KEYCLOAK_SECOND_APP_USER, KEYCLOAK_NON_CONJUR_APP_USER` and all the corresponding vars (password, email).

Do you thinhk it's better to have them initialized in `ci/test` & `dev/start` instead?

jtuttle: Ah okay. Nah this is fine.

PR #936, opened 2019-03-26
Refactored some command classes to fit standard

orenbm:
#### What does this PR do?
I found some command classes which i missed in the previous refactor. done it now.

### DIFF
--- a/app/domain/authentication/login.rb
+++ b/app/domain/authentication/login.rb
@@ -14,41 +14,59 @@ module Authentication
   ) do
 
     def call
-      login(@authenticator_input)
+      validate_authenticator_exists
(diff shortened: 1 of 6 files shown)


app/domain/authentication/login.rb
```diff
@@ -14,41 +14,59 @@ module Authentication
   ) do
 
     def call
-      login(@authenticator_input)
+      validate_authenticator_exists
+
```
jtuttle: You can remove all the blank spaces between these method calls. It's easy enough to read without the spaces.

orenbm: done

Issue #876, opened 2019-02-11
Verify Logs and error handling in OIDC flow


orenbm:
Before we release this feature we'd like to verify that the whole flow has sufficient logs and that errors are handled in a way that the user knows how to fix them. This will also help in solving cases. 

Logs & error handling are defined here:
https://ca-il-confluence.il.cyber-ark.com/display/rndp/Supporting+ID+Token+as+an+authentication+method+in+Conjur 

childof #838

nessiLahav:
Add an Info message in eace API request that this feature is POC

izgeri:
@orenbm can you add more info to the title and description for this issue? it's hard to see what it's about when you're looking at the larger board

InbalZilberman:
Whoever is talking this feature let's please spend some time also on making sure we have supportability in mind and that the troubleshooting section is already written together with refining the logs. Please consult me if needed

orenbm:
This is already implemented

PR #941, opened 2019-04-01
Extend and fine-tune OIDC error handling

moticless:
#### What does this PR do?
Extend and fine tune error handling of OIDC. Taking care to have proper log of the error and avoid from information disclosure by the response.

Connected to #876

authenticate_controller.rb
 
run request with expired token. Run request with invalid token. etc.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,16 +5,18 @@ class ApplicationController < ActionController::API
   include ::ActionView::Layouts
 
   class Unauthorized < RuntimeError
+    attr_accessor :response_with_message
 
+    def initialize(message = nil, respone_with_message = false)
+      super(message)
+      @response_with_message = respone_with_message
+    end
   end
 
   class GatewayTimeout < RuntimeError
   end
 
   class BadGateway < RuntimeError
-  end
-
-  class BadRequest < RuntimeError
   end
 
   class BadRequest < RuntimeError
@@ -39,7 +41,6 @@ class ApplicationController < ActionController::API
   rescue_from Unauthorized, with: :unauthorized
   rescue_from GatewayTimeout, with: :gateway_timeout
   rescue_from BadGateway, with: :bad_gateway
-  rescue_from BadRequest, with: :bad_request
   rescue_from Exceptions::NotImplemented, with: :not_implemented
   rescue_from Sequel::ValidationFailed, with: :validation_failed
   rescue_from Sequel::NoMatchingRow, with: :no_matching_row
@@ -198,7 +199,7 @@ class ApplicationController < ActionController::API
 
   def unauthorized e
     logger.debug "#{e}\n#{e.backtrace.join "\n"}"
-    if e.message == Authentication::AuthnOidc::IdTokenExpired.new.to_s
+    if e.response_with_message == true
       render json: {
         error: {
           code: :unauthorized,
@@ -206,7 +207,6 @@ class ApplicationController < ActionController::API
         }
       }, status: :unauthorized
     else
-      logger.debug "#{e}\n#{e.backtrace.join "\n"}"
       head :unauthorized
     end
   end
@@ -219,11 +219,6 @@ class ApplicationController < ActionController::API
   def bad_gateway e
     logger.debug "#{e}\n#{e.backtrace.join "\n"}"
     head :bad_gateway
-  end
-
-  def bad_request e
-    logger.debug "#{e}\n#{e.backtrace.join "\n"}"
-    head :bad_request
   end
 
   def not_implemented e
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -199,7 +199,7 @@ class ApplicationController < ActionController::API
 
   def unauthorized e
     logger.debug "#{e}\n#{e.backtrace.join "\n"}"
-    if e.response_with_message == true
+    if e.response_with_message
       render json: {
         error: {
           code: :unauthorized,
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,11 +5,11 @@ class ApplicationController < ActionController::API
   include ::ActionView::Layouts
 
   class Unauthorized < RuntimeError
-    attr_accessor :response_with_message
+    attr_reader :return_message_in_response
 
-    def initialize(message = nil, respone_with_message = false)
+    def initialize(message = nil, return_message_with_response = false)
       super(message)
-      @response_with_message = respone_with_message
+      @return_message_in_response = return_message_with_response
     end
   end
 
@@ -199,7 +199,7 @@ class ApplicationController < ActionController::API
 
   def unauthorized e
     logger.debug "#{e}\n#{e.backtrace.join "\n"}"
-    if e.response_with_message
+    if e.return_message_in_response
       render json: {
         error: {
           code: :unauthorized,
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,16 @@ class ApplicationController < ActionController::API
   include ::ActionView::Layouts
 
   class Unauthorized < RuntimeError
+
+  end
+
+  class GatewayTimeout < RuntimeError
+  end
+
+  class BadGateway < RuntimeError
+  end
+
+  class BadRequest < RuntimeError
   end
 
   class BadRequest < RuntimeError
@@ -27,6 +37,9 @@ class ApplicationController < ActionController::API
   rescue_from Exceptions::Forbidden, with: :forbidden
   rescue_from BadRequest, with: :bad_request
   rescue_from Unauthorized, with: :unauthorized
+  rescue_from GatewayTimeout, with: :gateway_timeout
+  rescue_from BadGateway, with: :bad_gateway
+  rescue_from BadRequest, with: :bad_request
   rescue_from Exceptions::NotImplemented, with: :not_implemented
   rescue_from Sequel::ValidationFailed, with: :validation_failed
   rescue_from Sequel::NoMatchingRow, with: :no_matching_row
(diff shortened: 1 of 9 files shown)


app/controllers/application_controller.rb
```diff
@@ -5,6 +5,16 @@ class ApplicationController < ActionController::API
   include ::ActionView::Layouts
 
   class Unauthorized < RuntimeError
+
```
orenbm: please remove blank line

app/controllers/application_controller.rb
```diff
@@ -5,6 +5,16 @@ class ApplicationController < ActionController::API
...
+  class GatewayTimeout < RuntimeError
+  end
+
+  class BadGateway < RuntimeError
+  end
+
+  class BadRequest < RuntimeError
   end
 
   class BadRequest < RuntimeError
```
orenbm: is `BadRequest` defined twice in this file? it could have happen in the merge

app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
```diff
@@ -72,7 +72,7 @@ def conjur_username
           id_token_username_field = oidc_secrets["id-token-user-property"]
 
           conjur_username = id_token_attributes[id_token_username_field]
-          raise IdTokenFieldNotFound, id_token_username_field unless conjur_username.present?
+          raise IdTokenFieldNotFound, id_token_username_field unless conjur_username
```
orenbm: why was this changed? i had the `present?` to verify that not only the field exists but that it has a value. is this not needed?

app/domain/authentication/errors.rb
```diff
@@ -22,9 +22,34 @@ module Authentication
...
+    ProviderDiscoveryTimeout = ::Util::ErrorClass.new(
+      "OIDC provider failed with timeout error (provider_uri='{0})'"
+    )
+
+    ProviderDiscoveryFailed = ::Util::ErrorClass.new(
+      "OIDC provider discovery failed (provider_uri='{0})'"
+    )
+
+    ProviderRetrieveCertificateFailed = ::Util::ErrorClass.new(
+      "Failed retrieve certificate from OIDC provider (provider_uri='{0})'"
```
orenbm: Failed to retrieve certificate from the OIDC provider

app/controllers/application_controller.rb
```diff
@@ -185,7 +198,32 @@ def bad_request e
...
-    head :unauthorized
+    if e.message == Authentication::AuthnOidc::IdTokenExpired.new.to_s
+      render json: {
+        error: {
+          code: :unauthorized,
+          message: e.message
+        }
+      }, status: :unauthorized
+    else
+      logger.debug "#{e}\n#{e.backtrace.join "\n"}"
```
orenbm: this is redundant. we already log in the beginning of the method

app/controllers/application_controller.rb
```diff
@@ -185,7 +198,32 @@ def bad_request e
 
   def unauthorized e
     logger.debug "#{e}\n#{e.backtrace.join "\n"}"
-    head :unauthorized
+    if e.message == Authentication::AuthnOidc::IdTokenExpired.new.to_s
```
orenbm: this is very specific. what if we decide that also for another scenario we want to return the error message too? we'll need to update both `authenticate_controller.rb` and this file. please find a way that doesn't include the message check.

a possible solution is to check that the message is not "ApplicationController::Unauthorized" which is the default value (and thus if we want some exception to return the message to the user we raise it with message in `authenticate_controller.rb` as we do for `Authentication::AuthnOidc::IdTokenExpired`.

please check if there is a neater solution. maybe enhance this class in `application_controller.rb`
```
  class Unauthorized < RuntimeError
  end
```

app/controllers/application_controller.rb
```diff
@@ -27,6 +37,9 @@ class RecordExists < Exceptions::RecordExists
   rescue_from Exceptions::Forbidden, with: :forbidden
   rescue_from BadRequest, with: :bad_request
   rescue_from Unauthorized, with: :unauthorized
+  rescue_from GatewayTimeout, with: :gateway_timeout
+  rescue_from BadGateway, with: :bad_gateway
+  rescue_from BadRequest, with: :bad_request
```
orenbm: this is already defined in row 38

app/controllers/application_controller.rb
```diff
@@ -185,7 +198,32 @@ def bad_request e
...
+    logger.debug "#{e}\n#{e.backtrace.join "\n"}"
+    head :gateway_timeout
+  end
+
+  def bad_gateway e
+    logger.debug "#{e}\n#{e.backtrace.join "\n"}"
+    head :bad_gateway
+  end
+
+  def bad_request e
```
orenbm: this method is already defined in row 193

orenbm:
@moticless did you add the log messages to the confluence table for review?

nessiLahav:
Please add some Debug info logs to the authenticate flow

app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
jtuttle: I think this was correct with `present?`. Also interested to know why it changed.

app/domain/authentication/authn_oidc/authenticate_id_token/decode_and_verify_id_token.rb
```diff
@@ -41,12 +42,24 @@ def decoded_attributes
...
             @id_token_jwt,
-            get_cert(@provider_uri)
+            get_cert
           )
+        rescue => e
+          raise IdTokenInvalidFormat, e.to_s
+        end
+
+        def discovered_provider
+          @discovered_provider ||= OpenIDConnect::Discovery::Provider::Config.discover!(@provider_uri)
```
jtuttle: The `discover!` method being called here should not have a `!`. In Ruby, the `!` on a method name signals that it's a mutator method that modifies the object that calls it. I'm guessing that's not happening here, though correct me if I'm wrong!

app/domain/authentication/authn_oidc/authenticate_id_token/decode_and_verify_id_token.rb
```diff
@@ -11,7 +11,8 @@ module AuthenticateIdToken
       ) do
 
         def call
-          decode_id_token
+          discovered_provider
+          decoded_id_token
```
jtuttle: Are you calling these methods early to make the `call` method look like an index of what's happening? I wouldn't bother doing so in this situation. It's forcing you to memoize an object that you could just be returning (`@discovered_provider`) and reads a bit strangely since these two methods are essentially 'getters'.  You could reword the methods to be a bit more action oriented like `discover_provider` and `decode_id_token` perhaps. Otherwise, I would just remove them and let them happen as helpers in `verify_id_token`.

app/domain/authentication/authn_oidc/authenticate_id_token/decode_and_verify_id_token.rb
jtuttle: Also if you  make the change I suggested above to the `call` method, I don't think you need to memoize `@ discovered_provider` and can just return it from the method for use in `get_cert`.

app/domain/authentication/authn_oidc/authenticate_id_token/decode_and_verify_id_token.rb
```diff
@@ -41,12 +42,24 @@ def decoded_attributes
...
+          raise ProviderDiscoveryTimeout, @provider_uri
+        rescue => e
+          raise ProviderDiscoveryFailed, @provider_uri
         end
 
-        def get_cert(provider_uri)
-          OpenIDConnect::Discovery::Provider::Config.discover!(provider_uri).jwks
+        def get_cert
+          discovered_provider.jwks
+        rescue => e
```
jtuttle: Any chance we get a useful error here that might tell us what went wrong? If so, we should try to preserve it.

app/domain/authentication/errors.rb
```diff
@@ -22,9 +22,34 @@ module Authentication
 
   module AuthnOidc
 
+    ProviderDiscoveryTimeout = ::Util::ErrorClass.new(
+      "OIDC provider failed with timeout error (provider_uri='{0})'"
```
jtuttle: Should be "provider **discovery** failed"

app/domain/authentication/authn_oidc/authenticate_id_token/decode_and_verify_id_token.rb
orenbm: the `discover!` method is in the 3rd party

app/domain/authentication/authn_oidc/authenticate_id_token/decode_and_verify_id_token.rb
orenbm: yes, that was the idea. would you rather have something like this?

```
def decode_id_token
  decoded_id_token
end
```
?

app/domain/authentication/authn_oidc/authenticate_id_token/decode_and_verify_id_token.rb
jtuttle: Oh weird, okay!

app/domain/authentication/authn_oidc/authenticate_id_token/decode_and_verify_id_token.rb
jtuttle: Hmm good question. @jonahx what's the "command class" philosophy here? It seems odd to be calling helper methods early just to produce a readable index. Should they be removed? Or is there a bigger restructure that needs to happen here to create a `call` method with meaningful line items?

app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
moticless: You are right. The only issue is that the error not fully reflect the error. I will modify the name and the log to IdTokenFieldNotFoundOrEmpty

PR #945, opened 2019-04-03
Oidc conf tests

orenbm:
#### What does this PR do?
Added tests for cases:
  - non-existing account in request
  - id-token-user-property variable missing in policy
  - provider-uri variable missing in policy
  - webservice missing in policy
  - admin user is denied

Also implemented behaviour for the "admin user is denied" case

Connected to #948

### DIFF
--- a/app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
+++ b/app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
@@ -73,7 +73,7 @@ module Authentication
 
           conjur_username = id_token_attributes[id_token_username_field]
           raise IdTokenFieldNotFoundOrEmpty, id_token_username_field unless conjur_username.present?
-          raise BlacklistedOidcUser, conjur_username if blacklisted_oidc_user?(conjur_username)
+          raise AdminAuthenticationAttempt if is_admin?(conjur_username)
 
           conjur_username
         end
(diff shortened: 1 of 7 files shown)


app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
```diff
@@ -73,6 +73,7 @@ def conjur_username
 
           conjur_username = id_token_attributes[id_token_username_field]
           raise IdTokenFieldNotFoundOrEmpty, id_token_username_field unless conjur_username.present?
+          raise BlacklistedOidcUser, conjur_username if blacklisted_oidc_user?(conjur_username)
```
jtuttle: Are there plans to blacklist other users? Otherwise, since we're only blacklisting the admin user, I would suggest being more specific here and name the error something like `AdminAuthenticationAttempt`. We could also rename the method being called `is_admin?`.

orenbm: will do. i had this more general for future use like we have in authn-ldap.

```
      # admin should only be able to login through plain Conjur authn
      def blacklisted_ldap_user?(login)
        login == 'admin'
      end
```

jtuttle: I would say make it non-generic for now. Seems like we may never need to blacklist other users as admin is a very special case.

cucumber/authenticators/features/authn_oidc_bad_policy.feature
```diff
@@ -0,0 +1,95 @@
+Feature: Users can authneticate with OIDC authenticator
```
jtuttle: Typo here: authenticate

PR #954, opened 2019-04-04
Fixed PR 946 issues

orenbm:
#### What does this PR do?
fixes comments from PR #946 

- Added log for certificate fetch
- changed `certs` method to `fetch-certs` to be action-oriented


app/domain/authentication/authn_oidc/authenticate_id_token/decode_and_verify_id_token.rb
```diff
@@ -42,7 +42,7 @@ def verify_decoded_id_token
           raise IdTokenVerifyFailed, e.inspect
         end
 
-        def certs
+        def fetch_certs
           @certs ||= @provider_certificate.fetch_certs(@provider_uri)
```
jtuttle: This doesn't actually need to be memoized. Just set it using `@certs =` and then refer to it as `@certs` in the method below. That should hopefully remove the CC warnings.

jtuttle:
I just spoke w/ Jonah a bit about the CodeClimate errors and he suggested turning `ProviderCertificate` into a `FetchCertificates` command class. This would allow you to inject `Rails.logger` as a dependency which would solve one of the errors. Mind doing that and then we'll deal with the other one?

app/domain/authentication/authn_oidc/authenticate_id_token/fetch_provider_certificate.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+
+        private
+
+        def discover_provider
+          @logger.debug("[OIDC] Discovering provider '#{@provider_uri}'")
+
+          @discovered_provider = OpenIDConnect::Discovery::Provider::Config.discover!(@provider_uri)
+        rescue HTTPClient::ConnectTimeoutError => e
+          raise ProviderDiscoveryTimeout.new(@provider_uri, e.inspect)
```
jtuttle: You could silence this `reek` error by creating a `raise_error` method:

```
def raise_error(error_class, original_error)
  raiser error_class.new(@provider_uri, original_error.inspect)
end
```

orenbm: done

PR #955, opened 2019-04-05
Remove expired links

typaulhus:
#### What does this PR do?
Removes links to out of date /reference pages in conur.org documentation.

We are removing the /reference pages from conjur.org, but some of those pages are linked in the API documentation. This PR removes those old links and updates w/ new links to docs.conjur.org

connected to cyberark/conjur-org#347

Run the API docs locally, and confirm that links to Conjur policy are updated in the policy pages:
https://www.conjur.org/api.html#policies

https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--conjur/detail/remove-expired-links/1/pipeline

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [1.3.7] - 2019-03-27
 ### Changed
+- Updated links to Policy & Cryptography reference in API documentation
 - Updated conjur-policy-parser to
   [v3.0.3](https://github.com/conjurinc/conjur-policy-parser/blob/possum/CHANGELOG.md#v303).
 - Replaced `changelog` entrypoint in `ci/test` with a separate script. Building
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Moved changelog validation up in CI pipeline
 
+## [1.3.8] - 2019-04-05
```
izgeri: Unless we're bumping the version also (and I don't think we necessarily need to) this line is not needed. also, this bullet can go with the other `Changed` item

VERSION
```diff
@@ -1 +1 @@
-1.3.7
+1.3.8
```
izgeri: I don't think we need to bump the version

typaulhus: changing back to 1.3.7

typaulhus:
@izgeri I made changes & rebased, based on your comments, so i think we're all good for a merge?

Issue #957, opened 2019-04-08
Test - webservice with read and w/o authenticate permission is denied


orenbm:
We want to test the scenario where a user who has read permission on a webservice, and without an authenticate permission on it is denied for OIDC authn

PR #958, opened 2019-04-08
No authenticate permission test

orenbm:
#### What does this PR do?
Adds a test for webservice with read and w/o authenticate permission is denied

Connected to #957

### DIFF
--- a/app/domain/authentication/security.rb
+++ b/app/domain/authentication/security.rb
@@ -60,7 +60,11 @@ module Authentication
 
       # Ensure user has access to the service
       has_access = user_role.allowed_to?('authenticate', webservice_resource)
-      raise NotAuthorizedInConjur, req.user_id unless has_access
+      unless has_access
+        Rails.logger.debug("[OIDC] User '#{req.user_id}' is not authorized to " \
+          "authenticate with webservice '#{webservice_resource[:resource_id]}'")
+        raise NotAuthorizedInConjur, req.user_id
+      end
     end
   end
 
--- a/app/domain/authentication/security.rb
+++ b/app/domain/authentication/security.rb
@@ -48,8 +48,10 @@ module Authentication
     end
 
     def validate_user_has_access(req)
+      resource_id = req.webservice.resource_id
+
       # Ensure webservice is defined in Conjur
-      webservice_resource = resource_class[req.webservice.resource_id]
+      webservice_resource = resource_class[resource_id]
       raise ServiceNotDefined, req.webservice.name unless webservice_resource
 
       # Ensure user is defined in Conjur
@@ -62,7 +64,7 @@ module Authentication
       has_access = user_role.allowed_to?('authenticate', webservice_resource)
       unless has_access
         Rails.logger.debug("[OIDC] User '#{req.user_id}' is not authorized to " \
-          "authenticate with webservice '#{webservice_resource[:resource_id]}'")
+          "authenticate with webservice '#{resource_id}'")
         raise NotAuthorizedInConjur, req.user_id
       end
     end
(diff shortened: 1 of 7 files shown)


cucumber/authenticators/features/authn_oidc_bad_policy.feature
```diff
@@ -93,3 +93,39 @@ Feature: Users can authenticate with OIDC authenticator
...
+    - !grant
+      role: !group conjur/authn-oidc/keycloak/users
+      member: !user alice
+    """
+    And I am the super-user
+    And I successfully set OIDC variables
+    Given I get authorization code for username "alice" and password "alice"
+    And I fetch an ID Token
+    When I authenticate via OIDC with id token
+    Then it is denied
```
InbalZilberman: @orenbm - Thank you!!
what is the meaning of denied? 401? 400? 501?

InbalZilberman: @orenbm this test is in OSS, do you plan to add it to appliance also?

orenbm: No. *if* we'll add tests to the appliance it will be only some vanilla flow

orenbm: > @orenbm - Thank you!!
> what is the meaning of denied? 401? 400? 501?

401 is unauthorized
400 is bad request
501 is not implemented (which we don't throw at all).

Do you think the step should be `Then it is unauthorized`?

InbalZilberman: @orenbm YES!

orenbm: done :)

app/domain/authentication/security.rb
```diff
@@ -60,7 +62,11 @@ def validate_user_has_access(req)
 
       # Ensure user has access to the service
       has_access = user_role.allowed_to?('authenticate', webservice_resource)
-      raise NotAuthorizedInConjur, req.user_id unless has_access
+      unless has_access
```
jtuttle: codeclimate says to use a guard clause here. This seems kind've silly but might as well do that instead of littering our code with suppression comments. So I believe that would make this:

```
return if has_access

Rails.logger.debug("[OIDC] User '#{req.user_id}' is not authorized to " \
          "authenticate with webservice '#{resource_id}'")
raise NotAuthorizedInConjur, req.user_id
```

app/domain/authentication/security.rb
```diff
@@ -60,7 +62,11 @@ def validate_user_has_access(req)
...
+        Rails.logger.debug("[OIDC] User '#{req.user_id}' is not authorized to " \
```
jtuttle: I think the other CodeClimate complaint can be addressed by sticking a helper var `user_id = req.user_id` at the top of this method and using that wherever necessary. See if that fixes it.

app/domain/authentication/security.rb
orenbm: i really don't like this peace of code. guard clauses are good IMO only when there is one outcome. with more than that it's just an abuse to the code style. don't you think that the current implementation is more readable?

jtuttle: Yep, it's much better the way it currently is. Feel free to put a suppression comment in instead.

app/domain/authentication/security.rb
orenbm: also here i wouldn;t take CC's advice. 
`req.user_id` is fine and adding the `user_id` method is redundant IMO

jtuttle: I think this one is actually somewhat valid. We refer to `req.user_id` four times in the method. Might as well stick it in a variable.

orenbm: ok

orenbm: so i started to fix this but it will be much easier once `security.rb` is a command class. I'll fix this issue in that effort.

PR #970, opened 2019-04-17
[WIP] Safe resource (design)

jvanderhoof:
Adds a design resource to capture the idea of a safe, a set of variables, groups and permissions to simplify managing related variables, like database credentials.

### DIFF
--- a/design/conjur-safe.md
+++ b/design/conjur-safe.md
@@ -27,3 +27,65 @@ When a Safe resource is created, the following resource primitives are created:
 - `<safe-id>/readers`: read/execute permission on above variables
 - `<safe-id>/writers`: update permission on above variables
 - `<safe-id>/admins`: read/execute/update permission on above variables
+
+### Enabling access
+As a Safe has three default groups associated with it, permissions are assigned slightly differently than a typical resource.
+
+```yml
+- !safe
+  id: foo-database
+  variables:
+    - url
+    - port
+    - username
+    - password
+
+- !group database-users
+- !group database-admins
+
+- !grant
+  member: !group database-users
+  role: !group foo-database/readers
+
+- !grant
+  member: !group database-admins
+  role: !group foo-database/admins
+```
+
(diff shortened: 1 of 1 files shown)


design/conjur-safe.md
```diff
@@ -0,0 +1,29 @@
...
+# policies/databases/foo.yml
+- !safe
+  id: foo-database
+  variables:
+    - url
+    - port
+    - username
+    - password
+```
+When a Safe resource is created, the following resource primitives are created:
```
izgeri: It might be worth adding here some thoughts on how we'll implement this - it seems to me that there are a wide range of possibilities.

I think of a safe (as you described it above) as essentially analogous to this policy snippet:

```yml
- !policy
  id: foo-database
  body:
    - &variables
      - !variable url
      - !variable port
      - !variable username
      - !variable password

    - !group readers
    - !group writers
    - !group admins

    # readers can read and execute
    - !permit
      resource: *variables
      privileges: [ read, execute ]
      role: !group readers

    # writers can update
    - !permit
      resource: *variables
      privileges: [ update ]
      role: !group writers

    # admins can read, execute, and update
    - !permit
      resource: *variables
      privileges: [ read, execute, update ]
      role: !group admins
```

design/conjur-safe.md
```diff
@@ -0,0 +1,29 @@
...
+**Variables:**
+- `<safe-id>/url`
+- `<safe-id>/port`
+- `<safe-id>/username`
+- `<safe-id>/password`
+
+**Groups:**
+- `<safe-id>/readers`: read/execute permission on above variables
+- `<safe-id>/writers`: update permission on above variables
+- `<safe-id>/admins`: read/execute/update permission on above variables
```
izgeri: there is a case here for a `viewers` group as well, that just has `read` and (as I understand it) can see the resource

micahlee:
@jvanderhoof, what do you think about marking this PR as `!WIP` if there isn't a plan to merge it eminently?

izgeri:
Maybe we should convert this to a gh issue at this point.

Issue #959, opened 2019-04-08
Errors have error codes


orenbm:
To enhance supportability, we want each error log to have an error code which will be its ID. This will help us to better support cases in the future.

We should also move all error logs to one class and call them from there.

Issue #920, opened 2019-03-18
Add error injection to Command class in oidc


orenbm:
Jonah enhanced the command_class to support ordinary ruby class syntax which allows you to embed error specific to the `CommandClass` directly inside the class.

We should use it in this way in the oidc-authn

PR #971, opened 2019-04-18
Error codes in error

orenbm:
#### What does this PR do?
- Updates command_class gem to a version which accepts errors
- Refactoring commandClasses to present the errors it raises at the top of the class
- Moves all authn errors to one place
- Creates an ErrorWithErrorCode class to raise errors with error codes
- Updates all authn errors to have also error codes

#### What doesn't this PR do?
Defines the error code template. In this PR the template is "CONJ0001E". This may change before the OIDC in GA but this PR is mainly for the infrastructure for this.

Connected to #920 
Connected to #959 

/conjur/app/domain/authentication/errors.rb
/conjur/app/domain/util/error_class_with_code


orenbm:
@jtuttle i think it's something to do with the branches, no worries :)

i saw that the tests are failing in GKE. i;m on it.

Issue #933, opened 2019-03-25
Add log validation to oidc tests


orenbm:
We'd like to validate the error logs are written in case of failure. as currently we don't have all the logs in the code, we'll need to add the validation once it's done

requires #876

child of #838

orenbm:
this is already implemented

PR #980, opened 2019-04-25
Logs validation tests

nessiLahav:
#### What does this PR do?
validate the error logs are written in case of failure in the cucumber tests 
Connected to #933 
Run oidc authenticator features tests
#### Jenkins build
[link](https://jenkins.conjur.net/job/cyberark--conjur/job/logs_validation_tests/)
 
Yes

### DIFF
--- a/cucumber/api/features/step_definitions/logs_steps.rb
+++ b/cucumber/api/features/step_definitions/logs_steps.rb
@@ -3,3 +3,8 @@
 Given(/^I save the amount of log lines into "([^"]*)"$/) do |bookmark|
   save_amount_of_log_lines(bookmark)
 end
+
+
+And(/^The log filtered from line number in "([^"]*)" should contains "([^"]*)" messages:$/) do |bookmark, occurrences, message|
+  expect(occurences_in_log_filtered_from_bookmark(bookmark, message)).to eq occurrences.to_i
+end
--- a/cucumber/api/features/step_definitions/logs_steps.rb
+++ b/cucumber/api/features/step_definitions/logs_steps.rb
@@ -1,9 +1,16 @@
 # frozen_string_literal: true
 
+Given(/^I save the log data from bookmark$/) do
+  save_log_data_from_bookmark()
+end
+
 Given(/^I save the log data from bookmark "([^"]*)"$/) do |bookmark|
   save_log_data_from_bookmark(bookmark)
 end
 
+And(/^The log filtered from bookmark contains messages:$/) do |message|
+  expect(occurences_in_log_filtered_from_bookmark(message)).to eq 1
+end
 
 And(/^The log filtered from bookmark "([^"]*)" contains "([^"]*)" messages:$/) do |bookmark, occurrences, message|
   expect(occurences_in_log_filtered_from_bookmark(bookmark, message)).to eq occurrences.to_i
--- a/cucumber/api/features/step_definitions/logs_steps.rb
+++ b/cucumber/api/features/step_definitions/logs_steps.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 Given(/^I save the log data from bookmark$/) do
-  save_log_data_from_bookmark()
+  lines_amount_in_bookmark()
 end
 
 Given(/^I save the log data from bookmark "([^"]*)"$/) do |bookmark|
-  save_log_data_from_bookmark(bookmark)
+  lines_amount_in_bookmark(bookmark)
 end
 
 And(/^The log filtered from bookmark contains messages:$/) do |message|
--- a/cucumber/api/features/step_definitions/logs_steps.rb
+++ b/cucumber/api/features/step_definitions/logs_steps.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+Given(/^I save the amount of log lines into "([^"]*)"$/) do |bookmark|
+  save_amount_of_log_lines(bookmark)
+end
--- a/cucumber/api/features/step_definitions/logs_steps.rb
+++ b/cucumber/api/features/step_definitions/logs_steps.rb
@@ -1,13 +1,9 @@
 # frozen_string_literal: true
 
-Given(/^I save the log data from bookmark$/) do
+Given(/^I save my place in the log file$/) do
   save_num_log_lines()
 end
 
-And(/^The log filtered from bookmark contains messages:$/) do |message|
+And(/^The following appears in the log after my savepoint:$/) do |message|
   expect(num_matches_since_savepoint(message)).to eq 1
-end
-
-And(/^The log filtered from bookmark "([^"]*)" contains "([^"]*)" messages:$/) do |occurrences, message|
-  expect(num_matches_since_savepoint(message)).to eq occurrences.to_i
 end
--- a/cucumber/api/features/step_definitions/logs_steps.rb
+++ b/cucumber/api/features/step_definitions/logs_steps.rb
@@ -1,17 +1,13 @@
 # frozen_string_literal: true
 
 Given(/^I save the log data from bookmark$/) do
-  lines_amount_in_bookmark()
-end
-
-Given(/^I save the log data from bookmark "([^"]*)"$/) do |bookmark|
-  lines_amount_in_bookmark(bookmark)
+  save_num_log_lines()
 end
 
 And(/^The log filtered from bookmark contains messages:$/) do |message|
-  expect(occurences_in_log_filtered_from_bookmark(message)).to eq 1
+  expect(num_matches_since_savepoint(message)).to eq 1
 end
 
-And(/^The log filtered from bookmark "([^"]*)" contains "([^"]*)" messages:$/) do |bookmark, occurrences, message|
-  expect(occurences_in_log_filtered_from_bookmark(bookmark, message)).to eq occurrences.to_i
+And(/^The log filtered from bookmark "([^"]*)" contains "([^"]*)" messages:$/) do |occurrences, message|
+  expect(num_matches_since_savepoint(message)).to eq occurrences.to_i
 end
--- a/cucumber/api/features/step_definitions/logs_steps.rb
+++ b/cucumber/api/features/step_definitions/logs_steps.rb
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
-Given(/^I save the amount of log lines into "([^"]*)"$/) do |bookmark|
-  save_amount_of_log_lines(bookmark)
+Given(/^I save the log data from bookmark "([^"]*)"$/) do |bookmark|
+  save_log_data_from_bookmark(bookmark)
 end
 
 
-And(/^The log filtered from line number in "([^"]*)" should contains "([^"]*)" messages:$/) do |bookmark, occurrences, message|
+And(/^The log filtered from bookmark "([^"]*)" contains "([^"]*)" messages:$/) do |bookmark, occurrences, message|
   expect(occurences_in_log_filtered_from_bookmark(bookmark, message)).to eq occurrences.to_i
 end
--- a/cucumber/api/features/support/logs_helpers.rb
+++ b/cucumber/api/features/support/logs_helpers.rb
@@ -2,34 +2,44 @@
 
 # Utility methods for Logs steps
 #
-require 'open3'
-
+# In order to be able validating messages in the conjur server log,
+# we chose to filter the log by save points instead of zeroing out the log,
+# due to the following reasons:
+# 1. save the log content for troubleshooting
+# 2. be able to parallelize the tests in the future
+#
 module LogsHelpers
   LOG_LOCATION = "/src/conjur-server/log/development.log"
 
-  def lines_amount_in_bookmark(bookmark = "bookmark")
-    @bookmarks ||= Hash.new
-
-    lines_cnt = File.new(LOG_LOCATION).readlines.size
-    raise "Log lines amount '#{lines_cnt}' should be a positive number" unless lines_cnt > 0
-
-    @bookmarks[bookmark] = lines_cnt
+  def num_log_lines
+    File.new(LOG_LOCATION).readlines.size
   end
 
-  def occurences_in_log_filtered_from_bookmark(bookmark = "bookmark", msg)
-    raise "Bookmark #{bookmark} doesn't exists" unless @bookmarks[bookmark].present?
+  # NOTE: Client code should explicitly call this to save current log size,
+  # rather than making it a side affect of getting the current size
+  def save_num_log_lines
+    @saved_num_lines = num_log_lines
+  end
 
-    lines_cnt = File.new(LOG_LOCATION).readlines.size
-    raise "Log lines amount '#{lines_cnt}' should be a positive number" unless lines_cnt > 0
-    raise "Current logs lines amount '#{lines_cnt}' is smaller then bookmark '#{bookmark}' value '#{@bookmarks[bookmark]}'" unless lines_cnt >= @bookmarks[bookmark]
+  def num_matches_since_savepoint(msg)
+    cur_num_lines = num_log_lines
 
-    num_lines_matching(start_line: @bookmarks[bookmark], end_line: lines_cnt, text: msg)
+    validate_savepoint_exists
+    validate_log_didnt_shrink(cur_num_lines)
+
+    start_line, end_line = @saved_num_lines, cur_num_lines
+    (File.readlines(LOG_LOCATION)[start_line..end_line].grep(/#{msg}/i)).count
   end
 
   private
 
-  def num_lines_matching(start_line:, end_line:, text:)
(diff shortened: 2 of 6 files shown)


cucumber/api/features/step_definitions/logs_steps.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Given(/^I save the amount of log lines into "([^"]*)"$/) do |bookmark|
```
orenbm: this step name isn't clear. can we change it to "I save the log data from bookmark (([^"]*)"?

cucumber/api/features/step_definitions/logs_steps.rb
```diff
@@ -0,0 +1,10 @@
...
+  save_amount_of_log_lines(bookmark)
+end
+
+
+And(/^The log filtered from line number in "([^"]*)" should contains "([^"]*)" messages:$/) do |bookmark, occurrences, message|
```
orenbm: "The log filtered from bookmark "([^"]*)" contains "([^"]*)" messages:"
i think this is better

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+# Utility methods for Logs steps
+#
+require 'open3'
+
+module LogsHelpers
+  @@log_location = "/src/conjur-server/log/development.log"
+
+  def save_amount_of_log_lines(bookmark)
```
orenbm: same here with the name

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+    @bookmarks ||= Hash.new
+    @bookmarks[bookmark] = amount_of_log_lines
+  end
+
+  def occurences_in_log_filtered_from_bookmark(bookmark, msg)
+    raise "Bookmark #{bookmark} doesn't exists" unless @bookmarks[bookmark].present?
+    amount = amount_of_log_lines
```
orenbm: why do we need this?

orenbm: can't we just use `amount_of_log_lines`?

orenbm: oh i see that each call to `amount_of_log_lines` reads the log and counts number. maybe just change to `current_amount` but not critical

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+    raise "Current logs amount is smaller then bookmark '#{bookmark}' value '#{@bookmarks[bookmark]}'" unless amount >= @bookmarks[bookmark]
```
orenbm: Current log lines amount '#{amount}' is smaller than in bookmark '#{bookmark}' value '#{@bookmarks[bookmark]}'

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+
+# Utility methods for Logs steps
+#
+require 'open3'
+
+module LogsHelpers
+  @@log_location = "/src/conjur-server/log/development.log"
+
+  def save_amount_of_log_lines(bookmark)
+    @bookmarks ||= Hash.new
```
orenbm: i think that `lines_amount_in_bookmark` is telling the story a bit better

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+    raise "Bookmark #{bookmark} doesn't exists" unless @bookmarks[bookmark].present?
+    amount = amount_of_log_lines
+    raise "Current logs amount is smaller then bookmark '#{bookmark}' value '#{@bookmarks[bookmark]}'" unless amount >= @bookmarks[bookmark]
+    occurences_in_log_section(from: @bookmarks[bookmark], to: amount, message: msg)
+  end
+
+  private
+
+  def amount_of_log_lines
+    wc_cmd = "wc -l #{@@log_location}"
```
orenbm: can we change this to the action name? `count_lines_cmd`

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+    stdout, stderr, status = Open3.capture3(wc_cmd)
+    raise "Failed to run command '#{wc_cmd}' with error '#{stderr}'" unless status.success?
```
orenbm: "Command '#{wc_cmd}' raised error '#{stderr}'"

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+    raise "Log file #{@@log_location} is empty" if stdout.to_s.empty?
+
+    amount = stdout.to_s.split(" ").first().to_i
+    raise "Log lines amount '#{amount}' should be a positive number" unless amount > 0
+    amount
+  end
+
+  def occurences_in_log_section(from:, to:, message:)
+    sed_cmd = "sed -n #{from},#{to}p #{@@log_location}"
```
orenbm: same here. explain what's happening and not the linux command

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+    stdout, stderr, status = Open3.capture3(sed_cmd)
+    raise "Failed to run command '#{sed_cmd}' with error '#{stderr}'" unless status.success?
```
orenbm: "Command '#{sed_cmd}' raised error '#{stderr}'"

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -81,8 +81,13 @@ Feature: Users can authneticate with OIDC authenticator
   Scenario: Non-existing username in ID token is denied
     Given I get authorization code for username "not_in_conjur" and password "not_in_conjur"
     And I fetch an ID Token
+    And I save the amount of log lines into "bookmark_not_in_conjur"
```
orenbm: i understand that the bookmark name isn't really needed as long as we don't run tests in parallel. as this is less readabe and makes the reader think why this is here, i suggest creating a step "I save the amount of log lines into bookmark" which will call the step above with some fixed value

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -81,8 +81,13 @@ Feature: Users can authneticate with OIDC authenticator
...
     When I authenticate via OIDC with id token
     Then it is unauthorized
+    And The log filtered from line number in "bookmark_not_in_conjur" should contains "1" messages:
```
orenbm: same here with the ammount of messages. if we use "1" all the time let's add a step named "The log filtered from line number in bookmark contains  message:"

and this will call the other step with the default bookmark and count = 1

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -91,31 +96,61 @@ Feature: Users can authneticate with OIDC authenticator
...
     And I fetch an ID Token
+    And I save the amount of log lines into "bookmark_alice"
     When I authenticate via OIDC with id token
     Then it is unauthorized
+    And The log filtered from line number in "bookmark_alice" should contains "1" messages:
+    """
+    Authentication Error: #<Authentication::AuthnOidc::IdTokenFieldNotFoundOrEmpty: Field 'non_existing_field' not found or empty in ID Token
+    """
 
   Scenario: Empty or missing id token is a bad request
```
orenbm: please split this into 2 scenarios

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -91,31 +96,61 @@ Feature: Users can authneticate with OIDC authenticator
...
 
   Scenario: non-existing account in request is denied
     Given I get authorization code for username "alice" and password "alice"
     And I fetch an ID Token
+    And I save the amount of log lines into "bookmark_not_existing_acnt"
     When I authenticate via OIDC with id token and account "non-existing"
     Then it is unauthorized
+    And The log filtered from line number in "bookmark_not_existing_acnt" should contains "1" messages:
+    """
+    Authentication Error: #<Conjur::RequiredResourceMissing: Missing required resource: non-existing:variable:conjur/authn-oidc/keycloak/provider-uri
```
orenbm: this will change in GA to "account does not exists" (or something like that) so we should keep in mind this test will crash once we merge

nessiLahav: added comment to test

cucumber/authenticators/features/authn_oidc.feature
nessiLahav: Thanks, i refactored please review

cucumber/authenticators/features/authn_oidc.feature
nessiLahav: Thanks, i refactored please review

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -81,8 +81,13 @@ Feature: Users can authneticate with OIDC authenticator
   Scenario: Non-existing username in ID token is denied
     Given I get authorization code for username "not_in_conjur" and password "not_in_conjur"
     And I fetch an ID Token
+    And I save the log data from bookmark
     When I authenticate via OIDC with id token
     Then it is unauthorized
+    And The log filtered from bookmark contains messages:
+    """
+    Authentication Error: #<Authentication::NotDefinedInConjur: User 'not_in_conjur' is not defined in Conjur
```
jonahx: I would match only on `Authentication::NotDefinedInConjur` and, if you want to be extra safe, `not_in_conjur`.  This way the test will be robust against (relatively likely) changes in the particular wording of the error message.

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
nessiLahav: Here i have a different opinion, due to the following:
1. I do not want to change the log file, i think the tests should not change anything that may affects other tests which maybe relaying on the log file 
2. imo the bookmarks are reflected what happening in manual tests, which i want to check the logs in specific timeline to ignore background noises 
3. i build here infrastructure that will work also if we will change the tests to run in parallel
4. with this approach you can easily validate the logs multiple time in the same scenario without a background noises 

@jonahx tell what do think?

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+    amount
+  end
+
+  def occurences_in_log_section(from:, to:, message:)
+    filter_log_cmd = "sed -n #{from},#{to}p #{@@log_location}"
+    stdout, stderr, status = Open3.capture3(filter_log_cmd)
+    raise "Command '#{filter_log_cmd}' raised error '#{stderr}'" unless status.success?
+    stdout.to_s.scan(message).count
+  end
+end
```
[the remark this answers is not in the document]
nessiLahav: Thanks, i changed it please review

cucumber/api/features/support/logs_helpers.rb
```diff
@@ -0,0 +1,43 @@
...
+  def amount_of_log_lines
+    count_lines_cmd = "wc -l #{@@log_location}"
+    stdout, stderr, status = Open3.capture3(count_lines_cmd)
+    raise "Command '#{count_lines_cmd}' raised error '#{stderr}'" unless status.success?
+    raise "Log file #{@@log_location} is empty" if stdout.to_s.empty?
+
+    amount = stdout.to_s.split(" ").first().to_i
+    raise "Log lines amount '#{amount}' should be a positive number" unless amount > 0
+    amount
+  end
```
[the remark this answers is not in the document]
nessiLahav: 1. done
2. I changed the location of the log empty error raising, because if i look for logs in a bookmark the log should not be empty

cucumber/api/features/step_definitions/logs_steps.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+Given(/^I save the log data from bookmark$/) do
+  lines_amount_in_bookmark()
+end
+
+Given(/^I save the log data from bookmark "([^"]*)"$/) do |bookmark|
```
jeniaSakirko: We should be able to give a name to the bookmark in case we want couple of bookmarks also when reading from bookmark

cucumber/api/features/support/logs_helpers.rb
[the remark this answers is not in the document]
jeniaSakirko: I agree that we shouldn't change anything for other tests but other tests shouldn't rely on the logs that were created in other tests, so I think we can clean the log file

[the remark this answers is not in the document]
jonahx: I agree with Jenia's point and still think zeroing is the best approach here.  That said, I'm not absolutely opposed to your approach either.  If you do decide to go with it, here are some changes I'd consider.  Was easier to just make them in vim vs try to describe each one, but basically I removed a few checks that didn't make sense to me, and changed bookmark from a Hash to a single value, since that's all we're using ("the previous number of log lines").  Remember each scenario will be run in its own world.

This is untested but should get across all the ideas, and you can make the required changes to the step definitions which should be simple:

```ruby
# frozen_string_literal: true

# Utility methods for Logs steps
#
# TODO: Add here comments explaining the reasons we're doing this, 
# to parallelize tests, etc.
#
module LogsHelpers
  LOG_LOCATION = "/src/conjur-server/log/development.log"

  def num_log_lines
    File.new(LOG_LOCATION).readlines.size
  end

  # NOTE: Client code should explicitly call this to save current log size,
  # rather than making it a side affect of getting the current size
  def save_num_log_lines
    @saved_num_lines = num_log_lines
  end

  def num_matches_since_savepoint(msg)
    cur_num_lines = num_log_lines

    validate_savepoint_exists
    validate_log_didnt_shrink(cur_num_lines)

    start_line, end_line = @saved_num_lines, cur_num_lines
    (File.readlines(LOG_LOCATION)[start_line..end_line].grep(/#{msg}/i)).count
  end

  private

  def validate_savepoint_exists
    raise "No savepoint exists.  'save_num_log_lines' must be called before this method" unless @saved_num_lines
  end

  def validate_log_didnt_shrink(cur_num_lines)
    return if cur_num_lines >= @saved_num_lines
    raise "The log has fewer lines (#{cur_num_lines}) than it did before (#{@saved_num_lines})"
  end
end

World(LogsHelpers)
```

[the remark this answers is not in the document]
nessiLahav: Thanks, i changed according to your comment.
Note: i used the hash in order to set infrastructure for multiple bookmarks in the scenario but i actually didn't used it.
By the way, do you think we should add the logs validation to other features ?

cucumber/api/features/step_definitions/logs_steps.rb
```diff
@@ -1,17 +1,13 @@
 # frozen_string_literal: true
 
 Given(/^I save the log data from bookmark$/) do
-  lines_amount_in_bookmark()
-end
-
-Given(/^I save the log data from bookmark "([^"]*)"$/) do |bookmark|
-  lines_amount_in_bookmark(bookmark)
+  save_num_log_lines()
```
jonahx: To make the language consistent with our changes in the helpers file, let's change:
```
Given(/^I save the log data from bookmark$/) do
```
to
```
Given(/^I save my place in the log file$/) do
```

cucumber/api/features/step_definitions/logs_steps.rb
```diff
@@ -1,17 +1,13 @@
...
 end
 
 And(/^The log filtered from bookmark contains messages:$/) do |message|
-  expect(occurences_in_log_filtered_from_bookmark(message)).to eq 1
+  expect(num_matches_since_savepoint(message)).to eq 1
```
jonahx: And here 
```
And(/^The log filtered from bookmark contains messages:$/) do |message|
```
will become
```
And(/^The following appears in the log after my savepoint:$/) do |message|
```

cucumber/api/features/step_definitions/logs_steps.rb
```diff
@@ -1,17 +1,13 @@
...
 end
 
-And(/^The log filtered from bookmark "([^"]*)" contains "([^"]*)" messages:$/) do |bookmark, occurrences, message|
-  expect(occurences_in_log_filtered_from_bookmark(bookmark, message)).to eq occurrences.to_i
+And(/^The log filtered from bookmark "([^"]*)" contains "([^"]*)" messages:$/) do |occurrences, message|
```
jonahx: IIRC we don't use this form.  We're only ever checking "number of matches since the last save."  So this one can be deleted.

PR #993, opened 2019-04-30
Revert command class version

orenbm:
#### What does this PR do?
Reverts the *usage* of command classes to the old syntax, using CommandClass.new.
We still declare at the beginning of each class which errors it might raise to maintain the command_class clarity. 

This makes our code less flakey and yet, still clear.

### DIFF
--- a/app/domain/authentication/authn_k8s/kubectl_exec.rb
+++ b/app/domain/authentication/authn_k8s/kubectl_exec.rb
@@ -36,7 +36,7 @@ module Authentication
 
         wait_for_close_message
 
-        raise CommandTimedOut.new(@container, @pod_name) unless @channel_closed
+        raise Authentication::AuthnK8s::CommandTimedOut.new(@container, @pod_name) unless @channel_closed
 
         # TODO: raise an `WebsocketServerFailure` here in the case of ws :error
 
--- a/app/domain/authentication/authn_k8s/kubectl_exec.rb
+++ b/app/domain/authentication/authn_k8s/kubectl_exec.rb
@@ -10,166 +10,122 @@ require 'authentication/errors'
 module Authentication
   module AuthnK8s
 
-    # Utility class for processing WebSocket messages.
-    class WebSocketMessage
-      class << self
-        def channel_byte(channel_name)
-          channel_number_from_name(channel_name).chr
-        end
+    # Possible Errors Raised:
+    # CommandTimedOut
 
-        def channel_number_from_name(channel_name)
-          channel_names.index(channel_name)
-        end
-
-        def channel_names
-          %w(stdin stdout stderr error resize)
-        end
-      end
-
-      def initialize(msg)
-        @msg = msg
-      end
-
-      def type
-        @msg.type
-      end
-
-      def data
-        @msg.data[1..-1]
-      end
-
-      def channel_name
-        self.class.channel_names[channel_number]
-      end
-
-      def channel_number
-        unless @msg.respond_to?(:data)
-          return self.class.channel_number_from_name('error')
-        end
-
-        @msg.data[0..0].bytes.first
-      end
-    end
-
-    class KubectlExec
-      extend CommandClass::Include
-
-      CommandTimedOut = Authentication::AuthnK8s::CommandTimedOut
-
-      command_class(
-        dependencies: { logger: Rails.logger,
-                        k8s_object_lookup: K8sObjectLookup,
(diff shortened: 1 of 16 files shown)


app/domain/authentication/authn_oidc/authenticate_oidc_conjur_token/validate_and_decrypt_oidc_conjur_token.rb
```diff
@@ -12,45 +12,42 @@ module AuthenticateOidcConjurToken
...
-          def user_name
-            @user_name ||= decoded_body.assoc('user_name').last
-          end
-
-          def expiration_time
-            @expiration_time ||= decoded_body.assoc('expiration_time').last
-          end
+
+      ValidateAndDecryptOidcConjurToken = CommandClass.new(
+        dependencies: {},
```
jonahx: You may want to do this separately, since it's not strictly related to the purpose of this PR,
but we should never have a CommandClass with no dependencies.  Or (a side point) names with the word `And` or `Or` in them.

In this particular case, it looks like the logic here belongs as a secondary constructor on `ConjurToken`.  That is, the primary or canonical way to construct a ConjurToken is by giving it an `id_token_encrypted`, `user_name` and `expiration_time`.

This is alternative way to construct using the contents of a request body (which contains those 3 bits of information, but in a different format).  Thus the secondary constructor would pull out the 3 bits of information, then pass them to the primary ctor, as we do here.

So I'd recommend creating `self.from_request_body` on `ConjurToken`, moving this logic there, and then deleting this class and file.

app/domain/authentication/authn_oidc/get_conjur_oidc_token/authenticator.rb
```diff
@@ -3,62 +3,59 @@
...
+      end
 
-          def call
-            validate_id_token_claims
-            validate_user_info
-            # TODO: validate_nonce
-            true
-          end
+      Authenticator = CommandClass.new(
+        dependencies: {
```
jonahx: This one does actually have a dependency, it's just hardcoded at the moment.  On line 43, `Resource` is a dep.  That should passed in here, which then allows this to be unit tested without hitting the actual db (via a mock for Resource)

app/domain/authentication/authn_oidc/get_conjur_oidc_token/conjur_oidc_token.rb
```diff
@@ -3,105 +3,102 @@
...
-
-          def audit_success
-            @audit_event.(input: @authenticator_input, success: true, message: nil)
-          end
-
-          def audit_failure(err)
-            @audit_event.(input: @authenticator_input, success: false, message: err.message)
-          end
+
+      ConjurOidcToken = CommandClass.new(
```
jonahx: The CommandClass should be named `GetConjurOidcToken` and the containing module can be deleted.

Since all CommandClass's will have some kind of side effects, we should always use verb phrase names, even when the return something.

app/domain/authentication/authn_k8s/kubectl_exec.rb
```diff
@@ -10,166 +10,122 @@
...
-
-      command_class(
-        dependencies: { logger: Rails.logger,
-                        k8s_object_lookup: K8sObjectLookup,
-                        timeout: 5.seconds },
-        inputs: %i( pod_namespace
+    # Possible Errors Raised:
+    # CommandTimedOut
+
+    KubectlExec = CommandClass.new(
```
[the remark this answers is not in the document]
orenbm: this was actually a command class before :)
i only refactored command classes in this PR

app/domain/authentication/authn_oidc/authenticate_oidc_conjur_token/validate_and_decrypt_oidc_conjur_token.rb
orenbm: this looks right but this class isn't used in the current OIDC scope (and is actually blocked in `routes.rb`. i created an issue for this in GH so we'll get to it when we start working on this.

app/domain/authentication/authn_oidc/get_conjur_oidc_token/conjur_oidc_token.rb
orenbm: this will also be addressed while we work on this. the reason we have the module is for separation inside OIDC but i don't want to put any effort in this now.

Issue #982, opened 2019-04-29
Logs are placed in one file


orenbm:
We'd like to have all the logs in one file for east supportablility

PR #998, opened 2019-05-01
Logs in one place

orenbm:
#### What does this PR do?
- Consolidates all log messages to 1 place for better supportability
- Creates a LogMessage & LogMessageWithCode classes
- Adds a module for Errors
- Moves `errors.rb` & `logs.rb` to `domain` level

Connected to #982


app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
```diff
@@ -92,12 +92,12 @@ def validate_security
               user_id: @authenticator_input.username,
               enabled_authenticators: @enabled_authenticators
           )
-          @logger.debug("[OIDC] Security validated")
+          @logger.debug(::LogMessages::Authentication::Security::SecurityValidated.new.to_s)
```
jonahx: Check if you need `to_s` here.  A lot of times (but not always) methods like this will auto-call to `to_s` when you pass them an object.

app/domain/logs.rb
```diff
@@ -0,0 +1,106 @@
...
+
+        PodChannelClosed = ::Util::LogMessageWithCodeClass.new(
+          msg: "Pod '{0}' : channel closed",
+          code: "CONJ00011D"
+        )
+
+        PodChannelData = ::Util::LogMessageWithCodeClass.new(
+          msg: "Pod '{0}', channel '{1}': {2}",
+          code: "CONJ00012D"
+        )
```
jonahx: Looking at this code out of context (ie, because we're not at the `raise` site which would show use the variables we're inputting into the numbered parameters), there's no way to tell what `{0}` and `{1}` and `{2}` mean, and in some of the cases it's not obvious.

I think I'd add comments to clarify.

Or, as an upgrade idea, we could allow syntax like:
```
msg: "Pod '{0-podname}', channel '{1-channel}': {2-message}",
```
And the util class would just ignore everything after the `-`, working exactly the way it does now, only allowing the extra info for clarity.

app/domain/util/log_message_class.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# A factory for creating a LogMessage
+#
+module Util
+  class LogMessageClass
+    def self.new(msg)
+      Class.new(String) do
```
jonahx: Why does it inherit from `String`?

app/domain/util/log_message_with_code_class.rb
```diff
@@ -0,0 +1,13 @@
...
+
+# A factory for creating a LogMessage with a code prefix
+#
+module Util
+  class LogMessageWithCodeClass
+    def self.new(msg:, code:)
+      LogMessageClass.new("#{code} #{msg}")
+    end
+  end
+end
```
jonahx: Small naming point:

What do you think about the name `TrackableLogMessageClass`?

`WithCode` is alright, but it reads a little clunky to me.  And the final purpose of the code is to allow the messages to be tracked and easily found (I think).

app/controllers/authenticate_controller.rb
```diff
@@ -120,29 +120,29 @@ def handle_authentication_error(err)
...
-      Authentication::AuthnOidc::IdTokenVerifyFailed,
-      Authentication::AuthnOidc::IdTokenInvalidFormat,
+    when Errors::Authentication::Security::UserNotAuthorizedInConjur,
+      Errors::Authentication::Security::NotWhitelisted,
+      Errors::Authentication::Security::ServiceNotDefined,
+      Errors::Authentication::Security::UserNotDefinedInConjur,
+      Errors::Authentication::Security::AccountNotDefined,
+      Errors::Authentication::AuthnOidc::IdTokenFieldNotFoundOrEmpty,
+      Errors::Authentication::AuthnOidc::IdTokenVerifyFailed,
+      Errors::Authentication::AuthnOidc::IdTokenInvalidFormat,
```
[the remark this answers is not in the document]
orenbm: that's interesting. the idea of having it in the `domain` scope was so in the future we don't have several files and then when you add a new error it's quite easy to know what will be it's code (as you need to look for the highest code in the file).

but i agree that writing
`raise MissingServiceAccountDir`
is much better than
`raise Errors::Authentication::AuthnK8s::KubeClientFactory::MissingServiceAccountDir`

i actually think that the second benefit is better so i'm tending to moving the files back to the `authentication` scope. i'll think about it a bit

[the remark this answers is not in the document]
orenbm: consulted with @nessiLahav and we got to the conclusion that it's important to have the files in `domain` scope as in the future a developer who adds an error/log can easily miss the other files and we'll have a duplicate error/log code.

i'll go with your suggestion of declaring the `Err` at the beginning of the class/module

app/domain/authentication/authn_oidc/authenticate_id_token/authenticate.rb
orenbm: checked it. without the `new.to_s` it prints `::LogMessages::Authentication::Security::SecurityValidated`

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
orenbm: i wrote some code for this:
```
module A
  module B
    class C
      Err = Errors::A::B::C
      # somwhere inside...
      raise Err::SomeErr
    end
  end
end
```

the problem is that some classes raise errors from different domains so it makes the code very error-prone. it is possible to explicitly call the errors in cases the call is to an error/log from another module but i'm afraid someone in the future will get confused why his error isn't working and is mapped to some other class

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
+# frozen_string_literal: true
+
+require 'util/error_class'
+
+unless defined? Errors::Authentication::AuthenticatorNotFound
+  # this wrapper is here so these classes will not be loaded by Rails
```
shuligh: This wrapper prevents classes from being loaded by Rails

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+  # auto-load. #TODO: fix this in a proper manner
```
shuligh: What does "Fix this in a proper manner" mean - is that a comment for me?

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+
+  module Errors
+    module Authentication
+
+      AuthenticatorNotFound = ::Util::ErrorClassWithCode.new(
+        msg: "'{0}' wasn't in the available authenticators",
```
shuligh: What does "**in** the authenticators" mean? just want to check the preposition is correct

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00006E"
+        )
+
+        UserNotDefinedInConjur = ::Util::ErrorClassWithCode.new(
+          msg: "User '{0}' is not defined in Conjur",
+          code: "CONJ00007E"
+        )
+
+        AccountNotDefined = ::Util::ErrorClassWithCode.new(
+          msg: "account '{0}' is not defined in Conjur",
```
shuligh: Account (capital A)

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00008E"
+        )
+
+      end
+
+      module RequestBody
+
+        MissingRequestParam = ::Util::ErrorClassWithCode.new(
+          msg: "field '{0}' is missing or empty in request body",
```
shuligh: Field (cap F)

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00009E"
+        )
+
+      end
+
+      module AuthnOidc
+
+        ProviderDiscoveryTimeout = ::Util::ErrorClassWithCode.new(
+          msg: "OIDC provider discovery failed with timeout error (provider_uri='{0}'). Reason: '{1}'",
```
shuligh: Provider (cap P)

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00010E"
+        )
+
+        ProviderDiscoveryFailed = ::Util::ErrorClassWithCode.new(
+          msg: "OIDC provider discovery failed (provider_uri='{0}'). Reason: '{1}'",
```
shuligh: Provider (cap P)

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00012E"
+        )
+
+        IdTokenFieldNotFoundOrEmpty = ::Util::ErrorClassWithCode.new(
+          msg: "Field '{0}' not found or empty in ID Token",
+          code: "CONJ00013E"
+        )
+
+        IdTokenInvalidFormat = ::Util::ErrorClassWithCode.new(
+          msg: "Invalid ID Token Format (3rdPartyError ='{0}')",
```
shuligh: format (lowercase f)

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00014E"
+        )
+
+        IdTokenVerifyFailed = ::Util::ErrorClassWithCode.new(
+          msg: "ID Token verification failed (3rdPartyError ='{0}')",
+          code: "CONJ00015E"
+        )
+
+        IdTokenExpired = ::Util::ErrorClassWithCode.new(
+          msg: "ID Token Expired",
```
shuligh: expired (lc e)

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          msg: "admin user is not allowed to authenticate with OIDC",
+          code: "CONJ00017E"
+        )
+
+      end
+
+      module AuthnIam
+
+        InvalidAWSHeaders = ::Util::ErrorClassWithCode.new(
+          msg: "'Invalid or Expired AWS Headers: {0}",
```
shuligh: Invalid or expired AWS headers - sentence caps (only first word and proper nouns/acronyms are capitalized)

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00022E"
+        )
+
+        CSRNamespaceMismatch = ::Util::ErrorClassWithCode.new(
+          msg: "Namespace in SPIFFE ID '{0}' must match namespace implied by common name '{1}'",
+          code: "CONJ00023E"
+        )
+
+        PodNotFound = ::Util::ErrorClassWithCode.new(
+          msg: "No Pod found for podname '{0}' in namespace '{1}'",
```
shuligh: is pod capilatised? Is it a name of a product? I think it should be small letter.
"No pod found for '{0}' in namespace '{1}'" - i also think podname may be superfluous in this message. We can discuss this

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00025E"
+        )
+
+        ControllerNotFound = ::Util::ErrorClassWithCode.new(
+          msg: "Kubernetes {0} {1} not found in namespace {2}",
+          code: "CONJ00026E"
+        )
+
+        CertInstallationError = ::Util::ErrorClassWithCode.new(
+          msg: "Cert could not be copied to pod: {0}",
```
shuligh: Certificate could not be .....
Is that what we mean here? Or was this specifically supposed to be "Cert"?

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00027E"
+        )
+
+        ContainerNotFound = ::Util::ErrorClassWithCode.new(
+          msg: "Container {0} was not found for requesting pod",
+          code: "CONJ00028E"
+        )
+
+        MissingClientCertificate = ::Util::ErrorClassWithCode.new(
+          msg: "The client SSL cert is missing from the header",
```
shuligh: Client SSL certificate...
(no need for "The" because we haven't used it anywhere else

app/domain/logs.rb
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+
+require 'util/log_message_with_code_class'
+
+unless defined? LogMessages::Authentication::Security::SecurityValidated
+  # this wrapper is here so these classes will not be loaded by Rails
```
shuligh: This wrapper prevent the classes from being loaded by Rails auto-load

app/domain/logs.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      module AuthnOidc
+
+        ExtractedUsernameFromIDToked = ::Util::LogMessageWithCodeClass.new(
+          msg: "Extracted username '{0}' from ID Token",
+          code: "CONJ00004D"
+        )
+
+        IDTokenDecodeSuccess = ::Util::LogMessageWithCodeClass.new(
+          msg: "ID Token Decode succeeded",
```
shuligh: I think decode can be with a small d

app/domain/logs.rb
```diff
@@ -0,0 +1,106 @@
...
+          code: "CONJ00005D"
+        )
+
+        IDTokenVerificationSuccess = ::Util::LogMessageWithCodeClass.new(
+          msg: "ID Token verification succeeded",
+          code: "CONJ00006D"
+        )
+
+        OIDCProviderUri = ::Util::LogMessageWithCodeClass.new(
+          msg: "Working with provider {0}",
```
shuligh: is this the OIDC Provider?

app/domain/logs.rb
```diff
@@ -0,0 +1,106 @@
...
+          code: "CONJ00007D"
+        )
+
+        OIDCProviderDiscoverySuccess = ::Util::LogMessageWithCodeClass.new(
+          msg: "Provider discovery succeeded",
```
shuligh: OIDC Provider?

app/domain/logs.rb
```diff
@@ -0,0 +1,106 @@
...
+          code: "CONJ00008D"
+        )
+
+        FetchProviderCertsSuccess = ::Util::LogMessageWithCodeClass.new(
+          msg: "Fetched provider certificates successfully",
```
shuligh: Fetched OIDC Provider

app/domain/logs.rb
```diff
@@ -0,0 +1,106 @@
...
+          code: "CONJ00013D"
+        )
+
+        PodError = ::Util::LogMessageWithCodeClass.new(
+          msg: "Pod '{0}' error : '{1}'",
+          code: "CONJ00014D"
+        )
+
+        CopySSLToPod = ::Util::LogMessageWithCodeClass.new(
+          msg: "Copying SSL cert to {0}/{1}",
```
shuligh: SSL certificate

app/domain/util/log_message_class.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# A factory for creating a LogMessage
```
shuligh: # A factory for creating a log message

app/domain/util/log_message_with_code_class.rb
orenbm: nice

app/domain/errors.rb
orenbm: this is actually a note for future developers but i'll change it :)

app/domain/errors.rb
orenbm: this is actually a note for future developers :)

app/domain/logs.rb
orenbm: great idea. done :)

app/domain/util/log_message_class.rb
orenbm: No good reason. removed it.

app/domain/errors.rb
orenbm: you're right

app/domain/logs.rb
orenbm: yep. I'll change to "Working with OIDC Provider"

app/domain/logs.rb
orenbm: yep

app/domain/errors.rb
orenbm: the available authenticators are all those which are implemented on conjur (oidc, ldap, k8s & iam)

orenbm: maybe change to "Authenticator '{0}' is not implemented in Conjur"?

app/domain/logs.rb
```diff
@@ -0,0 +1,106 @@
...
+          msg: "Origin validated",
+          code: "CONJ00003D"
+        )
+
+      end
+
+      module AuthnOidc
+
+        ExtractedUsernameFromIDToked = ::Util::LogMessageWithCodeClass.new(
+          msg: "Extracted username '{0}' from ID Token",
```
InbalZilberman: Can you please log also "according to id-token-user-property {1}" and give it the value?

app/domain/errors.rb
InbalZilberman: Can you please specify what is in the available authenticators add this " available authenticators are {1}"

app/domain/errors.rb
InbalZilberman: what do you mean by discovery?

app/domain/errors.rb
```diff
@@ -0,0 +1,213 @@
...
+          code: "CONJ00011E"
+        )
+
+        ProviderFetchCertificateFailed = ::Util::ErrorClassWithCode.new(
+          msg: "Failed to fetch certificate from OIDC provider (provider_uri='{0}'). Reason: '{1}'",
+          code: "CONJ00012E"
+        )
+
+        IdTokenFieldNotFoundOrEmpty = ::Util::ErrorClassWithCode.new(
+          msg: "Field '{0}' not found or empty in ID Token",
```
InbalZilberman: isnt it related to id-token-user-property? if so please add a message linking them

app/domain/errors.rb
orenbm: This is an OIDC terminology. it's an endpoint from which we get metadata on the Provider

app/domain/errors.rb
jonahx: You could say "wasn't in the list of available authenticators" if you wanted to be explicit.  Oren your version works grammatically but it want to maintain the distinction between "available" (you've turned it on and configured it) vs "implemented" (the devs at Conjur have built it and it exists).  We could take your version but just change that word:

```
Authenticator '{0}' is not currently available
```

But honestly any of these wording sound ok to me.

orenbm: i disagree on the definition. 

the installed authenticators are defined:
```
def installed_authenticators
    @installed_authenticators ||= Authentication::InstalledAuthenticators.authenticators(ENV)
  end

def self.authenticators(env, authentication_module: ::Authentication)
      loaded_authenticators(authentication_module)
        .select { |cls| valid?(cls) }
        .map { |cls| [url_for(cls), authenticator_instance(cls, env)] }
        .to_h
    end
```

the way we mark an authenticator as valid is:
```
def valid?
        valid_name? && valid_parent_name? && valid_interface?
end

      def valid_name?
        own_name == 'Authenticator'
      end

      def valid_parent_name?
        parent_name =~ /^Authn/
      end

      def valid_interface?
        @cls.method_defined?(:valid?)
      end
```

So, in conclusion, the available/installed authenticators are those who are implemented in conjur code (they have an `Authn-somehting::Authenticator` class with a `valid?` method). so it is exactly what you call "implemented" (the devs at Conjur have built it and it exists) and not your definition of "available" (you've turned it on and configured it).

am i missing something?

app/domain/errors.rb
```diff
@@ -101,17 +101,17 @@ module RequestBody
       module AuthnOidc
 
         ProviderDiscoveryTimeout = ::Util::TrackableErrorClass.new(
-          msg: "OIDC provider discovery failed with timeout error (provider_uri='{0}'). Reason: '{1}'",
+          msg: "OIDC provider discovery failed with timeout error (Provider URI: '{0}'). Reason: '{1}'",
```
shuligh: OIDC Provider - capital P for Provider... all through

app/domain/errors.rb
```diff
@@ -0,0 +1,252 @@
...
+
+      InvalidOrigin = ::Util::TrackableErrorClass.new(
+        msg: "Invalid origin",
+        code: "CONJ00003E"
+      )
+
+      module AuthenticatorClass
+
+        DoesntStartWithAuthn = ::Util::TrackableErrorClass.new(
+          msg: "'{0-authenticator-parent-name}' is not a valid authenticator parent module, because it does " +
```
shuligh: no comma needed before because it does....

app/domain/errors.rb
```diff
@@ -0,0 +1,252 @@
...
+        )
+
+        NotNamedAuthenticator = ::Util::TrackableErrorClass.new(
+          msg: "'{0-authenticator-name}' is not a valid authenticator name. " +
+            "The actual class implementing the authenticator must be named 'Authenticator'",
+          code: "CONJ00039E"
+        )
+
+        MissingValidMethod = ::Util::TrackableErrorClass.new(
+          msg: "'{0-authenticator-name}' is not a valid authenticator, because " +
```
shuligh: no comma before because

app/domain/errors.rb
shuligh: So then ho wabot "Authenticator '{0}' does not exist in Conjur"?

jonahx: @orenbm You are correct.  I read it wrong.

My reply was based on the assumption that this was the error raised when the authenticator wasn't in `CONJUR_AUTHENTICATORS`.  It looks like you're saying this is for when the authenticator doesn't even exist, so I like your message.

Issue #927, opened 2019-03-20
Optimize certificate retrieval


moticless:
Optimize provider certificate retrieval (aka JWKS) to happen only once and not per authentication request. If ID Token verification get failed only then try retrieve once again. Note that since the certificate become shared among requests there should be some synchronization mechanism during certificate refresh.

child of #974

orenbm:
@moticless @nessiLahav this was implemented with the caching mechanism, right?

PR #999, opened 2019-05-05
Optimize certificate cache

moticless:
#### What does this PR do?
Adding cache mechanism for provider certificates that will refresh them from provider only when client fails to decode with cached certificate.

Fix cucumber tests - fix tests to lookup for correct errors in log

OIDC authentication verifies given TokenID by retrieving certificate from OIDC provider and verifying token signature. Following this feature it won't retrieve certificate per each request but cache it. A new retrieval will be triggered only on failure.

Connected to #927 

read relevant section at :  https://ca-il-confluence.il.cyber-ark.com/display/rndp/Research++-+rails+caching+mechanism+and+cert+optimization#?lucidIFH-viewer-599ae143=1

Then start with certificate_cache.rb - it stands on its own.

- Run several oidc requests and verify only one time provider is accessed to retrieve certificate.
- Rotate provider certificate and verify OIDC handler knows how to recover and retrieve new certificate.


app/domain/authentication/authn_oidc/authenticate_id_token/fetch_provider_certificate.rb
```diff
@@ -12,21 +12,29 @@ module AuthenticateIdToken
 
       FetchProviderCertificate = CommandClass.new(
         dependencies: {
-          logger: Rails.logger
+          logger: Rails.logger,
+          cert_cache: lambda {::Authentication::AuthnOidc::CertificateCache.instance}
```
orenbm: @jonahx is this ok?

nessiLahav:
@moticless Please do the following:
1. add logs
2. also test it with Okta
3. crate a new exception on X failures for the last Y hours, because it should be a rare situation  (maybe an attacker or multiple mistakes in a row)

app/domain/authentication/authn_oidc/authenticate_id_token/fetch_provider_certificate.rb
```diff
@@ -8,28 +8,27 @@ module AuthenticateIdToken
...
         dependencies: {
-          logger: Rails.logger
+          logger: Rails.logger,
+          open_id_discovery_service: OpenIDConnect::Discovery::Provider::Config
         },
         inputs: %i(provider_uri)
       ) do
 
         def call
-          @logger.debug(Log::OIDCProviderUri.new(@provider_uri).to_s)
```
orenbm: was this removed intentionally? i re-added it

orenbm: added it like this so we won't have code in `call` method:
```
        def call
          log_provider_uri
          discover_provider
          fetch_certs
        end

        private

        def log_provider_uri
          @logger.debug(Log::OIDCProviderUri.new(@provider_uri).to_s)
        end
```

app/domain/authentication/authn_oidc/authenticate_id_token/fetch_provider_certificate.rb
```diff
@@ -38,9 +37,9 @@ def discover_provider
         end
 
         def fetch_certs
-          jwks = @discovered_provider.jwks
-          @logger.debug(Log::FetchProviderCertsSuccess.new.to_s)
-          jwks
+          @discovered_provider.jwks.tap do
+            @logger.debug(Log::FetchProviderCertsSuccess.new.to_s)
+          end
```
orenbm: that's nice

PR #1001, opened 2019-05-06
ErrorClass & LogMessageClass have UTs

orenbm:
#### What does this PR do?
- Add UTs for templating in ErrorClass 
- Add UTs for LogMessageClass
- Add UTs for TrackableErrorClass 
- Add UTs for TrackableLogMessageClass
- TrackableErrorClass composes ErrorClass instead of inheritance

Connected to #1000

### DIFF
--- a/spec/app/domain/util/log_message_spec.rb
+++ b/spec/app/domain/util/log_message_spec.rb
@@ -3,8 +3,8 @@
 require 'spec_helper'
 
 RSpec.describe 'Util::LogMessageClass' do
-  context 'object' do
-    let(:log_message) { "some log message" }
+  context 'pre-determined, static log message' do
+    let(:log_message) { 'some log message' }
     subject(:log_message_class) { Util::LogMessageClass.new(log_message) }
 
     it 'has the expected message' do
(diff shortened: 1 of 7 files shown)


orenbm:
@jonahx the build will fail on some OIDC tests. this is unrelated and is fixed in @moticless 's PR

spec/app/domain/util/trackable_log_message_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'Util::TrackableLogMessageClass' do
+  context 'object' do
```
jonahx: You can leave context out here.  You only need it when you have multiple contexts.

spec/app/domain/util/log_message_spec.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'Util::LogMessageClass' do
+  context 'object' do
```
jonahx:
```
context 'pre-determined, static log message'
```

spec/app/domain/util/log_message_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    let(:log_message) { "some log message" }
+    subject(:log_message_class) { Util::LogMessageClass.new(log_message) }
+
+    it 'has the expected message' do
+      expect(log_message_class.new.to_s).to eq(log_message)
+    end
+  end
+
+  context 'templating' do
+    let(:log_message_template) { "Variable is \#{0}." }
```
jonahx: small point: can use single quotes to avoid the backslash.  you might consider using a character other than `#` too, because at first I thought it was part of the required template string, but then I realized you were just using it as a marker to you could test more clearly against nil and empty strings.

spec/app/domain/util/log_message_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      expect(log_message_class.new(["aA", "bB"]).to_s).to eq("Variable is #[\"aA\", \"bB\"].")
+    end
+
+    it 'handles overflow args' do
+      expect(log_message_class.new("AA", "BB").to_s).to eq("Variable is #AA.")
+    end
+
+    it 'handles underflow args' do
+      expect(log_message_class.new.to_s).to eq("Variable is #\{0\}.")
+    end
```
jonahx: Seems like a missing template variable should error.

spec/app/domain/util/log_message_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      subject(:descriptive_log_message_class) { Util::LogMessageClass.new(descriptive_log_message_template) }
+
+      it 'ignores the description' do
+        expect(descriptive_log_message_class.new("var").to_s)
+          .to eq("Variable is var.")
+      end
+    end
+  end
+end
```
jonahx: Nice, clear tests 👍

spec/app/domain/util/log_message_spec.rb
orenbm: good to know, thanks!

The # was actually from the existing error_class.spec so i removed it from there too

spec/app/domain/util/log_message_spec.rb
orenbm: wanted to maintain the behaviour of error_class

jonahx:
Changes look good to me.  I think we can merge this one.  I did, however, notice something that merits a followup issue:
image

What I noticed at first was the duplication in the specs for this classes.  Both specs are thorough and good, but they test exactly the same thing, so there's a lot of duplication that should be removed.  Per the image above, fundamentally the only difference is that one of these is an error and one is not. 

So this is a good time to step back and ask "What is the *essence* of the shared functionality here?"

Fundamentally, what we've done here is built simple, tiny templating system.  

We can pull that out into it's own object, allowing us to solve the duplication with composition rather than inheritance:

```ruby
module Util

  class SimpleTemplate
    def initialize(template_str)
      @template = template_str
    end

    def render(*args)
      @args.each_with_index.reduce(msg) do |m, (x, arg_index)|
        x_stringified = x.nil? ? 'nil' : x.to_s
        m.gsub(Regexp.new("\\{#{arg_index}-?[a-zA-Z-]*}"), x_stringified)
      end
    end
  end

  class LogMessageClass
    def self.new(msg)
      template = SimpleTemplate.new(msg)

      Class.new do
        def initialize(*args)
          @args = args
        end
        define_method(:to_s) do
          template.render(@args)
        end
      end
    end
  end

  class ErrorClass
    def self.new(msg)
      template = SimpleTemplate.new(msg)

      Class.new(RuntimeError) do
        def initialize(*args)
          @args = args
        end
        define_method(:to_s) do
          template.render(@args)
        end
      end
    end
  end

end
```
With this change, the UTs that currently exist for `ErrorClass` and `LogMessageClass` can be moved to `simple_template_spec.rb`.  You can then replace the UTs for `ErrorClass` and `LogMessageClass` with very simple test cases (only 1 or 2 test assertions for each, just as a sanity check -- since all the real templating logic will now be tested in `simple_template_spec.rb`. 

You still have a little duplication left between `ErrorClass` and `LogMessageClass`, but that duplication is mere boilerplate now.  There is not real logical duplication.

You might ask: If the only difference between `ErrorClass` and `LogMessageClass` is that one inherits from `Object` and one from `RuntimeError`, why not make something like a `TemplateClassFactory` that can be used like: `ErrorClass = TemplateClassFactory.make_new_class(RuntimeError)` and `ErrorClass = TemplateClassFactory.make_new_class(Object)`?

And you could.  It wouldn't be *terrible*, but this is case of:

> All problems in computer science can be solved by another level of indirection ...except for the problem of too many layers of indirection.

Because we'd have a factory for creating classes that we in turn use to create the actual objects (error and log messages) that we actually care about in our application.  We'd be doing meta-meta-programming.  

In addition, the template abstraction is the natural one, and it just "felt right" once I realized it.  The fact we're using template strings to create log messages and errors is really secondary.  So it's less confusing, and makes re-use more likely as well.

orenbm:
@jonahx i added an issue for reomving the duplication. thanks!

PR #1002, opened 2019-05-06
Feature docs for authenticators endpoints

orenbm:
#### What does this PR do?
Adds a feature design for the `/authenticators` & `/authenticators/health` endpoints

While developing the OIDC Authenticator we came upon the need to verify that the authenticator is configured properly after the configuration process. This led us to define the Authenticator Health-Check API and during the research for the feature definition we found the `/authenticators` endpoint and its flaws.

#### Open Questions
- Is there real value in the `/authenticators` Endpoint? Do users need to check which authenticators exist before trying to login?

authenticators_api.md


design/authenticators_api.md
```diff
@@ -0,0 +1,178 @@
...
+(e.g they have an `Authn-somehting::Authenticator` class with a `valid?` method).
+
+2. configured: these authenticators have a webservice called 
+`conjur/authn-something/webservice`.
+
+3. enabled: these authenticators are whitelisted in the CONJUR_AUTHENTICATORS variable.
+
+Section 1 has important data but it shouldn't be here, but rather in the docs. It has
+nothing to do with the given Conjur instance and just describes the authenticators that
+are implemented in Conjur.
```
sgnn7: I would disagree here as keeping docs and codebase in lock-step would be incredibly difficult. Ideally you could use this data to somehow enable new authenticators (even though we currently have no method for doing this).

With that said, I think we should have different permissions to see each part of the data:
- `1` - admin account only
- `2` - high-privilege account (operator that may or may not be the main admin account)
- `3` - any user (including unauthenticated ones)

design/authenticators_api.md
```diff
@@ -0,0 +1,178 @@
...
+We can learn 2 things from the above:
+
+1. The Conjur user needs to authenticate according to the response. So this request must be applicable **without** a Conjur access token. 
+1. The Conjur operator needs to know which actions are needed for completing the configuration,
+which may consist sensitive data. So this request must be applicable **only with** a Conjur access token.
+
+This leads us to 2 API endpoints:
+
+1. `/authenticators` (for Conjur users)
+2. `/authenticators/health` (for Conjur operators)
```
sgnn7: Second endpoint `/authenticators/health` does not follow general guidelines for RESTful applications and is inconsistent with our endpoint definition patterns. I think there is also no need for an aggregated authenticator health check.

Anyways, to keep this short, this would fit better with our codebase patterns:
1. `/authenticators`
1. `/authenticators/<authenticator_id>`, where the `<authenticator_id>` is the individual authenticator information/status can be retrieved. We could even do `/authenticators/<authenticator_id>/status` if only status is needed

design/authenticators_api.md
```diff
@@ -0,0 +1,178 @@
...
+```
+{
+   “authenticators”:
+   {
+      "authn": "ok",
+      "authn-1": "ok",
+      "authn-2": "Authentication::Security::NotWhitelisted: CONJ00004E 'authn-2' is not whitelisted in CONJUR_AUTHENTICATORS", 
+   },
+   "ok": false
+}
```
sgnn7: If we do end up having an aggregated list, the custom format might not be advisable as you then need harder logic to monitor the endpoint. I think we probably should lean toward something easier for machine processing:
```json
{
   "authenticators":
   {
      "authn": {
        "status": "ok"
      },
      "authn-1": {
        "status": "ok"
      },
      "authn-2": {
        "status": "error",
        "error": "Authentication::Security::NotWhitelisted",
        "code": "CONJ00004E",
        "message": "'authn-2' is not whitelisted in CONJUR_AUTHENTICATORS",
      }
   },
   "status": "error"
}
```

design/authenticators_api.md
```diff
@@ -0,0 +1,178 @@
...
+```
+- The operator understands the issue from the response and fixes it
+- The operator runs the request once again until he gets a healthy response
+
+#### `/info` endpoint
+
+Another endpoint we should address is `evoke`'s `/info` endpoint, which contains the data of the `/authenticators` endpoint
+(with other info such as release, version, etc.). This endpoint calls `http://localhost/authenticators` and adds the output to the info
+json. Having the same data in 2 different places is not optimal as it can confuse the user (and developers), so it would make sense to move the
+implementation of the `/authenticators` endpoint to `evoke`, and remove the `/authenticators` endpoint.
```
sgnn7: I don't think we should move this endpoint to evoke because health checks will probably be needed/wanted in OSS in the future and also that will leave OSS without this `authenticators` functionality though I'm not sure how feasible this would be on a technical level.

design/authenticators_health_api.md
```diff
@@ -0,0 +1,231 @@
...
+This can be iterative as there might be several issues. It is possible to assume that in the case 
+above, the operator will solve this with the user and will ask him to try to authenticate again and 
+if it's still Unauthorized will solve the configuration issues one-by-one. 
+
+### Process Logic
+
+1. The Conjur operator logs into Conjur
+1. The Conjur operator configures the Authenticators according to the docs
+    1. We might have several authenticators configured
+1. The operator runs the Healthcheck request: `GET /authenticators/health`
```
sgnn7: Same comment as above about avoiding using `/health` as a suffix

design/authenticators_health_api.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+Let's go over the sections and decide which group should be the one we list in the response and 
+check their health.
+
+##### Installed
+The "installed" authenticators are those who meet to following criteria:
+
+- It's class name is "Authenticator"
+- It's parent module starts with "Authn"
```
sgnn7: Nit: `It's` -> `Its`

design/authenticators_health_api.md
```diff
@@ -0,0 +1,231 @@
...
+- 503 Service Unavailable
+    - Con: This doesn't really tell the story as the service that we're calling - The Authenticator health service - is actually available. 
+- 502 Bad Gateway & 504 Gateway Timeout
+    - Irrelevant as this is not a gateway.
+
+So response codes from the 4xx & the 5xx families aren't quite good.
+
+Another option is to return a 200 OK response. Yes, there is an error and 200 doesn't indicate that but the response body will. This also makes sense as the request itself succeeded and the server returned a response successfully so it makes sense to return a successful code.
+
+Conclusion: return 200 OK.
```
sgnn7: 200 should never be returned for faulty components. 500 is general but you can add messages to the response that will elaborate on what the issue is in both html and json formats.

design/authenticators_api.md
orenbm: i mentioned this in the doc itself but having 2 & 3 in the current way is very confusing. I'm a conjur dev and it took me a while to understand what is what so a user will be confused for sure. in my suggestion we will have just one piece of data - is the authenticator available for authentication. _how_ we define what is available (configured + enabled in the env) is too much data imo.

orenbm: regarding 1 - you have the version in the `info` endpoint so you can go to the documentation of that version and check which authenticators are implemented in that version.

design/authenticators_api.md
orenbm: can you elaborate on the RESTful guidelines or refer me to some place? i'd like to better understand this.

why are you against the aggregated authenticator health check? let's say it will be `/authenticators/status`.

design/authenticators_api.md
orenbm: i like it. changed in the doc

design/authenticators_api.md
orenbm: that's exactly what i wrote in the following sentence :)

design/authenticators_health_api.md
orenbm: correct, thanks

design/authenticators_health_api.md
orenbm: i get your point about 200. i'll move to 5xx.

i noticed now that the general health-check returns `502 Bad Gateway` in case one of the services is unavailable. While it makes sense there in some way as conjur is being used as a gateway to the services, here it seems pretty drunk so i have to agree with you about 500. I'll change it in the doc.

design/authenticators_api.md
izgeri: I don't think we have versioned documentation right now, so I'm not sure in practice someone would be able to discover implemented authenticators this way.

design/authenticators_api.md
```diff
@@ -0,0 +1,187 @@
...
+However, as `evoke` is not part of the OSS, we will leave the implementation in the `/authenticators` endpoint and will not change the implementation in `evoke`.
+
+### Security
+
+#### `/authenticators` Endpoint
+
+Although the response reveals which authenticators are configured and this request is prone
+for brute-force attacks, the value of this endpoint for the user is high and can
+provide the information needed for authentication. Furthermore, we don't say _why_
```
izgeri: from my reading above this endpoint only says the authenticators that are validly configured for use. can you explain more why you note:

> we don't say _why_ the authenticator is invalid

given that it doesn't show invalid authenticators at all, I do not follow how that is an argument for making the endpoint open (though I do agree it is fine for this endpoint to be open)

design/authenticators_health_api.md
```diff
@@ -0,0 +1,233 @@
...
+
+This feature lets the person who configures the Authenticator to get an immediate feedback 
+on the configuration, before any user needs to run an authentication request.
+
+For example, the OIDC Authenticator configuration includes whitelisting the authenticator in the CONJUR_AUTHENTICATORS variable:
+
+`CONJUR_AUTHENTICATORS=authn-oidc/adfs,authn`
+
+Let's say the Conjur Operator configured the OIDC Authenticator and forgot to whitelist 
+the authenticator in the CONJUR_AUTHENTICATORS variable. Without the Helthcheck API,
```
izgeri:
```suggestion
the authenticator in the CONJUR_AUTHENTICATORS variable. Without the Healthcheck API, 
```

design/authenticators_health_api.md
```diff
@@ -0,0 +1,233 @@
...
+    1. In case all the authenticators are healthy, a response with code 200 and the following body:
+        
+        ```
+        {
+           “authenticators”:
+           {
+              "authn-1": "ok",
+              "authn-2": "ok"
+           },
+           "ok": true
```
izgeri: should this be `"status"` instead of `"ok"` as above?

design/authenticators_health_api.md
```diff
@@ -0,0 +1,233 @@
...
+    1. In case some of the authenticators aren't healthy, a response with code 500 (more info [here](authenticators_health_api.md#response-code-for-unhealthy-authenticators)) and the following body:
+
+        ```
+        {
+           “authenticators”:
+           {
+              "authn-1": "ok",
+              "authn-2": "<configuration error>",
+           },
+           "ok": true
```
izgeri: should this be `"status": "error"` as above? can we copy the response format from the previous design doc here as well?

design/authenticators_health_api.md
```diff
@@ -0,0 +1,233 @@
...
+A possible response can look like this:
+
+```
+{
+   “authenticators”:
+   {
+      "authn": "ok",
+      "authn-1": "ok",
+      "authn-2": "not-implemented"
+      "authn-3": "not-implemented",
```
izgeri: it would be interested to see here what it will look like given the revised response format above as well

design/authenticators_api.md
sgnn7: https://restfulapi.net/resource-naming/

The general idea is that if you have an endpoint that aggregates results about a "resource", then it should end with `/<plural_resource_name>`.

If you are pulling information about a single resource, then the suffix should be `/<plural_resource_name>/<resource_id>` with an option for specialized pages: `/<plural_resource_name>/<resource_id>/<specific_info>`

sgnn7: > why are you against the aggregated authenticator health check? let's say it will be /authenticators/status.

Aggregated list is fine in some cases but then you create tight coupling between each component. 

With aggregated status page, `/authenticators` page needs to know:
- How to list authenticators
- How to query their individual statuses
- How to unify their statuses into a single response

This approach also cannot be run in parallel

With individual resource page, `/authenticators` page needs to know:
- Just how to list authenticators

Each individual page can then have self-contained code on how to evaluate its status

design/authenticators_api.md
sgnn7: > having 2 & 3 in the current way is very confusing. 

This might be something that we should look into by separating administrative endpoints with a specific prefix (`<host>/api/foo/bar` vs `<host>/api/admin/foo/bar`)

> regarding 1 - you have the version in the info endpoint so you can go to the documentation of that version and check which authenticators are implemented in that version.

If you think about how much work it is to get the relevant info they need to get things working to the developer vs a simple rest endpoint, I don't think it's a good developer experience. I always get reminded of how terrible it is to look for OpenShift functionality and after failing to get something running for a while, you notice that the google search linked you to an ancient version of APIs.

orenbm: ok, here's another approach. **As a user**, all that matters to me is which authentication are available for authentication. i have nothing to do with configuration, so when i hit the `/authenticators` endpoint I'll get
```
{
   “authenticators”:
   [
      "authn",
      "authn-1"
   ]
}
``` 
even though `authn-2` is implemented in conjur in the current version but not configured.

**As an operator** i do care about which authenticators i can configure so when i hit the `/authenticators/status` endpoint i'll get all the authenticators that are implemented in the current version of conjur.

having all the authenticators in the general `/info` endpoint is still too much information imo (not security-wise but ux) and can mislead a user. as a user i'm not aware of what makes an authenticator available for configuration so if i see `authn-2` in the response i can easily think that i can authenticate with it.

@izgeri @sgnn7 wdyt?

design/authenticators_api.md
orenbm: i was just stating that saying why an authenticator is invalid is insecure and because we don't do that then it is ok.

design/authenticators_health_api.md
orenbm: i was following the behaviour of our general health-check to not break ux:
```
{
  "services": {
    "ldap-sync": "ok",
    "possum": "ok",
    "ui": "ok",
    "ok": true
  },
  "database": {
    "ok": true,
    "connect": {
      "main": "ok"
    },
    "free_space": {
      "main": {
        "kbytes": 63672,
        "inodes": 128876
      }
    },
    "replication_status": {
      "pg_current_xlog_location": "0/23ECDD0",
      "pg_current_xlog_location_bytes": 37670352
    }
  },
  "ok": true
}
```

design/authenticators_health_api.md
orenbm: oh this should actually be "false", thanks!

orenbm: and same as above - using `"ok": false` instead of `"status": "error"` to maintain ux of general health-check

design/authenticators_health_api.md
orenbm: oh i see what happened now. i took srdjan's suggestion without looking at the status. i'll change it in the doc to be like the health-check

design/authenticators_health_api.md
orenbm: changed in the doc

InbalZilberman:
@orenbm canwe check first if info doesnt show the authenticators status? like @izgeri is saying?
@izgeri why dont we show this info?

design/authenticators_api.md
orenbm: @izgeri @sgnn7 would love to get your comments here :)

design/authenticators_api.md
orenbm: now that i think about it, it makes sense also UX-wise. We intend the operator to use this endpoint after he finished to configure a single authenticator. so if, for example, i just configured the `authn-oidc` authenticator, then i'd like to know the status/health of that authenticator. in the aggregated response the `authn-oidc` health is the signal, and the rest of the authenticators are just noise.

Thanks for this point :)

design/authenticators_health_api.md
orenbm: @sgnn7 i understand that here you're referring to not using a suffix at all but i also think you prefer `/authenticators/<authenticator_id>/status` over `/authenticators/<authenticator_id>/health`. can you elaborate on this point?

design/authenticators_api.md
sgnn7: @orenbm I like this approach 👍

design/authenticators_health_api.md
sgnn7: Yeah here, it's up to you what the suffix would be if we use the `/authenticators/<authenticator_id>` prefix. My main concern was that the `/authenticators` prefix does not work with `/health` suffix all by itself

orenbm: ok i'll consult with the architects and decide together. thanks for the input!

PR #1019, opened 2019-05-15
merge Oidc not poc -> master

orenbm:
#### What does this PR do?
Merges the OIDC work that was done between the POC and the Conjur release. We side branched so the release won't get flakey OIDC code


app/controllers/authenticate_controller.rb
```diff
@@ -120,28 +120,29 @@ def handle_authentication_error(err)
...
+    when Errors::Authentication::Security::UserNotAuthorizedInConjur,
+      Errors::Authentication::Security::NotWhitelisted,
+      Errors::Authentication::Security::ServiceNotDefined,
+      Errors::Authentication::Security::UserNotDefinedInConjur,
+      Errors::Authentication::Security::AccountNotDefined,
+      Errors::Authentication::AuthnOidc::IdTokenFieldNotFoundOrEmpty,
+      Errors::Authentication::AuthnOidc::IdTokenVerifyFailed,
+      Errors::Authentication::AuthnOidc::IdTokenInvalidFormat,
+      Errors::Conjur::RequiredSecretMissing,
+      Errors::Conjur::RequiredResourceMissing
```
jonahx: My question about this whole block above is:

Can each of the 5 controller actions that call this raise _all_ of these errors? If the answer is "yes", then the following is less important, but still worth saying...

If the answer is "no", we shouldn't be combining all of them and putting them in a helper method, because we lose important information that way.  This was probably driven by a good desire to avoid repetition, but you want to see _exactly_ which errors may be raised by the CommandClass you delegate to in each controller action.  That's key information. 
 And in theory it's possible you handle them differently in different actions, etc.

This is sort of a window into a deeper problem, which you may not want to fix now, but should be aware of.  I tried to answer the question I asked in the first sentence above, and  I went here:

https://github.com/cyberark/conjur/blob/f6f87a5fe33836c7af4ee8ab6c265ca4ba6b769c/app/domain/authentication/authn_oidc/get_conjur_oidc_token/conjur_oidc_token.rb#L29

but you'll see that there is not a way to see what set of errors this CommandClass is expecting.  To answer that question, you have to delve into _every one_ of its deps, and see what errors they raise.  Whenever understanding the code has you doing this -- looking at many classes to answer questions about _one parent class_, you know something has gone wrong.  

Instead, on line 29 above, we should list all the expected errors in the rescue clause, and then have a final generic rescue at the end that represents an "unexpected error."  That way,

1. Everything is explicit
2. Readers of the code can see all the assumptions in one file (remember, raised errors are part of a classes interface.  This is the same problem we've addressed previously by listing errors in comments.)
3. If some dep adds a new error, the parent class will break in the correct way, ie, it will report an "unexpected error".  You can even write UTs for this.

This is sort of a can of worms, and there's more to be said, but I've already rambled enough and hopefully that will make sense...

app/domain/authentication/audit_event.rb
```diff
@@ -1,30 +1,39 @@
...
     end
 
     private
 
-    def role(username, account)
+    def role
       return nil if username.nil?
 
-      @get_role_by_login.(username: username, account: account)
+      @role_cls.by_login(username, account: account)
```
jonahx: Style nitpick: Because account is only using once, I think I'd delete the `account` private method and just do:
```
@role_cls.by_login(username, account: @input.account)
```
I think the `username` method is more defensible.   Again, this is a total nitpick.

app/controllers/authenticate_controller.rb
orenbm: i don't think it's really important to know which class can throw which error in `authenticate_controller` but rather to know what happens in each case and what error code do we send to the user. i think that's it's pretty clear in the current implementation and a reader can better understand what's going on here, rather than going over the code and seeing different rescue blocks doing (pretty much) the same thing.

regarding the point of having rescue blocks in each class that raises the known errors and has a generic rescue error - i don't really like it. when a sees an error in the logs he will want to know the root cause and where it was thrown. adding data of an error that swallows the real error can mislead him. also, we already have the section at the top, stating the possible errors raised in this class. what we can do is add to that sections the errors this class can raise by rescuing them from deps. maybe even have 2 sections - one for errors raised in this class and one for errors raised in deps and re-reraised here. wdyt?

jonahx: > when a sees an error in the logs he will want to know the root cause and where it was thrown.

this is true

> adding data of an error that swallows the real error can mislead him. 

also true.  but this won't happen if you simply re-raise.  Eg, if you run this:

```ruby
def test1
  raise "Test 1"
end

def test2
  test1
rescue => e
  raise e
end

test2
```
you'll get:
```
"test_err.rb" 11L, 83C written                                                                                                                        
:!ruby /Users/jgoldstein/code/ruby/test_err.rb
/Users/jgoldstein/code/ruby/test_err.rb:2:in `test1': Test 1 (RuntimeError)
        from /Users/jgoldstein/code/ruby/test_err.rb:6:in `test2'
        from /Users/jgoldstein/code/ruby/test_err.rb:11:in `<main>'
```
> what we can do is add to that sections the errors this class can raise by rescuing them from deps. 

This solution is *ok*, but was missing from `conjur_oidc_token.rb`.  And, if you recall, we were both kinda "meh" about this solution, for good reason: it relies on programmers keeping things up to date.  But much worse than that: If that's our strategy, and you update a command class to throw a new error, you must update the comments for every other class that uses it, and then every class that uses any of those, and so on (I only realized this yesterday).  Whereas if the deps at least rescue and re-throw (without adding new data), then the new error will be correctly classified as unexpected (and you can include the stack trace of the original error).

orenbm: i think i didn't understand you at the first place, because this doesn't make sense to me. 
let's take `authenticate_id_token/authenticate.rb` for example. the current implementation for `call` is:

```
def call
          decode_and_verify_id_token
          validate_conjur_username
          add_username_to_input
          validate_security
          validate_origin
          audit_success
          new_token
        rescue => e
          audit_failure(e)
          raise e
        end
```

`DecodeAndVerifyIdToken` is a dep for this class and if we look there we can see that it may raise these errors: `IdTokenExpired, IdTokenVerifyFailed, IdTokenInvalidFormat`.

would you expect the new code to look like this:

```
def call
          decode_and_verify_id_token
          validate_conjur_username
          add_username_to_input
          validate_security
          validate_origin
          audit_success
          new_token
        rescue IdTokenExpired, IdTokenVerifyFailed, IdTokenInvalidFormat => e
          audit_failure(e)
          raise e
        rescue => e
          audit_failure(e)
          raise e
        end
```

i assume not because you mentioned earlier that you would raise a generic error but we removed that so i'm not sure i understand your suggestion.

jonahx: So what you have above in the snippet is very close to what I meant.  except that the final `rescue` would look something like:

```ruby
        rescue => e
          unexpected_err = UnexpectedError.new(e)
          audit_failure(unexpected_err)
          raise unexpected_err
```
where we have:
```ruby
class UnexepectedError < RuntimeError
  def initialize(orig_error)
    @orig_error = orig_error
  end

  def to_s
    <<~EOS
      UnexepectedError: #{@orig_error}.\n
      Original backtrace: #{@orig_error.backtrace.join("\n")}
    EOS
  end
end
```
So this is not a generic error, per so.  It's a very specific error which says, "Something happened that our code didn't believe was possible, and was not prepared for."  Which is meaningful and helpful when debugging.

Wdyt?

orenbm: ok i see. so in the future if i add an error to `DecodeAndVerifyIdToken` i need to add it to the rescue clause of every class that consumes `DecodeAndVerifyIdToken` as a dependency. this is a bit inconvenient, dont you think?

i don't strongly disagree with you here, just want to get your opinion.

jonahx: > ok i see. so in the future if i add an error to DecodeAndVerifyIdToken i need to add it to the rescue clause of every class that consumes DecodeAndVerifyIdToken as a dependency. this is a bit inconvenient, dont you think?

@orenbm ah, so this is this thing!

yes, it is inconvenient.  but remember: the errors are part of the public interface of a class!  Or more specifically, part of the "return value" of a method.

So you're question is like asking: "If change the type my method is returning, isn't it inconvenient that I'd have to udpate all its clients?"  And the answer is also yes, but you've broken the public interface... that's what happens when you break a public interface.

orenbm: really liked the analogy :)

how can we make this a hard constraint that will break if someone adds an error without adding it to rescue clauses of consuming classes?

jonahx: @orenbm 

> how can we make this a hard constraint that will break if someone adds an error without adding it to rescue clauses of consuming classes?

let's continue our analogy...

Q: How can I add a hard constraint so that when I change my return type from string to array, all client classes will break?

A: You can't.  This is a major downside of not having static types...

Q: That's a pretty damn big problem.  How does ruby deal with this?

A: Tests.   What static types handle automatically in other languages, the dev is expected to handle in ruby by writing tests.  

Q: Ok I could see having tests that break when the return type changes, but how do I create tests when a new error type is _added_, as in our case here?

This is the real question.  And the answer is, it depends what your UTs are testing, and when the new error arises.   If your tests are truly thorough and test all the relevant paths that you care about, and the the new error arises as part of one of those paths, it *will* break your client class.  (Note: This is why it's important for UTs to only rescue *expected* errors -- you want them to break when some new error comes up.)  If the new error is not part of the behavior your client relies on, your client won't break, and that's fine -- you're client is only relying on *part* of the public interface, and that part wasn't affected.

The more deadly -- and more likely -- case is that you didn't actually write tests for all the paths you care about, and your client won't break until someone hits one of those paths in production.  

But note: The approach we're discussing above is *still* preferable even in that worst case scenario, but at least it gets reported correctly as an `UnexpectedError` and figuring out what happened will be easier.  And it's also preferable because we retain the massive readability benefit we're after, which is that anyone can understand the class's contract merely by looking at it.  In this case, to understand the class fully, we need only look at:

1. Its dependencies, listed in the signature
2. Its inputs, listed in the signature
3. Its errors, listed in the rescue clause of the "main" method, ie `call`.

orenbm: ok, i see. i created an issue which we can address after the GA 
https://app.zenhub.com/workspaces/appliance-development-5c9cf04bfa1c204b63eddcc6/issues/cyberark/conjur/1022

Issue #1008, opened 2019-05-08
Implement Performance tests oidc-authn automation


nessiLahav:
requires #988
child of #974

PR #1021, opened 2019-05-20
Added tests to check the performance of authn oidc

jeniaSakirko:
#### What does this PR do?
Performance test for authn oidc
Connected to #1008
cucumber/authenticators/features/authn_oidc.feature 
Run the cucumber tests for the authenticators

### DIFF
--- a/cucumber/authenticators/features/authn_oidc.feature
+++ b/cucumber/authenticators/features/authn_oidc.feature
@@ -163,3 +163,13 @@ Feature: Users can authneticate with OIDC authenticator
     When I authenticate "1000" times in "10" threads via OIDC with id token
     Then The "max" response time should be less than "1" seconds
     And The "avg" response time should be less than "0.25" seconds
+
+  Scenario: Load test
+    Given I get authorization code for username "alice" and password "alice"
+    And I fetch an ID Token
+    And I save my place in the log file
+    When I authenticate "2000" times in "20" threads via OIDC with id token
+    Then The following appears "2000" times in the log after my savepoint:
+    """
+    Completed 200 OK
+    """
--- a/cucumber/authenticators/features/authn_oidc.feature
+++ b/cucumber/authenticators/features/authn_oidc.feature
@@ -156,3 +156,10 @@ Feature: Users can authneticate with OIDC authenticator
     """
     Errors::Authentication::AuthnOidc::AdminAuthenticationDenied
     """
+
+  Scenario: Performance test
+    Given I get authorization code for username "alice" and password "alice"
+    And I fetch an ID Token
+    When I authenticate "1000" times in "10" threads via OIDC with id token
+    Then The "max" response time should be less than "1" seconds
+    And The "avg" response time should be less than "0.25" seconds
--- a/cucumber/authenticators/features/step_definitions/authn_oidc_steps.rb
+++ b/cucumber/authenticators/features/step_definitions/authn_oidc_steps.rb
@@ -38,8 +38,8 @@ When(/^I authenticate via OIDC with empty id token$/) do
   authenticate_id_token_with_oidc(service_id: 'keycloak', account: 'cucumber', id_token: "")
 end
 
-When(/^I authenticate "([^"]*)" times in "([^"]*)" threads via OIDC with id token$/) do |num_of_requests, num_of_threads|
-  measure_oidc_performance(num_of_requests.to_i, num_of_threads.to_i, service_id: 'keycloak', account: 'cucumber')
+When(/^I authenticate "([^"]*)" times in "([^"]*)" threads via OIDC with id token$/) do |num_requests, num_threads|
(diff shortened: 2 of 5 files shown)


cucumber/authenticators/features/support/authn_oidc_helper.rb
```diff
@@ -104,6 +127,21 @@ def parse_oidc_id_token
...
+    }
+
+    results.each do |res|
+      real_time = res.real
+      @performance_result[:max] = real_time unless @performance_result[:max] > real_time
+      @performance_result[:avg] += real_time
+    end
+
+    @performance_result[:avg] /= results.length
+  end
```
jonahx: So I actually think this would clean up nicely, and have a more ruby-esque spirit, if we separated the two calculations, rather than trying to combine inside one loop.  The performance difference will be negligible, so we should focus on readability.  And if you think about it, in the above we are re-implementing the functionality provided by `Enumerable.max`.

In fact, the calculation of `max` and `avg` are less than 1-liners.  

And we can actually remove the above method and `ensure_performance_results` altogether.

Instead, at end of the method currently named `run_performance_with_oidc` (which I suggested renaming to `measure_oidc_performance`), rather than calling `calculate_performance_results(results)`, we'll simply save the results:
```ruby
@oidc_perf_results = results.map(&:real)
```

And then we'll update the previous step definition as follows:

```ruby
Then(/^The "{string}" response time should be less than "{float}" seconds$/) do |type_str, threshold|
  type = type_str.downcase.to_sym
  raise "Unexpected Type" unless [:max, :avg].include?(type)
  results = @oidc_perf_results
  actual_time = (type == :avg) ? results.sum.fdiv(results.size) : results.max
  expect(actual_time).to be < threshold
end
```
Allowing us to delete two methods.

cucumber/authenticators/features/step_definitions/authn_oidc_steps.rb
```diff
@@ -37,3 +37,11 @@
 When(/^I authenticate via OIDC with empty id token$/) do
   authenticate_id_token_with_oidc(service_id: 'keycloak', account: 'cucumber', id_token: "")
 end
+
+When(/^I authenticate "([^"]*)" times in "([^"]*)" threads via OIDC with id token$/) do |num_of_requests, num_of_threads|
+  run_performance_with_oidc(num_of_requests.to_i, num_of_threads.to_i, service_id: 'keycloak', account: 'cucumber')
```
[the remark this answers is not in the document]
orenbm: Love it!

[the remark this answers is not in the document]
orenbm: @jeniaSakirko can you please go over our helpers (oidc) and use this mechanism? i guess `{int}` & `{string}/{word}` will do.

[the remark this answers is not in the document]
orenbm: as discussed in slack, we will delay the cucumber upgrade until after the GA as it breaks our tests and the fix isn't trivial

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -156,3 +156,20 @@ Feature: Users can authneticate with OIDC authenticator
...
+    When I authenticate "1000" times in "10" threads via OIDC with id token
+    Then The "max" response time should be less than "1" seconds
+    And The "avg" response time should be less than "0.25" seconds
+
+  Scenario: Load test
+    Given I get authorization code for username "alice" and password "alice"
+    And I fetch an ID Token
+    And I save my place in the log file
+    When I authenticate "2000" times in "20" threads via OIDC with id token
+    Then The following appears in the log after my savepoint:
```
orenbm: please add number of occurences

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -156,3 +156,20 @@ Feature: Users can authneticate with OIDC authenticator
...
+    """
+    Completed 200 OK
```
orenbm: i'm not sure this is enough. maybe add also the line of sending the request. should be something like `POST .../authn-oidc/authenticate`

cucumber/authenticators/features/step_definitions/authn_oidc_steps.rb
```diff
@@ -37,3 +37,11 @@
 When(/^I authenticate via OIDC with empty id token$/) do
   authenticate_id_token_with_oidc(service_id: 'keycloak', account: 'cucumber', id_token: "")
 end
+
+When(/^I authenticate "([^"]*)" times in "([^"]*)" threads via OIDC with id token$/) do |num_of_requests, num_of_threads|
```
orenbm: pls update to `num_request` & `num_threads`

cucumber/authenticators/features/authn_oidc.feature
jeniaSakirko: Checking `POST .../authn-oidc/authenticate` isn't enough to know that the request was successful

Issue #981, opened 2019-04-29
Enhance OIDC logs


orenbm:
For better supportablity, we want good loggging. we should follow the guide here: https://ca-il-confluence.il.cyber-ark.com/display/VSSHD/Logging+Best+Practices

Add thread id to each message printed in the log

childof #974

orenbm:
this is good too: https://watchitlater.com/blog/2009/12/logging-guidelines/

InbalZilberman:
Hi, In this ticket please think of cases like the following:
session ID is a must
we cannot log the ID Token 
details regarding the OIDC provider failure in call are wanted

PR #1027, opened 2019-05-26
Logging extra info

orenbm:
#### What does this PR do?
Logs extra info in each log line
 
Each log line consists of:
- Severity
- Timestamp
- Process ID
- Thread ID
- Origin IP
- Request ID: a unique identifier for this request that gets used throughout the call stack 

An example line looks like this:
` INFO [2019/05/26 11:36:32 +0000] [pid:1546] [tid:9831860] [origin:127.0.0.1] [req_id:4565c64d-47d9-4426-a569-c111ca5c59ab] Completed 200 OK in 21ms (Views: 8.5ms)`

Connected to #981


shaharglazner:
@orenbm great work :)

few thoughts - 

**Can we eliminate the "[]"** 
Instead of 
`INFO [2019/05/26 11:36:32 +0000] [pid:1546] [tid:9831860] [origin:127.0.0.1] [req_id:4565c64d-47d9-4426-a569-c111ca5c59ab] Completed 200 OK in 21ms (Views: 8.5ms)`
To use
`INFO 2019/05/26 11:36:32 +0000 pid:1546 tid:9831860 origin:127.0.0.1 req_id:4565c64d-47d9-4426-a569-c111ca5c59ab Completed 200 OK in 21ms (Views: 8.5ms)`

The reason is that log analyzing products such as splunk knows to handle this format out-of-the-box. Another minor change could be to replace ":" with "="

Second - can we aggregate multiple request id's based on session id?

orenbm:
> **Can we eliminate the "[]"**
This is a built-in separator in rails. I tried to change the separator without success.

> Second - can we aggregate multiple request id's based on session id?
session-id is a pickle in rails. we used the request-id instead. what's do you mean by aggregate here? would you want them to be in a bulk even if they aren't really one after the other?

InbalZilberman:
@orenbm can you please explain what origin ip means?
why do you think it is relevant for each log? is it going to be changed between calls? when?

orenbm:
origin ip is the host from which the request was sent. it could help us to identify which host requested something when troubleshooting.

PR #1033, opened 2019-05-28
Load test with cache

nessiLahav:
#### What does this PR do?
Adjust load test to validate also cache functionality according to test plan

### DIFF
--- a/cucumber/authenticators/features/authn_oidc.feature
+++ b/cucumber/authenticators/features/authn_oidc.feature
@@ -164,12 +164,21 @@ Feature: Users can authneticate with OIDC authenticator
     Then The "max" response time should be less than "1" seconds
     And The "avg" response time should be less than "0.25" seconds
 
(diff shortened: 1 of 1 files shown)


cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -164,12 +164,21 @@ Feature: Users can authneticate with OIDC authenticator
...
+    # Load while the cache contains OIDC provider certificate
     When I authenticate "2000" times in "20" threads via OIDC with id token
     Then The following appears "2000" times in the log after my savepoint:
     """
     Completed 200 OK
     """
+    # Validate cache functionality
+    And The following appears "0" times in the log after my savepoint:
+    """
+    CONJ00016D Rate limited cache updated successfully
```
jeniaSakirko: We should also check for other errors in the logs

PR #1037, opened 2019-05-29
fixed tid & pid

orenbm:
#### What does this PR do?
fixes the thread id & process id in the logs


shaharglazner:
Maybe it worth to edit the commit message and explain what is being fixed

PR #1073, opened 2019-06-24
Cache roles only for single authentication request

apotterri:
#### What does this PR do?
Currently, the class MemoizedRole, used by authenticators, keeps a cache
of the Roles seen in the class itself. This cache is never cleared or
updated, so its entries become stale if the Roles in the database
change. For example, if a cache entry is added for a non-existent Role,
but the Role is subsequently added (by loading a policy), the cache
entry will still indicate that the Role does not exist.

With these changes, the user of MemoizedRole must create an instance of
the class, and use it to cache Roles.

AQR is currently experiencing problems caused by this bug.

https://github.com/conjurinc/appliance/issues/853
yes
no
no
no


app/domain/authentication/authn/authenticator.rb
```diff
@@ -31,9 +31,13 @@ def valid?(input)
...
 
       def credentials(input)
-        role_id = role_cls.roleid_from_username(input.account, input.username)
+        role_id = role_cache.roleid_from_username(input.account, input.username)
         credentials_cls[role_id]
       end
+
+      def role_cache
+        @role_class ||= role_cls.new
+      end
```
jonahx: Per our Slack convo, we can just inject `Role` here directly and do away with the cached version.  It turns out we can also fix my old `Dry::Struct` misuse here with very little effort.  Here's the updated class in its entirety.

Note I also I change the local variable names from `credentials` to `creds` so as to avoid any possible confusion with the `credentials` method.  

Clients of this class can still call it exactly as it was being called before.

Also, this may require minor modifications to a unit test or two somewhere.

```ruby
# frozen_string_literal: true

# default conjur authenticator
#
module Authentication
  module Authn

    class Authenticator

      def self.requires_env_arg?
        false
      end

      def initialize(role_cls: Role, credentials_cls: ::Credentials)
        @role_cls = role_cls
        @credentials_cls = credentials_cls
      end

      # Authenticates a Conjur using their username and password
      def login(input)
        creds = credentials(input)
        success = creds&.authenticate(input.password)

        success ? creds.api_key : nil
      end

      # Authenticates a Conjur role using its id and API key
      def valid?(input)
        creds = credentials(input)
        creds&.valid_api_key?(input.password)
      end

      def credentials(input)
        role_id = @role_cls.roleid_from_username(input.account, input.username)
        credentials_cls[role_id]
      end
    end
  end
end
```

app/domain/authentication/authn_ldap/authenticator.rb
```diff
@@ -64,6 +64,10 @@ def ldap_configuration(input)
       def blacklisted_ldap_user?(login)
         login == 'admin'
       end
+
+      def role_cache
+        @role_cache ||= @role_cls.new
+      end
```
jonahx: We can then eliminate this entirely.

app/domain/authentication/authn_ldap/authenticator.rb
```diff
@@ -38,7 +38,7 @@ def login(input)
         return nil unless bind_results
 
         # Return Conjur API key
-        role_id = @role_cls.roleid_from_username(input.account, input.username)
+        role_id = role_cache.roleid_from_username(input.account, input.username)
```
jonahx: In the initializer, change:

```ruby
role_cls: ::Authentication::MemoizedRole,
```
to:
```ruby
role_cls: Role,
```
and then this line will become:
```ruby
role_id = role_cls.roleid_from_username(input.account, input.username)
```

app/domain/authentication/validate_security.rb
```diff
@@ -68,15 +68,15 @@ def validate_user_has_access
       end
 
       def user_role_id
-        @user_role_id ||= @role_class.roleid_from_username(@account, @user_id)
+        role_cache.roleid_from_username(@account, @user_id)
```
jonahx: In this class, too, we can do away with `MemoizedRole` and just use standard `||=` memoization.  Again, in the initializer, change it to `Role`, and then here change to:

```ruby
@user_role_id ||= @role_class.roleid_from_username(@account, @user_id)
```

app/domain/authentication/validate_security.rb
```diff
@@ -68,15 +68,15 @@ def validate_user_has_access
...
       end
 
       def user_role
-        @role_class[user_role_id]
+        role_cache[user_role_id]
```
jonahx: same here:
```ruby
@user_role ||= @role_class[user_role_id]
```

app/domain/authentication/validate_security.rb
```diff
@@ -68,15 +68,15 @@ def validate_user_has_access
...
       end
 
       def account_admin_role
-        @role_class["#{@account}:user:admin"]
+        role_cache["#{@account}:user:admin"]
```
jonahx: And here:
```ruby
@account_admin_role ||= @role_class["#{@account}:user:admin"]
```

app/domain/authentication/validate_security.rb
```diff
@@ -93,6 +93,10 @@ def whitelisted_webservices
           @enabled_authenticators || Authentication::Common.default_authenticator_name
         )
       end
+
+      def role_cache
+        @role_cache ||= @role_class.new
+      end
```
jonahx: And now we can delete this.

spec/app/domain/authentication/validate_security_spec.rb
```diff
@@ -251,4 +256,57 @@ def resource_class(returned_resource)
...
+    let(:user_role_double) { double('user_role_double') }
+
+    before do 
+      allow(Role).to receive(:[]).with(admin_roleid).and_return(admin_role_double)
+      allow(Role).to receive(:roleid_from_username).and_return(user_roleid)
+      
+      allow(Resource).to receive(:[]).with(ws_resourceid).and_return(ws_double)
+
+      allow(user_role_double).to receive(:allowed_to?).with('authenticate', ws_double).and_return(true)
+    end
```
jonahx: We shouldn't have references to `Role` or `Resource` in these tests.  If this is necessary, it means that we have hardcoded deps that shouldn't be there.  I tried to look to find what was going on, but....

see my comment below....

spec/app/domain/authentication/validate_security_spec.rb
```diff
@@ -251,4 +256,57 @@ def resource_class(returned_resource)
...
+      # Simulate two validations. For the first, the role should not
+      # be found.
+      allow(Role).to receive(:[]).with(user_roleid).and_return(nil)
+      expect { subject.(
+                   webservice: webservice(ws_name),
+                   account: test_account,
+                   user_id: user_id,
+                   enabled_authenticators: two_authenticator_env
+                 )
+      }.to raise_error(Errors::Authentication::Security::UserNotDefinedInConjur)
```
jonahx: I'm confusing how this is working because I couldn't find the definition for this `subject`.  The definitions above it in the file seem to be for subjects defined in other scopes, which are always out of scope by the time we get here.  

Related: We should use named subjects for clarity.

apotterri: When you `Rspec.describe` a Class, as this test does, the subject of the test is an instance of the Class. It's not clear to me that it's necessary to repeat this.

spec/app/domain/authentication/validate_security_spec.rb
apotterri: I think to test this properly, we need to stub out database access only, and not anything above that. That's why there are currently references to these two classes.

To remove them, we'll need to ask the subject what its `role_class` and `webservice_resource_class` are and stub out methods on those, instead.

That seems a little weird to me, but only a little. WDYT?

apotterri: Actually, looks like this is already taken care of in the specs. I'll update them and remove these references.

jonahx: So `ValidateSecurity` (and all the CommandClasses) were designed specifically to make this kind of unit testing easy:

```ruby
    ValidateSecurity = CommandClass.new(
      dependencies: {
        role_class: ::Authentication::MemoizedRole,
        webservice_resource_class: ::Resource,
        logger: Rails.logger
      },
```
Simply inject the doubles you need for `role_class` and `webservice_resource_class`.  Currently the subject is being instantiated with default deps, ie, the ones we want in prod.  We'll just be using mocks here instead.

spec/app/domain/authentication/validate_security_spec.rb
jonahx: Ah right, forgot about that.  Ok, I hereby *amend* the above comment to:

We should treat this consistently.  Ie, in the test cases above this one we're explicitly creating the subject again with the mocks.  I believe if once you make the change below (discussed on slack) this will happen automatically.

app/domain/authentication/authn/authenticator.rb
```diff
@@ -5,34 +5,34 @@
...
 
       # Authenticates a Conjur using their username and password
       def login(input)
-        credentials = credentials(input)
-        success = credentials&.authenticate(input.password)
+        creds = credentials(input)
+        success = creds&.authenticate(input.password)
 
-        success ? credentials.api_key : nil
+        success ? creds.api_key : nil
```
jonahx: So to *really* get rid of nil checks we'd need to be using sum types, but that's a bigger change we won't be doing (and don't do anywhere).   That said, I think we can make reek happy (and make the code more explicit) with this change:
```ruby
      def login(input)
        creds = credentials(input)
        return nil unless creds

        success = creds.authenticate(input.password)
        success ? creds.api_key : nil
      end
```

spec/app/domain/authentication/validate_security_spec.rb
```diff
@@ -251,4 +244,52 @@ def resource_class(returned_resource)
...
+    # above, because it doesn't constrain the arguments to +:[]+.
+    let(:role_class) { double('role_class') }
+
+    before do
+      allow(role_class).to receive(:roleid_from_username).and_return(user_roleid)
+      allow(role_class).to receive(:[])
+        .with("#{test_account}:user:admin")
+        .and_return(user_role(is_authorized: true))
+      allow(user_role_double).to receive(:allowed_to?).and_return(true)
+    end
```
jonahx: In general I'm not a fan of `before` blocks: They can always be replaced by simply using the correct objects defined by `let`s and `def`s in your tests.  They basically exist to facilitate testing code that isn't unit testable (which is not the case here).  

So here, we can simply remove it and move the setup into the `let(:role_class)` block above.

spec/app/domain/authentication/validate_security_spec.rb
```diff
@@ -251,4 +244,52 @@ def resource_class(returned_resource)
...
+      allow(role_class).to receive(:[]).with(user_roleid).and_return(nil)
+      expect { subject }.to raise_error(Errors::Authentication::Security::UserNotDefinedInConjur)
+
+      # For the second, the role should be found, and validation
+      # should succeed.
+
+      # Note that, because the arguments are the same, this +allow+
+      # overwrites the previous one.
+      allow(role_class).to receive(:[]).with(user_roleid).and_return(user_role_double)
+      expect { subject }.not_to raise_error
```
jonahx: I believe this is not testing what we want it to test.  `subject` will recreate a new instance each time it's called, whereas we want a single subject with the behaivor of `role_class` changing.  

Instead you'll need to create an instance of `described_class` and assign it to a local variable at the top of this `it` block.  Then explicitly call it
```ruby
call.(
          webservice: webservice('service1'),
          account: test_account,
          user_id: user_id,
          enabled_authenticators: two_authenticator_env
        )
```
two times, chancing the `role_class` in between.  I *think* that last part will work, but you'll need to try it....

spec/app/domain/authentication/validate_security_spec.rb
apotterri: I'm ok with moving this to a `let`, I guess. I don't usually think of them as places to put much code.

FWIW, I found the `def`s in this spec really hard to work with. They've resulted in (what seems like) a lot of repeated code, which feels un-rspec-ish.

spec/app/domain/authentication/validate_security_spec.rb
apotterri: No, that's not right. The original bug was that *two different instances* of the validator were sharing the same memoization of a `Role` lookup (one that failed the first time it was attempted).

This tests that two different instances with access to the same role class implementation don't share state that way.

spec/app/domain/authentication/validate_security_spec.rb
jonahx: I agree these specs could be made clearer.  There shouldn't be a lot of repeated code.  That's a separate issue from the `before` though.  

I'd recommend going through this whole file, with an eye to cleaning up repetition and simplifying.  That seems out of scope at this point, though.

PR #1074, opened 2019-06-25
Authn status design

orenbm:
#### What does this PR do?
adds design pages to the authenticator status api


design/authenticators-status/authn_status_general.md
```diff
@@ -0,0 +1,113 @@
...
+
+- The requesting user has access to the authenticator status route webservice
+    - more info in the Security section of the [epic](https://github.com/cyberark/conjur/issues/1062)
+- The authenticator is implemented
+- The account exists
+    - although a wrong account in an `/authenticate` request will indicate a client error and
+    not a server error, we still need to verify it exists in order to verify that the webservice exists
+    under that account
+- The webservice exists
+- The authenticator is enabled in the ENV
```
rafis3: @orenbm the authenticator is configured on the master, but the enabled authenticators environment variable is set for each an every Conjur instance. Will we tell the user that after configuring the authenticator on the master, checking the status should be done on the relevant Conjur instances?

design/authenticators-status/authn_status_general.md
```diff
@@ -0,0 +1,113 @@
...
+- [Specific authenticator requirements](#implementing_specific_authenticator_status_check)
+ 
+***Note:*** The default authenticator (`authn`) is always healthy
```
rafis3: @orenbm what if it's not enabled in the authenticators environment variable?
For example, in K8s, I would not enable `authn` for the followers

design/authenticators-status/authn_status_oidc.md
```diff
@@ -0,0 +1,74 @@
...
+`call` method that will perform the following checks:
+
+- The following variables exist and have value
+    - `provider-uri`
+    - `id-token-user-property`
+- The group `conjur/authn-oidc/<service-id>/users` exists and has `read` & `authenticate` 
+permissions on the webservice
+- The OIDC Provider (defined in the `provider-uri` variable) is responsive
+
+## Things we will not check
```
rafis3: @orenbm also the authenticators environment variable is not mentioned here, not in `things we will not check` and not in what we do plan to implement. Can you share why this was left out?

design/authenticators-status/authn_status_general.md
orenbm: some of the configuration is done on the master (i.e loading the policy with a conjur cli which runs against master) but some of the configuration is run against other conjur instances. For example, enabling the authenticator in the ENV is done in any instance that users will authenticate against (and it's documented).

This is a good point for documentation though, thanks!

design/authenticators-status/authn_status_general.md
```diff
@@ -0,0 +1,113 @@
...
+ 
+ def authenticator
+  # The `@authenticators` map includes all authenticator classes that are implemented in 
+  # Conjur (`Authentication::AuthnOidc::Authenticator`, `Authentication::AuthnLdap::Authenticator`, etc.). 
+  
+  @authenticator = @authenticators[@authenticator_input.authenticator_name]
+ end
+ ```
+ 
+ # Test Plan
```
InbalZilberman: @orenbm  I believe we need to add a test for a valid authenticator and a valid access token but the authenticator returns status false then the result should be 500?
what would be the text of other fields:  "error": "<configuration_error>",
    "message": "<conjur_error_message>",

InbalZilberman: @orenbm i believe that the Then part of all the other test case should also include  "error": "<configuration_error>",
    "message": "<conjur_error_message>",

sigalsax: absolutely! will add this to the doc

design/authenticators-status/authn_status_general.md
orenbm: @rafis3 this is something that i noticed sometime earlier. even if you don't mention `authn` in the CONJUR_AUTHENTICATORS env var it is still enabled. we should double-check this because it contradicts the documentation

orenbm: so if we want to have the ability to disable `authn` then we have a bug

design/authenticators-status/authn_status_oidc.md
sigalsax: It is mentioned in the general status check that the authenticator is enabled in the ENV. This check is shared across all authenticators. (https://github.com/cyberark/conjur/blob/authn-status-design/design/authenticators-status/authn_status_general.md#implementation-details)

design/authenticators-status/authn_status_general.md
rafis3: Yes, I think I noticed this behavior as well. In that case, yes it is a bug.

design/authenticators-status/authn_status_oidc.md
```diff
@@ -0,0 +1,74 @@
...
+ end
+ ``` 
+
+The new CommandClass `Authentication::AuthnOidc::Status` will consist of a
+`call` method that will perform the following checks:
+
+- The following variables exist and have value
+    - `provider-uri`
+    - `id-token-user-property`
+- The group `conjur/authn-oidc/<service-id>/users` exists and has `read` & `authenticate`
```
InbalZilberman: @orenbm we need read? not only authenticate? blackout here

InbalZilberman: I am actually not sure that this is a valid check...
Today we provide freedom to our users to provide any name they want. they can create a layer called conjur/authn-oidc/<service-id>/myfavoritehosts` with `read` & `authenticate`permissions and that is fine.
is there a prossibility to test it the other way around - meaning that other then the oidc authn owner there are  users or hosts that have `read` & `authenticate`permissions?
if not maybe we should drop this check

design/authenticators-status/authn_status_oidc.md
```diff
@@ -0,0 +1,74 @@
...
+        error
+    - Cons
+        - We lose generalisation of the HTTP request between the OIDC authenticator and the rest (as we changed the method to POST and added an id token as an input).
+        - We need an id token as an input. Who's id token? this can add an extra layer of complexity 
+    - Decision: We will not validate the `id-token-user-property` value as the cons are 
+    stronger than the pros. As we will still log the error, the operator will be able to 
+    see the configuration failure and may proceed accordingly. This is relevant for all the 
+    checks but this check's price is too high.
+
+# Test Plan
```
InbalZilberman: @sigalsax @orenbm I believe that the Then part of all the other test case should also include "error": "<configuration_error>",
"message": "<conjur_error_message>"

design/authenticators-status/authn_status_general.md
orenbm: > @orenbm I believe we need to add a test for a valid authenticator and a valid access token but the authenticator returns status false then the result should be 500?
> what would be the text of other fields: "error": "<configuration_error>",
> "message": "<conjur_error_message>",

The message will be the message of the relevant error. if the status check failed on `WebserviceNotFound` then the message will be `Webservice '{0-webservice-name}' wasn't found` (which is its built-in message)

design/authenticators-status/authn_status_general.md
```diff
@@ -0,0 +1,113 @@
...
+
+We will create a new route in `routes.rb` for `/authenticators/:authenticator(/:service_id)/:account/status`.
+This route will lead to `AuthenticateController` which will consist of a new `status` method.
+In this method we will call a new CommandClass `Authentication::Status`. This class's
+`call` method will perform the general checks that apply to all authenticators
+and those specific to each authenticator:
+
+- The requesting user has access to the authenticator status route webservice
+    - more info in the Security section of the [epic](https://github.com/cyberark/conjur/issues/1062)
+- The authenticator is implemented
```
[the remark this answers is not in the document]
orenbm: no. an authenticator is implemented if it has a class that:
1. Its name is `Authenticator`
2. Its parent name is `AuthnSomething`
3. Its main module is `Authentication`
4. It has a `valid?` method

it doesn't have anything to do with the account.

design/authenticators-status/authn_status_general.md
```diff
@@ -0,0 +1,113 @@
...
+- The authenticator is enabled in the ENV
+- [Specific authenticator requirements](#implementing_specific_authenticator_status_check)
+ 
+***Note:*** The default authenticator (`authn`) is always healthy
+
+We will also need to audit access success & failures to this endpoint.
+ 
+## Implementing specific authenticator status check
+ 
+Some authenticators need extra validation. As mentioned in the Epic, if
```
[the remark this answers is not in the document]
orenbm: done

design/authenticators-status/authn_status_oidc.md
sigalsax: > @orenbm we need read? not only authenticate? blackout here
We do need the privileges of `authenticate` AND `read` because with `read`, the user that is a member of the group will be able to view the permitted webservices (in the UI for example).

orenbm: regarding the name - the group name above is defined in the oidc policy. it has the policy id and the name `users`.

as you can see in the docs (https://docs-staging.conjur.org/staging_a/en/Content/OIDC/OIDC.htm?tocpath=Integrations%7CAuthenticators%7C_____1) we later tell the customers to add groups/users to that group. this is the best practice to grant permissions to `webservice-users` & `secrets-users`.

design/authenticators-status/authn_status_oidc.md
sigalsax: added to the doc

design/authenticators-status/authn_status_oidc.md
orenbm: after chatting in live - i understand what you mean @InbalZilberman . i removed this check and explained in the docs why we don't check it

design/authenticators-status/authn_status_oidc.md
```diff
@@ -62,7 +60,12 @@ permissions on the webservice
     stronger than the pros. As we will still log the error, the operator will be able to 
     see the configuration failure and may proceed accordingly. This is relevant for all the 
     checks but this check's price is too high.
-
+- The group `conjur/authn-oidc/<service-id>/users` exists and has `read` & `authenticate` 
+permissions on the webservice
+    - Although it is a best practice to create a group with this name and add members who we want to have permissions to that group
```
sigalsax: We want to be more detailed as to what 'group' we are talking about. Our goal is to explain that no, this additional group is not required but yes, this is a better option in terms of organizing your space.

For organizational purposes, we recommend that a new group is created for each authenticator.  That way, you can easily give the necessary `read` and `authenticate` privileges to multiple groups that already exist in Conjur. Although this is best practice, we do not enforce this in the product. Therefore, we cannot consider an authenticator 'unhealthy' if that extra grouping does not exist (or has insufficient permissions on the webservice) and will not include this as one of the status checks.

InbalZilberman:
@orenbm @sigalsax  i cant find my comment regarding the users group. Have we checked if it is possible to see that there is someone out there in Conjur that is NOT the owner of the authenticator who has read and authenticate permissions on the authenticator?
That way we are not looking at naming conventions just making sure someone can use it and that Conjur operator did not forget this step?

design/authenticators-status/authn_status_general.md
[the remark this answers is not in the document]
shaharglazner: @orenbm So why the route is `/authenticators/:authenticator(/:service_id)/:account/status` and not just `/authenticators/:authenticator(/:service_id)/status`? (I've read https://conjurhq.slack.com/archives/C0A2NF134/p1561461773196100)

design/authenticators-status/authn_status_general.md
```diff
@@ -0,0 +1,153 @@
...
+ 
+ ```
+{
+   "authn-name": {
+     "status": "ok"
+   }
+}
+```
+
+ ### Failure
```
shaharglazner: Wouldn't it be better to return full health check results?
```
{
  "authn-name": {
      "permissions": 
             {
                 "status": "error",
                 "error": "<configuration_error>",
                 "code": "<conjur_error_code>",
                 "message": "<conjur_error_message>"
             }
      "implemented": 
           {
                 "status": "pass"
                 "error": "none"
                 ...
           }
      "account_exist": ...
      "webservice_exists": ...
      "webservice_enabled": ...
      "other": {}
  }
}
```

design/authenticators-status/authn_status_general.md
```diff
@@ -0,0 +1,153 @@
...
+  # The `@authenticators` map includes all authenticator classes that are implemented in 
+  # Conjur (`Authentication::AuthnOidc::Authenticator`, `Authentication::AuthnLdap::Authenticator`, etc.). 
+  
+  @authenticator = @authenticators[@authenticator_input.authenticator_name]
+ end
+```
+ 
+ # Test Plan
+ 
+ ## Response bodies
```
shaharglazner: It should be under the `Implementation details` and not under the `Test Plain`

sigalsax:
> @orenbm @sigalsax i cant find my comment regarding the users group. Have we checked if it is possible to see that there is someone out there in Conjur that is NOT the owner of the authenticator who has read and authenticate permissions on the authenticator?
> That way we are not looking at naming conventions just making sure someone can use it and that Conjur operator did not forget this step?

We need to understand what defines a “unhealthy” authenticator. If no users are permissioned to `read` and `authenticate` to the authenticator does that _actually_ mean that the authenticator isn’t healthy (healthy being accessible, live, and ready to be authenticated by *any user who is permitted to it*)? After discussion with Oren on this point, we have defined healthy as being accessible, live, and is ready to be authenticated by *any user who is permitted to it* but open to hear your thoughts on how you would define healthy.

orenbm:
furthermore, as of the latest release, in case a user is not permitted to the webservice he will get a 403 response and not the general 401 response. what we wanted to achieve in the status check is that if a user gets the general 401 response then it will actually be because of a client error (e.g something is wrong in the client’s request). A user will get a 401 response if the authenticator is not enabled in the env and that’s a problem because there is nothing wrong with the client’s request (so there is no client error which the 401 indicates).

in this case - where the requesting user is not permitted to the webservice - he will get a 403 Forbidden response. That fine! because he really isn’t permitted to the webservice so the error is in teh client request. This is correct to any case that a user is not permitted to a resource. For example, earlier this week i tried to get a secret to run summon. i got a 403 response. so i spoke with gary and he gave me permissions. that’s a valid and healthy procedure - I got a 403 Forbidden on a resource -> I asked an admin to give me permissions -> I got the permission and got the resource in the next request. if i would have gotten a 401 Unauthtorized response, but the reason for that was unrelated at all, then the behaviour is drunk and there is a problem, but that’s not the case. In our case the behaviour is super sober :)

orenbm:
@shaharglazner 
>@orenbm So why the route is `/authenticators/:authenticator(/:service_id)/:account/status` and not >just `/authenticators/:authenticator(/:service_id)/status`? (I've read >https://conjurhq.slack.com/archives/C0A2NF134/p1561461773196100)

we need the account in the route to get the webservice resource from the db

design/authenticators-status/authn_status_general.md
orenbm: i think this enlarges the effort and the value is not worth it. every check requires implementation in parsing the check to the  response while in the current design we need only to translate the raised error to the response json.

furthermore, we can't really run all the checks if one fails. for example, if the account doesn't exist then we won't find the webservice too.

design/authenticators-status/authn_status_general.md
orenbm: I put it here just to not have them in every row of the test plan. these bodies are actually part of the feature design.

sigalsax:
Some comments made during discussion:
1. should the service-id in the route be optional? 
2. is the 503 authenticator not implemented error message sensitive? Should this check be done after checking that the user/operator can navigate to the status check? @shaharglazner 
3. How to check if OIDC provider is responsive? W/ certs? Discovery?
4. We need to create checks against not only the OSS but against Enterprise (could have some troubles with Nginx)
5. We need to remove the checks that have to do with the group/user existence under the OIDC test plan 
6. Update OIDC test plans with an error body with the relevant error message that explains where the user failed

design/authenticators-status/authn_status_general.md
orenbm: @rafis3 thanks, i opened a bug - https://github.com/cyberark/conjur/issues/1076

design/authenticators-status/authn_status_general.md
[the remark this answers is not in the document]
orenbm: @shaharglazner 
we need the account in the route to get the webservice resource from the db

orenbm:
> Some comments made during discussion:
> 
> 1. should the service-id in the route be optional?
> 2. is the 503 authenticator not implemented error message sensitive? Should this check be done after checking that the user/operator can navigate to the status check? @shaharglazner
> 3. How to check if OIDC provider is responsive? W/ certs? Discovery?
> 4. We need to create checks against not only the OSS but against Enterprise (could have some troubles with Nginx)
> 5. We need to remove the checks that have to do with the group/user existence under the OIDC test plan
> 6. Update OIDC test plans with an error body with the relevant error message that explains where the user failed

PR #1078, opened 2019-07-02
Separated ValidateSecurity to CommandClasses per responsibility

orenbm:
#### What does this PR do?
Separated ValidateSecurity to CommandClasses per responsibility

ValidateSecurity has 2 responsibilities:
- Validate that the webservice is whitelisted in the env
- Validate that the user is defined & authtorized to the webservice

In this commit we separate ValidateSecurity into 2 CommandClass, each with its own
responsibility. This way we can use these CommandClasses in other places where
we want to do only one of the actions (i.e validate a webservice is whitelisted),
instead of copying code.

### DIFF
--- a/app/domain/authentication/validate_webservice_access.rb
+++ b/app/domain/authentication/validate_webservice_access.rb
@@ -1,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'authentication/webservices'
+require 'logs'
+
+module Authentication
+
+  module Security
+
+    Log = LogMessages::Authentication::Security
+    Err = Errors::Authentication::Security
+    # Possible Errors Raised:
+    # AccountNotDefined, ServiceNotDefined,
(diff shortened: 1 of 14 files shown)


app/domain/authentication/validate_webservice_access.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      def validate_user_is_defined
+        raise Err::UserNotDefinedInConjur, @user_id unless user_role
+      end
+
+      def validate_user_has_access
+        # Ensure user has access to the service
+        has_access = user_role.allowed_to?('authenticate', webservice_resource)
+        unless has_access
+          @logger.debug(Log::UserNotAuthorized
```
jonahx: The `to_s` shouldn't be needed there, but check to be sure.

orenbm: You are correct :)

PR #1087, opened 2019-07-08
Upgrade baseimages

dividedmind:
#### What does this PR do?

Upgrade the docker base images.

Ubuntu 16.04 is EOL. We've been using a third-party brightbox PPA to bring in modern rubies into it, but it has been removed, breaking our builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,8 @@ RUN apt-get update -y && \
                        unattended-upgrades \
                        update-notifier-common \
                        # needed to build some gem native extensions:
-                       libz-dev
+                       libz-dev \
+    && rm -rf /var/lib/apt/lists/*
 
 RUN gem install -N -v 1.16.1 bundler
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,10 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 ENV DEBIAN_FRONTEND=noninteractive
 
 ENV PORT 80
 
 EXPOSE 80
-
-RUN apt-get update -y && \
-    apt-get install -y software-properties-common && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -20,7 +16,10 @@ RUN apt-get update -y && \
                        ruby2.5 ruby2.5-dev \
                        tzdata \
                        unattended-upgrades \
-                       update-notifier-common
+                       update-notifier-common \
+                       # needed to build some gem native extensions:
+                       libz-dev \
+    && rm -rf /var/lib/apt/lists/*
```
apotterri: Just out of curiosity, what are the implications (e.g. on future uses of `apt`) of removing the files in this directory? Is this something we should have been doing all along, or is it a new requirement with 18.04?

dividedmind: No, something we should have been doing all along. The implications are that `apt-get install` will just fail outright telling you to `apt-get update` instead of mysteriously trying and failing to download packages that no longer exist.

sgnn7:
Seems reasonable to me though I don't know if there's other stuff that needs verifying that this won't break stuff so I'll leave the final approval to @apotterri / @jvanderhoof

PS: Build failed though so I'm good to approve on my side after that.

dividedmind:
@apotterri You think we should address those failing tests before merging? They're in authenticator tests which are fragile anyway, all seem to say 'phantomjs has crashed' without any more details. I've always wondered why these tests should even need phantomjs -- this is a purely API project after all.

The way I see it, it might turn out to be difficult to figure out what's wrong with these and there's likely to be a better solution than just hacking them into working; disabling them seems preferable than leaving this repo in an utterly unbuildable state until we solve this, so my instinct is to do that and open another ticket to revisit these tests.

Issue #1079, opened 2019-07-03
Remove OIDC login flow code 


orenbm:
As this is not planned to be developed at the moment, having this code in the repo can be confusing for other devs. we can always get it from git history.

- Remove login flow code
- Add git version to confluence

PR #1095, opened 2019-07-22
Removed OIDC login flow until it will be relevant again

orenbm:
#### What does this PR do?
Removes the OIDC login flow. It is not planned to be developed at the moment, so having this code in the repo can be confusing for other devs. we can always get it from git history.

Connected to #1079


nessiLahav:
@orenbm we should remove this file as well [conjur_token](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_oidc/conjur_token.rb)

orenbm:
> @orenbm we should remove this file as well [conjur_token](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_oidc/conjur_token.rb)

Correct, thanks!

Issue #1098, opened 2019-07-25
Modify log level with specific env var instead of RAILS_ENV


orenbm:
We should have the ability to configure the log level with a separate environment variable other than RAILS_ENV. changing the latter to development has side effects other than changing the log level to debug and we should avoid changing that for this purpose.

Changes should be in:
- evoke
  - change `LOG_LEVEL` to `CONJUR_LOG_LEVEL` to be the variable in `conjur.conf`
  - change default level to `warn`?
- possum
  - set default value to `info/warn`
  - consume `ENV['CONJUR_LOG_LEVEL']` in all environment files and set default values in case they don't show
- appliance
  - fix the cucumber step that enables debug level

jtuttle:
@orenbm Can we close this?

orenbm:
sure, thanks!

PR #1099, opened 2019-07-25
Log level env

orenbm:
#### What does this PR do?
Set log level to the value defined under the `CONJUR_LOG_LEVEL` env var. 
If the var doesn't exist it defaults to its current value.

Until now, the only way to enable debug logs was to change the `RAILS_ENV` var. 
this had 2 caveats:
1. It had side effects other than changing the log level
2. It wrote the logs to a different location than in appliance RAILS_ENV

With this PR the user can change only the log level.

Connected to #1098 1098

### DIFF
--- a/config/environments/appliance.rb
+++ b/config/environments/appliance.rb
@@ -6,7 +6,7 @@ require 'syslog/logger'
 
 Rails.application.configure do
   config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new('conjur-possum'))
-  config.log_level = :info
+  config.log_level = ENV['CONJUR_LOG_LEVEL'] || :info
   config.middleware.use Rack::RememberUuid
   config.audit_socket = '/run/conjur/audit.socket'
   config.audit_database ||= 'postgres://:5433/audit'
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -29,7 +29,7 @@ Rails.application.configure do
   config.serve_static_files = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
   # Use log level "warn" in prod to avoid logging parameters.
-  config.log_level = :warn
+  config.log_level = ENV['CONJUR_LOG_LEVEL'] || :warn
   config.log_formatter = ConjurFormatter.new
 
   # Specifies the header that your server uses for sending files.
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,7 @@ COPY . .
 RUN ln -sf /opt/conjur-server/bin/conjurctl /usr/local/bin/
 
 ENV RAILS_ENV production
+ENV CONJUR_LOG_LEVEL info
 
 # The Rails initialization expects the database configuration
 # and data key to exist. We supply placeholder values so that
--- a/ci/docker-compose.yml
+++ b/ci/docker-compose.yml
@@ -17,6 +17,7 @@ services:
       CONJUR_ACCOUNT: cucumber
       CONJUR_DATA_KEY:
       RAILS_ENV:
+      CONJUR_LOG_LEVEL: debug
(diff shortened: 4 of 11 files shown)


config/environments/appliance.rb
```diff
@@ -6,7 +6,7 @@
 
 Rails.application.configure do
   config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new('conjur-possum'))
-  config.log_level = :info
+  config.log_level = ENV['CONJUR_LOG_LEVEL'] || :info
```
alexkalish: Did you test to see what happens if a undefined log level value is provided here?  Is a reasonable error returned?

config/environments/production.rb
```diff
@@ -29,7 +29,7 @@
   config.serve_static_files = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
   # Use log level "warn" in prod to avoid logging parameters.
-  config.log_level = :warn
+  config.log_level = ENV['CONJUR_LOG_LEVEL'] || :warn
```
alexkalish: Is the ENV used for OSS production?  If so, why would the level be different from the appliance?

orenbm: good call, missed that file. added to new PR

config/environments/appliance.rb
orenbm: nice one. I added a new PR that fixes this

orenbm:
> @orenbm Could you provide a list of the different support log levels for customers and how each should be used? It would be great if those could make it back to the Confluence page @sigalsax made to improve the documentation.

Sure. The log levels are listed here: https://guides.rubyonrails.org/debugging_rails_applications.html#log-levels
@sigalsax please add them to the doc

orenbm:
@alexkalish I added a new PR which addresses your comments. you can find it here: https://github.com/cyberark/conjur/pull/1100

thanks!

sigalsax:
> > @orenbm Could you provide a list of the different support log levels for customers and how each should be used? It would be great if those could make it back to the Confluence page @sigalsax made to improve the documentation.
> 
> Sure. The log levels are listed here: https://guides.rubyonrails.org/debugging_rails_applications.html#log-levels
> @sigalsax please add them to the doc

Added the log levels and their descriptions to the doc

alexkalish:
@sigalsax Thank you!

I do wonder if all of those log levels are actually used (by Rails or our code).  If not, we should provide some guidance to the customer regarding which levels they might actually find valuable beyond their general meaning.

orenbm:
At this point we log mostly in debug level but there are a few logs in `info` & `warn`. 

why would you guide users like that? this will constrain us in the future to not use log levels.

I think it's best to say that the default log level in info and that for debugging purposes they may move to debug. but i won't say anything about the other levels (regardless to wether it is likely to be written)

alexkalish:
I don't want to constrain R&D.  It's fine to leave all log levels operable.  But if certain log levels are essentially unused, I don't want to recommend that customers use them.  It will be confusing for them to essentially see no difference in the output.

If default is `info` and they should switch to `debug` for troubleshooting, we should document those use cases.  If they want less logging, maybe we should recommend which level would best suit that purpose, while still not missing critical issues.

sigalsax:
@alexkalish @orenbm I do believe the info about the logs is important but we could technically remove 'unknown', no? I don't really imagine that a user would set a log level to 'unknown'...

orenbm:
I wouldn't remove the support in these log levels but i would write in our docs that the default log level  is INFO and that some messages are written in debug mode

alexkalish:
> I wouldn't remove the support in these log levels but i would write in our docs that the default log level is INFO and that some messages are written in debug mode

Agreed, plus maybe documenting `warn` if they want to reduce logging.

orenbm:
like i wrote in the other pr - we need to have a conversation about how we log and what we log to build guidelines. after we have that we can decide on what we document

PR #1100, opened 2019-07-27
Log level assertion

orenbm:
#### What does this PR do?
- Assert environment variable `CONJUR_LOG_LEVEL` is a valid Rails log level (if it exists)
- Set default log level value of production env to info to align with DAP

Connected to #1098

### DIFF
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -3,7 +3,7 @@
 LOG_LEVELS = ['debug', 'info', 'warn', 'error', 'fatal', 'unknown']
 LOG_LEVEL_ERROR = "Environment variable 'CONJUR_LOG_LEVEL' must be a valid Rails log level: #{LOG_LEVELS.inspect}"
 
-def assert_conjur_log_level
+def assert_valid_conjur_log_level
   log_level_env = ENV['CONJUR_LOG_LEVEL']
   if log_level_env
     raise LOG_LEVEL_ERROR unless LOG_LEVELS.include? log_level_env
@@ -13,7 +13,7 @@ end
 # Load the Rails application.
 require File.expand_path('../application', __FILE__)
 
-assert_conjur_log_level
+assert_valid_conjur_log_level
 
 # Initialize the Rails application.
 Rails.application.initialize!
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -1,7 +1,19 @@
 # frozen_string_literal: true
 
+LOG_LEVELS = ['debug', 'info', 'warn', 'error', 'fatal', 'unknown']
+LOG_LEVEL_ERROR = "Environment variable 'CONJUR_LOG_LEVEL' must be a valid Rails log level: #{LOG_LEVELS.inspect}"
+
+def assert_conjur_log_level
+  log_level_env = ENV['CONJUR_LOG_LEVEL']
+  if log_level_env
+    raise LOG_LEVEL_ERROR unless LOG_LEVELS.include? log_level_env
+  end
+end
+
 # Load the Rails application.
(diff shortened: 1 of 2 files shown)


config/environment.rb
```diff
@@ -1,7 +1,19 @@
 # frozen_string_literal: true
 
+LOG_LEVELS = ['debug', 'info', 'warn', 'error', 'fatal', 'unknown']
+LOG_LEVEL_ERROR = "Environment variable 'CONJUR_LOG_LEVEL' must be a valid Rails log level: #{LOG_LEVELS.inspect}"
+
+def assert_conjur_log_level
```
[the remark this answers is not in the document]
orenbm: sure, done.

config/environment.rb
```diff
@@ -1,7 +1,19 @@
 # frozen_string_literal: true
 
+LOG_LEVELS = ['debug', 'info', 'warn', 'error', 'fatal', 'unknown']
```
alexkalish: Can you not programmatically get a list of valid log levels from Rails, so as to avoid making this list yourself?

config/environment.rb
```diff
@@ -1,7 +1,19 @@
...
+LOG_LEVEL_ERROR = "Environment variable 'CONJUR_LOG_LEVEL' must be a valid Rails log level: #{LOG_LEVELS.inspect}"
+
+def assert_valid_conjur_log_level
+  log_level_env = ENV['CONJUR_LOG_LEVEL']
+  if log_level_env
+    raise LOG_LEVEL_ERROR unless LOG_LEVELS.include? log_level_env
```
alexkalish: How would the user correct an invalid log level?  Would they have to restart Rails?  Will this exception kill the Rails process (I forget, as it's been a while)?  We have to make sure that this is properly documented in @sigalsax's new doc, so that it gets in front of customers.

config/environment.rb
orenbm: couldn't get it. i don't think there's a risk here. for this to break we need that rails will add a severity _and_ that we will write logs in that severity. not likely to happen in any way.

config/environment.rb
orenbm: Yes, they will have to restart rails. this error will occur on rails initialization so it's fine that it kills the process. the user tried to run rails and failed with a descriptive message

config/environment.rb
alexkalish:
```ruby
2.3.8 :004 > Logger::Severity::constants
 => [:DEBUG, :INFO, :WARN, :ERROR, :FATAL, :UNKNOWN]
```
It's from plain Ruby and not Rails.

InbalZilberman:
@orenbm the log level info is not something that is common out there. usually is is error or warning.
@alexkalish, @garkler have you heard any complaint about this?

orenbm:
In the Synchronizer our default log level is INFO. From looking in the web, i see that the same is in Spring logging (https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-logging.html#boot-features-logging-console-output). 

However, in python the default is WARN (https://docs.python.org/2/howto/logging.html).

At this point we log in INFO level only 2 messages so there isn't much that will be added. I think this subject should be discussed thoroughly and get to a decision that will be a guideline for all our projects and devs.

What do we log? in which log level? This decision is independent from this PR, once we get to a decision we can change the default level of `production` (for oss) and `appliance` (for DAP)

alexkalish:
I agree with @orenbm that `info` is pretty standard.  This was our default at my last company.  I also agree that it depends how we use those levels.  For some apps, `info` could be crazy chatty.

PR #1116, opened 2019-08-05
User conjur log level in follower env

jtuttle:
#### What does this PR do?

Use new `CONJUR_LOG_LEVEL` variable in the follower environment config. Also add gitleaks config to whitelist earlier commits.

### DIFF
--- a/config/environments/follower.rb
+++ b/config/environments/follower.rb
@@ -7,7 +7,7 @@ load File.expand_path '../production.rb', __FILE__
 require 'rack/remember_uuid'
 
 Rails.application.configure do
-  config.log_level = :info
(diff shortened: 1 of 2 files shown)


config/environments/follower.rb
```diff
@@ -7,7 +7,7 @@
 require 'rack/remember_uuid'
 
 Rails.application.configure do
-  config.log_level = :info
+  config.log_level = ENV['CONJUR_LOG_LEVEL'] || :info
```
orenbm: why do we need this line if it's configured already in `production.rb`? don't you specify here only fields you're overriding or adding?

jtuttle:
This was taken care of in another PR.

PR #1149, opened 2019-08-26
Make log level configurable in the `follower` environment

micahlee:
Each of the other Conjur environments support modifying the log level using
the `CONJUR_LOG_LEVEL` environment variable. We add it here as well to
support easier debugging of Conjur follower containers.


InbalZilberman:
thanks Micha! nice and simple!

PR #1167, opened 2019-09-05
WIP - add audit to db method  [DO NOT MERGE]

uCatu:
This description is not yet in the formal format of PR

### DIFF
--- a/app/models/audit/event.rb
+++ b/app/models/audit/event.rb
@@ -10,8 +10,22 @@ module Audit
     progname 'conjur'
 
     def log_to logger
+      log_to_db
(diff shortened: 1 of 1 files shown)


app/models/audit/event.rb
```diff
@@ -10,9 +10,23 @@ class Event < Util::Struct
     progname 'conjur'
 
     def log_to logger
+      log_to_db
       logger.log logger_severity, self, progname
     end
 
+    def log_to_db
```
jvanderhoof: What about passing these arguments into the `log_to_db` method?  It would make unit testing this a snap.

PR #1194, opened 2019-09-26
Oss add audit deployment

aloncarmel111:
What does this PR do?
Update changelog, readme and audit.md files with the current audit design

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
```
uCatu: Version bump is required to 1.5.0
@jvanderhoof WDYT

sashaCher: We haven't added functionality from end user point of view. No APIs have been changed/added so I think that it should be 1.4.3 rather 1.5.0.

aloncarmel111: As discussed with @garkler, we need to add our changelog to the unreleased version in first phase.

uCatu:
Reverted Audit US.

PR #1196, opened 2019-10-01
Skip failing OIDC tests

jvanderhoof:
We've seen intermittent issues with all three of these OIDC authenticator features.  I've added a skip to all three until we can address these in a meaningful way.


cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -173,12 +174,14 @@ Feature: Users can authneticate with OIDC authenticator
     Then it is bad gateway
 
   Scenario: Performance test
+    Given I skip
```
Tovli: We've changed this test to not include the "max" part. I suggest to not skip it so we can see if the issue is resolved

PR #1209, opened 2019-10-17
Authenticate specific jenkins jobs during run time (authn-jenkins)

AndrewCopeland:
#### What does this PR do?
It allows jenkins jobs to authenticate to conjur without the need of a stored api key. Also it allows the ability to specifically authenticate a job rather than an entire host.

Open jenkins, import the jenkins plugin that supports Jenkins JIT authentication.
Create a jenkins 'conjur authn' user.
Load the authn-jenkins policy and populate the variables 'jenkinsURL', 'jenkinsUsername' & 'jenkinsPassword'.
Create a host that has access to use the authn-jenkins policy and ends with the appropriate job name.
Create this job in jenkins and enter in the correct conjur metadata (URL, account, ssl cert, etc)
Create a pipeline that fetches a secret using the Conjur Credential Plugin.
Run the pipeline. 

No
It has unit tests, integration tests have not been created
Yes
No
Yes (integration/authentication)

### DIFF
--- a/app/domain/authentication/authn_jenkins/authenticator.rb
+++ b/app/domain/authentication/authn_jenkins/authenticator.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'json'
+
 module Authentication
   module AuthnJenkins
     class Authenticator
@@ -9,7 +10,7 @@ module Authentication
         @env = env
       end
 
-      def webservice()
+      def webservice
         @webservice ||= ::Authentication::Webservice.new(
           account:            @account,
           authenticator_name: @authenticator_name,
@@ -24,56 +25,53 @@ module Authentication
           http.read_timeout = 5
           request = Net::HTTP::Get.new(uri.request_uri)
           request.basic_auth(username, password)
-          response = http.request(request)
+          http.request(request)
         rescue Net::OpenTimeout => e
-          Rails.logger.error("Timeout to Jenkins host Exception #{e.to_s}")
-          Rails.logger.error("Jenkins host '#{url}' is most likely invalid. Validate this url can be accessed from the conjur server.")
+          Rails.logger.error("Timeout to Jenkins host Exception #{e}. Jenkins host '#{url}' is most likely invalid. Validate this url can be accessed from the conjur server.")
           raise Err::HostNotFound, url
         end
-
-        return response
       end
 
-      def get_build_info(jenkinsURL, jobPath, buildNumber, username, password)
-        jenkinsBuildEndpoint = "/job/#{jobPath}/#{buildNumber}/api/json"
-        jenkinsBuildURL = "#{jenkinsURL}#{jenkinsBuildEndpoint}"
-        response = http_get_request(jenkinsBuildURL, username, password)
-        return response
+      def get_build_info(jenkins_url, job_path, build_number, username, password)
+        jenkins_build_endpoint = "/job/#{job_path}/#{build_number}/api/json"
+        jenkins_build_url = "#{jenkins_url}#{jenkins_build_endpoint}"
+        http_get_request(jenkins_build_url, username, password)
       end
 
       def build_running?(response)
-        Rails.logger.debug("Jenkins Build Response: #{response.body}")
+        body = response.body
+        status_code = response.code
 
-        if response.code != '200'
-          raise Err::BuildInfoError.new(jenkinsBuildURL, "#{response.code} - #{response.body}")
+        Rails.logger.debug("Jenkins Build Response: #{body}")
+        if status_code != '200'
+          raise Err::BuildInfoError.new(jenkinsBuildURL, "#{status_code} - #{body}")
         end
 
-        json = JSON.parse(response.body)
-        return json['building']
+        JSON.parse(body)['building']
       end
 
       def jenkins_public_key(response)
-        publicKey = response['X-Instance-Identity']
+        public_key = response['X-Instance-Identity']
 
-        publicKey = "-----BEGIN PUBLIC KEY-----\n#{publicKey}\n-----END PUBLIC KEY-----"
-        Rails.logger.debug("Returned public key: #{publicKey}")
-        return OpenSSL::PKey::RSA.new(publicKey)
+        public_key = "-----BEGIN PUBLIC KEY-----\n#{public_key}\n-----END PUBLIC KEY-----"
+        Rails.logger.debug("Returned public key: #{public_key}")
+        OpenSSL::PKey::RSA.new(public_key)
       end
 
       def parse_metadata(username, password)
-        jsonBody = JSON.parse password
+        json_body = JSON.parse password
 
-        buildNumber = jsonBody["buildNumber"]
-        signature = Base64.decode64(jsonBody["signature"])
-        jobProperty_hostPrefix = jsonBody["jobProperty_hostPrefix"]
-        if jsonBody.key?("jobProperty_hostPrefix")
-          jobName = username.sub("host/#{jobProperty_hostPrefix}/", "")
+        build_number = json_body["buildNumber"]
+        signature = Base64.decode64(json_body["signature"])
+        job_property_host_prefix = json_body["jobProperty_hostPrefix"]
+        if json_body.key?("jobProperty_hostPrefix")
+          job_name = username.sub("host/#{job_property_host_prefix}/", "")
         else
-          jobName = username.sub("host/", "")
+          job_name = username.sub("host/", "")
         end
-        jobPath = jobName.split("/").join("/job/")
+        job_path = job_name.split("/").join("/job/")
 
-        return jobName, jobPath, buildNumber, signature
+        return job_name, job_path, build_number, signature
       end
 
       def valid?(input)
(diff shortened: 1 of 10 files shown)


jvanderhoof:
@AndrewCopeland: a great place to start on this PR is to bring syntax inline with the our Ruby Style Guide: https://github.com/cyberark/ruby.  A style guide sets forth a common set of conventions so code looks similar between authors.  CodeClimate flags these for us: https://codeclimate.com/github/cyberark/conjur/pull/1209 (please let me know if you can't see the above link).

It looks like you're using camel case (`fooBar`) for a lot of variable names. The Ruby convention is for snake case (`foo_bar`).

dev/files/authn-jenkins/myapp.yml
```diff
@@ -0,0 +1,35 @@
...
+  # Give the `secrets-users` group permission to retrieve variables
+  - !permit
+    role: !group secrets-users
+    privilege: [ read, execute ]
+    resource: *variables
+
+  # Create a layer to hold this application's hosts
+  - !layer
+
+  # Create a host using the namespace `aws` to identify this as an AWS resource.
```
jvanderhoof: Looks like this comment might be referencing the `Authn-IAM` sample policy.

spec/app/domain/authentication/authn_jenkins/authenticator_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  def valid_public_key
+    double('HTTPResponse', 
+            code: 200,
+            body: 'MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAhU/Sg1ubdTES9qVSj45US8SRKh3PCvXwezMzJ8Xp7Rhi0qpPQmSKireaJ+R7+yUkY9xsZ7eTxFw2gCJz5fQyWoryjwJaA9ZsDU6jFZIv+SYHvTf3LTZ+TWeYi6A/gxF6JMckawgviQ1MV9VuBan5D8B/P4GR7TbmqiZIvBWfjSayz3Yj+85/PraBJJNC8lTYK61XjDd981nGNddcJ1KG1ZUarsBKXEX45cNEQJ21ZpisELVpXySk4czOSA52bADsaLJDX5MKNjbQgG+jd9tWo+8w3J7rKPThLRroMGukqMT7l535YnwQ0IocbvrM5uX3okydwcEADNYr8QCW1BhxUwIDAQAB'
+    )
+  end
+
+  def valid_login
```
jvanderhoof: Is this method used anywhere?  If not, we should get rid of it.

app/domain/authentication/authn_jenkins/authenticator.rb
```diff
@@ -0,0 +1,117 @@
...
+        else
+          jobName = username.sub("host/", "")
+        end
+        jobPath = jobName.split("/").join("/job/")
+
+        return jobName, jobPath, buildNumber, signature
+      end
+
+      def valid?(input)
+        @account = input.account
```
jvanderhoof: Try to avoid setting class level variables like this. It causes a couple of challenges. First, it means the `webservice` method interacts with this method (`valid?` in a non-intuitive way.  It also makes it harder to test the `webservice` method because class level variables need to be set outside the method being tested.

A possibility for resolving this could be to instantiate the webservice object in the method:
```ruby
def webservice(account, authenticator_name, service_id)
  @webservice ||= ::Authentication::Webservice.new(
    account:            account,
    service_id:         service_id
  )
end
...
def valid?(input)
  variables = webservice(
    input.account,
    input.authenticator_name,
    input.service_id
  )

  # Get needed secrets to connect into the jenkins API
  jenkinsUsername = variables.variable("jenkinsUsername").secret.value
...
```

app/domain/authentication/authn_jenkins/authenticator.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      def jenkins_public_key(response)
+        public_key = response['X-Instance-Identity']
+
+        public_key = "-----BEGIN PUBLIC KEY-----\n#{public_key}\n-----END PUBLIC KEY-----"
+        Rails.logger.debug("Returned public key: #{public_key}")
+        OpenSSL::PKey::RSA.new(public_key)
+      end
+
+      def parse_metadata(username, password)
```
jvanderhoof: Although the key information comes in through the `password` attribute, what do you think about changing it to something like `message_body` to make it more clear that it's a JSON string?

app/domain/authentication/authn_jenkins/authenticator.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      def webservice
+        @webservice ||= ::Authentication::Webservice.new(
+          account:            @account,
+          authenticator_name: @authenticator_name,
+          service_id:         @service_id
+        )
+      end
+
+      def http_get_request(url, username, password)
```
jvanderhoof: Let's try to get all the functionality related Jenkins out into it's own class.  This accomplishes a couple of things:
- Provides better abstraction
- Simplifies authenticator tests as you can mock the requests to Jenkins

Something like:
```ruby
class JenkinsClient
  def initialize(url, username, password)
    @url = url
    @username = username
    @password = password
  end

  def public_key()
    response = request('')
    create_public_key(response['X-Instance-Identity'])
  end

  def build(job, build_number)
    request("/job/#{job}/#{build_number}/api/json")
  end

  private
  def create_public_key(identity)
    public_key = [
      "-----BEGIN PUBLIC KEY-----",
      identity,
      "-----END PUBLIC KEY-----"
    ].join("\n")
    Rails.logger.debug("Returned public key: #{public_key}")
    OpenSSL::PKey::RSA.new(public_key)
  end

  def request(path)
    # Uses the HTTP gem: https://github.com/httprb/http
    begin
      HTTP.basic_auth(user: @username, pass: @password).get("#{@url}/#{path}")
    rescue Net::OpenTimeout => e
      Rails.logger.error("Timeout to Jenkins host Exception #{e}. Jenkins host '#{url}' is most likely invalid. Validate this url can be accessed from the conjur server.")
      raise Err::HostNotFound, url
    end
  end
end
```

app/domain/authentication/authn_jenkins/authenticator.rb
```diff
@@ -0,0 +1,111 @@
...
+        http_get_request(jenkins_build_url, username, password)
+      end
+
+      def build_running?(response)
+        body = response.body
+        status_code = response.code
+
+        Rails.logger.debug("Jenkins Build Response: #{body}")
+        if status_code != '200'
+          raise Err::BuildInfoError.new(jenkinsBuildURL, "#{status_code} - #{body}")
```
jvanderhoof: The variable `jenkinsBuildURL` does not exist in the scope of this method.  You'd get an exception trying to throw this exception :stuck_out_tongue_closed_eyes

app/domain/authentication/authn_jenkins/authenticator.rb
```diff
@@ -0,0 +1,111 @@
...
+            return true
+          else
+            Rails.logger.error("AUTHENTICATION FAILED: Data tampered or private-public key mismatch.")
+            raise Err::InvalidSignature
+          end
+        else
+          Rails.logger.error("AUTHENTICATION FAILED: Job '#{job_name} ##{build_number}' is currently not running.")
+          raise Err::RunningJobNotFound "#{job_name} ##{build_number}"
+        end
+      end
```
jvanderhoof: With the above `JenkinsClient`, we could simplify the `valid?` method to something like:
```ruby
def jenkins_client(account, authenticator_name, service_id)
  @jenkins_client ||= begin
    variables = webservice(
      input.account,
      input.authenticator_name,
      input.service_id
    )
    JenkinClient.new(
      variables.variable("jenkinsURL").secret.value,
      variables.variable("jenkinsUsername").secret.value,
      variables.variable("jenkinsPassword").secret.value
    )
  end
end

def valid?(input)
  # Parse the body
  # e.g {"buildNumber": 5, "signature": "<base64 signature>", "jobProperty_hostPrefix": "myapp"}
  job_name, job_path, build_number, signature = parse_metadata(input.username, input.password)

  Rails.logger.debug("Job Name: #{job_name} | Build Number: #{build_number} | Signatature: #{signature}")

  # Validate job is running and signature was signed with the jenkins identities private key
  if build_running?(jenkins_client.build(job_path, build_number))
    public_key = jenkins_client.public_key
    message = "#{job_name}-#{build_number}"

    unless public_key.verify(OpenSSL::Digest::SHA256.new, signature, message)
      Rails.logger.error("AUTHENTICATION FAILED: Data tampered or private-public key mismatch.")
      raise Err::InvalidSignature
    end
  else
    Rails.logger.error("AUTHENTICATION FAILED: Job '#{job_name} ##{build_number}' is currently not running.")
    raise Err::RunningJobNotFound "#{job_name} ##{build_number}"
  end
  true
end
```
We can now mock the `jenkins_client` method, which allows us to easily create the three possible scenarios we can expect from the `valid?` method:
- success
- exception: job not running
- exception: signature mismatch

app/domain/authentication/authn_jenkins/authenticator.rb
jvanderhoof: `parse_metadata` isn't a super descriptive. Is there something more descriptive we can call it?  Maybe `extract_job_name_path_build_signature`?  I also wonder if there is a better way to handle the extraction that's happening inside this method.

jvanderhoof: Not sure if it's a net win, but it could be abstracted out into another class:
```ruby
class JenkinsLoad
  attr_reader :build_number, :signature, :job_name, :job_path

  def initialize(raw_body, username)
    body = JSON.parse(raw_body)
    @build_number = body['buildNumber']
    @signature = Base64.decode64(body['signature'])
    @job_name = if json_body.key?("jobProperty_hostPrefix")
      username.sub("host/#{job_property_host_prefix}/", "")
    else
      username.sub("host/", "")
    end
    @job_path = @job_name.split("/").join("/job/")
  end
end
```
Which would allow us to do something like:
```ruby
load = JenkinsLoad.new(password)
load.build_number
...
```

dev/files/authn-jenkins/myapp.yml
AndrewCopeland: Resolved in this commit: https://github.com/AndrewCopeland/conjur/commit/b69b6de44c479288cb5515f8493775bef157c1db

dev/files/authn-jenkins/policy.yml
```diff
@@ -0,0 +1,15 @@
+- !policy
+  id: conjur/authn-jenkins/prod
+  body:
+  - !webservice
+  - !group clients
+  # these will be used to connect to the jenkins server and see if a specific job is running
+  - !variable jenkinsURL
```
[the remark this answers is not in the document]
AndrewCopeland: I like this idea, I will try to implement this. In this example is 'ci/jenkins/url' an absolute path or a relative path to the webservice? It would be nice to have it absolute so we could onboard the jenkins account into cyberark and have cyberark automatically rotate the credential and then update via the synchronizer. Also do permissions need to be granted to the webservice to fetch specific secrets?

spec/app/domain/authentication/authn_jenkins/authenticator_spec.rb
AndrewCopeland: Nope they are not used and can be removed. Resolved in this commit: https://github.com/AndrewCopeland/conjur/commit/9badd0d64ef96774aa946dc1add0e6e6bd6dc0ec

dev/files/authn-jenkins/policy.yml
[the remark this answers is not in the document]
jvanderhoof: An absolute path feels like the best option. We could always force it with a leading `/`.  A webservice should be entitled as part of the policy, otherwise resources can gain access to credentials without explicitly being granted access.

app/domain/authentication/authn_jenkins/authenticator.rb
AndrewCopeland: That makes sense. Resolved in this commit: https://github.com/AndrewCopeland/conjur/commit/f3972bb8aa834dc3c0037ec010425a6b916680e7

app/domain/authentication/authn_jenkins/authenticator.rb
AndrewCopeland: Added jenkins client in this commit: https://github.com/AndrewCopeland/conjur/commit/d16c40c290d4f1ac139b101cd089a1635f20db29

app/domain/authentication/authn_jenkins/authenticator.rb
AndrewCopeland: Removed jenkinsBuildURL from the exception in this commit: https://github.com/AndrewCopeland/conjur/commit/d16c40c290d4f1ac139b101cd089a1635f20db29

app/domain/authentication/authn_jenkins/authenticator.rb
AndrewCopeland: Resolved in commit: https://github.com/AndrewCopeland/conjur/commit/d16c40c290d4f1ac139b101cd089a1635f20db29

dev/files/authn-jenkins/policy.yml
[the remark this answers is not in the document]
AndrewCopeland: So this behavior is possible but it requires jumping through some hoops.
```ruby
url_id = variables.annotation("url") 
username_id = variables.annotation("username")
password_id = variables.annotation("password")

url = variables.resource_class["#{input.account}:variable:#{url_id}"].secret.value
username = variables.resource_class["#{input.account}:variable:#{username_id}"].secret.value
password = variables.resource_class["#{input.account}:variable:#{password_id}"].secret.value
```

Lets discuss this further

app/domain/authentication/authn_jenkins/authenticator.rb
AndrewCopeland: Added a JenkinsLoad class. 
https://github.com/AndrewCopeland/conjur/commit/6a6526c2d914cc0d24e48569fcd9b61312c46577
And removed the parse_metadata:
https://github.com/AndrewCopeland/conjur/commit/91e297962396780b27521b9b9b0b8f73773449a8

jvanderhoof:
@AndrewCopeland, now that we have feature flags, I'm pulling this in. I'll implement the feature flag bits. I'm going to add you to the PR to make sure I don't futz anything up.

jtuttle:
This PR addresses #1210.

PR #1219, opened 2019-11-12
Move authenticator status to db

jtuttle:
Change the way we enable authenticators from the `CONJUR_AUTHENTICATORS` environment variable to making a request against the API and storing the enabled status in the database. This allows authenticators to be enabled without restarting Conjur and also allows enablement state to propagate to standby databases automatically in high availability setups.

Remaining TODOs:
- [x] Add test for authenticator not found
- [x] Add test for insufficient permissions to update authenticator
- [x] Add test for attempting to edit a webservice that isn't an authenticator (e.g. the `!webservice status` nested under ODIC authenticators).


app/controllers/authenticate_controller.rb
```diff
@@ -100,8 +120,24 @@ def k8s_inject_client_cert
 
   private
 
+  def handle_login_error(err)
+    logger.warn("Authentication Error: #{err.inspect}")
```
orenbm: please change this to "Login Error" so we can differentiate it from the "Authentication Error" below

orenbm: also, did we decide this should be a warn and not a debug? 
Are we sure the `err` doesn't contain any sensitive information? 
@shaharglazner do you see any concern here?

config/routes.rb
```diff
@@ -21,7 +21,9 @@ def matches?(request)
 
     constraints account: /[^\/\?]+/ do
       constraints authenticator: /authn-?[^\/]*/, id: /[^\/\?]+/ do
-        get '/:authenticator/:service_id/:account/status' => 'authenticate#status'
+        get '/:authenticator(/:service_id)/:account/status' => 'authenticate#status'
```
orenbm: Why is `service_id` optional? do we have a status check for `authn`?

app/domain/authentication/installed_authenticators.rb
```diff
@@ -5,59 +5,80 @@ class InstalledAuthenticators
...
-      pass_env = ::Authentication::AuthenticatorClass.new(cls).requires_env_arg?
-      pass_env ? cls.new(env: env) : cls.new
-    end
+      def env_enabled_authenticators(env)
+        authenticators = env["CONJUR_AUTHENTICATORS"]
+        authenticators.present? ? authenticators.split(',') : nil
+      end
+      
+      def db_enabled_authenticators
+        # For now, always include 'authn' when enabling authenticators via CLI
```
orenbm: What do you mean by "For now"? what can change in the future that will remove `authn` from the enabled authenticators?

orenbm:
@micahlee @jtuttle 

Please add to your TODO:
- [ ] Update [docs](https://docs.conjur.org/Latest/en/Content/Operations/Services/authentication-types.htm#Whitelis)

app/domain/authentication/installed_authenticators.rb
```diff
@@ -5,59 +5,80 @@ class InstalledAuthenticators
...
-        .where(kind => "webservice")
-        .select_map(identifier)
-        .map { |id| id.sub %r{^conjur\/}, "" }
-        .push(::Authentication::Common.default_authenticator_name)
-    end
+        Resource
+          .where(identifier.like("#{AUTHN_RESOURCE_PREFIX}%"))
+          .where(kind => "webservice")
+          .select_map(identifier)
+          .map { |id| id[%r{^conjur\/(authn(?:-[^\/]+)?(?:\/[^\/]+)?)$}, 1] }
```
orenbm: please add a comment to what is happening here to enhance readability

app/controllers/authenticate_controller.rb
```diff
@@ -25,9 +26,25 @@ def status
     )
     render json: { status: "ok" }
   rescue => e
-    render status_failure_response(e)
+    render status_response(e)
   end
 
+  def update
```
orenbm: i think that `update_config` will tell the story better

app/controllers/authenticate_controller.rb
```diff
@@ -25,9 +26,25 @@ def status
...
+    body_params = Rack::Utils.parse_nested_query(request.body.read)
+    
+    Authentication::UpdateAuthenticatorConfig.new.(
+      account: params[:account],
+      authenticator: params[:authenticator],
```
orenbm: in the rest of the file, this param is called `authenticator_name`. please align this and update `UpdateAuthenticatorConfig` accordingly

orenbm:
@jtuttle @micahlee can you please update the description of this PR so it explains what it does? Not everyone will know your hackathon project :)

app/domain/authentication/update_authenticator_config.rb
```diff
@@ -0,0 +1,60 @@
...
+    private
+
+    def validate_resource_visible
+      raise Exceptions::RecordNotFound, resource_id \
+        unless resource.visible_to?(@current_user)
+    end
+
+    def validate_resource_is_authenticator
+      raise Errors::Authentication::AuthenticatorNotFound, resource_id \
+        unless Authentication::InstalledAuthenticators.configured_authenticators
```
orenbm: `Authentication::InstalledAuthenticators.configured_authenticators` can be a dependency

app/domain/authentication/update_authenticator_config.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    private
+
+    def validate_resource_visible
+      raise Exceptions::RecordNotFound, resource_id \
+        unless resource.visible_to?(@current_user)
+    end
+
+    def validate_resource_is_authenticator
```
orenbm: can we change this method name to `validate_webservice_exists`? it seems that this is what it does

orenbm: And if this is really what's happening, consider using [ValidateWebserviceExists](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/validate_webservice_exists.rb)

You can see a usage [here](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/validate_status.rb)

app/domain/authentication/update_authenticator_config.rb
```diff
@@ -0,0 +1,60 @@
...
+      raise Errors::Authentication::AuthenticatorNotFound, resource_id \
+        unless Authentication::InstalledAuthenticators.configured_authenticators
+            .include?("#{@authenticator}/#{@service_id}")
+    end
+    
+    def validate_resource_writable
```
orenbm: [this](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/validate_webservice_access.rb) might help you here.

app/domain/authentication/update_authenticator_config.rb
```diff
@@ -0,0 +1,60 @@
...
+  UpdateAuthenticatorConfig = CommandClass.new(
+    dependencies: {
+      webservice_class: ::Authentication::Webservice,
+      resource_class: ::Resource,
+      authenticator_config_class: ::AuthenticatorConfig
+    },
+    inputs: %i(account authenticator service_id enabled current_user)
+  ) do
+    
+    def call
```
orenbm: what happens if you send an `update` request with an account that doesn't exist? if it fails with a misleading message (e.g `resource some-non-existing-account:webservice:conjur/authn-blah`) please use [ValidateAccountExists](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/validate_account_exists.rb).

 You can see a usage [here](https://github.com/cyberark/conjur/blob/aac94ad69d2d13e07ac9a90e5e9288e01121139b/app/domain/authentication/validate_webservice_access.rb#L32)

Also, after the fix (if needed) please add tests for this. You can see an example [here](https://github.com/cyberark/conjur/blob/aac94ad69d2d13e07ac9a90e5e9288e01121139b/cucumber/authenticators/features/authn_oidc.feature#L131)

cucumber/authenticators/features/authn_config.feature
```diff
@@ -0,0 +1,86 @@
...
+    And authenticator "cucumber:webservice:conjur/authn-ldap/test" is enabled
+
+    When I successfully PATCH "/authn-ldap/test/cucumber" with body:
+    """
+    enabled=false
+    """
+    Then the HTTP response status code is 204
+    And authenticator "cucumber:webservice:conjur/authn-ldap/test" is disabled
+
+  Scenario: Authenticator does not exist
```
orenbm: in this case it is not that the authenticator does not exist, rather than the account does not exist. Please change this test to `Account does not exist` and add another test for `Authenticator does not exist`. Both tests will return the same response but we can verify that we write the correct error to the log. 

please see [this example](https://github.com/cyberark/conjur/blob/aac94ad69d2d13e07ac9a90e5e9288e01121139b/cucumber/authenticators/features/authn_oidc.feature#L135) so see how we can check the logs.

cucumber/authenticators/features/authn_config.feature
```diff
@@ -0,0 +1,86 @@
...
+
+    - !grant
+      role: !group conjur/authn-ldap/test/readers
+      member: !user authn-reader
+
+    - !grant
+      role: !group conjur/authn-ldap/test/writers
+      member: !user authn-writer
+    """
+
```
orenbm: is the authenticator enabled or disabled when it is created? can we add a scenario to test this?

app/controllers/authenticate_controller.rb
```diff
@@ -100,8 +120,24 @@ def k8s_inject_client_cert
...
+      Errors::Authentication::Security::AccountNotDefined
+      raise Unauthorized
+    else
+      raise err
+    end
+  end
+
   def handle_authentication_error(err)
-    logger.debug("Authentication Error: #{err.inspect}")
+    logger.warn("Authentication Error: #{err.inspect}")
```
orenbm: i don't really like this happening under the hood. Are we sure the err doesn't contain any sensitive information?
@shaharglazner do you see any concern here?

orenbm: @sgnn7 can you weigh in? I remember talking to you about this after the hackathon

app/domain/authentication/update_authenticator_config.rb
```diff
@@ -0,0 +1,60 @@
...
+      validate_resource_visible
```
orenbm: a. can you elaborate on why do we need `validate_resource_visible`?
b. You can see my comments below but it seems that you can user here [ValidateAccountExists](https://github.com/cyberark/conjur/blob/aac94ad69d2d13e07ac9a90e5e9288e01121139b/app/domain/authentication/validate_account_exists.rb), [ValidateWebserviceExists](https://github.com/cyberark/conjur/blob/aac94ad69d2d13e07ac9a90e5e9288e01121139b/app/domain/authentication/validate_webservice_exists.rb) & [ValidateWebserviceAccess](https://github.com/cyberark/conjur/blob/aac94ad69d2d13e07ac9a90e5e9288e01121139b/app/domain/authentication/validate_webservice_access.rb)

app/domain/authentication/update_authenticator_config.rb
```diff
@@ -0,0 +1,60 @@
...
+      validate_resource_is_authenticator
+      validate_resource_writable
```
orenbm: a. can you elaborate on why do we need `validate_resource_visible`?
b. You can see my comments below but it seems that you can user here [ValidateAccountExists](https://github.com/cyberark/conjur/blob/aac94ad69d2d13e07ac9a90e5e9288e01121139b/app/domain/authentication/validate_account_exists.rb), [ValidateWebserviceExists](https://github.com/cyberark/conjur/blob/aac94ad69d2d13e07ac9a90e5e9288e01121139b/app/domain/authentication/validate_webservice_exists.rb) & [ValidateWebserviceAccess](https://github.com/cyberark/conjur/blob/aac94ad69d2d13e07ac9a90e5e9288e01121139b/app/domain/authentication/validate_webservice_access.rb)

orenbm: Please note that if you do use these classes, you will need to update the `rescue` clause of the `update` method in `authenticate_controller`

orenbm:
While you're at it - can you take a look at [this](https://github.com/cyberark/conjur/issues/1220) and see if your change fixes it?

app/controllers/authenticate_controller.rb
jtuttle: These are the errors it can produce:

```
    rescue Errors::Authentication::InvalidCredentials
      raise ApplicationController::Unauthorized, "Invalid username or password"
    rescue Errors::Authentication::InvalidOrigin
      raise ApplicationController::Forbidden, "User is not authorized to login from the current origin"
    rescue Errors::Authentication::Security::UserNotAuthorizedInConjur
      raise ApplicationController::Forbidden, "User is not authorized to login to Conjur"
    end
```

config/routes.rb
jtuttle: @micahlee do you remember why we did this? IIRC it was a bug we discovered during the hackathon but I can't recall the details.

app/controllers/authenticate_controller.rb
sgnn7: @orenbm yeah - this action is unknown if it will leak any credentials. we'd need to investigate the matrix of possible errors and authenticators to confirm that this won't dump stuff into the logs that shouldn't be there :/

orenbm: @jtuttle please change this to `debug` and create another issue to change this to `warn`. let's not delay this PR from being merged because of this. same goes for the warn in `handle_login_error`. thanks!

config/routes.rb
micahlee: Yes, it's because the built-in authenticator is just `authn` without an additional `/:service_id` component.

orenbm: sure, but why would someone want to know the status of the `authn` authenticator? isn't it always working?

The purpose of the `status` endpoint is for the operator to validate that the authenticator is configured properly, to get an indication that users can authenticate against it. it doesn't make much sense to have that for `authn`.

Having said that, if the `status` check for authn passes then it's fine. I didn't check that when we developed the status check because i didn't see any need to. If it works and doesn't fail on some weird error that assumes `service-id` is not blank then this change is fine. Please validate this before making this change.

jtuttle: Going to remove this change because YAGNI. I'd be surprised if there was a reason we had to do this and it may have just been for API consistency.

app/domain/authentication/installed_authenticators.rb
jtuttle: Removed the "For now".

app/domain/authentication/update_authenticator_config.rb
jtuttle: The purpose of this method is to validate that the webservice is a configured authenticator instead of (for example) a nested status webservice. Without this check you can configure the nested status webservice, which doesn't really make sense. I'll change the name to be more explicit.

app/domain/authentication/update_authenticator_config.rb
jtuttle: Refactored to use `ValidateWebserviceAccess`.

app/domain/authentication/update_authenticator_config.rb
jtuttle: Added validation using `ValidateAccountExists`.

cucumber/authenticators/features/authn_config.feature
jtuttle: Right you are, good catch. Fixed this test and added a separate test for non-existent webservice.

cucumber/authenticators/features/authn_config.feature
jtuttle: Added a test to confirm that an unconfigured authenticator is disabled.

app/domain/authentication/update_authenticator_config.rb
jtuttle: Changed validations to use the suggested classes and updated controller method to use the existing error handling logic.

cucumber/authenticators/features/authn_config.feature
```diff
@@ -0,0 +1,97 @@
...
+    Then the HTTP response status code is 204
+    And authenticator "cucumber:webservice:conjur/authn-ldap/test" is disabled
+
+  Scenario: Account does not exist
+    When I am the super-user
+    And I PATCH "/authn-ldap/test/nope" with body:
+    """
+    enabled=true
+    """
+    Then the HTTP response status code is 401
```
orenbm: can you verify that we log the message we expect?

you can see a usage here: https://github.com/cyberark/conjur/blob/master/cucumber/authenticators/features/authn_oidc.feature#L80

spec/app/domain/authentication/security/validate_webservice_is_authenticator_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  
+  let(:webservice_id) { "#{fake_authenticator_name}/#{fake_service_id}" }
+  let(:same_authenticator) { "#{fake_authenticator_name}/same-authn" }
+  let(:different_authenticator) { "diff-authn/service-id" }
+
+  let(:webservice) {
+    mock_webservice(test_account, fake_authenticator_name, fake_service_id)
+  }
+  
+  context "webserve is an authenticator" do
```
orenbm: webservice is ...

spec/app/domain/authentication/security/validate_webservice_is_authenticator_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+        webservice: webservice
+      )
+    }
+
+    it "validates without error" do
+      expect { subject }.to_not raise_error
+    end
+  end
+
+  context "webserve is not authenticator" do
```
orenbm: webservice is...

Issue #1189, opened 2019-09-19
Authn-k8s requires authenticated apps to be in specific "apps" policy


rafis3:
## Issue Description
The authn-k8s authenticator requires the applications to be specifically defined under the policy `conjur/authn-k8s/<service_id>/apps`. This behavior does not align with the other authenticators and is somewhat limiting the control for the user.

## Expected Behavior
Applications that get authenticated with authn-k8s should be able to reside anywhere in the policy tree. The authenticator should only require that they have `authenticate` permissions on the authenticator instance `!webservice` resource.

DoD:
- [ ] Applications can reside anywhere in the policy tree
- [ ] The solution is backward compatible i.e. old authenticators can work with new conjur and new authenticators can work with old conjur
- [ ] The Core and Community team is familiar with this solution and its design is available to them

izgeri:
Notes from when I experienced the problem:

I assumed if the layer `conjur/authn-k8s/{AUTHENTICATOR_ID}/apps` had the correct privileges (eg has  `[read, authenticate]` on the authenticator webservice and has `[read, execute]` on the needed secrets) that I could just load a host that’s a member of this layer wherever I want in policy and it could use the k8s authn sidecar:
```
- !host
  id: {APPLICATION_NAMESPACE}/service_account/{APPLICATION_SERVICE_ACCOUNT}
  annotations:
    kubernetes/authentication-container-name: secretless
    kubernetes: "true"

- !grant
  role: !layer conjur/authn-k8s/{AUTHENTICATOR_ID}/apps
  member: !host {APPLICATION_NAMESPACE}/service_account/{APPLICATION_SERVICE_ACCOUNT}
```

I learned yesterday that this is not true, and when I try to authenticate as `host/{APPLICATION_NAMESPACE}/service_account/{APPLICATION_SERVICE_ACCOUNT}` using the authn sidecar the Conjur server processes the request as if I authenticated as `host/conjur/authn-k8s/{AUTHENTICATOR_ID}/app/{APPLICATION_NAMESPACE}/service_account/{APPLICATION_SERVICE_ACCOUNT}`
that is - Conjur automatically prepends the host ID I give the authenticator client with a very specific policy branch ID for that specific authenticator service

this breaks some expectations:
- the host ID I provide the authenticator is the fully qualified ID of the host (it’s actually not)
- membership to the layer with the correct privileges is the only thing that matters (it’s not, the layer needs to be in a very specific policy branch and the host must also be defined in that branch)

[...]

PR #1228, opened 2019-11-25
Allow hosts to be defined outside apps

orenbm:
#### What does this PR do?

In order to authenticate hosts from every policy branch, we need the
authn-client to send the full host-id to the server, and in turn the server builds the conjur host-id 
from the common-name instead of using the prefix `{@account}:host:conjur/authn-k8s/#{@service_name}/apps`.

To maintain backwards compatibility we need to support also client versions that do not send the
full host-id, and assume that the server adds the host-id prefix to it. in this PR we handle this case and add the prefix to the common-name in the server, before validation.

We want to give our customers the freedom to create hosts anywhere in the policy branch, and not only under the `conjur/authn-k8s/#{@service_name}/apps` branch as we do now.

Connected to #1189 

Yes.

### DIFF
--- a/app/domain/authentication/authn_k8s/inject_client_cert.rb
+++ b/app/domain/authentication/authn_k8s/inject_client_cert.rb
@@ -16,10 +16,11 @@ module Authentication
         kubectl_exec: KubectlExec,
         validate_pod_request: ValidatePodRequest.new
       },
-      inputs: %i(conjur_account service_id csr)
+      inputs: %i(conjur_account service_id csr host_id_prefix)
     ) do
 
       def call
+        update_csr_common_name
         validate
         install_signed_cert
       end
@@ -50,6 +51,22 @@ module Authentication
         validate_cert_installation(resp)
       end
 
+      # In the old version of the authn-client we assumed that the host is under the "apps" policy branch.
+      # Now we send the host-id in 2 parts:
+      #   suffix - the last 3 parts that indicate the machine identity (e.g ${TEST_APP_NAMESPACE_NAME}/*/*)
+      #   prefix - all the rest (e.g host/conjur/authn-k8s/${AUTHENTICATOR_ID})
+      # We update the CSR's common_name to have the full host-id. This way, the validation
+      # that happens in the "authenticate" request will work, as the signed certificate
+      # contains the full host-id.
+      def update_csr_common_name
+        prefix = @host_id_prefix.nil? || @host_id_prefix.empty? ? apps_host_id_prefix : @host_id_prefix
+        smart_csr.common_name = prefix + "." + smart_csr.common_name
+      end
+
+      def apps_host_id_prefix
+        "host.conjur.authn-k8s.#{@service_id}.apps"
+      end
+
       def pod_request
         PodRequest.new(
           service_id: @service_id,
@@ -62,7 +79,7 @@ module Authentication
         @k8s_host ||= Authentication::AuthnK8s::K8sHost.from_csr(
           account: @conjur_account,
           service_name: @service_id,
-          csr: @csr
+          csr: smart_csr
         )
       end
 
@@ -132,7 +149,7 @@ module Authentication
 
       def cert_to_install
         ca_for_webservice.signed_cert(
-          @csr,
+          smart_csr,
           subject_altnames: [spiffe_id.to_altname]
         )
       end
--- a/app/domain/util/open_ssl/x509/smart_csr.rb
+++ b/app/domain/util/open_ssl/x509/smart_csr.rb
@@ -29,6 +29,15 @@ module Util
           smart_subject.common_name
         end
 
+        def common_name=(common_name)
+          smart_subject.common_name = common_name
+        end
+
+        # We don't just return subject as it might have been updated
+        def subject_to_s
+          smart_subject.to_s
+        end
+
         private
 
         def smart_subject
--- a/cucumber/kubernetes/features/pki.feature
+++ b/cucumber/kubernetes/features/pki.feature
@@ -1,8 +1,16 @@
 Feature: Certificates issued by the login method.
 
-  Scenario: The subject name is the dot-separated relative Conjur host ID.
+  Scenario: The subject name is the dot-separated, with hard-coded apps prefix, Conjur host ID when the "Host-Id-Prefix" header isn't present.
     When I can login to pod matching "app=inventory-pod" to authn-k8s as "*/*"
-    Then the certificate subject name is "/CN=@namespace@.*.*"
+    Then the certificate subject name is "/CN=host.conjur.authn-k8s.minikube.apps.@namespace@.*.*"
+
+  Scenario: The subject name is the dot-separated given Conjur host ID in apps when the "Host-Id-Prefix" header is present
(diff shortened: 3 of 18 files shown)


orenbm:
@garkler we need to release a version with this change. What is the process? Do I just update `VERSION` & `CHANGELOG`?

orenbm:
@jvanderhoof can you have someone to review this?
@nessiLahav will review too

jvanderhoof:
`CHANGELOG` will need to have an entry. Don't worry about the `VERSION` file. We'll take care of that as part of the release.

app/domain/util/open_ssl/x509/smart_csr.rb
```diff
@@ -29,6 +29,15 @@ def common_name
           smart_subject.common_name
         end
 
+        def common_name=(common_name)
+          smart_subject.common_name = common_name
+        end
+
+        # We don't just return subject as it might have been updated
+        def subject_to_s
```
jvanderhoof: This method name doesn't read very intuitively.  Can we just call it `subject`?

orenbm:
> CHANGELOG will need to have an entry.
@jvanderhoof what do you mean here? I added an entry in the CHANGELOG for the new version. am i missing something?

orenbm:
> `CHANGELOG` will need to have an entry. Don't worry about the `VERSION` file. We'll take care of that as part of the release.

is it ok now? If i understand you correctly, I need to add an entry to the `Unreleased` section. @jvanderhoof am i correct?

app/domain/util/open_ssl/x509/smart_csr.rb
orenbm: well, it can't be `subject` because that is another parameter - the one that smart_subject wraps (https://github.com/cyberark/conjur/blob/528c2eac8096807a19245ad3bbed6c7e472014f2/app/domain/util/open_ssl/x509/smart_csr.rb#L35)

I thought of `updated_subject` but that can be misleading because it might not have been updated.

`printable_subject` is also a candidate but that's very similar to `subject_to_s`. 

what do you think?

cucumber/kubernetes/features/pki.feature
```diff
@@ -1,8 +1,16 @@
 Feature: Certificates issued by the login method.
 
-  Scenario: The subject name is the dot-separated relative Conjur host ID.
```
jvanderhoof: Does this mean our old behavior is changing?

orenbm: yes but the change is affected also in the `/authenticate` request that follows this response. 

The client sends a CSR in the `/inject_client_cert` request. This CSR is signed and is sent back to the client. After that the certificate is sent again to the server in the `/authenticate` request and is validated.

We changed the CSR that is returned at the end of the first request, _and_ the way it is treated in the second request so it is seamless to the client - old or new.

jvanderhoof:
@orenbm (cc: @Tovli), as a start, can we please add the following tests to this repo:
 **Unit Tests**
 - [ ] Old style reference (scoped by authenticator)
 - [ ] New style reference (not scoped by authenticator)
 - [ ] Certificate w/ old style modifies certificate to expected cert format
 - [ ] Certificate w/ new style generates expected cert format

**Integration Tests**
 - [ ] `inject_client_cert` flow old format works in OCP 3.9
 - [ ] `inject_client_cert` flow new format works in OCP 3.9

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -50,6 +51,22 @@ def install_signed_cert
...
       end
 
+      # In the old version of the authn-client we assumed that the host is under the "apps" policy branch.
+      # Now we send the host-id in 2 parts:
+      #   suffix - the last 3 parts that indicate the machine identity (e.g ${TEST_APP_NAMESPACE_NAME}/*/*)
+      #   prefix - all the rest (e.g host/conjur/authn-k8s/${AUTHENTICATOR_ID})
+      # We update the CSR's common_name to have the full host-id. This way, the validation
+      # that happens in the "authenticate" request will work, as the signed certificate
+      # contains the full host-id.
+      def update_csr_common_name
```
jvanderhoof: Is it possible for the `smart_csr` to generate the correct common name for itself?  It feels like the abstraction is leaking a bit.

orenbm:
> * [ ]  `inject_client_cert` flow old format works in OCP 3.9
> * [ ]  `inject_client_cert` flow new format works in OCP 3.9
@jvanderhoof  do we have the infra to test with OCP 3.9? Do all the other tests run against OCP 3.9?

orenbm:
> **Unit Tests**
> 
> * [ ]  Old style reference (scoped by authenticator)
> * [ ]  New style reference (not scoped by authenticator)

@jvanderhoof didn't really understand these. Can you elaborate on the unit we are testing here?

app/domain/authentication/authn_k8s/inject_client_cert.rb
orenbm: I don't understand your suggestion. 

The logic here is:
1. understand wether the client sent a custom prefix
2. Deciding which is the corresponding prefix according to (1)
3. Update the common-name to have the prefix chosen in (2)

Are you suggesting to move this logic into `SmartCsr`?

jvanderhoof:
@orenbm, I updated the the above unit test description. Essentially I want to make sure we're testing both the new and old style `inject_client_cert` formats.  Currently it looks like they have only been manually tested.

We have a OCP 3.9 cluster (we're using it to run the new [follower tests](https://github.com/conjurinc/container-dap)).  As this feature is primarily for our enterprise customers, it struck me that we're not really testing in the environment they will be using the feature inside.

orenbm:
> Essentially I want to make sure we're testing both the new and old style inject_client_cert formats.

@jvanderhoof please see
  - UTs
    - [Old way](https://github.com/cyberark/conjur/blob/allow-hosts-outside-apps/spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb#L308) 
    - [New way](https://github.com/cyberark/conjur/blob/allow-hosts-outside-apps/spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb#L323) 
  - Integration tests
    -  Old way: [here](https://github.com/cyberark/conjur/blob/allow-hosts-outside-apps/cucumber/kubernetes/features/login.feature#L23) & [here](https://github.com/cyberark/conjur/blob/allow-hosts-outside-apps/cucumber/kubernetes/features/pki.feature#L3)
    - New way: [here](https://github.com/cyberark/conjur/blob/allow-hosts-outside-apps/cucumber/kubernetes/features/login.feature#L26) & [here](https://github.com/cyberark/conjur/blob/allow-hosts-outside-apps/cucumber/kubernetes/features/pki.feature#L7)

Do you think we need more than that?

orenbm:
> We have a OCP 3.9 cluster (we're using it to run the new follower tests). As this feature is primarily for our enterprise customers, it struck me that we're not really testing in the environment they will be using the feature inside.

@jvanderhoof where can I find how tests are run there? I never done it before

orenbm:
@jvanderhoof can you approve the PR please?

PR #1229, opened 2019-11-25
Updated test LDAP certificates

sgnn7:
The old ones expired and needed to be replaced. README was added to
ensure that it's a repeatable procedure.

N/A


ci/ldap-certs/README.md
```diff
@@ -0,0 +1,7 @@
+# Regenerating Test Certificates
```
micahlee: Thanks for adding this!

PR #1230, opened 2019-11-25
Upgrade the container base image

sgnn7:
This change upgrades the base image that Conjur OSS depends on and flattens the image as well to reduce layers. Small cleanups were done as part of this process since the Dockerfile seemed neglected for a while. Nokogiri + bundler were also upgraded to remove a security CVE issue.

Connected to https://github.com/conjurinc/container-dap/issues/137

[Jenkins Build](https://jenkins.conjur.net/job/cyberark--conjur/job/upgrade-base-image/)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
 FROM ubuntu:20.04
 
-ENV DEBIAN_FRONTEND=noninteractive
-
-ENV PORT 80
+ENV DEBIAN_FRONTEND=noninteractive \
+    PORT=80
 
 EXPOSE 80
 
@@ -15,8 +14,6 @@ RUN apt-get update -y && \
                        postgresql-client \
                        ruby2.5 ruby2.5-dev \
                        tzdata \
-                       unattended-upgrades \
-                       update-notifier-common \
                        # needed to build some gem native extensions:
                        libz-dev \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -15,18 +17,21 @@ RUN apt-get update -y && \
...
     && rm -rf /var/lib/apt/lists/*
 
-RUN gem install -N -v 1.16.1 bundler
-
-RUN mkdir -p /opt/conjur-server
+RUN gem install -N -v 1.17.3 bundler
 
 WORKDIR /opt/conjur-server
 
+# Ensure few required GID0-owned folders to run as a random UID (OpenShift requirement)
```
jtuttle: Why move this here? We still have to build the Dockerfile in container-dap to insert the custom `run-conjur` entrypoint right? Isn't it cleaner to have OpenShift-specific functionality added in another repo?

build.sh
```diff
@@ -19,8 +19,26 @@ esac
 shift # past argument or value
 done
 
+# Flatten resulting image.
+function flatten() {
+  local image="$1"
+  echo "Flattening image '$image'..."
+
+  # We have to rebuild some values since we're flattening with
```
jtuttle: Can you take another stab at this comment? It doesn't really help me understand what's going on here.

Dockerfile
sgnn7: This container ("Conjur OSS") should be capable of running in OC without any changes if we want wide adoption and we make that happen with the changes in the container-dap repo. By keeping the changes there, we are not upstreaming our fixes to the proper repo which is kind of backwards of what we should be doing so I moved some of the more critical ones here (though a few more might be needed for running OSS as a limited user in k8s but that was out of this scope).

PR #1233, opened 2019-12-04
Use postgres advisory lock to ensure that only one server does the rotations

ismarc:
#### What does this PR do?
Enable advisory locks around secret rotation to prevent conflicts and race conditions on multiple servers trying to rotate secrets.
When Conjur does secret rotations, it fetches all secrets that need rotation from the database, then processes the list one by one, rotating each then updating the DB.  If multiple Conjur instances point to the same DB, this allows for race conditions and multiple updates to happen on each rotation, even if each secret is locked individually.  This provides a lock around the rotation process itself so only one server will do the rotation each time.  In the future, the process of determining which secret to rotate, and fetching them from the DB one at a time could remove the need for the advisory lock.
Provide the same CONJUR_DATA_KEY and postgres URL to two different Conjur instances, configure a rotation and observe the secret only being rotated by one server on each increment
Covered by existing tests
Maybe?  Wouldn't be very exciting
This particular change doesn't have any needed associated documentation.  What it enables (multiple Conjur instances to a single Postgres db) should be documented once fully verified.
No

### DIFF
--- a/app/domain/rotation/master_rotator.rb
+++ b/app/domain/rotation/master_rotator.rb
@@ -13,14 +13,13 @@ module Rotation
     def initialize(avail_rotators:,
                    rotation_model: ::Secret,
                    secret_model: ::Secret,
-                   facade_cls: ::Rotation::ConjurFacade,
-                   db: Sequel::Model.db)
+                   facade_cls: ::Rotation::ConjurFacade)
       @avail_rotators = avail_rotators
       @rotation_model = rotation_model
       @secret_model = secret_model
       @facade_cls = facade_cls
-      @db = db
-      @db.extension :pg_advisory_locking
+
+      Sequel::Model.db.extension :pg_advisory_locking
     end
 
     def rotate_every(seconds)
@@ -36,7 +35,7 @@ module Rotation
       # The same `id` needs to be used across all servers talking to the same
       # database, but can be any unused integer.
       id = Rails.configuration.rotator_lock_name.to_i(36)
-      @db.try_advisory_lock(id) do
+      Sequel::Model.db.try_advisory_lock(id) do
         scheduled_rotations.each(&:run)
       end
     end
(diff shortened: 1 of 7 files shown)


app/domain/rotation/master_rotator.rb
```diff
@@ -28,7 +31,17 @@ def rotate_every(seconds)
     end
 
     def rotate_all
-      scheduled_rotations.each(&:run)
+      # Attempt an advisory lock to prevent other servers from trying to rotate
+      # at the same time or immediately after another server rotated the secret.
+      # The same `id` needs to be used across all servers talking to the same
+      # database, but can be any unused integer.
+      id = 8675309
```
sgnn7: This needs to be something static and definitely not within the rotate function itself. Can we take a hash of something stable to generate it?

app/domain/rotation/master_rotator.rb
```diff
@@ -28,7 +31,17 @@ def rotate_every(seconds)
...
+      @db.try_advisory_lock(id) do
+        scheduled_rotations.each(&:run)
+      end
+    rescue => e
+      set_retry_expiration
+      log_error(e)
```
sgnn7: This loop seems a tad bit odd - we still want to propagate an error up if the rotation itself fails

sgnn7: Also, this really should be a warning if it was just the locking failure, not an error probably

app/domain/rotation/master_rotator.rb
ismarc: Updated to use a string specified under initializers for the rotator that gets converted to integer at point in time (used the string "possum" for now).

app/domain/rotation/master_rotator.rb
ismarc: After reviewing the history, this was a bad copy/paste from deeper in the stack where the error handling was needed, it's been pulled out.

config/initializers/rotator.rb
```diff
@@ -0,0 +1,4 @@
+Rails.application.configure do
+  config.rotator_lock_name = "conjur"
+end
+
```
sgnn7: @jtuttle / @ismarc Drive-by comment: Stray newline

PR #1234, opened 2019-12-05
Coverage data is submitted to Code Climate

hughsaunders:
#### What does this PR do?
**Summary:**
Collect code coverage data from all the tests, merge them, generate an html report to store in Jenkins and also submit the coverage data to code climate.

**Detail:**
* Adds two stages to the Jenkinsfile, one to prepare for coverage data before the tests run, and another to collect results after tests complete.
* Modifies the simplecov configuration `.simplecov`:
  * Adds a custom command name including a random integer to prevent
    collisions between branches when merging reports
  * Overrides the at_exit hook to to add an optional sleep. The sleep is controlled by an environment variable `SIMPLECOV_SLEEP`. The purpose of the sleep is to keep the conjur process running after the simplecov report has been generated, which allows the report to be retrieved from a k8s pod, before the pod is killed.
* Adds a `REQUIRE_SIMPLECOV` env variable, previously this was governed by the `RAILS_ENV` variable but the `cucumber_authenticator` tests fail when `RAILS_ENV` is set to test so a new var was needed to enable coverage recording without changing the rails environment.
* Ensures that conjur is killed before the container its running in is killed. SimpleCov uses an `at_exit` hook to write out the report. If the container is killed the `at_exit` hook doesn't fire and the report isn't written. 
* Adds a new dirctory `ci/coverage-report-generator` which contains a ruby script (`generate_report.rb`) and supporting files for generating an HTML report from SimpleCov JSON output. Normally the HTML report is generated during the `at_exit` hook mentioned above, but this repo requires manually merged test results, so a new HTML report must be generated to reflect the merged JSON results. On merging: SimpleCov locks its storage dir and handles merging of most of the test sets itself, the only test results that require manual merging are the GKE tests as they can't mount the coverage directory from `$WORKSPACE`. 
* Set `WEB_CONCURRENCY: 0`. With this value `>0` `puma` (the webserver conjur uses) creates child processes to handle requests, SimpleCov cannot track calls in those processes, so coverage data is innaccurate. Setting this value to 0 ensures that requests are handled within the main process and SimpleCov tracks code usage succesfully. To compensate for the lack of child processes, `RAILS_MAX_THREADS` is increased to 10 from the default of 5 (previously there were 2x5 threads).
* Updates SimpleCov config in spec_helper files to ensure reports are generated and can be safely merged with other reports.

This is required for the updated definition-of-done. 
Connected to conjurinc/container-dap#107
* [A successful build](https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--conjur/detail/cyberark-conjur-107/48/pipeline/)
* [JSON report artifact, showing 7 top level keys for each of the 7 parallel test sets](https://jenkins.conjur.net/job/cyberark--conjur/job/cyberark-conjur-107/48/artifact/coverage/.resultset.json): 

    ```
    $ jq keys <resultset.json 
    [
      "SimpleCov 115551",
      "SimpleCov 296886",
      "SimpleCov 348898",
      "SimpleCov 41905",
      "SimpleCov 73534",
      "SimpleCov 799924",
      "SimpleCov 937420"
    ]
    ```
* [HTML report artifact](https://jenkins.conjur.net/job/cyberark--conjur/job/cyberark-conjur-107/48/Coverage_20Report/)

* [Example report received by CodeClimate](https://codeclimate.com/repos/5b460c731a0e37241a000488/settings/test_reports/5ded442c09a64f48f7002a52). Note that this doesn't show in the main coverage section as this is not yet merged to master, and I can't find a way of getting codeclimate to display coverage for non-default branches.

Nope, just CI updates here.

This work is entirely improvements to test infrastructure to ensure that coverage data is correctly stored and published.

Possibly, theres a neat jq dict merge which people might be unfamiliar with, not sure thats enough for a whole blog post though.

Not oustide the code.

Nope.

### DIFF
--- a/ci/coverage-report-generator/Gemfile
+++ b/ci/coverage-report-generator/Gemfile
@@ -1,0 +1,2 @@
+source 'https://rubygems.org'
+gem 'simplecov', '~> 0.17.1'
--- a/ci/coverage-report-generator/generate_report.rb
+++ b/ci/coverage-report-generator/generate_report.rb
@@ -1,0 +1,50 @@
+#!/usr/bin/env ruby
+
+# This script generates an html coverage report from simplecov json output.
+# Normally simplecov generates the html report at the end of a run, however
+# the conjur tests produce multiple simplecov reports, each with json data
+# and an html report. The json files are easy enough to merge, the html reports
+# are much harder. So instead we merge the json files then use this script
+# to generate a new html report. See ci/submit-coverage for the merge.
+
+require 'simplecov'
+require 'json'
+
+# Override at_exit callback as we don't want this program to hang forever
+# (.simplecov adds infinite sleep to keep containers alive after writing the
+# coverage report)
+# We also don't actually want a coverage report for this script ;-)
+SimpleCov.at_exit do
+end
+
+if ARGV.size != 2
+  puts "Usage: generate_report.rb <project root dir> <report json file>"
+  exit!
+end
+
+# Simplecov filters exclude all files outside the current project root.
+# The project root defaults to the working directory, so we have to move
+# the root up a couple of levels so all the source files can be included.
+# Use the first argument so the user can specify the approprite dir
+SimpleCov.root ARGV[0]
+
+# Read the result file, path passed in as second arg.
+jsonraw = File.open(ARGV[1]).read
+
+# Parse JSON to create ruby object
+jsonobj = JSON.parse(jsonraw)
+
+# Create result object for each subresult
+resultobjs = jsonobj.keys.map do |key|
+  SimpleCov::Result.from_hash(key => jsonobj[key])
+end
+
+# Merge the sub-results.
+# This is actually the second merge (ci/submit-coverage uses jq to merge
+# multiple result files together, this merges separate sub reports within the
+# one json structure.)
+mergedresult = SimpleCov::ResultMerger.merge_results(*resultobjs)
+
+# Format the result using the html formatter
+formatter = SimpleCov::Formatter::HTMLFormatter.new
+formatter.format(mergedresult)
--- a/ci/coverage-report-generator/generate_report.rb
+++ b/ci/coverage-report-generator/generate_report.rb
@@ -7,8 +7,8 @@
 # are much harder. So instead we merge the json files then use this script
 # to generate a new html report. See ci/submit-coverage for the merge.
 
-require 'simplecov'
 require 'json'
+require 'simplecov'
 
 # Override at_exit callback as we don't want this program to hang forever
 # (.simplecov adds infinite sleep to keep containers alive after writing the
--- a/ci/coverage-report-generator/run.sh
+++ b/ci/coverage-report-generator/run.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# Script to run generate_report.rb within docker so that the gems don't
+# need to be installed locally.
+# generate_report.rb generates an html report from simplecov json output.
+
+set -xeu
+
+repo_root=$(git rev-parse --show-toplevel)
+# Use the first arg, or if not supplied use the simplecov default report location
+report_file="${1:-${repo_root}/coverage/.resultset.json}"
+image="ruby:2.6.5-stretch"
+
+docker run \
+    --rm \
+    -v "${repo_root}":"${repo_root}" \
+    -w "${repo_root}/ci/coverage-report-generator" \
+    "${image}" \
+        bash -cx "bundle install --path gems
+                  bundle list
+                  bundle exec ./generate_report.rb \
+                     ${repo_root} \
+                     ${report_file}
+                 "
--- a/ci/coverage-report-generator/run.sh
+++ b/ci/coverage-report-generator/run.sh
@@ -6,19 +6,21 @@
 
 set -xeu
 
-repo_root=$(git rev-parse --show-toplevel)
+IMAGE="ruby:2.6.5-stretch"
+
+REPO_ROOT=$(git rev-parse --show-toplevel)
+
 # Use the first arg, or if not supplied use the simplecov default report location
-report_file="${1:-${repo_root}/coverage/.resultset.json}"
-image="ruby:2.6.5-stretch"
+REPORT_FILE="${1:-${REPO_ROOT}/coverage/.resultset.json}"
 
 docker run \
     --rm \
-    -v "${repo_root}":"${repo_root}" \
-    -w "${repo_root}/ci/coverage-report-generator" \
-    "${image}" \
-        bash -cx "bundle install --path gems
-                  bundle list
-                  bundle exec ./generate_report.rb \
-                     ${repo_root} \
-                     ${report_file}
-                 "
+    -v "${REPO_ROOT}":"${REPO_ROOT}" \
+    -w "${REPO_ROOT}/ci/coverage-report-generator" \
+    "${IMAGE}" \
+        bash -cex "bundle install --path gems
+                   bundle list
+                   bundle exec ./generate_report.rb \
+                       ${REPO_ROOT} \
+                       ${REPORT_FILE}
+                  "
--- a/ci/docker-compose.yml
+++ b/ci/docker-compose.yml
@@ -17,6 +17,7 @@ services:
       CONJUR_ACCOUNT: cucumber
       CONJUR_DATA_KEY:
       RAILS_ENV:
+      REQUIRE_SIMPLECOV: "true"
       CONJUR_LOG_LEVEL: debug
       CONJUR_AUTHENTICATORS: authn-ldap/test,authn-ldap/secure,authn-oidc/keycloak
       LDAP_URI: ldap://ldap-server:389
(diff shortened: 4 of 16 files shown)


ci/coverage-report-generator/run.sh
```diff
@@ -0,0 +1,24 @@
...
+# Use the first arg, or if not supplied use the simplecov default report location
+report_file="${1:-${repo_root}/coverage/.resultset.json}"
+image="ruby:2.6.5-stretch"
+
+docker run \
+    --rm \
+    -v "${repo_root}":"${repo_root}" \
+    -w "${repo_root}/ci/coverage-report-generator" \
+    "${image}" \
+        bash -cx "bundle install --path gems
```
sgnn7: Doesn't this need a `e` as ell in the `-cx`?

ci/submit-coverage
```diff
@@ -0,0 +1,68 @@
...
+# The rspec and cucumber reports are all written to repo_root/coverage
+# Simplecov handles locking and merging for parallel tests.
+# The GKE tests can't mount that dir, so they produce a separate report
+# which is merged into repo_root/coverage by this script.
+
+set -eux
+
+dir="coverage"
+bin="cc-test-reporter"
+report="${dir}/.resultset.json"
```
sgnn7: these are constants - I think we normally use all-caps for those

ci/submit-coverage
```diff
@@ -0,0 +1,68 @@
...
+
+if [[ ! -e ${report} ]]; then
+    echo "SimpleCov report (${report}) not found"
+    ls -laR ${dir}
+    return 1
```
sgnn7: Hmm, is there a reason why we don't use `exit <code>` instead of return?

ci/submit-coverage
```diff
@@ -0,0 +1,68 @@
...
+fi
+
+if [[ ! -x ${bin} ]]; then
+    echo "cc-test-reporter binary not found, not reporting coverage data to code climate"
+    ls -laR ${dir}
+    # report is present but reporter binary is not, definitely a bug, exit error.
+    return 1
+fi
+
+gke_report="ci/authn-k8s/output/simplecov-resultset-authnk8s-gke.json"
```
sgnn7: nit: this is a const and can probably go up top of the file (or should be within the `if`)

ci/submit-coverage
```diff
@@ -0,0 +1,68 @@
...
+    # be able to read all the files referenced within the report. As the reports
+    # are generated in containers, the absolute paths contained in the report
+    # are not valid outside that container. This sed fixes the paths
+    # So they are correct relative to the Jenkins workspace.
+    sed -i -E "s+/(opt|src)/conjur-server+${WORKSPACE}+g" "${dir}/.resultset.json"
+
+    # Now need to regenerate the html report, as it was generated before
+    # the GKE results were merged in.
+    pushd ci/coverage-report-generator
+        ./run.sh
```
sgnn7: nit: Would be good to have an echo to show that we're trying to run this - it's easier to debug if run can't be launched or is missing.

ci/submit-coverage
```diff
@@ -0,0 +1,68 @@
...
+    popd
+else
```
sgnn7: nit: If you do this case first, the other part doesn't need to be indented

ci/submit-coverage
```diff
@@ -0,0 +1,68 @@
...
+    exit 1
+fi
+
+echo "Coverage reports prepared, submitting to CodeClimate."
+# vars GIT_COMMIT, GIT_BRANCH & TRID are set by ccCoverage.dockerPrep
+
+./${bin} after-build \
+    --coverage-input-type "simplecov"\
+    --id "${TRID}" \
+    && echo "Successfully Reported Coverage Data"
```
sgnn7: `bash -e` should prevent the need to chain these two with `&&`

ci/test
```diff
@@ -105,6 +105,7 @@ rspec() {
     rm -rf $REPORT_ROOT/spec/reports
     bundle exec env CI_REPORTS=$REPORT_ROOT/spec/reports rspec --format progress --require ci/reporter/rspec --format CI::Reporter::RSpecFormatter
   "
+
```
sgnn7: nit: Stray newline I think

ci/docker-compose.yml
```diff
@@ -17,19 +17,23 @@ services:
...
+      REQUIRE_SIMPLECOV: "true"
       CONJUR_LOG_LEVEL: debug
       CONJUR_AUTHENTICATORS: authn-ldap/test,authn-ldap/secure,authn-oidc/keycloak
       LDAP_URI: ldap://ldap-server:389
       LDAP_BASE: dc=conjur,dc=net
       LDAP_FILTER: '(uid=%s)'
       LDAP_BINDDN: cn=admin,dc=conjur,dc=net
       LDAP_BINDPW: ldapsecret
+      WEB_CONCURRENCY: 0
+      RAILS_MAX_THREADS: 10
```
sgnn7: I think we may want to document somewhere that the ci-based runs now will be single-threaded so that someone doesn't try to benchmark the app based on this setup.

ci/coverage-report-generator/run.sh
```diff
@@ -0,0 +1,24 @@
...
+# Script to run generate_report.rb within docker so that the gems don't
+# need to be installed locally.
+# generate_report.rb generates an html report from simplecov json output.
+
+set -xeu
+
+repo_root=$(git rev-parse --show-toplevel)
+# Use the first arg, or if not supplied use the simplecov default report location
+report_file="${1:-${repo_root}/coverage/.resultset.json}"
+image="ruby:2.6.5-stretch"
```
sgnn7: nit: Maybe pull this up to make sure it's visible since it's not the same ruby version that Conjur natively runs

ci/coverage-report-generator/generate_report.rb
```diff
@@ -0,0 +1,50 @@
...
+
+# This script generates an html coverage report from simplecov json output.
+# Normally simplecov generates the html report at the end of a run, however
+# the conjur tests produce multiple simplecov reports, each with json data
+# and an html report. The json files are easy enough to merge, the html reports
+# are much harder. So instead we merge the json files then use this script
+# to generate a new html report. See ci/submit-coverage for the merge.
+
+require 'simplecov'
+require 'json'
```
sgnn7: super tiny nit - the requires should be swapped (alphabetization)

ci/coverage-report-generator/Gemfile
```diff
@@ -0,0 +1,2 @@
+source 'https://rubygems.org'
+gem 'simplecov', '~> 0.17.1'
```
sgnn7: Are we still using [this](https://github.com/cyberark/conjur/blob/master/Gemfile#L54) thing? If not, it probably should be deleted but I'm guessing that we may want to consider just a separate deployment type and having simplecov as part of conjur gemset maybe? WDYT?

ci/submit-coverage
hughsaunders: `set -x` will handle that

ci/submit-coverage
hughsaunders: Nope, that should be `exit`, thanks.

ci/submit-coverage
hughsaunders: Handled by `set -x`

ci/docker-compose.yml
hughsaunders: Any suggestions where?

sgnn7: Hrmm... not sure tbh - maybe [here](https://github.com/cyberark/conjur#set-up-a-development-environment)?

ci/submit-coverage
sgnn7: Good call - totally missed it

ci/coverage-report-generator/Gemfile
hughsaunders: I'm not a bundler expert, but it looks like for that to work, every gem would need to be in a group so they could be excluded, probably easier to just keep the separate `Gemfile`. Unless theres a super strong reason for conjur and this random script to have to use the same verison of `simplecov`?

sgnn7: Yeah fair enough

sgnn7: Then we should remove the one in main gemfile then

ci/docker-compose.yml
hughsaunders: Added a note

ci/coverage-report-generator/run.sh
```diff
@@ -0,0 +1,26 @@
...
+docker run \
+    --rm \
+    -v "${REPO_ROOT}":"${REPO_ROOT}" \
+    -w "${REPO_ROOT}/ci/coverage-report-generator" \
+    "${IMAGE}" \
+        bash -cex "bundle install --path gems
+                  bundle list
+                  bundle exec ./generate_report.rb \
+                     ${REPO_ROOT} \
+                     ${REPORT_FILE}
```
sgnn7: super tiny nit: the trailing lines are off-by-one

ci/submit-coverage
sgnn7: Still needs fixing

PR #1240, opened 2019-12-08
Validate that the given account exists in Conjur with authn oidc

orenbm:
When users try to authenticate with Conjur using the authn-oidc they
might give a non-existing account in the request URL. This PR adds
validation that the account exists to the beginning of the authentication
flow so that the users will get the correct reason for the error in the
log.

Before this change, the users would have seen the following message
in the log: Authentication Error: #<Errors::Conjur::RequiredResourceMissing:
CONJ00036E Missing required resource:
non-existing:variable:conjur/authn-oidc/keycloak/provider-uri>

The error above is not a good indicator the the account "non-existing"
doesn't exist in Conjur.

### DIFF
--- a/app/domain/authentication/authn_oidc/authenticate.rb
+++ b/app/domain/authentication/authn_oidc/authenticate.rb
@@ -13,6 +13,7 @@ module Authentication
         enabled_authenticators:     ENV['CONJUR_AUTHENTICATORS'],
         fetch_oidc_secrets:         AuthnOidc::Util::FetchOidcSecrets.new,
         token_factory:              TokenFactory.new,
+        validate_account_exists:    ::Authentication::Security::ValidateAccountExists.new,
         validate_security:          ::Authentication::Security::ValidateSecurity.new,
         validate_origin:            ValidateOrigin.new,
         audit_event:                AuditEvent.new,
@@ -23,6 +24,7 @@ module Authentication
     ) do
 
       def call
+        validate_account_exists
         decode_and_verify_id_token
         validate_conjur_username
         add_username_to_input
@@ -36,6 +38,12 @@ module Authentication
       end
 
       private
+
+      def validate_account_exists
+        @validate_account_exists.(
+          account: @authenticator_input.account
+        )
+      end
 
       def decode_and_verify_id_token
         @id_token_attributes = @decode_and_verify_id_token.(
--- a/cucumber/authenticators/features/authn_oidc.feature
+++ b/cucumber/authenticators/features/authn_oidc.feature
@@ -128,15 +128,13 @@ Feature: Users can authneticate with OIDC authenticator
     Errors::Authentication::RequestBody::MissingRequestParam
     """
 
-  # Should crash in GA, update the message to "account does not exists"
   Scenario: non-existing account in request is denied
-    And I fetch an ID Token for username "alice" and password "alice"
-    And I save my place in the log file
+    Given I save my place in the log file
     When I authenticate via OIDC with id token and account "non-existing"
     Then it is unauthorized
     And The following appears in the log after my savepoint:
     """
-    Errors::Conjur::RequiredResourceMissing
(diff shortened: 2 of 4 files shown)


app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -13,6 +13,7 @@ module AuthnOidc
         enabled_authenticators:     ENV['CONJUR_AUTHENTICATORS'],
         fetch_oidc_secrets:         AuthnOidc::Util::FetchOidcSecrets.new,
         token_factory:              TokenFactory.new,
+        validate_account_exists:    ::Authentication::Security::ValidateAccountExists.new,
```
sigalsax: Beyond the scope of this PR but, I see that inside the actual `validate_account_exists.rb` file, there is:
```
      def account_admin_role
        @role_class["#{@account}:user:admin"]
```
Why do we want to check the role of the account is admin? It can be any type of user no?

cucumber/authenticators/features/authn_oidc.feature
```diff
@@ -136,7 +136,7 @@ Feature: Users can authneticate with OIDC authenticator
     Then it is unauthorized
     And The following appears in the log after my savepoint:
     """
-    Errors::Conjur::RequiredResourceMissing
+    Errors::Authentication::Security::AccountNotDefined
```
sigalsax: I think you might need to update the other `Scenerios` to accept a valid account (something like "I authenticate via OIDC with id token and account "existing"". Otherwise, they are all failing on `AccountNotDefined` when that is what we want to check in other tests

app/domain/authentication/authn_oidc/authenticate.rb
orenbm: we don't check that the role of the account is an admin. We check that there is a user admin in the given account. This is the way we check that an account exists as when an account is created, an admin user is created in the DB and that's it. So an account exists if and only if there is a user `<account>:user:admin`

cucumber/authenticators/features/authn_oidc.feature
orenbm: good catch.

orenbm: but just to clarify - the file that you are referring to is an integration test. The failing tests were UTs.

nessiLahav: You do not need to fetch the ID Token in the beginning of the scenario

orenbm: cool, removed it from the test.

app/domain/authentication/authn_oidc/authenticate.rb
sigalsax: oh that's right. There needs to be at least 1 admin for an acc to exist

spec/app/domain/authentication/authn-oidc/authn_oidc_spec.rb
```diff
@@ -140,6 +143,7 @@
             enabled_authenticators: oidc_authenticator_name,
             token_factory: token_factory,
             validate_security: mocked_security_validator,
+            validate_account_exists: mocked_account_validator,
```
sigalsax: how here do we mock here that an acc has an user that is an admin?

orenbm: You can see it here: https://github.com/cyberark/conjur/blob/master/spec/support/oidc_spec_helper.rb

for each test, the `mocked_account_validator` returns true when called upon (we don't actually check that there is an account, we just mock the validation success). additionally, in [line 117](https://github.com/cyberark/conjur/blob/master/spec/support/oidc_spec_helper.rb#L117) we have a test that verifies that if the account validation fails then the authenticator fails with the same error.

sigalsax: טיל נפץ @nessiLahav

PR #1242, opened 2019-12-11
GKE coverage report is reliably generated

hughsaunders:
#### What does this PR do?
This commit removes the unnecessary `-it` flags to a kubectl exec
command for killing puma in the conjur container.

Conjur must be killed to ensure the simplecov report is written by
an at_exit hook. `-it` doesn't work in a Jenkins environment
as a tty isn't allocated for each executor.

This was bug was introduced by my recent coverage pr #1234 

Related: conjurinc/ops#470


hughsaunders:
I've ran three builds against this branch and all three were ✅, so I'm confident the issue is fixed.

PR #1252, opened 2019-12-22
Bump version to 1.4.5

orenbm:
#### What does this PR do?
Bump version to 1.4.5. This version includes the change that the k8s authenticator doesn't constraint the host to be under the "apps" policy branch, and hosts can authenticate with authn-k8s from anywhere in the policy branch.


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [1.4.5] - 2019-12-22
+
 ### Changed
 - The k8s host id does not use the "{@account}:host:conjur/authn-k8s/#{@service_name}/apps"
   prefix and takes the full host-id from the CSR. We also handle backwards-compatibility and use
```
sigalsax: Minor nitpick but "We also handle backwards-compatibility and use..." could be part of a new ### Added section because we are adding backwards compatibility

CHANGELOG.md
```diff
@@ -240,7 +242,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - The first tagged version.
 
-[Unreleased]: https://github.com/cyberark/conjur/compare/v1.4.3...HEAD
+[Unreleased]: https://github.com/cyberark/conjur/compare/v1.4.4...HEAD
```
sigalsax: shouldn't the head be `1.4.5...HEAD`? 
The 1.4.4 one was 
```
[Unreleased]: https://github.com/cyberark/conjur/compare/v1.4.4...HEAD
[1.4.4]: https://github.com/cyberark/conjur/compare/v1.4.3...v1.4.4
```
https://github.com/cyberark/conjur/pull/1251/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eL219

CHANGELOG.md
orenbm: i don't agree here. The backwards compatibility is part of the change and doesn't have enough context outside of it.

CHANGELOG.md
orenbm: correct. fixed it.

PR #1257, opened 2020-01-09
initialize platform in a separate file

orenbm:
In order to locally login to GKE/OCP to get pods, exec, etc. we want to
have the ability to do that with a built-in script.

Now to do that we can run `summon ./ci/authn-k8s/init_k8s.sh gke`.

Note: This ability was not added to the README as we don't even have a section for GKE tests. I created [an issue](https://github.com/cyberark/conjur/issues/1258) for it.


orenbm:
thanks @hughsaunders , I added documentation in the script. Is it ok?

PR #1259, opened 2020-01-09
Add GKE tests logs

orenbm:
When running the GKE tests we deploy a Conjur cluster and run "inject_client_cert" & "authenticate" requests against it. To help developers investigate failures, we print the Conjur log to the console.

The logs are wrapped in 2 lines of "===========================" so you can look for that in the console to find the beginning of the Conjur log.

### DIFF
--- a/ci/authn-k8s/test_gke_entrypoint.sh
+++ b/ci/authn-k8s/test_gke_entrypoint.sh
@@ -32,6 +32,11 @@ function finish {
       # Rails.logger writes the logs to the environment log file
       kubectl exec $pod_name -- bash -c "cat /opt/conjur-server/log/test.log" >> output/gke-authn-k8s-logs.txt
 
+      echo "Printing Logs from Conjur to the console"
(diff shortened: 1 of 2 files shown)


ci/authn-k8s/test_gke_entrypoint.sh
```diff
@@ -29,6 +29,14 @@ function finish {
       echo "Logs from Conjur Pod $pod_name:"
       kubectl logs $pod_name > output/gke-authn-k8s-logs.txt
 
+      # Rails.logger writes the logs to the environment log file
+      kubectl exec $pod_name -- bash -c "cat /opt/conjur-server/log/test.log" >> output/gke-authn-k8s-logs.txt
```
hughsaunders: Do you want a separator to show that there are two log files being combined here? Some kind of header may be useful.

orenbm: it doesn't really matter

Issue #1244, opened 2019-12-15
Hosts can authenticate with authn-k8s with their machine identity in annotations


orenbm:
At this point, hosts have their machine identity in their ID (e.g `test-namespace/service-account/test-sa`). We would like to let users configure hosts with the machine identity stored as annotations. This will:
1. Improve UX as users can give their hosts a name that identifies their purpose.
2. Let users constraint their hosts with more than 1 machine identity property. Currently you can have a constraint on one property (service account, deployment, pod name, etc.) whereas with annotations we can have more than one.

A host will be defined as follows:
```
- !host
  id: test-app
  annotations:
    kubernetes/namespace: test-namespace
    kubernetes/service_account: test-sa
    kubernetes/deployment: test-deployment
    kubernetes/authentication-container-name: test-container
```

Note that in the example above, we have a constraint on both the service account and the deployment. This is not available at this moment where the machine identity is in the host id.

DoD
- [x] Solution designed is approved by team
- [x] Hosts can authenticate with authn-k8s with their machine identity in annotations
- [x] Hosts can authenticate with authn-k8s with their machine identity in the ID
- [x] Backwards compatibility is held
  - [x] new client with an old server
  - [x] old client with a new server
- [x] Test plan is reviewed by PO and QA
- [x] Tests are implemented according to the test plan
- [x] Logs were added according to the design
- [x] Github tickets are created for documentation with instructions
- [x] Versions are bumped in all relevant projects
  - [x] conjur
    - [x] appliance
  - [x] conjur-authn-k8s-client
    - [x] secrets-provider-for-k8s
    - [x] secretless-broker

rafis3:
@orenbm, @InbalZilberman do you mind if we change the `kubernetes/` prefix to the authenticator prefix? So it would looks something like that:

```
- !host
  id: test-app
  annotations:
    authn-k8s/namespace: test-namespace
    authn-k8s/service_account: test-sa
    authn-k8s/deployment: test-deployment
    authn-k8s/authentication-container-name: test-container
```

I think it's important to keep the same terminology so our users would easily understand which authentication it belongs to. This same name is also used by the app when calling the API endpoint of that authentication.

orenbm:
2 comments:
1. Do we also need the service-id? @InbalZilberman is checking this/
2. currently we have the annotation `kubernetes/authentication-container-name`. I guess we can search for both and leave the current one working for backwards compatibility.

PR #1260, opened 2020-01-09
Allow defining application identity in host annotations

orenbm:
#### What does this PR do?
Prior to this commit, the application identity (formerly known as `machine identity`)
was defined only in the host id. This commit enables users to define the application identity
in the host's annotations.

As part of this PR, the application identity is not extracted from
the CSR (that has the host id in it) but is extracted from either the
host's annotation or the host id.

Connected to #1244

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 [quay]: https://quay.io/repository/cyberark/conjur "Conjur container image on Quay.io"
 [twitter]: https://twitter.com/intent/user?screen_name=ConjurInc "Follow Conjur on Twitter"
 
-Conjur provides secrets management and machine identity for modern infrastructure:
+Conjur provides secrets management and application identity for modern infrastructure:
 
 * **Machine Authorization Markup Language ("MAML")**, a role-based
   access policy language to define system components & their roles,
--- a/app/domain/authentication/authn_k8s/application_identity.rb
+++ b/app/domain/authentication/authn_k8s/application_identity.rb
@@ -4,7 +4,7 @@ module Authentication
     Log = LogMessages::Authentication::AuthnK8s
     Err = Errors::Authentication::AuthnK8s
     # Possible Errors Raised: MissingNamespaceConstraint, IllegalConstraintCombinations,
-    # ScopeNotSupported, ArgumentError
+    # ScopeNotSupported, InvalidHostId
 
     # This class defines an application identity of a given conjur host.
     # The constructor initializes an ApplicationIdentity object and validates that
@@ -149,10 +149,7 @@ module Authentication
         Rails.logger.debug(Log::ValidatingHostId.new(@host_id))
 
         valid_host_id = host_id_suffix.length == 3
-        unless valid_host_id
-          raise ArgumentError, "Invalid K8s host id: #{@host_id}. " \
-          "Must end with namespace/resource_type/resource_id"
-        end
+        raise Err::InvalidHostId, @host_id unless valid_host_id
 
         return if host_id_namespace_scoped?
 
--- a/app/domain/authentication/authn_k8s/application_identity.rb
+++ b/app/domain/authentication/authn_k8s/application_identity.rb
@@ -1,6 +1,7 @@
 module Authentication
   module AuthnK8s
 
+    Log = LogMessages::Authentication::AuthnK8s
     Err = Errors::Authentication::AuthnK8s
     # Possible Errors Raised: MissingNamespaceConstraint, IllegalConstraintCombinations,
     # ScopeNotSupported, ArgumentError
@@ -84,9 +85,7 @@ module Authentication
         controllers = %i(deployment deployment_config stateful_set)
 
         controller_constraints = constraints.keys & controllers
-        unless controller_constraints.length <= 1
-          raise Err::IllegalConstraintCombinations, controller_constraints
-        end
+        raise Err::IllegalConstraintCombinations, controller_constraints unless controller_constraints.length <= 1
       end
 
       def constraint_value constraint_name
@@ -100,7 +99,12 @@ module Authentication
 
       def annotation_value name
         annotation = @host_annotations.find { |a| a.values[:name] == name }
-        annotation ? annotation[:value] : nil
+
+        # return the value of the annotation if it exists, nil otherwise
+        if annotation
+          Rails.logger.debug(Log::RetrievedAnnotationValue.new(name))
+          annotation[:value]
+        end
       end
 
       def constraint_from_id constraint_name
@@ -117,6 +121,8 @@ module Authentication
       end
 
       def validate_prefixed_permitted_annotations prefix
+        Rails.logger.debug(Log::ValidatingAnnotationsWithPrefix.new(prefix))
+
         prefixed_k8s_annotations(prefix).each do |annotation|
           annotation_name = annotation[:name]
           unless prefixed_permitted_annotations(prefix).include?(annotation_name)
@@ -140,6 +146,8 @@ module Authentication
       end
 
       def validate_host_id
+        Rails.logger.debug(Log::ValidatingHostId.new(@host_id))
+
         valid_host_id = host_id_suffix.length == 3
         unless valid_host_id
           raise ArgumentError, "Invalid K8s host id: #{@host_id}. " \
--- a/app/domain/authentication/authn_k8s/application_identity.rb
+++ b/app/domain/authentication/authn_k8s/application_identity.rb
@@ -1,0 +1,175 @@
+module Authentication
+  module AuthnK8s
+
+    Err = Errors::Authentication::AuthnK8s
+    # Possible Errors Raised: MissingNamespaceConstraint, IllegalConstraintCombinations,
+    # ScopeNotSupported, ArgumentError
+
+    # This class defines an application identity of a given conjur host.
+    # The constructor initializes an ApplicationIdentity object and validates that
+    # it is configured correctly.
+    # The difference between the validation in this constructor and
+    # the validation in ValidateApplicationIdentity is that here we validate that
+    # the application identity is configured correctly, and thus is a valid application
+    # identity. In ValidateApplicationIdentity we validate that the defined application
+    # identity is actually the correct one in kubernetes.
+    # For example, an application identity `some-namepsace/blah/some-value` is not
+    # valid and will fail the validation here. However, an application identity
+    # `some-namespace/service-account/some-value` is a valid application identity
+    # and will pass the validation here. If the host is actually running from
+    # a pod with service account `some-other-value` then it will fail the
+    # validation of ValidateApplicationIdentity
+    class ApplicationIdentity
+
+      def initialize(host_id:, host_annotations:, service_id:)
+        @host_id          = host_id
+        @host_annotations = host_annotations
+        @service_id       = service_id
+
+        validate
+      end
+
+      def namespace
+        @namespace ||= application_identity_in_annotations? ? constraint_from_annotation("namespace") : host_id_suffix[-3]
+      end
+
+      def constraints
+        @constraints ||= {
+          pod:               constraint_value("pod"),
+          service_account:   constraint_value("service_account"),
+          deployment:        constraint_value("deployment"),
+          deployment_config: constraint_value("deployment_config"),
+          stateful_set:      constraint_value("stateful_set")
+        }.compact
+      end
+
+      def container_name
+        annotation_name = "authentication-container-name"
+        annotation_value("authn-k8s/#{@service_id}/#{annotation_name}") ||
+          annotation_value("authn-k8s/#{annotation_name}") ||
+          annotation_value("kubernetes/#{annotation_name}") ||
+          "authenticator"
+      end
+
+      # returns true if the only constraint is on the namespace, false otherwise
+      def namespace_scoped?
+        constraints.empty?
+      end
+
+      private
+
+      # Validates that the application identity is defined correctly
+      def validate
+        validate_permitted_scope
+
+        # validate that a constraint exists on the namespace
+        raise Err::MissingNamespaceConstraint unless namespace
+
+        validate_constraint_combinations
+      end
+
+      # If the application identity is defined in:
+      #   - annotations: validates that all the constraints are
+      #                  valid (e.g there is no "authn-k8s/blah" annotation)
+      #   - host id: validates that the host-id has 3 parts and that the given
+      #              constraint is valid (e.g the host id is not
+      #              "namespace/blah/some-value")
+      def validate_permitted_scope
+        application_identity_in_annotations? ? validate_permitted_annotations : validate_host_id
+      end
+
+      # Validates that the application identity doesn't include logical constraint
+      # combinations (e.g deployment & deploymentConfig)
+      def validate_constraint_combinations
+        controllers = %i(deployment deployment_config stateful_set)
+
+        controller_constraints = constraints.keys & controllers
+        unless controller_constraints.length <= 1
+          raise Err::IllegalConstraintCombinations, controller_constraints
+        end
+      end
+
+      def constraint_value constraint_name
+        application_identity_in_annotations? ? constraint_from_annotation(constraint_name) : constraint_from_id(constraint_name)
+      end
+
+      def constraint_from_annotation constraint_name
+        annotation_value("authn-k8s/#{@service_id}/#{constraint_name}") ||
+          annotation_value("authn-k8s/#{constraint_name}")
+      end
+
+      def annotation_value name
+        annotation = @host_annotations.find { |a| a.values[:name] == name }
+        annotation ? annotation[:value] : nil
+      end
+
+      def constraint_from_id constraint_name
+        host_id_suffix[-2] == constraint_name ? host_id_suffix[-1] : nil
+      end
+
+      def host_id_suffix
+        @host_id_suffix ||= @host_id.split('/').last(3)
+      end
+
+      def validate_permitted_annotations
+        validate_prefixed_permitted_annotations("authn-k8s/")
+        validate_prefixed_permitted_annotations("authn-k8s/#{@service_id}/")
+      end
+
+      def validate_prefixed_permitted_annotations prefix
+        prefixed_k8s_annotations(prefix).each do |annotation|
+          annotation_name = annotation[:name]
+          unless prefixed_permitted_annotations(prefix).include?(annotation_name)
+            raise Err::ScopeNotSupported, annotation_name
+          end
+        end
+      end
+
+      def prefixed_k8s_annotations prefix
+        @host_annotations.select do |a|
+          annotation_name = a.values[:name]
+
+          annotation_name.start_with?(prefix) &&
+            # Verify we take only annotations from the same level
+            annotation_name.split('/').length == prefix.split('/').length + 1
+        end
+      end
+
+      def prefixed_permitted_annotations prefix
+        permitted_annotations.map { |k| "#{prefix}#{k}" }
+      end
+
+      def validate_host_id
+        valid_host_id = host_id_suffix.length == 3
+        unless valid_host_id
+          raise ArgumentError, "Invalid K8s host id: #{@host_id}. " \
+          "Must end with namespace/resource_type/resource_id"
+        end
+
+        return if host_id_namespace_scoped?
+
+        constraint       = host_id_suffix[-2]
+        valid_constraint = permitted_constraints.include?(constraint)
+        raise Err::ScopeNotSupported, constraint unless valid_constraint
+      end
+
+      def permitted_constraints
+        @permitted_constraints ||= %w(
+          namespace service_account pod deployment stateful_set deployment_config
+        )
+      end
+
+      def permitted_annotations
+        @permitted_annotations ||= permitted_constraints << "authentication-container-name"
+      end
+
+      def application_identity_in_annotations?
+        @application_identity_in_annotations ||= @host_annotations.select { |a| a.values[:name].start_with?("authn-k8s/") }.any?
+      end
+
+      def host_id_namespace_scoped?
+        host_id_suffix[-2] == '*' && host_id_suffix[-1] == '*'
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_k8s/inject_client_cert.rb
+++ b/app/domain/authentication/authn_k8s/inject_client_cert.rb
@@ -53,8 +53,8 @@ module Authentication
 
       # In the old version of the authn-client we assumed that the host is under the "apps" policy branch.
       # Now we send the host-id in 2 parts:
-      #   suffix - the last 3 parts that indicate the machine identity (e.g ${TEST_APP_NAMESPACE_NAME}/*/*)
-      #   prefix - all the rest (e.g host/conjur/authn-k8s/${AUTHENTICATOR_ID})
+      #   suffix - the host id
+      #   prefix - the policy id
       # We update the CSR's common_name to have the full host-id. This way, the validation
       # that happens in the "authenticate" request will work, as the signed certificate
       # contains the full host-id.
@@ -99,12 +99,6 @@ module Authentication
 
       def host
         @host ||= @resource_repo[host_id]
-      end
-
-      def container_name
-        name = 'kubernetes/authentication-container-name'
-        annotation = host.annotations.find { |a| a.values[:name] == name }
-        annotation[:value] || 'authenticator'
       end
 
       def validate_csr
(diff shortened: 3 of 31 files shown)


spec/app/domain/authentication/authn_k8s/application_identity_spec.rb
```diff
@@ -0,0 +1,380 @@
...
+            expect { subject }.to raise_error(::Errors::Authentication::AuthnK8s::MissingNamespaceConstraint)
+          end
+        end
+
+        context "with a non existing resource" do
+          context "in a global constraint" do
+            let(:host_annotations) { [namespace_annotation, invalid_annotation, container_name_annotation] }
+
+            it "raises an error" do
+              expect { subject }.to raise_error(::Errors::Authentication::AuthnK8s::ScopeNotSupported)
```
nessiLahav: - Add test of annotation name with authn-k8sInvalid
- Add test of annotation name with authn-k8sInvalid/blah
- Add test of annotation name with authn-k8sInvalid/pod

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,171 @@
...
+      def constraint_from_id constraint_name
+        host_id_suffix[-2] == constraint_name ? host_id_suffix[-1] : nil
+      end
+
+      def host_id_suffix
+        @host_name_suffix ||= @host_id.split('/').last(3)
+      end
+
+      def validate_permitted_annotations
+        validate_prefixed_permitted_annotations("authn-k8s")
```
nessiLahav: please test an host with following annotation:
authn-k8s/namspace = blah
authn-k8sInvalid/namspace = blah
what is the expected behavior ?

spec/app/domain/authentication/authn_k8s/application_identity_spec.rb
orenbm: not sure where to put these tests as in all the cases above these annotations will be ingnored as they are not interpreted as part of the `authn-k8s` annotations (as expected)

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: This is the same test as the above. This annotation will be ignored.

spec/app/domain/authentication/authn_k8s/application_identity_spec.rb
orenbm: ran this test manually to verify it is ok but i won't add a UT for this.

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: Added a UT: https://github.com/cyberark/conjur/blob/host-annotations/spec/app/domain/authentication/authn_k8s/application_identity_spec.rb#L348

spec/app/domain/authentication/authn_k8s/application_identity_spec.rb
orenbm: Added a UT: https://github.com/cyberark/conjur/blob/host-annotations/spec/app/domain/authentication/authn_k8s/application_identity_spec.rb#L348

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+          stateful_set:      constraint_value("stateful_set")
+        }
+      end
+
+      def container_name
+        annotation_name = "authentication-container-name"
+        annotation_value("authn-k8s/#{@service_id}/#{annotation_name}") ||
+          annotation_value("authn-k8s/#{annotation_name}") ||
+          annotation_value("kubernetes/#{annotation_name}") ||
+          "authenticator"
```
jvanderhoof: Why is `authenticator` a valid container name?

jvanderhoof: I understand from looking at the sample policy below.  Instead of `authenticator`, how about something like `kubernetes-authenticator` to make it more explicit about what authenticator we're referencing. 

The syntax would look something like:
```
    ...
    annotations:
        kubernetes/authentication-container-name: kubernetes-authenticator
```

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+      end
+
+      def namespace_scoped?
```
jvanderhoof: Is this method used to determine if we use the old style vs new style for lookup?

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+      #   - host id: validates that the host-id has 3 parts and that the given
+      #              constraint is valid (e.g the host id is not
+      #              "namesapce/blah/some-value")
+      def validate_permitted_scope
+        application_identity_in_annotations? ? validate_permitted_annotations : validate_host_id
+      end
+
+      # This method assumes that 'validate_permitted_scope' has already run.
+      # It uses the value of 'host_id_suffix' after it is validated.
+      def validate_namespace_constraint_exists
```
jvanderhoof: Thanks for adding this comment to point out a critical dependency in flow.  Could we move the body of this function up into to `validate_permitted_scope` method to reduce the chance that a future developer will make a change which breaks the dependency?

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+      end
+
+      # Validates that the application identity doesn't include logical constraint
+      # combinations (e.g deployment & deploymentConfig)
+      def validate_constraint_combinations
+        return if permitted_constraint_combinations?
+        raise Err::IllegalConstraintCombinations
+      end
+
+      def permitted_constraint_combinations?
```
jvanderhoof: Why not raise the `Err::IllegalConstraintCombinations` message in this method rather than adding a second method (above) to perform the check?

README.md
```diff
@@ -10,7 +10,7 @@
 [quay]: https://quay.io/repository/cyberark/conjur "Conjur container image on Quay.io"
 [twitter]: https://twitter.com/intent/user?screen_name=ConjurInc "Follow Conjur on Twitter"
 
-Conjur provides secrets management and machine identity for modern infrastructure:
+Conjur provides secrets management and application identity for modern infrastructure:
```
sigalsax: changed machine -> application identity b/c machine identity will no longer be part of the host name?

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+
+    # This class defines a application identity of a given conjur host.
+    # The constructor initializes an ApplicationIdentity object and validates that
+    # it is configured correctly.
+    # The difference between the validation in this constructor and
+    # the validation in ValidateApplicationIdentity is that here we validate that
+    # the application identity is configured correctly, and thus is a valid application
+    # identity. In ValidateApplicationIdentity we validate that the defined application
+    # identity is actually the correct one in kubernetes.
+    # For example, a application identity `some-namepsace/blah/some-value` is not
```
sigalsax: NIT: "an application"

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+    # valid and will fail the validation here. However, a application identity
```
sigalsax: NIT: "an application"

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+        validate_namespace_constraint_exists
+        validate_constraint_combinations
+      end
+
+      # If the application identity is defined in:
+      #   - annotations: validates that all the constraints are
+      #                  valid (e.g there is no "authn-k8s/blah" annotation)
+      #   - host id: validates that the host-id has 3 parts and that the given
+      #              constraint is valid (e.g the host id is not
+      #              "namesapce/blah/some-value")
```
sigalsax: `"namesapce/blah/some-value"` -> "namespace/blah/some-value")

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+      # combinations (e.g deployment & deploymentConfig)
+      def validate_constraint_combinations
+        return if permitted_constraint_combinations?
+        raise Err::IllegalConstraintCombinations
+      end
+
+      def permitted_constraint_combinations?
+        # validate that we have only one constraint from the following
+        controllers = %i(deployment deployment_config stateful_set)
+
```
sigalsax: NIT: extra enter here

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+      def constraint_from_id constraint_name
+        host_id_suffix[-2] == constraint_name ? host_id_suffix[-1] : nil
+      end
+
+      def host_id_suffix
+        @host_id_suffix ||= @host_id.split('/').last(3)
+      end
+
+      def validate_permitted_annotations
+        validate_prefixed_permitted_annotations("authn-k8s/")
```
sigalsax: in terms of naming, do we want to make this more general? because I imagine we will also have `authn-oidc` etc in the future

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: this was actually there before my change. I left it for backwards-compatibility.

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: no. It is used to determine if the application identity has a namespace granularity. For example:
```
    - !host
      id: test-app-namespace
      annotations:
        authn-k8s/namespace: {{CONJUR_AUTHN_K8S_TEST_NAMESPACE}}
```
is namespace scoped while
```
    - !host
      id: test-app-service-account
      annotations:
        authn-k8s/namespace: {{CONJUR_AUTHN_K8S_TEST_NAMESPACE}}
        authn-k8s/service_account: inventory-pod-only
```
is not.

Same goes for the old style - the host `.../test-namespace/*/*` is namespace scoped while `.../test-namespace/service_account/test-sa` is not.

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: good idea, thanks.

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: I added a comment in the code:
```
      # returns true if the only constraint is on the namespace, false otherwise
      def namespace_scoped?
        constraints.compact.empty?
      end
```

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: i left only the method above and moved the body into it:
```
      # Validates that the application identity doesn't include logical constraint
      # combinations (e.g deployment & deploymentConfig)
        # validate that we have only one constraint from the following
        controllers = %i(deployment deployment_config stateful_set)

        unless (constraints.compact.keys & controllers).length <= 1
          raise Err::IllegalConstraintCombinations
        end
      end
```

README.md
orenbm: we renamed `machine identity` to `application identity` altogether

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: thanks. missed this in the `machine` -> `application` refactor.

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: this is generic as it is. the `authn-k8s/` is a parameter. 

But anyway, we will make this whole thing more generic once we know the needs from azure.

cucumber/kubernetes/features/authenticate_with_annotations.feature
```diff
@@ -0,0 +1,27 @@
+Feature: An authorized client can authenticate as a permitted role with the host's
```
sigalsax: It isn't clear to me what is "permitted role" and therefore not clear what we are testing here. Do you mean that the client has necessary Conjur permissions to authenticate and application identity is defined in its annotations?

cucumber/kubernetes/features/authenticate_with_annotations.feature
```diff
@@ -0,0 +1,27 @@
...
+
+  Scenario: Authenticate as a Deployment.
+    Given I can login to pod matching "app=inventory-deployment" to authn-k8s as "test-app-deployment" with prefix "host/some-policy"
+    Then I can authenticate pod matching "pod/inventory-deployment" with authn-k8s as "test-app-deployment" with prefix "host/some-policy"
+
+  Scenario: Authenticate as a StatefulSet.
+    Given I can login to pod matching "app=inventory-stateful" to authn-k8s as "test-app-stateful-set" with prefix "host/some-policy"
+    Then I can authenticate pod matching "pod/inventory-stateful" with authn-k8s as "test-app-stateful-set" with prefix "host/some-policy"
+
+  @k8s_skip
```
sigalsax: Not sure how familiar we expect devs to be with k8s/openshift differences so might be helpful here to add a comment that we are skipping b.c this test is for openshift only

cucumber/kubernetes/features/login_errors.feature
```diff
@@ -18,6 +18,10 @@ Feature: Errors emitted by the login method.
     When I login to authn-k8s as "@namespace@/pod/inventory-unauthorized" with prefix "host/conjur/authn-k8s/minikube/apps"
```
sigalsax: why do we have 2 `@` here? 
Ex: `@namespace@/`

cucumber/kubernetes/features/login_errors.feature
```diff
@@ -18,6 +18,10 @@ Feature: Errors emitted by the login method.
...
     Then the HTTP status is "401"
 
-  Scenario: It's an error to login as a service account host which does not match the calling pod.
+  Scenario: It's an error to login with a service account which does not match the calling pod.
```
sigalsax: "It returns/raises an error when logging in with a ..."

cucumber/kubernetes/features/step_definitions/authenticate_steps.rb
```diff
@@ -56,10 +56,14 @@ def authenticate_k8s(host, cert, key, conjur_id)
   end
 end
 
-Then(/^I( can)? authenticate pod matching "([^"]*)" with authn-k8s as "([^"]*)"( without cert and key)?$/) do |success, objectid, hostid, nocertkey|
+Then(/^I( can)? authenticate pod matching "([^"]*)" with authn-k8s as "([^"]*)"( with prefix "([^"]*)")?( without cert and key)?$/) do |success, objectid, hostid_suffix, has_custom_prefix, hostid_prefix, nocertkey|
```
sigalsax: What does `"([^"]*)")` actually mean here? Is that regex?

cucumber/kubernetes/features/login_with_annotations.feature
```diff
@@ -0,0 +1,47 @@
...
+
+  Scenario: It's an error to login from a namespace which does not match the one configured in the host
+    When I login to pod matching "app=inventory-deployment" to authn-k8s as "test-app-incorrect-namespace" with prefix "host/some-policy"
+    Then the HTTP status is "401"
+
+  Scenario: It's an error to login with a non-existing K8s resource
+    When I login to pod matching "app=inventory-pod" to authn-k8s as "test-app-non-existing-resource" with prefix "host/some-policy"
+    Then the HTTP status is "401"
+
+  Scenario: It's an error to login from a K8s resource which does not match the one configured in the host
```
sigalsax: Is the purpose of "application identity" to just allow the operation _login_ from pods that originate from a certain namespace, using a service acc, etc or does it include other k8s operations as well? Ex: with a application identity of namespace, I can only fetch secrets

app/domain/logs.rb
```diff
@@ -110,6 +110,20 @@ module AuthnK8s
           code: "CONJ00015D"
         )
 
+        RetrievedAnnotationValue = ::Util::TrackableLogMessageClass.new(
+          msg: "Retrieved value of annotation {0-annotation-name}",
```
sigalsax: should these be reviewed by a TW? that is what we did in our previous task

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -84,9 +85,7 @@ def validate_constraint_combinations
         controllers = %i(deployment deployment_config stateful_set)
 
         controller_constraints = constraints.keys & controllers
```
sigalsax: why do we defined controllers and constraints as 2 separate entities? It looks like it holds the same information

sigalsax: and what is the `&` here for?

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -100,7 +99,12 @@ def constraint_from_annotation constraint_name
 
       def annotation_value name
         annotation = @host_annotations.find { |a| a.values[:name] == name }
```
sigalsax: in terms of syntax `:name`, is to access to attribute `name`?

spec/app/domain/authentication/authn_k8s/application_identity_spec.rb
```diff
@@ -0,0 +1,511 @@
...
+    allow(container_name_annotation_service_id_prefix).to receive(:values)
+                                                            .and_return(container_name_annotation_service_id_prefix)
+    allow(container_name_annotation_service_id_prefix).to receive(:[])
+                                                            .with(:name)
+                                                            .and_return("authn-k8s/#{good_service_id}/authentication-container-name")
+    allow(container_name_annotation_service_id_prefix).to receive(:[])
+                                                            .with(:value)
+                                                            .and_return("ServiceIdContainerName")
+
+    allow(container_name_annotation_kubernetes_prefix).to receive(:values)
```
sigalsax: I remember we discussed this before but what are we doing here with allow() to do something? We are specifying what parameters it can accept and once it does, return the relevant data values defined in `and_return`?

spec/app/domain/authentication/authn_k8s/application_identity_spec.rb
```diff
@@ -0,0 +1,511 @@
...
+        context "when is namespace scoped" do
+          let(:k8s_resource_name) { "*" }
+          let(:k8s_resource_value) { "*" }
+
+          it "does not raise an error" do
+            expect { subject }.not_to raise_error
+          end
+        end
+
+        context "when is not namespace scoped" do
```
sigalsax: shouldn't it be the inverse? when it is service account, then it does not raise an error? or are we specifying hierarchy that if we are not namespace scoped, then we can do operations when we use a service_account?

app/domain/authentication/authn_k8s/validate_application_identity.rb
```diff
@@ -0,0 +1,119 @@
...
+      def call
+        validate_namespace
+        validate_pod_properties
+        validate_container
+      end
+
+      private
+
+      def validate_namespace
+        return if application_identity.namespace == @spiffe_id.namespace
```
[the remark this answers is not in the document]
sigalsax: from our convo, the spiffe id holds extra pod information for validation purposes

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -101,18 +101,8 @@ def host
         @host ||= @resource_repo[host_id]
```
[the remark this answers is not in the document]
orenbm: refactored to `resource_class`

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -157,6 +147,23 @@ def cert_to_install
...
       end
+
+      # This code is implemented similarly also in application_identity.rb
+      # We have it here too as we need the container name for the injection
+      # and it simplifies the code to have this specific 2 methods duplicated
+      # rather than passing around the ApplicationIdentity object
+      def container_name
+        container_annotation_value("authn-k8s/#{@service_id}") ||
+          container_annotation_value("authn-k8s") ||
+          container_annotation_value("kubernetes") ||
```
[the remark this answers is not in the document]
orenbm: no, and I answered this question here: https://github.com/cyberark/conjur/pull/1260#discussion_r367306609

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -0,0 +1,181 @@
...
+      def permitted_annotations
+        @permitted_annotations ||= permitted_constraints << "authentication-container-name"
+      end
+
+      def application_identity_in_annotations?
+        @application_identity_in_annotations ||= @host_annotations.select { |a| a.values[:name].start_with?("authn-k8s/") }.any?
+      end
+
+      def host_id_namespace_scoped?
+        host_id_suffix[-2] == '*' && host_id_suffix[-1] == '*'
```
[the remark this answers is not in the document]
orenbm: no. `host_id_suffix` is an array of strings which is separated by slashes.

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -157,6 +147,23 @@ def cert_to_install
...
+          "authenticator"
+      end
+
+      def container_annotation_value prefix
```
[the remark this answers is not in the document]
orenbm: no, this is a function that gets different prefixes. Please see usage here: https://github.com/cyberark/conjur/blob/host-annotations/app/domain/authentication/authn_k8s/inject_client_cert.rb#L155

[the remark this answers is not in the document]
sigalsax: yeah didnt see that function- good point

app/domain/logs.rb
sigalsax: might be a bit overkill now that I think about it b/c it's only a few of them

cucumber/kubernetes/features/authenticate_with_annotations.feature
orenbm: fixed

cucumber/kubernetes/features/authenticate_with_annotations.feature
orenbm: the tag `@k8s_skip` tells the story

cucumber/kubernetes/features/login_errors.feature
orenbm: this is a placeholder that is used for text replacements - https://github.com/cyberark/conjur/blob/host-annotations/cucumber/kubernetes/features/support/world.rb#L24

cucumber/kubernetes/features/step_definitions/authenticate_steps.rb
orenbm: yes

cucumber/kubernetes/features/login_with_annotations.feature
orenbm: only for authentication

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: controllers is a hard-coded array of the controllers that can't be combined. `constraints.keys` is the actual constraints in the application identity. 

`constraints.keys & controllers` is the intersection of the 2. So we validate that the intersection of the 2 arrays have at most one value.

spec/app/domain/authentication/authn_k8s/application_identity_spec.rb
orenbm: we have a separation between cases that the application identity is namespace scoped and cases that the application identity is *not* namespace scoped. the service_account is just an example of a resource that is defined in the app id along with the namespace

app/domain/authentication/authn_k8s/application_identity.rb
sigalsax: nice!

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -157,6 +151,23 @@ def cert_to_install
...
       end
+
+      # This code is implemented similarly also in application_identity.rb
+      # We have it here too as we need the container name for the injection
+      # and it simplifies the code to have this specific 2 methods duplicated
+      # rather than passing around the ApplicationIdentity object
+      def container_name
+        container_annotation_value("authn-k8s/#{@service_id}") ||
+          container_annotation_value("authn-k8s") ||
+          container_annotation_value("kubernetes") ||
```
sigalsax: what is the difference here from 
```
          container_annotation_value("authn-k8s") ||
          container_annotation_value("kubernetes") ||
```
I know you are referencing the application_identity.rb file but please explain to me why we also need `container_annotation_value("kubernetes")` (:

orenbm: it is for backwards compatibility as before this PR the container name was in the `kubernetes/authentication-container-name` annotation.

spec/app/domain/authentication/authn_k8s/validate_application_identity_spec.rb
```diff
@@ -197,7 +197,7 @@
         allow(host_annotation_2).to receive(:values)
                                       .and_return({ :name => "notimportant" })
 
-        expected_message = /Container authenticator was not found for requesting pod/
+        expected_message = /Container authenticator was not found in the pod/
```
sigalsax: what is this syntax? "/" instead of a string?

orenbm: regex

sigalsax: cool!

PR #1267, opened 2020-01-20
Add ability to test log output in GKE

orenbm:
#### What does this PR do?
In order to test log output in our GKE tests we first need to:
1. Create a PersistentVolume & a PersistentVolumeClaim so we have
   a shared volume between the Conjur & Cucumber deployments. This
   enables us to run tests in cucumber that read the Conjur log.
2. Add the logs helper files to the docker image so they are present
   in the cucumber container.

### DIFF
--- a/ci/authn-k8s/dev/dev_conjur_log_volume.template.yaml
+++ b/ci/authn-k8s/dev/dev_conjur_log_volume.template.yaml
@@ -1,0 +1,27 @@
+---
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: {{ LOG_VOLUME_NAME }}
+  labels:
+    type: local
+spec:
+  storageClassName: manual
+  capacity:
+    storage: 1Gi
+  accessModes:
+    - ReadWriteMany
+  hostPath:
+    path: "/tmp/conjur-log"
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: conjur-log-pv-claim
+spec:
+  storageClassName: manual
+  accessModes:
+    - ReadWriteMany
+  resources:
+    requests:
+      storage: 1Gi
--- a/ci/authn-k8s/test_gke_entrypoint.sh
+++ b/ci/authn-k8s/test_gke_entrypoint.sh
@@ -54,6 +54,10 @@ function finish {
   echo '-----'
   kubectl --ignore-not-found=true delete namespace $CONJUR_AUTHN_K8S_TEST_NAMESPACE
 
+  echo 'Removing Conjur log PersistentVolume'
+  echo '-----'
+  kubectl delete --ignore-not-found=true pv $LOG_VOLUME_NAME &
+
   delete_image $CONJUR_TEST_AUTHN_K8S_TAG
   delete_image $CONJUR_AUTHN_K8S_TAG
   delete_image $INVENTORY_TAG
@@ -73,6 +77,7 @@ function main() {
 
   pushDockerImages
 
+  createLogPersistentVolume
   launchConjurMaster
   
   copyNginxSSLCert
@@ -117,6 +122,17 @@ function pushDockerImages() {
   gcloud docker -- push $CONJUR_TEST_AUTHN_K8S_TAG
   gcloud docker -- push $INVENTORY_TAG
   gcloud docker -- push $NGINX_TAG
+}
+
+# We create a PersistentVolume so cucumber tests can read Conjur logs to test the output
+function createLogPersistentVolume() {
+  echo 'Creating PersistentVolume for Conjur log'
+
+  LOG_VOLUME_NAME="conjur-log-volume"
+  kubectl delete --ignore-not-found=true pv $LOG_VOLUME_NAME
+
+  sed -e "s#{{ LOG_VOLUME_NAME }}#$LOG_VOLUME_NAME#g" dev/dev_conjur_log_volume.${TEMPLATE_TAG}yaml |
+    kubectl create -f -
 }
 
 function launchConjurMaster() {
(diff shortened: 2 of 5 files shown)


rafis3:
Hi @orenbm,
I like this solution. It's really an easy win that keeps things relatively uncomplicated. However, there are two concerns that come to my mind with this approach:
1. This solution doesn't deal with a cluster of multiple nodes. The `hostPath` relies on the fact that the two Deployments run on the same node, to access the same filesystem. If we have one node anyway, there shouldn't be a problem. But if we run our Pods on multiple nodes, it would randomly fail. You can solve this with the `nodeSelector` parameter. For more information, see this: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
You can add to your script a command that would check the node of your Conjur, after its deployed and then make sure with the parameter above, that cucumber runs on the same node.
1. What will happen if multiple tests run at the same time? Since the `hostPath` is generic, won't the two instances of Conjur override each other?

Thanks!

ci/authn-k8s/dev/dev_conjur_log_volume.template.yaml
```diff
@@ -0,0 +1,27 @@
+---
+apiVersion: v1
+kind: PersistentVolume
```
hughsaunders: Looking at the GKE docs, you probably don't need to create the PV explicitly. Create the PVC and the PV will be provisoned by GKE: https://cloud.google.com/kubernetes-engine/docs/concepts/persistent-volumes#dynamic_provisioning

ci/authn-k8s/test_gke_entrypoint.sh
```diff
@@ -54,6 +54,10 @@ function finish {
   echo '-----'
   kubectl --ignore-not-found=true delete namespace $CONJUR_AUTHN_K8S_TEST_NAMESPACE
 
+  echo 'Removing Conjur log PersistentVolume'
+  echo '-----'
+  kubectl delete --ignore-not-found=true pv $LOG_VOLUME_NAME &
```
hughsaunders: I'm not sure if the PVC will be removed automatically with the pods, please can you check and add a remove line for the PVC if required?

ci/authn-k8s/dev/dev_conjur_log_volume.template.yaml
rafis3: @hughsaunders do we want to make this solution generic? So it would work the same in our OpenShift environments as well?

ci/authn-k8s/test_gke_entrypoint.sh
orenbm: the PVCs are removed when we remove the namespace

orenbm:
> What will happen if multiple tests run at the same time? Since the hostPath is generic, won't the two instances of Conjur override each other?

@rafis3 thanks for this comment. I added the namespace to the end of the path so it will be unique: `path: "/tmp/conjur-log-{{ CONJUR_AUTHN_K8S_TEST_NAMESPACE }}"`

ci/authn-k8s/dev/dev_conjur_log_volume.template.yaml
orenbm: I agree with Rafi here. It seems that dynamic provisioning will work only for GKE.

orenbm:
> This solution doesn't deal with a cluster of multiple nodes. The hostPath relies on the fact that the two Deployments run on the same node, to access the same filesystem. If we have one node anyway, there shouldn't be a problem. But if we run our Pods on multiple nodes, it would randomly fail. You can solve this with the nodeSelector parameter. For more information, see this: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/
You can add to your script a command that would check the node of your Conjur, after its deployed and then make sure with the parameter above, that cucumber runs on the same node.

@rafis3 i'm not sure that this is the correct approach. Labelling and unlabelling the nodes in each test run can be quite troublesome. If I understand correctly, I can set the PV & PVC `accessMode` to `ReadWriteMany` which means that the Volume can be mounted as read-write by many nodes.

hughsaunders:
@orenbm You are right about the lack of dynamic volume provisioners configured for our openshift clusters :(

OC3.11:
```
$kubectl get storageclass
No resources found.
```

GKE:
```
$ kubectl get storageclass
NAME                 PROVISIONER            AGE
standard (default)   kubernetes.io/gce-pd   705d
```

I've created https://app.zenhub.com/workspaces/infrastructure-5cd9b1bae995480d64c12f0d/issues/conjurinc/ops/485 to address the lack of dynamic storage in our OC clusters. 

However I'm not sure that changing the mode on a hostPath volume will solve your problem, as hostPaths are not exported across the network, so if two pods that need access to the volume land on different nodes either one won't be able to access the volume, or they'll both be accessing the filesystem on the individual node and won't be able to share data.

ci/authn-k8s/dev/dev_conjur_log_volume.template.yaml
hughsaunders: This is a bug with our openshift deployment, but I do agree that we should use a method that works for both.

ci/authn-k8s/test_gke_entrypoint.sh
hughsaunders: Ah yes thanks.

ci/authn-k8s/dev/dev_conjur_log_volume.template.yaml
orenbm: i get a page not found for the link above. So @hughsaunders do you still think we should use dynamic provisioning?

orenbm:
@hughsaunders thanks for the explanation. Do you think that dynamic provisioning will solve the node issue?

orenbm:
Closing this PR as the solution in it will not help us with a multi-node environment.

PR #1268, opened 2020-01-20
Read log file of the current environment

orenbm:
Before this commit we always read the `development.log` file.
In order to read the file when running in test env we need to
extract the filename from the env.


orenbm:
Closing this PR as we will not investigate in this in the near future.

PR #1270, opened 2020-01-21
Bump version to 1.4.6

orenbm:
This version includes the change:
K8s hosts' application identity is extracted from annotations or id. If it is
  defined in annotations it will taken from there and if not, it will be taken
  from the id.


orenbm:
@alexkalish FYI

PR #1271, opened 2020-01-22
Root-based hosts can authenticate with k8s

orenbm:
#### What does this PR do?
Fixes an issue where hosts that are defined in the root policy were not
able to authenticate with k8s.

This PR also adds tests for such hosts.

### DIFF
--- a/app/domain/authentication/authn_k8s/application_identity.rb
+++ b/app/domain/authentication/authn_k8s/application_identity.rb
@@ -112,7 +112,13 @@ module Authentication
       end
 
       def host_id_suffix
-        @host_id_suffix ||= @host_id.split('/').last(3)
+        @host_id_suffix ||= hostname.split('/').last(3)
+      end
+
+      # Return the last part of the host id (which is the actual hostname).
+      # The host id is build as "account_name:kind:identifier" (e.g "org:host:some_hostname").
+      def hostname
+        @hostname ||= @host_id.split(':')[2]
       end
 
       def validate_permitted_annotations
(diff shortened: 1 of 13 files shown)


app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -112,7 +112,9 @@ def constraint_from_id constraint_name
       end
 
       def host_id_suffix
-        @host_id_suffix ||= @host_id.split('/').last(3)
+        # We first split the host id by ':' and take the last part to remove the '<account>:host:' prefix.
+        # Leaving it in a one-liner for memoization.
+        @host_id_suffix ||= (@host_id.split(':')[2]).split('/').last(3)
```
Tovli: Consider extracting a method. ExtractSuffixFromHostId ?
Can we have cases where hostId does not include ':' ?

orenbm:
@shuligh can you review please the log messages in `logs.rb`?

spec/app/domain/authentication/authn_k8s/application_identity_spec.rb
```diff
@@ -136,7 +136,7 @@
 
     context "Application identity in host id" do
       let(:host_annotations) { [] }
-      let(:host_id) { "#{namespace}/#{k8s_resource_name}/#{k8s_resource_value}" }
+      let(:host_id) { "accountName:host:#{namespace}/#{k8s_resource_name}/#{k8s_resource_value}" }
```
sigalsax: I suggest we factor this out like we have for namespace, resource_name, and resource_value for the sake of uniformity

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: created a new method

orenbm: The host id will always include `account:kind:identifier`. It is built here: `host_id = "#{@account}:" + @common_name.k8s_host_name.sub('host/', 'host:')`

PR #1272, opened 2020-01-22
Improving flows and rules around user creation

nahumcohen:
What does this PR do?
Improves flows and rules around user creation.

Install appliance once this is merged to master.

The version has not the Changelog has updated.

yes
no
no
no

### DIFF
--- a/app/controllers/credentials_controller.rb
+++ b/app/controllers/credentials_controller.rb
@@ -35,9 +35,9 @@ class CredentialsController < ApplicationController
   # This method requires a PUT request. The new password is in the request body.
   def update_password
     password = request.body.read
-    
+
     @role.credentials.password = password
-    @role.credentials.save
+    @role.credentials.save(raise_on_save_failure: true)
     head 204
   end
   
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -19,6 +19,12 @@ unless defined? Errors::Authentication::AuthenticatorNotFound
         code: "CONJ00037E"
       )
 
+      InsufficientPasswordComplexity = Util::TrackableErrorClass.new(
+        msg: "The password you have chosen does not meet the complexity requirements. " \
+             "Choose a password that includes: 12-128 characters, 2 uppercase letters, 2 lowercase letters, 1 digit, 1 special character",
+        code: "CONJ00046E"
+      )
+
     end
 
     module Authentication
--- a/app/models/credentials.rb
+++ b/app/models/credentials.rb
@@ -7,6 +7,13 @@ class Credentials < Sequel::Model
   # Bcrypt work factor, minimum recommended work factor is 12
   BCRYPT_COST = 12
 
+  # special characters according to https://www.owasp.org/index.php/Password_special_characters
+  VALID_PASSWORD_REGEX = %r{^(?=.*?[A-Z].*[A-Z])                             # 2 uppercase letters
+                             (?=.*?[a-z].*[a-z])                             # 2 lowercase letters
+                             (?=.*?[0-9])                                    # 1 digit
+                             (?=.*[ !"#$%&'()*+,-.\/:;<=>?@\[\\\]^_`{|}~]).  # 1 special character
+                             {12,128}$}x                                     # 12-128 characters
+
   plugin :validation_helpers
 
   unrestrict_primary_key
@@ -23,7 +30,7 @@ class Credentials < Sequel::Model
       Slosilo::Random.salt.unpack("N*").map{|i| Base32::Crockford::encode(i)}.join.downcase
     end
   end
-  
+
   def as_json
     { }
   end
@@ -63,8 +70,7 @@ class Credentials < Sequel::Model
 
     validates_presence [ :api_key ]
 
-    errors.add(:password, 'must not be blank') if @plain_password && @plain_password.empty?
-    errors.add(:password, 'cannot contain a newline') if @plain_password && @plain_password.index("\n")
+    errors.add(:password, ::Errors::Conjur::InsufficientPasswordComplexity.new.to_s) if @plain_password && (@plain_password.index("\n") || @plain_password !~ VALID_PASSWORD_REGEX)
   end
   
   def before_validation
--- a/ci/docker-compose.yml
+++ b/ci/docker-compose.yml
@@ -13,7 +13,7 @@ services:
     image: "conjur:${TAG}"
     environment:
       DATABASE_URL: postgres://postgres@pg/postgres
-      CONJUR_ADMIN_PASSWORD: admin
+      CONJUR_ADMIN_PASSWORD: ADmin123!!!!
       CONJUR_ACCOUNT: cucumber
       CONJUR_DATA_KEY:
       RAILS_ENV:
--- a/cucumber/api/features/change_password.feature
+++ b/cucumber/api/features/change_password.feature
@@ -10,14 +10,26 @@ Feature: Change the password of a role
 
   Scenario: With basic authentication, users can update their own password using the current password.
 
-    Given I set the password for "alice" to "my-password"
-    When I successfully PUT "/authn/cucumber/password" with username "alice" and password "my-password" and plain text body "new-password"
-    Then I can GET "/authn/cucumber/login" with username "alice" and password "new-password"
+    Given I set the password for "alice" to "My-Password1"
+    When I successfully PUT "/authn/cucumber/password" with username "alice" and password "My-Password1" and plain text body "New-Password1"
+    Then I can GET "/authn/cucumber/login" with username "alice" and password "New-Password1"
 
   Scenario: With basic authentication, users can update their own password using the current API key.
 
-    When I successfully PUT "/authn/cucumber/password" with username "alice" and password ":alice_api_key" and plain text body "new-password"
-    Then I can GET "/authn/cucumber/login" with username "alice" and password "new-password"
+    When I successfully PUT "/authn/cucumber/password" with username "alice" and password ":alice_api_key" and plain text body "New-Password1"
+    Then I can GET "/authn/cucumber/login" with username "alice" and password "New-Password1"
+
+  Scenario: Succeed to set password with escaped special character applying to password complexity.
+
+    Given I set the password for "alice" to "My-Password1"
+    When I successfully PUT "/authn/cucumber/password" with username "alice" and password "My-Password1" and plain text body "NewPassword1/"
+    Then I can GET "/authn/cucumber/login" with username "alice" and password "NewPassword1/"
+
+  Scenario: Fail to set password with 11 characters not applying to password complexity.
+
+    Given I set the password for "alice" to "My-Password1"
+    When I PUT "/authn/cucumber/password" with username "alice" and password "My-Password1" and plain text body "My-Passwor1"
+    Then the HTTP response status code is 422
 
   Scenario: Bearer token cannot be used to change the password
 
--- a/cucumber/api/features/login.feature
+++ b/cucumber/api/features/login.feature
@@ -7,8 +7,8 @@ Feature: Exchange a role's password for its API key
     Given I create a new user "alice"
 
   Scenario: Password can be used to obtain API key
-    Given I set the password for "alice" to "my-password"
-    Then I can GET "/authn/cucumber/login" with username "alice" and password "my-password"
+    Given I set the password for "alice" to "My-Password1"
+    When I can GET "/authn/cucumber/login" with username "alice" and password "My-Password1"
     Then the result is the API key for user "alice"
 
   @logged-in
--- a/cucumber/api/features/rotate_api_key.feature
+++ b/cucumber/api/features/rotate_api_key.feature
@@ -9,8 +9,8 @@ Feature: Rotate the API key of a role
     Given I create a new user "alice"
 
   Scenario: Password can be used to rotate API key
-    Given I set the password for "alice" to "my-password"
-    Then I can PUT "/authn/cucumber/api_key" with username "alice" and password "my-password"
+    Given I set the password for "alice" to "My-Password1"
+    When I can PUT "/authn/cucumber/api_key" with username "alice" and password "My-Password1"
     Then the result is the API key for user "alice"
 
   @logged-in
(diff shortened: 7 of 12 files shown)


cucumber/api/features/change_password.feature
```diff
@@ -10,14 +10,43 @@ Feature: Change the password of a role
...
+    Then I can GET "/authn/cucumber/login" with username "alice" and password "New-Password1"
 
   Scenario: With basic authentication, users can update their own password using the current API key.
 
-    When I successfully PUT "/authn/cucumber/password" with username "alice" and password ":alice_api_key" and plain text body "new-password"
-    Then I can GET "/authn/cucumber/login" with username "alice" and password "new-password"
+    When I successfully PUT "/authn/cucumber/password" with username "alice" and password ":alice_api_key" and plain text body "New-Password1"
+    Then I can GET "/authn/cucumber/login" with username "alice" and password "New-Password1"
+
+  Scenario: Succeed to set password applying to password complexity.
```
orenbm: I believe this should be split into 3 scenarios

cucumber/api/features/change_password.feature
```diff
@@ -10,14 +10,43 @@ Feature: Change the password of a role
...
+    Given I set the password for "alice" to "My-Password1"
+    When I successfully PUT "/authn/cucumber/password" with username "alice" and password "My-Password1" and plain text body "NewPassword1/"
+    Then I can GET "/authn/cucumber/login" with username "alice" and password "NewPassword1/"
+
+    # with mixed languages (Russian and French)
+    Given I set the password for "alice" to "My-Password1"
+    When I successfully PUT "/authn/cucumber/password" with username "alice" and password "My-Password1" and plain text body "New!Password1БбИиàâæ"
+    Then I can GET "/authn/cucumber/login" with username "alice" and password "New!Password1БбИиàâæ"
+
+  Scenario: Fail to set password not applying to password complexity.
```
orenbm: Same here. each set of `Given-When-Then` should have its own scenario

cucumber/api/features/rotate_api_key.feature
```diff
@@ -9,8 +9,8 @@ Feature: Rotate the API key of a role
     Given I create a new user "alice"
 
   Scenario: Password can be used to rotate API key
-    Given I set the password for "alice" to "my-password"
-    Then I can PUT "/authn/cucumber/api_key" with username "alice" and password "my-password"
+    Given I set the password for "alice" to "My-Password1"
+    Then I can PUT "/authn/cucumber/api_key" with username "alice" and password "My-Password1"
     Then the result is the API key for user "alice"
```
orenbm: change the `And`

cucumber/api/features/login.feature
```diff
@@ -7,8 +7,8 @@ Feature: Exchange a role's password for its API key
     Given I create a new user "alice"
 
   Scenario: Password can be used to obtain API key
-    Given I set the password for "alice" to "my-password"
-    Then I can GET "/authn/cucumber/login" with username "alice" and password "my-password"
+    Given I set the password for "alice" to "My-Password1"
+    Then I can GET "/authn/cucumber/login" with username "alice" and password "My-Password1"
     Then the result is the API key for user "alice"
```
orenbm: Change to `And`

lib/util/password.rb
```diff
@@ -0,0 +1,13 @@
+module Util
+  # provides helper methods for validating password complexity
+  class Password
+    def self.validate_complexity(password)
+      # special characters according to https://www.owasp.org/index.php/Password_special_characters
+      #                    (2 uppercase letters)                 (  1 digit )                                              (12-128 characters)
+      #                                       (2 lowercase letters)          (          1 special character              )
+      unless password =~ /^(?=.*?[A-Z].*[A-Z])(?=.*?[a-z].*[a-z])(?=.*?[0-9])(?=.*[ !"#$%&'()*+,-.\/:;<=>?@\[\\\]^_`{|}~]).{12,128}$/
```
orenbm: can we move the regex to a const member of the class? This way it will not be calculated each time and this method will be more readable.

orenbm: If you do so, you can change this to a one-liner (raise error unless condition)

spec/controllers/credentials_controller_spec.rb
```diff
@@ -2,6 +2,8 @@
 
 require 'spec_helper'
```
orenbm: remove this line

spec/util_spec.rb
```diff
@@ -2,7 +2,81 @@
 
 require 'spec_helper'
```
orenbm: remove this line

app/models/role.rb
```diff
@@ -68,6 +68,7 @@ def by_login login, account:
   end
 
   def password= password
+    Util::Password.validate_complexity password
```
jvanderhoof: Validations like this should be added as a validation: https://guides.rubyonrails.org/active_record_validations.html. Let's refactor and move the logic down to the model responsible for the password.

app/domain/errors.rb
```diff
@@ -19,6 +19,12 @@ module Conjur
         code: "CONJ00037E"
       )
 
+      PasswordComplexityBreak = Util::TrackableErrorClass.new(
```
micahlee: Could we change the name of this error type to something like:

```suggestion
      InsufficientPasswordComplexity = Util::TrackableErrorClass.new(
```

Or even just `PasswordComplexityError`. The `Break` part is a little odd.

app/controllers/credentials_controller.rb
```diff
@@ -35,10 +35,16 @@ class CredentialsController < ApplicationController
   # This method requires a PUT request. The new password is in the request body.
   def update_password
     password = request.body.read
-    
-    @role.credentials.password = password
-    @role.credentials.save
-    head 204
+
+    begin
+      Util::Password.validate_complexity password
```
micahlee: If this is occurring in `password=`, we should not be duplicating it here.

micahlee: Although, Jason's suggestion to move this into the ActiveModel validation is even better.

micahlee:
Thanks @nahumcohen! This is looking good, just a couple comments for you. 

The Code Climate issues also look relatively easy to address (syntax formatting issues). Would you mind taking care of those as well?

Thanks!

jvanderhoof:
@nahumcohen, once we get the validation moved into the `Role` class, let's get a couple of unit test in place to validate functionality.

app/models/role.rb
micahlee: Just clarifying here that this should be a `sequel` validation, rather than active record:
- http://sequel.jeremyevans.net/rdoc/files/doc/validations_rdoc.html
- http://guides.rubyonrails.org/active_record_validations.html

cucumber/api/features/change_password.feature
nahumcohen: Remove tests covered in UT and left only one in integration.

cucumber/api/features/change_password.feature
nahumcohen: Remove tests covered in UT and left only one in integration.

lib/util/password.rb
nahumcohen: moved to credentials and done

micahlee:
@nahumcohen, thanks for making these updates! This is looking good.

These last code climate issues look like they're worth fixing, and then I think this should be good to go!

app/controllers/credentials_controller.rb
```diff
@@ -35,10 +35,18 @@ class CredentialsController < ApplicationController
...
-    @role.credentials.password = password
-    @role.credentials.save
-    head 204
+
+    begin
+      @role.credentials.password = password
+      @role.credentials.save(raise_on_save_failure: true)
+      head 204
+    rescue
+      # render json with unprocessable_entity only for insufficient_password_complexity otherwise exception is thrown up
```
uCatu: Suggestion, change "thrown up" to "otherwise, throw an exception"

app/controllers/credentials_controller.rb
```diff
@@ -35,10 +35,18 @@ class CredentialsController < ApplicationController
...
-    
-    @role.credentials.password = password
-    @role.credentials.save
-    head 204
+
+    begin
+      @role.credentials.password = password
+      @role.credentials.save(raise_on_save_failure: true)
+      head 204
+    rescue
```
uCatu: I would love to see rescue for a specific error that is thrown during save action instead of searching for message string.

app/controllers/credentials_controller.rb
uCatu: I don't see any "rerase" after we rescue from catch all, are we sure that we are not swallowing other errors here?
 
e.g: https://stackoverflow.com/questions/23770924/reraise-same-exception-after-catching-an-exception-in-ruby

app/domain/errors.rb
```diff
@@ -19,6 +19,12 @@ module Conjur
         code: "CONJ00037E"
       )
 
+      InsufficientPasswordComplexity = Util::TrackableErrorClass.new(
+        msg: "The password you have chosen does not meet the complexity requirements. " \
+             "Choose a password that includes: 12-128 characters, 2 uppercase letters, 2 lowercase letters, 1 digit, 1 special character",
+        code: "CONJ00046E"
```
uCatu: Is the code documeneted somewhere?

app/models/credentials.rb
```diff
@@ -63,8 +70,7 @@ def validate
 
     validates_presence [ :api_key ]
 
-    errors.add(:password, 'must not be blank') if @plain_password && @plain_password.empty?
```
uCatu: Are we relaying on "must no be blank" or "cannot contain a newline" somewhere? just like we are now with ":insufficient_password_complexity"

ci/docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
     image: "conjur:${TAG}"
     environment:
       DATABASE_URL: postgres://postgres@pg/postgres
-      CONJUR_ADMIN_PASSWORD: admin
+      CONJUR_ADMIN_PASSWORD: ADmin123!!!!
```
uCatu: Can we check if we inject environment variable from Jenkins or other places that will override new compilent password with other one?

dev/docker-compose.yml
```diff
@@ -50,7 +50,7 @@ services:
       LDAP_BASE: dc=conjur,dc=net
       CONJUR_APPLIANCE_URL: http://conjur:3000
       DATABASE_URL: postgres://postgres@pg/postgres
-      CONJUR_ADMIN_PASSWORD: admin
+      CONJUR_ADMIN_PASSWORD: ADmin123!!!!
```
uCatu: Same comment as injection of environment variable

spec/controllers/credentials_controller_spec.rb
```diff
@@ -92,17 +82,7 @@
...
-                message: "password cannot contain a newline",
-                details: [
-                  {
-                    code: "validation_failed",
-                    target: "password",
-                    message: "cannot contain a newline"
-                  }
-                ]
-              }
+              Reason: Errors::Conjur::InsufficientPasswordComplexity.new.to_s
```
uCatu: You have two tests that are now redundent - empty password and one containing new line, since we consolidated those errors to one InsufficientPasswordComplexity, yielding same error message that don't need to be seperated.
However! there is something we need to check - why is new line is not escaped? are there any other issues for example: `<script>123!@#Test` and are we now enabline \n in password? I think that `Test\nWillWork123' and thus allowing passwords that we didn't allow before, maybe breaking UI also.

spec/models/credentials_spec.rb
```diff
@@ -57,10 +57,82 @@
...
+        # with all acceptable special characters
+        credentials.password = "New-Password1 !\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~]"
+        expect { credentials.save }.not_to raise_error
+      end
+      it 'allows passwords with mixed languages (Russian and French)' do
+        # with mixed languages (Russian and French)
+        credentials.password = "New!Password1БбИиàâæ"
+        expect { credentials.save }.not_to raise_error
+      end
       it 'disallows passwords with newlines' do
```
uCatu: not corrent. we don't check that `\n` is not allowed since this use case `foo\nbar` is not answering required complexity.

spec/models/credentials_spec.rb
```diff
@@ -57,10 +57,82 @@
...
+      it 'disallows passwords with 129 characters long' do
+        # 129 characters long (Requires: At least 12 characters and maximum of 128)
+        credentials.password = "My-Password1345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789"
+        expect { credentials.save }.to raise_error(Sequel::ValidationFailed) do |e|
+          expect(e.errors.to_h).to eq({ password: [ :insufficient_password_complexity ]})
+        end
+      end
+      it 'disallows passwords with only 1 upper case character' do
+        # only 1 upper case character (Requires: At least 2 upper case characters)
+        credentials.password = "My-assword12"
```
uCatu: asswrd 👎

app/models/credentials.rb
```diff
@@ -63,8 +70,7 @@ def validate
...
-    errors.add(:password, 'cannot contain a newline') if @plain_password && @plain_password.index("\n")
+    errors.add(:password, ::Errors::Conjur::InsufficientPasswordComplexity.new.to_s) if @plain_password && (@plain_password.index("\n") || @plain_password !~ VALID_PASSWORD_REGEX)
```
jvanderhoof: Let's  use the Sequel Gem validation: https://github.com/jeremyevans/sequel/blob/master/doc/validations.rdoc#validates_format
It would be something like:
```
validates_format VALID_PASSWORD_REGEX, : plain_password
```

nahumcohen: The scope of `validate` method includes both `api_key` and `plain_password`. If we add `validate_format` it will not separate between the two use-cases. 

See existing password validation logic is to verify existence of `plain_password` parameter, and only after, verify complexity, blank & new line.

We also wanted to utilize `validates_presence`, but neglect the idea for the same reason above.

PR #1273, opened 2020-01-22
Increase private key size to 2048 bits

orenbm:
#### What does this PR do?
We had an issue that GKE tests raised an `SSL_CTX_use_certificate: ee key too small (OpenSSL::SSL::SSLError)` error when using the private key used in the `inject_client_cert` request. This PR increases the private key size to 2048 bits to fir the requirements.

### DIFF
--- a/cucumber/kubernetes/features/step_definitions/login_steps.rb
+++ b/cucumber/kubernetes/features/step_definitions/login_steps.rb
@@ -35,7 +35,7 @@ end
 
 def login_with_username request_ip, username, success, headers = {}
   begin
-    @pkey = OpenSSL::PKey::RSA.new 1048
+    @pkey = OpenSSL::PKey::RSA.new 2048
     login(username, request_ip, authn_k8s_host, @pkey, headers)
   rescue
     raise if success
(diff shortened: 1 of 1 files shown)


cucumber/kubernetes/features/step_definitions/login_steps.rb
```diff
@@ -35,7 +35,7 @@ def login_with_custom_prefix request_ip, host_id_suffix, host_id_prefix, success
 
 def login_with_username request_ip, username, success, headers = {}
   begin
-    @pkey = OpenSSL::PKey::RSA.new 1048
```
shaharglazner: it's weird because 1048 isn't a power of 2...

sgnn7: lol indeed - this is weird

sgnn7:
> The change looks good. Can you please add a changelog entry?

@jvanderhoof This is just a test change so changelog might not be needed

orenbm:
@sgnn7 i added another small change from 1048 to 2048. It's in a step that is used by a skipped test so it didn't affect the build.

sgnn7:
> @sgnn7 i added another small change from 1048 to 2048. It's in a step that is used by a skipped test so it didn't affect the build.

I noticed :)
I already re-approved

micahlee:
I would also recommend this not be included in the `CHANGELOG`.

PR #1276, opened 2020-01-27
New scripts for cluster

oburstein-hub:
#### What does this PR do?
1. Adding files / scripts for open-shift installation
2. Adding support for remote debugger in Dockerfile
All these are needed for developer manuals
NA
On first file
Running procedures according to the documents
No


examples/basic_openshift/conjur-java-api-example.yaml
```diff
@@ -0,0 +1,84 @@
+---
+apiVersion: apps/v1
```
rafis3: If the example works for Kubernetes and not just for OpenShift, I would change the folder name and all relevant references to Kubernetes, because it's a more generic name.

examples/basic_openshift/custom-values.yaml
```diff
@@ -0,0 +1,15 @@
+authenticators: "authn-k8s/minikube,authn"
+dataKey: "iMldjotGK9DcKUjKfCv+k6vzVSoL/6ODTpS0CAoKio8="
+account: "myaccount"
```
rafis3: I think it would be great if there would be one place to define the constants that are used in more than one place. For example, the `account` name is currently hard coded in multiple places. If someone wants to make a change, it will not be straight forward to change it.

examples/basic_openshift/conjur-java-api-example.yaml
```diff
@@ -0,0 +1,84 @@
...
+          - name: CONJUR_AUTHN_URL
+            value: https://conjur-oss.ofira-conjur-test.svc.cluster.local/authn-k8s/minikube
+          - name: CONJUR_APPLIANCE_URL
+            value: https://conjur-oss.ofira-conjur-test.svc.cluster.local
+          - name: CONJUR_ACCOUNT
+            value: myaccount
+          - name: CONJUR_AUTHN_LOGIN
+            value: admin
+          - name: CONJUR_AUTHN_API_KEY
+            value: 2naavj9kywx73q7zr5q2qz1e1q1f8e7spg3g7e3tddtv133v95pt
```
rafis3: We can't leave clear API keys in GitHub. In addition, will that work if I run it in my environment or will it only work with your setup? Please find a way to remove this dependency.

Dockerfile
```diff
@@ -35,7 +35,13 @@ RUN mkdir -p $TMP_DIR \
...
      Gemfile.lock ./
 
-RUN bundle --without test development
+ENV WEB_CONCURRENCY=2
+
+RUN if [ "$debug" = "true" ] ; then  \
+	WEB_CONCURRENCY=0 ; \
+fi
+
+RUN if [ "$debug" != "true" ] ; then bundle --without test development ; else bundle --without test ; fi
```
rafis3: Can you please use multiline for this `if` the same way you did above?

Also what do you think about this way, to align the indention?

```Dockerfile
RUN if [ "$debug" = "true" ] ; then  \
      WEB_CONCURRENCY=0 ; \
    fi
```

examples/basic_openshift/conjur-java-api-example.yaml
```diff
@@ -0,0 +1,84 @@
...
+          - name: MY_POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: MY_POD_IP
+            valueFrom:
+              fieldRef:
+                fieldPath: status.podIP
+          - name: CONJUR_AUTHN_URL
+            value: https://conjur-oss.ofira-conjur-test.svc.cluster.local/authn-k8s/myauthenticator
```
rafis3: Can you please remove references of your name and use more generic names in this demo?

examples/basic_openshift/conjur-java-api-example.yaml
```diff
@@ -0,0 +1,84 @@
...
+          - name: CONJUR_APPLIANCE_URL
+            value: https://conjur-oss.ofira-conjur-test.svc.cluster.local
+          - name: CONJUR_ACCOUNT
+            value: myaccount
+          - name: CONJUR_AUTHN_LOGIN
+            value: admin
+          - name: CONJUR_AUTHN_API_KEY
+            value: <API Key>
```
rafis3: Why is this still required? The pod authenticates without the API key anyway, no?

examples/basic_openshift/custom-values.yaml
```diff
@@ -0,0 +1,15 @@
+authenticators: "authn-k8s/myauthenticator,authn"
+dataKey: "<datakey>"
```
rafis3: Since this is a placeholder, how is this value filled before it's used by Helm?

jtuttle:
@oburstein-hub Will you be able to finish this up in the near future? The PR has been untouched for quite some time. Thanks!

Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM ubuntu:20.04
-
+ARG debug
```
izgeri: I'm concerned about adding a `debug` argument to the primary Dockerfile for this project without tying it to the deeper strategy for debugging. Is this an indication that you should be using the [dev Dockerfile](https://github.com/cyberark/conjur/blob/master/dev/Dockerfile.dev) instead? Do we expect end-users to leverage this `debug` flag? In my opinion, the main Dockerfile should be clean and representative of the primary final image we build for end users that gets published to Dockerhub, etc.

Dockerfile
```diff
@@ -35,7 +35,13 @@ RUN mkdir -p $TMP_DIR \
 COPY Gemfile \
      Gemfile.lock ./
 
-RUN bundle --without test development
+ENV WEB_CONCURRENCY=2
```
izgeri: Is this consistent with the default `WEB_CONCURRENCY` value, or is this a change in behavior? If we proceed with this change, then I think it needs to be clearly called out in the PR description - and it would be useful to also describe the motivation, so we can understand the context around why this change was made when trying to understand it later.

Dockerfile
izgeri: this change makes me think that what @oburstein-hub is really looking for is the dev Dockerfile [here](https://github.com/cyberark/conjur/blob/master/dev/Dockerfile.dev)

app/controllers/policies_controller.rb
```diff
@@ -31,7 +31,6 @@ def post
 
   def load_policy perform_automatic_deletion:, delete_permitted:, update_permitted:
     policy_text = request.raw_post
-
```
izgeri: can this change be dropped from the PR? doesn't seem relevant / connected

examples/basic_openshift/policy1.yaml
```diff
@@ -0,0 +1,39 @@
+# Filename: policy_for_human_users.yml
```
izgeri: this is inconsistent with the actual policy name, and I like it better than the actual filename (since it's clearer what it contains). can the actual policy name be updated?

examples/basic_openshift/policy2.yaml
```diff
@@ -0,0 +1,23 @@
+---
+# Filename: policy_for_authenticator_identities.yml
```
izgeri: same comment about filename as above

examples/basic_openshift/policy3.yaml
```diff
@@ -0,0 +1,31 @@
+# Filename: policy_for_k8s_authenticator_service.yml
```
izgeri: same comment about filename as above

examples/basic_openshift/policy4.yaml
```diff
@@ -0,0 +1,14 @@
+---
+# Filename: policy_for_k8s_authenticator_service.yml
```
izgeri: same comment about filename as above

examples/basic_openshift/conjur-java-api-example.yaml
```diff
@@ -0,0 +1,84 @@
...
+          - name: CONJUR_SSL_CERTIFICATE
+            valueFrom:
+              configMapKeyRef:
+                name: conjur-cert
+                key: ssl-certificate
+        volumeMounts:
+          - mountPath: /run/conjur
+            name: conjur-access-token
+      - name: ofira-conjur-java-client
+        image: docker-registry.default.svc:5000/ofira-conjur-test/ofira-conjur-java-client:latest
```
izgeri: we have a test Java app that we use with Summon (it's the [pet store demo](https://github.com/conjurdemos/pet-store-demo)) - maybe that could be updated to have a Java client flow so that we don't have to refer to your personal image in this example?

PR #1279, opened 2020-01-27
Raise proper error if host is nil

orenbm:
In case the host was nil then we got some
`NoMethodError: undefined method 'role' for nil:NilClass`. This Commit
raises a proper error so it is easier to fix issues around this.

### DIFF
--- a/app/domain/authentication/authn_k8s/validate_pod_request.rb
+++ b/app/domain/authentication/authn_k8s/validate_pod_request.rb
@@ -73,6 +73,8 @@ module Authentication
 
       def host
         @host ||= @resource_class[k8s_host.conjur_host_id]
+        raise Err::HostNotFound(k8s_host.conjur_host_id) if @host.nil?
+        @host
       end
 
       def pod


app/domain/authentication/authn_k8s/validate_pod_request.rb
```diff
@@ -72,7 +72,11 @@ def host_can_access_service?
       end
 
       def host
-        @host ||= @resource_class[k8s_host.conjur_host_id]
+        unless defined? @host
+          @host = @resource_class[k8s_host.conjur_host_id]
+        end
```
jonahx: I believe you can replace this with the single line:
```ruby
@host ||= @resource_class[k8s_host.conjur_host_id]
```
Also even though it would be removed by the change suggested above, note that `defined? @host` should be `defined?(@host)` per [the ruby style guide's recommendation for method calls](https://github.com/rubocop-hq/ruby-style-guide#method-invocation-parens).  This should be getting caught by code climate but it may not be working.

orenbm: this is the whole fix. your suggestion was the original code but it's not good because `@resource_class[k8s_host.conjur_host_id]` can be nil. I want to raise an error in that case.

app/domain/authentication/authn_k8s/validate_pod_request.rb
```diff
@@ -72,7 +72,11 @@ def host_can_access_service?
...
+        return @host unless @host.nil?
+        raise Err::HostNotFound(k8s_host.conjur_host_id)
```
jonahx: Btw, even though we haven't followed this in our ruby codebase everywhere, I do like the Go convention of "happy path to the left",  and there no reason not to use it in ruby as well, so here you could swap the order of these two lines.

PR #1283, opened 2020-01-28
Generalize errors so they are consumed in all authenticators

orenbm:
This PR re-arranges some errors so they can be consumed easily in all authenticators.

This PR also changes the behaviour of the Kubernetes authenticator so it now returns 403 on unpermitted hosts instead of a 401. This was done to align it with the other authenticators.

### DIFF
--- a/app/controllers/concerns/basic_authenticator.rb
+++ b/app/controllers/concerns/basic_authenticator.rb
@@ -19,7 +19,7 @@ module BasicAuthenticator
       raise ApplicationController::Unauthorized, "Invalid username or password"
     rescue Errors::Authentication::InvalidOrigin
       raise ApplicationController::Forbidden, "User is not authorized to login from the current origin"
-    rescue Errors::Authentication::Security::UserNotAuthorizedInConjur
+    rescue Errors::Authentication::Security::RoleNotAuthorizedOnWebservice
       raise ApplicationController::Forbidden, "User is not authorized to login to Conjur"
     end
   end
--- a/app/domain/authentication/authn_k8s/validate_pod_request.rb
+++ b/app/domain/authentication/authn_k8s/validate_pod_request.rb
@@ -6,8 +6,9 @@ module Authentication
   module AuthnK8s
 
     Err = Errors::Authentication::AuthnK8s
+    SecurityErr = Errors::Authentication::Security
     # Possible Errors Raised:
-    # WebserviceNotFound, HostNotAuthorized, PodNotFound
+    # WebserviceNotFound, RoleNotAuthorizedOnWebservice, PodNotFound
     # ContainerNotFound, ScopeNotSupported, K8sResourceNotFound
 
     ValidatePodRequest = CommandClass.new(
@@ -32,12 +33,12 @@ module Authentication
       private
 
       def validate_webservice_exists
-        raise Err::WebserviceNotFound, service_id unless webservice.resource
+        raise SecurityErr::WebserviceNotFound, service_id unless webservice.resource
       end
 
       def validate_host_can_access_service
         return if host_can_access_service?
-        raise Err::HostNotAuthorized.new(host.role.id, service_id)
+        raise SecurityErr::RoleNotAuthorizedOnWebservice.new(host.role.id, service_id)
       end
 
       def validate_pod_exists
@@ -73,7 +74,7 @@ module Authentication
 
       def host
         @host ||= @resource_class[k8s_host.conjur_host_id]
-        raise Err::HostNotFound(k8s_host.conjur_host_id) if @host.nil?
+        raise SecurityErr::RoleNotFound(k8s_host.conjur_host_id) if @host.nil?
         @host
       end
 
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -240,11 +240,6 @@ unless defined? Errors::Authentication::AuthenticatorNotFound
           code: "CONJ00034E"
         )
 
-        MissingEnvVar = ::Util::TrackableErrorClass.new(
-          msg: "Expected ENV variable '{0}' is not set",
-          code: "CONJ00035E"
-        )
-
         UnknownK8sResourceType = ::Util::TrackableErrorClass.new(
           msg: "Unknown Kubernetes resource type '{0}'",
           code: "CONJ00041E"
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -73,23 +73,23 @@ unless defined? Errors::Authentication::AuthenticatorNotFound
 
       module Security
 
-        NotWhitelisted = ::Util::TrackableErrorClass.new(
-          msg: "'{0-authenticator-name}' is not whitelisted in CONJUR_AUTHENTICATORS",
+        AuthenticatorNotWhitelisted = ::Util::TrackableErrorClass.new(
+          msg: "'{0-authenticator-name}' is not enabled",
           code: "CONJ00004E"
         )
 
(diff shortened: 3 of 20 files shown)


app/domain/authentication/authn_k8s/validate_pod_request.rb
```diff
@@ -73,7 +73,7 @@ def host_can_access_service?
 
       def host
         @host ||= @resource_class[k8s_host.conjur_host_id]
-        raise Err::HostNotFound(k8s_host.conjur_host_id) if @host.nil?
+        raise ::Errors::Authentication::HostNotFound(k8s_host.conjur_host_id) if @host.nil?
```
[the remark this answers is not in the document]
sigalsax: ^ I am not sure what that means or why that might be an issue. do you?

sigalsax:
Good idea! If we already here, let's think about errors that currently exist in AuthnK8s and if they can be applicable to AuthnAzure. 

For example:
"Err::HostNotAuthorized" to check if the resource can authenticate to Conjur? 
"WebserviceNotFound" to check if the authenticator exists in Conjur as a webservice. Aka if the Azure authn was loaded as a webservice in Conjur

app/domain/authentication/authn_k8s/validate_pod_request.rb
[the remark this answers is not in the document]
orenbm: you can ignore this comment

app/domain/errors.rb
```diff
@@ -73,24 +73,24 @@ module AuthenticatorClass
 
       module Security
 
-        NotWhitelisted = ::Util::TrackableErrorClass.new(
+        AuthenticatorNotWhitelisted = ::Util::TrackableErrorClass.new(
           msg: "'{0-authenticator-name}' is not whitelisted in CONJUR_AUTHENTICATORS",
```
sigalsax: Because the whitelisting happens in DB, this message ("whitelisted in CONJUR_AUTHENTICATORS") I believe is not correct. How does enabling look in the DB?

app/domain/errors.rb
```diff
@@ -73,24 +73,24 @@ module AuthenticatorClass
...
-          code: "CONJ00006E"
+        RoleNotFound = ::Util::TrackableErrorClass.new(
+          msg: "'{0-role-name}' wasn't found",
+          code: "CONJ00007E"
         )
 
-        UserNotDefinedInConjur = ::Util::TrackableErrorClass.new(
-          msg: "User '{0-user-name}' is not defined in Conjur",
-          code: "CONJ00007E"
+        RoleNotAuthorizedOnWebservice = ::Util::TrackableErrorClass.new(
```
sigalsax: please move this above "RoleNotFound" so it is in the proper numerical order

app/controllers/concerns/basic_authenticator.rb
```diff
@@ -19,7 +19,7 @@ def perform_basic_authn
       raise ApplicationController::Unauthorized, "Invalid username or password"
     rescue Errors::Authentication::InvalidOrigin
       raise ApplicationController::Forbidden, "User is not authorized to login from the current origin"
-    rescue Errors::Authentication::Security::UserNotAuthorizedInConjur
```
sigalsax: Please follow suite here by setting  "Errors::Authentication::Security" outside so that it is "Err = Errors::Authentication::Security"

sigalsax: *below*

cucumber/kubernetes/features/login_errors.feature
```diff
@@ -15,12 +15,12 @@ Feature: Errors emitted by the login method.
...
     on the webservice.
     When I login to authn-k8s as "pod/inventory-unauthorized"
-    Then the HTTP status is "401"
+    Then the HTTP status is "403"
 
   Scenario: it raises an error when logging in with a custom prefix as a host which does not
     hold "authenticate" privilege on the webservice.
     When I login to authn-k8s as "@namespace@/pod/inventory-unauthorized" with prefix "host/conjur/authn-k8s/minikube/apps"
-    Then the HTTP status is "401"
+    Then the HTTP status is "403"
```
sigalsax: how did our tests pass previously if the error status wasn't correct?

spec/app/domain/authentication/authn_k8s/validate_pod_request_spec.rb
```diff
@@ -87,15 +87,15 @@
     subject(:validator) { Authentication::AuthnK8s::ValidatePodRequest
                             .new(dependencies: dependencies) }
 
-    it 'raises WebserviceNotFound error when webservice is missing' do
+    it 'raises WebserviceNotDefined error when webservice is missing' do
```
sigalsax: Why was this renamed from WebserviceNotFound -> WebserviceNotDefined here? Did a search and "WebserviceNotDefined" doesn't exist in other places but this one

spec/app/domain/authentication/validate_webservice_access_spec.rb
```diff
@@ -92,7 +92,7 @@ def role_class(returned_role)
       )
     end
     it "raises a NotDefinedInConjur error" do
```
sigalsax: This needs to be changed here b/c we changed it to "Role not found"

spec/app/domain/authentication/validate_webservice_access_spec.rb
```diff
@@ -112,7 +112,7 @@ def role_class(returned_role)
     end
 
     it "raises a NotAuthorizedInConjur error" do
```
sigalsax: same here. We need to change the expectation here b/c we changed it to RoleNotAuthorizedOnA*Webservice*

app/domain/errors.rb
orenbm: changed to `'{0-authenticator-name}' is not enabled`

app/controllers/concerns/basic_authenticator.rb
orenbm: This class doesn't use this method. We do this only when we have a lot of the same error module in the class so we don't want to repeat ourselves.

cucumber/kubernetes/features/login_errors.feature
orenbm: This PR actually changes the returned error to a better one. 403 Forbidden is a proper error in this use-case.

spec/app/domain/authentication/authn_k8s/validate_pod_request_spec.rb
orenbm: that's a mistake, it should be `WebserviceNotFound`

app/domain/errors.rb
```diff
@@ -73,23 +73,23 @@ module AuthenticatorClass
...
-        UserNotAuthorizedInConjur = ::Util::TrackableErrorClass.new(
-          msg: "User '{0-user-name}' is not authorized in the Conjur policy",
+        RoleNotAuthorizedOnWebservice = ::Util::TrackableErrorClass.new(
+          msg: "'{0-role-name}' does not have 'authenticate' privilege on {1-service-name}",
           code: "CONJ00006E"
         )
 
-        UserNotDefinedInConjur = ::Util::TrackableErrorClass.new(
-          msg: "User '{0-user-name}' is not defined in Conjur",
+        RoleNotFound = ::Util::TrackableErrorClass.new(
```
sigalsax: Maybe we should leave "In Conjur" to be clear

orenbm: I removed it because we don't want the word `Conjur` in our logs anyway

PR #1284, opened 2020-01-29
Split policy to smaller ones so we don't get a timeout on load

orenbm:
Our build was unstable because of a timeout in the GKE policy load. This PR splits the policy to smaller ones so we don't have this issue and the build will be stable.


orenbm:
This actually didn't solve the issue. I ran ~15 builds and some still failed on the policy load :(

PR #1285, opened 2020-01-29
Generalize FetchOidcSecrets so we can use it in other authenticators

orenbm:
In order to use this util in other authenticators we refactor
it to FetchAuthenticatorSecrets and moved it under the `authenticators`
directory


app/domain/authentication/util/fetch_authenticator_secrets.rb
```diff
@@ -5,11 +5,11 @@ module Authentication
   module AuthnOidc
```
sigalsax: If we are making it more general, then let's remove the AuthnOidc module

app/domain/authentication/authn_oidc/validate_status.rb
```diff
@@ -26,9 +26,10 @@ def validate_secrets
       end
 
       def fetch_oidc_secrets
-        @oidc_secrets ||= @fetch_oidc_secrets.(
+        @oidc_secrets ||= @fetch_authenticator_secrets.(
```
sigalsax: Let's change this to "@oidc_authenticator_secrets" because that is how we have changed it in authn_oidc/authenticate.rb

app/domain/authentication/util/fetch_authenticator_secrets.rb
orenbm: nice catch

app/domain/authentication/util/fetch_authenticator_secrets.rb
```diff
@@ -0,0 +1,38 @@
...
+
+        required_secrets = @fetch_secrets.(resource_ids: required_resource_ids)
+
+        @required_variable_names.each do |variable_name|
+          full_variable_name     = full_variable_name(variable_name)
+          secrets[variable_name] = required_secrets[full_variable_name]
+        end
+
+        secrets
+      end
```
jonahx: This can be cleaned up a bit and made more consistent with the style we use elsewhere (very simple call, named memoized methods for any relevant concept.  This should work (untested though):
```ruby
#... 
      def call
        @required_variable_names.reduce({}) do |secrets, variable_name|
          full_variable_name     = full_variable_name(variable_name)
          secrets[variable_name] = required_secrets[full_variable_name]
          secrets
        end
      end

      private

      def required_secrets
        @required_secrets ||= @fetch_secrets.(resource_ids: required_resource_ids)
      end
#... 
```

jonahx: Since you're using `each_with_object` rather than `reduce` (which is a good call) you can delete line 18: `secrets`.

PR #1286, opened 2020-01-29
Generalize credentials field of AuthenticatorInput

orenbm:
We would like to make this field more generic so it will make sense in all of its uses. For example, in the authn-iam
authenticator, we use the AuthenticatorInput's credentials field
to extract the AWS header which are located in the request's
body. Calling that field `password` just because the body happens
to include the password in `authn` and `authn-ldap` is wrong.
This will also be the case in `authn-azure` where the body
will have the Azure access token and not a password.
Also, calling it `request_body` is too generic so the field
is called `credentials`

### DIFF
--- a/app/controllers/concerns/basic_authenticator.rb
+++ b/app/controllers/concerns/basic_authenticator.rb
@@ -40,7 +40,7 @@ module BasicAuthenticator
       service_id:         params[:service_id],
       account:            params[:account],
       username:           username,
-      request_body:       password,
+      credentials:        password,
       origin:             request.ip,
       request:            request
     )
--- a/app/domain/authentication/authenticator_input.rb
+++ b/app/domain/authentication/authenticator_input.rb
@@ -9,7 +9,7 @@ module Authentication
     attribute :service_id, ::Types::NonEmptyString.optional
     attribute :account, ::Types::NonEmptyString
     attribute :username, ::Types::NonEmptyString.optional
-    attribute :request_body, ::Types::String.optional
+    attribute :credentials, ::Types::String.optional
     attribute :origin, ::Types::NonEmptyString
     attribute :request, ::Types::Any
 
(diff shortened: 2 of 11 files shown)


app/controllers/concerns/basic_authenticator.rb
```diff
@@ -40,7 +40,7 @@ def login_input(username, password)
       service_id:         params[:service_id],
       account:            params[:account],
       username:           username,
-      request_body:       password,
+      credentials:       password,
```
sigalsax: please fix spacing here

app/domain/authentication/authenticator_input.rb
```diff
@@ -9,7 +9,7 @@ class AuthenticatorInput < ::Dry::Struct
     attribute :service_id, ::Types::NonEmptyString.optional
     attribute :account, ::Types::NonEmptyString
     attribute :username, ::Types::NonEmptyString.optional
-    attribute :request_body, ::Types::String.optional
+    attribute :credentials, ::Types::String.optional
```
sigalsax: let's feel aligning here for this collection of attributes

spec/app/domain/authentication/authenticate_spec.rb
```diff
@@ -76,7 +76,7 @@ def authenticator(pass:)
         service_id:         nil,
         account:            'my-acct',
         username:           'my-user',
-        request_body:       'my-pw',
+        credentials:       'my-pw',
```
sigalsax: fix spacing here

spec/app/domain/authentication/authenticate_spec.rb
```diff
@@ -133,7 +133,7 @@ def authenticator(pass:)
           service_id:         nil,
           account:            'my-acct',
           username:           'my-user',
-          request_body:       'my-pw',
+          credentials:       'my-pw',
```
sigalsax: and here (:

app/domain/authentication/authenticator_input.rb
orenbm: what? if you mean to align the `::Types` then i prefer not to as it is not configurable.

PR #1288, opened 2020-01-30
WIP Authn-azure solution doc is open to community

sigalsax:
#### What does this PR do?
This PR pushes the authn-azure solution doc to GH to get early feedback from the team and open-source community

MD file: https://github.com/cyberark/conjur/blob/azure-authn-design/design/authn_azure/authn_azure_solution_design.md

Connected to #1293


design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,242 @@
...
+request is sent to Conjur, the admin will load the authenticator policy: 
+
+```yaml
+# policy id needs to match the convention `conjur/authn-azure/<service ID>`
+- !policy
+  id: conjur/authn-azure/prod
+  body:
+  - !webservice
+    - !variable
+        id: provider-uri # https://sts.windows.net/TENANT_ID
```
AndrewCopeland: Is there any way this could be an annotation? Or is this limited because of OIDC implementation?
If it were an annotation it would help with UX.

InbalZilberman:
Please link any design to the feature doc. In our case a git issue.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+  id: azure-apps 
+  body: 
+  - !host 
+    id: test-app 
+    annotations: 
+    authn-azure/subscription-id: test-subscription 
+    authn-azure/resource-group: test-group 
+    authn-azure/user-assigned-identity: test-app-pipeline 
+OR 
+    authn-azure/system-assigned-identity: ie_qWCXhXxt1zIEsu4c7acQVGn4
```
shaharglazner: I would change it to a real system assigned identity (which is GUID) to avoid confusion

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+    
+    1. validate_authenticator_exists ensures that the authenticator sent as part of the request URI is one that we support (authn-k8s, authn-azure, etc.).
+
+    1. validate_security validates that the Azure authenticator has been defined correctly in policy and the host is permitted to authenticate to the webservice.
+
+1. The next validation, validate_credentials calls the authenticator's valid? function. If so, Azure-specific authenticator validations will be triggered. 
+
+    1. validate_azure_token (see below for a more detailed explanation of the function).
+    
+    1. validate_application_identity, the Conjur identity host annotations for the Azure VM and the properties extracted from the "xms_mirid" field of the Azure token provided by the Azure VM are compared (see below for a more detailed explanation of the function).
```
shaharglazner: It would be great to add "xms_mirid" to the Glossary and explain what it is and why it's Ok we use it.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+| Service principle                      | object representation of VM instance in AD. This object holds information defining its permissions on resources                                                                                 |
+| Client ID                              | unique identifier generated by Azure AD tied to application and service principal during initial provisioning. This client ID is used to identify the object that is requesting an access token |
+| System-assigned identity               | an identity for an instance in AD that is trusted. The identity is tied to the resource and abides to its lifecycle. In other words, if the resource is delete so to will the identity.         |
+| User-assigned identity                 | an identity for an instance in AD that is trusted. This identity can be assigned to 1+ Azure services instances and is independent of instance lifecycle                                        |
+| Azure Instance Metadata Service (IMDS) | a REST endpoint accessible to all VMs created via the Azure Resource Manager. Resources in Azure request access tokens from the Azure Instance Metadata service endpoint                        |
```
shaharglazner: Worth to mention that while, at first, we support VM, IMDS supports other services also, and we can leverage it too.

sigalsax:
> Hi @orenbm
> First let me say how clear and useful this document is !
> I already stole a few pages for the presentation :-)
> 
> A few of items are missing for me.
> 
> 1. Can you please show  a **detailed** flow of the developer code (which includes fetching the token), for both the system assigned and the user assigned flow. Specifically, how does the developer get the string which s/he needs to put in the REST header?
> 2. Similarly can you show a **detailed**  pipeline admin code.  How does the pipeline admin know which fields to load into the annotation? Again for both user and system assigned.
>    Speaking of that, in your example
>    ` authn-azure/user-assigned-identity: test-app-pipeline`
>    how is it that the identity is a string and how is this string resolved against what is found in the JWT?
> 3. Finally (for now :-) would be helpful to show the  content of  the "xms_mirid" as @shaharglazner  also pointed out. Maybe a table that maps those fields to annotations ?
>    Thanks!
>    Boaz

@boazmichaely glad you think so! Will add the feedback first thing when we get back

boazmichaely:
@sigalsax I read @orenbm email wrong, did not realize it was your work :flushed: 
My praize stays the same LOL .. Keep up the great work 🥇!!

design/authn_azure/authn_azure_solution_design.md
sigalsax: why would it add confusion? b/w what is in the azure token (guid) and how we reference it here?

design/authn_azure/authn_azure_solution_design.md
sigalsax: Good point- Adding this!

design/authn_azure/authn_azure_solution_design.md
sigalsax: If I am understanding correctly, you are proposing adding the `provider-uri` as an annotation in the webservice instead of housing it as variable in Conjur. I appreciate you challenging this!  

In terms of security, I am not sure how sensitive this variable is and this is something we should definitely check @shaharglazner 

In terms of UX, this is how we configure the OIDC authenticator. I imagine that our users use multiple authenticators across their environments and so I am wary to suddenly change this configuration _just_ for Azure and not also have this change cascaded to OIDC.

@InbalZilberman @Tovli @orenbm opinions on this one?

sigalsax:
> @sigalsax I read @orenbm email wrong, did not realize it was your work 😳
> My praize stays the same LOL .. Keep up the great work 🥇!!

Joint effort! 💯 
Regarding points 1 and 2, these are great questions! 
This doc is more technical. A detailed explanation of each REST call can be found in Inbal's Feature Doc under "App Developer David & Azure Authenticator Logic".

The feature doc can be found here: https://app.zenhub.com/workspaces/appliance-development-5c9cf04bfa1c204b63eddcc6/issues/cyberark/conjur/1266

In response to your third point, I will add that mapping to the doc (:

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+ 
+#### validate_azure_token
+
+The Azure token validation checks the signature and claims of the Azure token to ensure that the token was issued by Azure Active Directory. To do this we need to:
+1. Discover the Oauth 2.0 provider - in this case the Azure AD is the Oauth 2.0 provider. Its URl will be retrieved from the "provider-uri" variable defined in the authenticator policy
+1. Retrieve JWKs from the discovery endpoint - These keys are used for the token validation
+1. Decode and verify the token using the JWKs.
+
+Steps 1 & 2 may seem familiar as they are the exact steps that we perform in the OIDC Authenticator. This is because OIDC is Oauth 2.0 based. We will take advantage on this and use the 3rd party used in the OIDC Authenticator to perform steps 1 & 2 in our authenticator.
+Unfortunately, the 3rd party used in the OIDC Authenticator does not implement an access token decoding. Thus, to perform step 3, we will need another 3rd party - the "jwt" gem.
```
sigalsax: @orenbm 
I think we should further make the difference b/w OIDC id token and Azure access token, fleshing out that the OIDC 3rd party offers a decoding for an ID token and not access token. Therefore, we need to look elsewhere.

boazmichaely:
> > @sigalsax I read @orenbm email wrong, did not realize it was your work 😳
> > My praize stays the same LOL .. Keep up the great work 🥇!!
> 
> Joint effort! 💯
> Regarding points 1 and 2, these are great questions!
> This doc is more technical. A detailed explanation of each REST call can be found in Inbal's Feature Doc under "App Developer David & Azure Authenticator Logic".
> 
> The feature doc can be found here: https://app.zenhub.com/workspaces/appliance-development-5c9cf04bfa1c204b63eddcc6/issues/cyberark/conjur/1266
> 
> In response to your third point, I will add that mapping to the doc (:

Hi @sigalsax 
Thanks for referring me to the feature doc. I'm looking at it again but I think that more detail is needed

> Authenticate the application against Azure authenticator (REST API) -
> Fetch the Access token provided by Azure ← exact details will be filled in in design phases.

design/authn_azure/authn_azure_solution_design.md
InbalZilberman: @AndrewCopeland this may have also performance aspects.
From UX I agree with Andrew it would be easier for the user.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+#### Authentication Flow
+
+This flow assumes that the VM already exists in Azure, and that an authn-azure authenticator and a Conjur host were loaded.
+
+![Authn Azure Flow](authn-azure-flow.png)
+
+
+1. Azure VM sends the following request to Conjur:
+    ```yaml
+    POST https://<DAP-server-hostname>/authn-azure/<service-id>/<account>/<host-id>authenticate
```
InbalZilberman: missing slash between host id and authenticate

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+    Header 
+    
+    Content-Type: application/x-www-form-urlencoded
+    Body 
+    
+     token: "eyJhbGciOiJSUzI1NiIs......uTonCA"
```
InbalZilberman: On second thought lets change the param name to "azureAccessToken"
WDYT? @orenbm @sigalsax

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+
+- Conjur
+
+- DAP
+
+## Test Plan
+
+TBT
+
+## Logs
```
InbalZilberman: Can you please add also the level of logs?

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+| 3  | WebserviceNotFound (in Conjur)                                        | '{0-role-name}' does not have 'authenticate' privilege on {1-service-name}             |                                                                                      |        | Yes                                           |
+| 4  | RoleNotFound (Host/User not defined in Conjur)                        | '{0-role-name}' wasn't found                                                           |                                                                                      |        | Yes                                           |
+| 5  | ProviderDiscoveryTimeout (couldn't make connection with provider-uri) | Azure Identity Provider failed with timeout error (Provider URI: '{0}'). Reason: '{1}' |                                                                                      |        | Yes (needed to be generalized)                |
```
InbalZilberman: Maybe Azure authenticator?

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+| 7  | MissingRequestParam                                                   | Field '{0-field-name}' is missing or empty in request body                             | "Token" not in body of request                                                       |        | Yes (must extract authn-request body)         |
+| 8  | TokenFieldNotFoundOrEmpty                                             | Field '{0-field-name}' not found or empty in token                                     |                                                                                      |        | No (for OIDC but need to be more generalized) |
+| 9  | InvalidApplicationIdentity                                            | Azure application identity field '{0-field-name}' does not match Azure token           | 1+ of the fields in the Azure token don't match the Conjur host identity annotations |        | No                                            |
```
InbalZilberman: Can we specify the values of the annotation and the token? maybe in a debug level only?

InbalZilberman: Do we have a positive log for the success of Azure authenticator? specifying the host it served successfully?

design/authn_azure/authn_azure_solution_design.md
nessiLahav: @orenbm make sure to use the same cache capabilities we used for oidc implementation, for example: we used concurrency_limited_cache to limit the amount of threads that can access in parallel to Azure, and in case of timeout between conjur to Azure we do not suffocate the server

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,280 @@
...
+- [ ] Logs were reviewed by TW and PO
+- [ ] Documentation has been written and reviewed by PO and TW for both DAP and Conjur
+- [ ] Engineer(s) that were not involved in the project use documentation to authenticate with Conjur using Azure
+- [ ] Versions are bumped in all relevant projects
+   - [ ] Conjur
+   - [ ] DAP
+ 
+## Open questions
+
+- Do we want to develop a mechanism to support a request body with more than one field (currently, we have only "token", but we may want to add more fields later).
```
InbalZilberman: 1. What would happen if a user will try to use Azure authenticator?
2. Do we need to validate the source of the token or provider to be Azure? 
3. Do you know of or think on any reason (other than tests) that an application would not be able to run from Azure Functions or App Service or  VM scale set?

boazmichaely:
Is it Access token or ID token ?

design/authn_azure/authn_azure_solution_design.md
boazmichaely: Is it Access token or ID token?

InbalZilberman: Access Token

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,357 @@
...
+
+- DAP
+
+## Test Plan
+
+TBT
+
+### Integration tests
+
+#### Functionality
```
eladkug: 1. Need to verify more 4xx scenarios such as:
- 400 - size to large , invalid chars
- 408 - request timeout if we support it. 
- 429 - too many requests 
2. VM sends authentication request with a **invalid** Azure token

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,357 @@
...
+| 1  | *Vanilla flow* - a Conjur identity exists for Azure VM and added to privileged layer         | VM sends authentication request with a valid Azure token | Conjur access token is returned in body and action is logged                                  | Integration               |            |
+| 2  | A Conjur Host identity for Azure VM, added to privileged layer and has permissions on secret | VM sends request with valid Azure token to fetch secret  | Secret is successfully returned and action is logged                                          | Integration               |            |
+| 3  | No Conjur Host identity exists for Azure VM                                                  | VM sends authentication request with a valid Azure token | 403 Forbidden response / RoleNotFound in error body is returned and action is logged          | Integration               |            |
+| 4  | Conjur identity exists for Azure VM but doesn't have permissions on secret                   | VM sends request with valid Azure token to fetch secret  | 401 Unauthorized response is returned and action is logged                                    | Integration               |            |
```
eladkug: Need to play with permissions :
1. No permissions at all (maybe this is the one you meant)
2. read permission only
3. authentication permission only

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Error handling
```
eladkug: Im not sure all the errors here are 500.  According to this [document](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) .
Some of them may be 501,503.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,357 @@
...
+
+### Unit tests
+
+1. Validate that `validate_azure_token` properly decodes the of JWT token
+
+1. Validate that `validate_application_identity` accurately compares the Conjur host identity and Azure token correctly
+
+1. If `validate_azure_token` or `validate_application_identity` raise an error then it propagates to calling `Authentication::Authenticate` class
+
+### Performance
```
eladkug: 1. How many sessions we can run in parallel ? need to test it
2. How we can declare pass for performance ? verify 3000 sessions and verify the average is 1 sec per call ?
3. Stability - how many requests we send ? do we verify during this test some failover and see recovery ?

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,357 @@
...
+
+### Integration tests
+
+#### Functionality
+
+This suite of tests assumes that the Azure authenticator has been configured correctly
+
+|    | Given                                                                                        | When                                                     | Then                                                                                          | Manual / UT / Integration | Status     |
+|----|----------------------------------------------------------------------------------------------|----------------------------------------------------------|-----------------------------------------------------------------------------------------------|---------------------------|------------|
+| 1  | *Vanilla flow* - a Conjur identity exists for Azure VM and added to privileged layer         | VM sends authentication request with a valid Azure token | Conjur access token is returned in body and action is logged                                  | Integration               |            |
```
eladkug: VM sends authentication request with an **invalid** Azure token

design/authn_azure/authn_azure_solution_design.md
AndrewCopeland: The only information that could be sensitive is the tenet ID, the azure STS URI looks like:
`https://sts.windows.net/<tenant ID>/`

It appears the tenant ID is similar to your account number in AWS.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,357 @@
...
+  id: azure-apps 
+  body: 
+  - !host 
+    id: test-app 
+    annotations: 
+    authn-azure/subscription-id: test-subscription 
+    authn-azure/resource-group: test-group 
+    authn-azure/user-assigned-identity: test-app-pipeline 
+OR 
+    authn-azure/system-assigned-identity: ie_qWCXhXxt1zIEsu4c7acQVGn4
```
AndrewCopeland: This fixes policy syntax for annotations.
```suggestion
      authn-azure/subscription-id: test-subscription 
      authn-azure/resource-group: test-group 
      authn-azure/user-assigned-identity: test-app-pipeline 
OR 
      authn-azure/system-assigned-identity: ie_qWCXhXxt1zIEsu4c7acQVGn4
```

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,357 @@
...
+
+```yaml
+# policy id needs to match the convention `conjur/authn-azure/<service ID>`
+- !policy
+  id: conjur/authn-azure/prod
+  body:
+  - !webservice
+  
+    - !variable
+      id: provider-uri # https://sts.windows.net/TENANT_ID
```
AndrewCopeland: This fixes policy syntax
```suggestion
  - !variable
    id: provider-uri # https://sts.windows.net/TENANT_ID
```

design/authn_azure/authn_azure_solution_design.md
sigalsax: I also thought AzureToken/azureAccessToken would be more fitting but how to refer to it in OIDC? is it `idToken` or just `token` as well?

design/authn_azure/authn_azure_solution_design.md
sigalsax: What do you mean here? Do you suggest we define what each level of log is (WARN, DEBUG, etc)? or which logs will be returned for each log level?

design/authn_azure/authn_azure_solution_design.md
sigalsax: This timeout would be raised as a result of inability to talk to the `provider-uri` (https://sts.windows.net/TENANT_ID) so I don't think it would be a timeout of our Azure authn, rather the actual tenant Identity Provider. What do you think?

design/authn_azure/authn_azure_solution_design.md
sigalsax: > Can we specify the values of the annotation and the token? maybe in a debug level only?
In terms of security, would that be offering too much info into what is in the Conjur policy? @shaharglazner

sigalsax: > Do we have a positive log for the success of Azure authenticator? specifying the host it served successfully?

Nice catch! Adding the positive log messages now!

design/authn_azure/authn_azure_solution_design.md
orenbm: @shaharglazner @InbalZilberman if this data isn't sensitive then I'd be happy to change it to an annotation of the webservice

design/authn_azure/authn_azure_solution_design.md
orenbm: why do we need `azure_access_token`? the request is sent to the `authn-azure` authenticator so it's implied to be an azure token. I can go with `access_token` instead of `token`

design/authn_azure/authn_azure_solution_design.md
orenbm: 1. @InbalZilberman i don't understand the question. Can you elaborate?
2. We validate the token in `validate_azure_token` as part of the decoding.
3. Not that i can think of.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Affected Components
+
+- Conjur
+
+- DAP
+
+## Test Plan
+
+TBT
```
orenbm: this can be removed

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Integration tests
+
+#### Functionality
```
orenbm: i don't understand the difference between `Functionality` and `Error Handling`

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+
+|    | Given                                                                                        | When                                                     | Then                                                                                          | Manual / UT / Integration | Status     |
+|----|----------------------------------------------------------------------------------------------|----------------------------------------------------------|-----------------------------------------------------------------------------------------------|---------------------------|------------|
+| 1  | *Vanilla flow* - a Conjur identity exists for Azure VM and added to privileged layer         | VM sends authentication request with a valid Azure token | Conjur access token is returned in body and action is logged                                  | Integration               |            |
+| 2  | A Conjur Host identity for Azure VM, added to privileged layer and has permissions on secret | VM sends request with valid Azure token to fetch secret  | Secret is successfully returned and action is logged                                          | Integration               |            |
```
orenbm: I'm missing something. how is the secret related to this use-case? The VM sends request with valid Azure token to *authenticate with Conjur* or *to get a Conjur access token*. We can have a test that verifies that the retrieved Conjur access token is valid for retrieving a secret in another request but that's a different test.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+| 3  | No Conjur Host identity exists for Azure VM                                                  | VM sends authentication request with a valid Azure token | 403 Forbidden response / RoleNotFound in error body is returned and action is logged          | Integration               |            |
```
orenbm: 1. Why would we return a 403 in this case? In case the user does not exist we return a 401. 403 is in case the user exists but is not permitted on the webservice.
2. We do not return the error in the response body, only write it to the log.

design/authn_azure/authn_azure_solution_design.md
orenbm: same here - i'm guessing you mean that the user does not have permissions on the webservice, not the secret.

design/authn_azure/authn_azure_solution_design.md
sigalsax: Don't we want to give similar experiences across our authenticators?

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+
+#### Error handling
+
+|    | Given                                                                                     | When                                                                | Then                                                                                                                                         | Manual / UT / Integration | Status     |
+|----|-------------------------------------------------------------------------------------------|---------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------|---------------------------|------------|
+| 1  | authn-azure is not whitelisted in ENV or enabled in DB                                    | VM sends authentication request to Conjur with valid Azure token    | 500 Internal Server Error response with an error body with the relevant error message (AuthenticatorNotWhitelisted) is returned and logged   | Integration               |            |
```
orenbm: In this case we return 401

orenbm: And also here we don't return the error message in the body, only in the log

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+| 2  | Policy is loaded w/o "provider-uri" variable or value is missing                          | VM sends authentication request to Conjur with valid Azure token    | 500 Internal Server Error response with an error body with the relevant error message (RoleNotAuthorizedOnWebservice) is returned and logged | Integration               |            |
```
orenbm: 1. In this case we return 401
2. We don't return the error message in the body, only in the log

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+| 3  | Policy is loaded w/o defining authenticator webservice                                    | VM sends authentication request to Conjur with valid Azure token    | 500 Internal Server Error response with an error body with the relevant error message (WebserviceNotFound) is returned and logged            | Integration               |            |
```
orenbm: 1. In this case we return 401
2. We don't return the error message in the body, only in the log

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+| 4  | An incorrectly/undefined provider-uri (Conjur couldn't connect to provider)               | VM sends authentication request to Conjur with valid Azure token    | 500 Internal Server Error response with an error body with the relevant error message (ProviderDiscoveryTimeout) is returned and logged      | Integration               |            |
```
orenbm: 1. In this case we return 401
2. We don't return the error message in the body, only in the log
3. what is `undefined provider-uri`? Is the variable value missing?

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+| 5  | Conjur can make contact with provider but couldn't confirm token signature                | VM sends authentication request to Conjur with invalid Azure token  | 500 Internal Server Error response with an error body with the relevant error message (ProviderTokenInvalid) is returned and logged          | Integration               |            |
```
orenbm: 1. In this case we return [502](https://www.lifewire.com/502-bad-gateway-error-explained-2622939)
2. We don't return the error message in the body, only in the log

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+| 6  | One or more annotations are missing from the host policy **                               | VM sends authentication request to Conjur with valid Azure token    | 500 Internal Server Error response with an error body with the relevant error message (MissingRequestParam) is returned and logged           | Integration               |            |
```
orenbm: I disagree with the `MissingRequestParam` error. The request itself is fine but the missing param is in the host policy definition. We should return an 401 Unauthorized response in this case and log which annotation is missing.

design/authn_azure/authn_azure_solution_design.md
InbalZilberman: @eladkub 
1. We can start with ±20 the raise the bar in GA.
3. Yes, such scenarios should be tested but not necessarily with each and every authenticator. A scenario for this authenticator can be unavailability of the Azure AD or restart of Application VM.

design/authn_azure/authn_azure_solution_design.md
InbalZilberman: @orenbm  - no performance aspects?

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+| 7  | One or more fields in Azure token is missing                                              | VM sends authentication request to Conjur with invalid Azure token  | 500 Internal Server Error response with an error body with the relevant error message (TokenFieldNotFoundOrEmpty) is returned and logged     | Integration               |            |
```
orenbm: let's pinpoint this to the `xms_mirid` field as it's the only one that we are looking for.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+| 8  | One or more fields in the Azure token don't match the Conjur host annotations             | VM sends authentication request to Conjur with valid   Azure token  | 500 Internal Server Error response with an error body with the relevant error message (InvalidApplicationIdentity) is returned and logged    | Integration               |            |
```
orenbm: 1. In this case we return 401
2. We don't return the error message in the body, only in the log

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Unit tests
+
+1. Validate that `validate_azure_token` properly decodes the of JWT token
+
+1. Validate that `validate_application_identity` accurately compares the Conjur host identity and Azure token correctly
+
+1. If `validate_azure_token` or `validate_application_identity` raise an error then it propagates to calling `Authentication::Authenticate` class
```
orenbm: So this is a UT on `Authentication::Authenticate`? This class doesn't know `validate_azure_token` or `validate_application_identity` so it's not a valid UT.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Performance
+
+|    | Type      | Given                                                | When                                                                      | Then                                             | Environment      | Status     |
+|----|-----------|------------------------------------------------------|---------------------------------------------------------------------------|--------------------------------------------------|------------------|------------|
+| 1  | Load      | Multiple Azure VMs defined in Conjur                 | Each Azure VM sends authentication requests                               | Check load on Azure and check load on Conjur     | Prod and OSS     |            |
+| 2  | Load      | Multiple Azure VMs defined in Conjur                 | Each Azure VM has invalid Azure tokens and sends authentication requests  | Check load on Azure and check load on Conjur     | Prod and OSS     |            |
+| 3  | Stability | Multiple Azure VMs defined in Conjur                 | Each Azure VM sends authentication requests for a few days                | Ensure no errors are returned                    | Prod and OSS     |            |
```
orenbm: what's Prod? is it DAP?

design/authn_azure/authn_azure_solution_design.md
InbalZilberman: @orenbm as Conjur has access token and Azure has access token this is very confusing. 
As per ID token and access token the design is supporting only access token. we can change it to 'azure_token' to be open for extension in the future or 'azure_jwt'. WDYT?

design/authn_azure/authn_azure_solution_design.md
orenbm: I didn't test this but I don't think there should be any performance issue.

design/authn_azure/authn_azure_solution_design.md
InbalZilberman: Each level of log is. @sigalsax

design/authn_azure/authn_azure_solution_design.md
orenbm: I think that if we decide that this is a better experience than the one defined in `authn-oidc` we should go for it. In the future we can add this ability to oidc as well (with backwards-compatibility with variables).

orenbm: @InbalZilberman wdyt?

design/authn_azure/authn_azure_solution_design.md
InbalZilberman: @orenbm 
1. What would happen if a user, not a host, that is recognized in Azure would use the Azure Authenticator?
2. Do we validate that the provider is an Azure one? can it also be Okta for example? Can it be GKE? (any Oauth 2.0 provider)
3. @shaharglazner  WDYT on the third question?

design/authn_azure/authn_azure_solution_design.md
orenbm: I can live with `azure_token`

design/authn_azure/authn_azure_solution_design.md
orenbm: 1. So the admin defined everything correctly but instead of using `!host` he used `!user`? With all the annotations and permissions?
2. We do not validate such a thing. We do validate that the `xms_mirid` field is present and has the correct application identity so while the validation of the token will pass (as it's an Oauth 2.0 token) it will not pass Conjur's authentication process.

design/authn_azure/authn_azure_solution_design.md
orenbm: also - if `https://sts.windows.net/` is constant, we can let the user input only the tenant id. So we can have 
```
- !variable
        id: tenant-id 
```
And have `https://sts.windows.net/` hard-coded in the authenticator. @InbalZilberman wdyt?

design/authn_azure/authn_azure_solution_design.md
sigalsax: I check that in number 7 ( One or more fields in Azure token is missing which is an invalid token

design/authn_azure/authn_azure_solution_design.md
sigalsax: Removing

design/authn_azure/authn_azure_solution_design.md
sigalsax: > We can have a test that verifies that the retrieved Conjur access token is valid for retrieving a secret in another request but that's a different test.

I will break it up into 2 separate test. I don't know yet how we would go about validation the Conjur access token but if you suggest we check that the access token is valid for a retrieving a secret, why don't we actually do the secret retrieval?

orenbm: You can look at how we did this in OIDC - https://github.com/cyberark/conjur/blob/master/cucumber/authenticators/features/authn_oidc.feature#L43

design/authn_azure/authn_azure_solution_design.md
sigalsax: Updating to "invalid/non-existent provider-uri"  because if it was missing a value that would fail on `RequiredSecretMissing` which is another test

design/authn_azure/authn_azure_solution_design.md
sigalsax: I am confused by your note here. Are we checking the conjur policy being sent in request body or the actual request being sent by the Azure resource. Ex: the resource sends Azure token without "token"

design/authn_azure/authn_azure_solution_design.md
sigalsax: And update the actual log message for `TokenFieldNotFoundOrEmpty` to be 
"xms_mirid field not found or empty"?

sigalsax: I am wary to be specific here b/c I imagine the future we will want to check other fields of tokens, not just `xms_mirid`

design/authn_azure/authn_azure_solution_design.md
orenbm: If i understand correctly the scenario you described, then the host was loaded without all the required annotations. Then an authentication request is sent with a valid Azure token. So the request itself is valid so `MissingRequestParam` seems wrong.

design/authn_azure/authn_azure_solution_design.md
orenbm: I was referring to the `Given` section. The error should still be `TokenFieldNotFoundOrEmpty` as it's a field in the token that is missing, not the field `token`. Let's make sure that we log also the field that is missing.

design/authn_azure/authn_azure_solution_design.md
sigalsax: From my understanding, `Authentication::Authenticate` runs the general authenticator validations and part of this sequence of validations includes `validate_credentials` which calls `valid?` for the relevant authenticator (authn-azure). This function triggers the authenticator-specific validations. 

So my thought was that we need to ensure that when one of the authenticator-specific checks fails, it is propagated. 

Is the act of propagation an integration test?

orenbm: this test is already implemented here: https://github.com/cyberark/conjur/blob/master/spec/app/domain/authentication/authenticate_spec.rb

design/authn_azure/authn_azure_solution_design.md
sigalsax: So here I think this should be 2 separate tests,
1. Check if the 3 annotations we expect to be in Host policy are in fact there (RoleMissingAnnotations)
2. Ensure Azure access token has the fields we expect MissingRequestParam

orenbm: 1. I agree - this test should be in `ValidateApplicationIdentity`. I would name the error with something related to Application Identity just so it's clear what the annotations are.
2. Not quite. `MissingRequestParam` should be raised if the request body is lacking the `token` field, or if it is empty. If the token exists but doesn't have the `xms_mirid` field it should raise something similar to this:
```
TokenFieldNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
          msg: "Field '{0-field-name}' not found or empty in token",
          code: "CONJ00013E"
        )
```

design/authn_azure/authn_azure_solution_design.md
orenbm: I'd like to summarize the options we have:
1. provider uri as variable:
```
- !policy
  id: conjur/authn-azure/prod
  body:
  - !webservice

  - !variable
    id: provider-uri
```
2. Provider URI as annotation:
```
```
- !policy
  id: conjur/authn-azure/prod
  body:
  - !webservice
    annotations:
      provider-uri: https://sts.windows.net/TENANT_ID
```
Some notes:
- UX: In option 1 the user also needs to load a value to the `provider-uri` variable so the UX is better in option 2. 
- Security: Option 1 holds the data as a secret but I don't see a security issue here as this data is not sensitive (but maybe @shaharglazner can correct me). So i think option 1 is fine with its security.
- Performance: It is mainly choosing wether to retrieve a secret from the DB or to retrieve a webservice's annotation (the webservice is already retrieved). I don't see a difference here.
- Implementation: Both options are generally developed. We may need to tweak some of the code but it shouldn't be too much.

In conclusion, i think we should go with option 2. 
@InbalZilberman @sigalsax @shaharglazner i'd like to get your comments on this.

File: design/authn_azure/authn_azure_solution_design.md [hunk 33]
```diff
@@ -0,0 +1,382 @@
...
+
+```yaml
+# policy id needs to match the convention `conjur/authn-azure/<service ID>`
+- !policy
+  id: conjur/authn-azure/prod
+  body:
+  - !webservice
+  
+  - !variable
+    id: provider-uri # https://sts.windows.net/TENANT_ID
```
orenbm: @InbalZilberman @shaharglazner @sigalsax wdyt about changing this from `provider-uri` to `tenant-id`. This is relevant only id the `https://sts.windows.net/` is always constant. This way `https://sts.windows.net/` will be hard-coded in the authenticator and the user will only need to input the tenant-id.

design/authn_azure/authn_azure_solution_design.md
orenbm: I think that @shaharglazner means that the example should be a GUID. Something like:
```
authn-azure/system-assigned-identity: 853b9a84-5bfa-4b22-a3f3-0b9a43d9ad8a
```

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,382 @@
...
+Steps 1 & 2 may seem familiar as they are the exact steps that we perform in the 
+OIDC Authenticator. This is because OIDC is Oauth 2.0 based. We will take advantage 
+on this and use the 3rd party used in the OIDC Authenticator to perform steps 
+1 & 2 in our authenticator.
+
+Unfortunately, the 3rd party used in the OIDC Authenticator does not implement 
+an access token decoding, and only an ID Token decoding. Thus, to perform step 3, we will need another 3rd 
+party - the ["jwt" gem](https://github.com/jwt/ruby-jwt). This gem has an MIT License which is aligned with our policy.
+
+As you can see in the following blueprint for the `validate_azure_token` implementation, it will use `AuthnOidc`'s `FetchProviderCertificate` class to retrieve the JWKs and will use them to decode the token, using `JWT`'s `decode` function:
```
sigalsax: I am missing here what the next steps are after decoding. We have a  decoded token and now we must verify it. How do we verify this against the Azure AD?

orenbm:
@boazmichaely please see comments:
> Can you please show a detailed flow of the developer code (which includes fetching the token), for both the system assigned and the user assigned flow. Specifically, how does the developer get the string which s/he needs to put in the REST header?

This is more of a functional flow than a part of a solution design. You can take this with @InbalZilberman . You can also find more details here: https://ca-il-confluence.il.cyber-ark.com/display/rndp/Conjur+Azure+Authenticator+-+Research+page#/

> Similarly can you show a detailed pipeline admin code. How does the pipeline admin know which fields to load into the annotation? Again for both user and system assigned.

Same here. Please take it with @InbalZilberman .

> Finally (for now :-) would be helpful to show the content of the "xms_mirid" as @shaharglazner also pointed out. Maybe a table that maps those fields to annotations ?

@sigalsax kindly did that :)

design/authn_azure/authn_azure_solution_design.md
orenbm: thanks @nessiLahav. I didn't add it to the doc to leave the code simple but i'll add it.

design/authn_azure/authn_azure_solution_design.md
orenbm: @sigalsax @InbalZilberman let's get to a decision here.

design/authn_azure/authn_azure_solution_design.md
orenbm: I'll elaborate in the doc itself for visibility but the verification is part of the decoding process. The `JWT.decode` also validates the token with the keys retrieved from the provider and the issuer.

design/authn_azure/authn_azure_solution_design.md
shaharglazner: @orenbm @sigalsax @InbalZilberman @andrew
As far as I can find, tenant id is not sensitive - 

- https://stackoverflow.com/questions/45661109/are-azure-subscription-id-aad-tenant-id-and-aad-app-client-id-considered-secret

- https://stackoverflow.com/questions/56016004/are-the-azure-client-id-tenant-and-key-vault-uri-considered-secrets

design/authn_azure/authn_azure_solution_design.md
orenbm: also - if we change this we shouldn't forget to change it in the diagram

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,449 @@
...
+| 1  | authn-azure is not whitelisted in ENV or enabled in DB                                       | VM sends authentication request to Conjur with valid Azure token    | Return 401 Unauthorized error and `AuthenticatorNotWhitelisted` message appears in log   | Integration               | - [ ]            |
+| 2  | Policy is loaded w/o `provider-uri` variable                                                 | VM sends authentication request to Conjur with valid Azure token    | Return 401 Unauthorized error and `RequiredResourceMissing` message appears in log       | Integration               | - [ ]            |
+| 2  | Policy is loaded w/o `provider-uri` value                                                    | VM sends authentication request to Conjur with valid Azure token    | Return 401 Unauthorized error and `RequiredSecretMissing` message appears in log         | Integration               | - [ ]            |
```
orenbm: nit: The policy is never loaded with a value. You can change to `Policy is loaded with `provider-uri` variable but a value is never loaded to it`. You can play with the wording.

design/authn_azure/authn_azure_solution_design.md
AndrewCopeland: As a heads up, you cannot add annotations to a resources that does not have an ID 😢 
```
- !webservice
  annotations:
    provider-uri: https://sts.io
```
When the policy is loaded we get the following error:
```
{"error":{"code":"validation_failed","message":"policy_text webservice has a blank id","details":[{"code":"validation_failed","target":"policy_text","message":"webservice has a blank id"}]}}
```

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,242 @@
...
+
+1. Azure VM sends the following request to Conjur:
+    ```yaml
+    POST https://<DAP-server-hostname>/authn-azure/<service-id>/<account>/<host-id>authenticate
+    Header 
+    
+    Content-Type: application/x-www-form-urlencoded
+    Body 
+    
+     token: "eyJhbGciOiJSUzI1NiIs......uTonCA"
```
AndrewCopeland: How is this token obtained?
From this endpoint `http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https://management.azure.com/`

And if so are we parsing the `access_token` from this response and sending it to conjur?
If thats the case the access_token appears to be an azure specific JWT.
Since it is a JWT why not just call it `jwt`?

AndrewCopeland: Nevermind. I guess there is a difference between access_token and id_token mentioned here:
https://developer.okta.com/docs/guides/validate-access-tokens/overview/#access-tokens-vs-id-tokens

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,474 @@
...
+
+- !grant
+  role: !group conjur/authn-azure/prod/apps
+  member: !host azure-apps/test-app
+```
+
+Note that such a host can have only one of the annotations 
+`authn-azure/user-assigned-identity` & `authn-azure/system-assigned-identity`
+
+#### Multiple-assigned identity for host - NOT FOR NOW
```
Tovli: Is this planned for any future phase?

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,474 @@
...
+
+We may allow hosts to be defined with both "authn-azure/user-assigned-identity" and "authn-azure/system-assigned-identity". This way such a host can 
+authenticate a token with a user-assigned id as well as a token with a system-assigned id. As a first step we will not develop - nor test - such a use-case.
+
+### Design
+
+#### Authentication Flow
+
+This flow assumes that the VM already exists in Azure, and that an authn-azure authenticator and a Conjur host were loaded.
```
Tovli: Worth mentioning that while we expect the app to be on the VM we currently won't officially support having conjur or the DAP follower in Azure

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,474 @@
...
+provider. Its URl will be retrieved from the "provider-uri" variable defined in the authenticator policy
+1. Retrieve JWKs from the discovery endpoint - These keys are used for the token validation
+1. Decode and verify the token using the JWKs.
+
+Steps 1 & 2 may seem familiar as they are the exact steps that we perform in the 
+OIDC Authenticator. This is because OIDC is Oauth 2.0 based. We will take advantage 
+on this and use the 3rd party used in the OIDC Authenticator to perform steps 
+1 & 2 in our authenticator.
+
+Unfortunately, the 3rd party used in the OIDC Authenticator does not implement
```
Tovli: Does that mean you don't plan to consolidate that part of the flow?

design/authn_azure/authn_azure_solution_design.md
sigalsax: 1. I would imagine having !user instead of !host will not break anything because in conjur user and host and group and layer are interchangeable and are more constructs to assist the customer

design/authn_azure/authn_azure_solution_design.md
sigalsax: You don't have to give the webservice an ID if you have a policy like this
```
- !policy 
  id: policytest
  body:
    - !webservice
      annotations:
        provider-uri: https://sts.io
```
B/c the webservice will take the name of the policy ID
Ex: `conjur list` 
>"cucumber:webservice:policytest"

design/authn_azure/authn_azure_solution_design.md
InbalZilberman: @sigalsax Thank you for this clarification. I agree.

design/authn_azure/authn_azure_solution_design.md
InbalZilberman: @sigalsax @orenbm @Tovli let's go with a simple jwt

design/authn_azure/authn_azure_solution_design.md
orenbm: @sigalsax so we will go with the annotation?

design/authn_azure/authn_azure_solution_design.md
orenbm: I updated this in the doc

orenbm:
@boazmichaely regarding your last question - 
> Speaking of that, in your example
authn-azure/user-assigned-identity: test-app-pipeline
how is it that the identity is a string and how is this string resolved against what is found in the JWT?

@sigalsax addressed this in the design.

design/authn_azure/authn_azure_solution_design.md
orenbm: @sigalsax - final decision with @InbalZilberman is `jwt`.

design/authn_azure/authn_azure_solution_design.md
orenbm: The general validation are ignorant to the role type (host/user). We can strive to do that also in Azure if there is a business need for that. I don't see any implementation blocker on this.

@sigalsax - please update this in the doc when you have the time. Don't forget about test plan changes. 
@InbalZilberman  ^

design/authn_azure/authn_azure_solution_design.md
orenbm: Updating for visibility - we will stay with provider-uri because in some cases the prefix will not be `https://sts.windows.net/`.

@shaharglazner @InbalZilberman do you have any examples for this?

design/authn_azure/authn_azure_solution_design.md
orenbm: Not really. Do you think it's better to remove this altogther? I believe that having it here let's the reader understand that this use-case was considered and was decided not to be implemented atm.

design/authn_azure/authn_azure_solution_design.md
orenbm: @sigalsax ^

orenbm: No need to mention the DAP Follower (but it's ok to mention it if you think it's better)

design/authn_azure/authn_azure_solution_design.md
orenbm: @Tovli i don't understand your question

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,508 @@
...
+request is sent to Conjur, the admin will load the authenticator policy: 
+
+```yaml
+# policy id needs to match the convention `conjur/authn-azure/<service ID>`
+- !policy
+  id: conjur/authn-azure/prod
+  body:
+  - !webservice
+    annotations: 
+      provider-uri: https://sts.windows.net/TENANT_ID
```
orenbm: @sigalsax i would change this to 
```
provider-uri: <Azure AD URI> # For example: https://sts.windows.net/df242c82-fe4a-47e0-b0f4-e3cb7f8104f2/
```

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,508 @@
...
+    role: !group apps
+    privilege: [ read, authenticate ]
+    resource: !webservice
+```
+
+and add the `provider-uri` annotation to the webservice (e.g `https://sts.windows.net/TENANT_ID`). 
+
+As part of our effort to improve the user experience for our authenticators, we discussed two options for defining the `provider-uri`:
+ 
+1. `provider-uri` as a variable_+
```
orenbm: @sigalsax is `variable_+` a typo?

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,508 @@
...
+    - !policy
+      id: conjur/authn-azure/prod
+      body:
+        - !webservice
+          annotations: 
+            provider-uri: https://sts.windows.net/TENANT_ID
+          
+    ```     
+
+Because the `provider-uri` does not hold sensitive information, it does not need to saved in Conjur. Adding the `provider-uri` as a variable demands later that the
```
orenbm: ... to be saved as a secret in Conjur...

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,508 @@
...
+admin user populate its value through the CLI. Supplying `provider-uri` as annotation entry makes it easier to set it as a key-value pair and is therefore more readable. 
+For these reasons, we have decided to add provide the `provider-uri`.
```
orenbm: I would also add that there is no increase in performance. Because the webservice itself is already retrieved  then it's a difference between retrieving an annotation from the DB and retrieving a secret from the DB.

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,508 @@
...
+ 
+### Azure Application Identity
+
+To authenticate with Conjur, Azure-specific fields will need to be provided in the host annotations of the Conjur host identity for the Azure resource- subscription-id, resource-group, and user/system-assigned-identity. Specifically, If the Azure resource is assigned a system-assigned-identity, 
+we will also need to validate the `oid` field of the Azure access token. These fields will be used to confirm the identity of the Azure resource that is
+sending the authentication request. This will be done by comparing the Azure access token information with the host annotations that are defined in the policy below.
```
orenbm: In this section i would add the levels of granularity that we support:
1. Subscription ID + Resource Group
2. Subscription ID + Resource Group + System Assigned Identity
3. Subscription ID + Resource Group + User Assigned Identity

We do not support a combination of System Assigned Identity & User Assigned Identity

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,508 @@
...
+    ```yaml
+    POST https://<conjur-server-hostname>/authn-azure/<service-id>/<account>/<host-id>/authenticate
+    Header 
+    
+    Content-Type: application/x-www-form-urlencoded
+    Body 
+    
+     jwt: "eyJhbGciOiJSUzI1NiIs......uTonCA"
+    ```
+
```
orenbm: let's change to this (just played with newlines):

```yaml
    POST https://<conjur-server-hostname>/authn-azure/<service-id>/<account>/<host-id>/authenticate

    Header  
        Content-Type: application/x-www-form-urlencoded

    Body 
        jwt: "eyJhbGciOiJSUzI1NiIs......uTonCA"
```

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,508 @@
...
+      end
+
+      private
+
+      ...
+
+    end
+ ```
+ 
+![Authn Azure Flow](authn-azure-flow.png)
```
orenbm: this diagram is duplicated

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,508 @@
...
+    for greater security granularity, a third host annotation can exist - `authn-azure/system-assigned-identity` or `authn-azure/user-assigned-identity`. 
+
+1. If a third annotation, `authn-azure/system-assigned-identity` or `authn-azure/user-assigned-identity` is present, the process of 
+validating this annotation entry differs depending on the identity assigned to the Azure resource. 
+Because of this, we need to handle each case differently. The two types of identities that can be assigned to an Azure resource include the following:
+
+    1. system-assigned identity
+
+    1. user-assigned identity
+
```
orenbm: Beautiful!

design/authn_azure/authn_azure_solution_design.md
```diff
@@ -0,0 +1,508 @@
...
+1. As you can see in the "Examples" section below, the suffix of the `xms_mirid` field includes
+the assigned identity. If the suffix includes `Microsoft.ManagedIdentity/userAssignedIdentities/<user_assigned_identity>`
+then it is a user assigned identity. In this case we will compare its value (i.e `<user_assigned_identity>`) 
+with the `authn-azure/user-assigned-identity` annotation defined in the Conjur host.
+
+1. Otherwise, we will check if `providers/Microsoft.Compute/virtualMachines/` is present in the `xms_mirid` claim. 
+If so, we will compare the `oid` field from the Azure access token with the 
+`authn-azure/user-assigned-identity` annotation in the Conjur host.
+  1. Note: At this point we support only VMs. In the future we can check for other Azure resources in the `xms_mirid` suffix.
```
orenbm: This shouldn't be numbered. It's just a note, not a step in the flow

design/authn_azure/authn_azure_solution_design.md
InbalZilberman: Regarding 1 let's decide if this use case is in or out of scope.

orenbm: let's go for it. Should be an easy win.

design/authn_azure/authn_azure_solution_design.md
sigalsax: Changing it back to variable as per solution design meeting

design/authn_azure/authn_azure_solution_design.md
sigalsax: @InbalZilberman for stability tests do we want to measure requests over a few days?

sigalsax: Also, 20 requests per vm?

design/authn_azure/authn_azure_solution_design.md
sigalsax: We decided against annotation for maintenance concerns and will readdress this at a later date

design/authn_azure/authn_azure_solution_design.md
sigalsax: I think we need both 
1. Check read only and check authn only
2. Check if group has permissions

design/authn_azure/authn_azure_solution_design.md
sigalsax: Added!

design/authn_azure/authn_azure_solution_design.md
sigalsax: If I am understanding correctly, there is no increase in performance because regardless we need to retrieve info from the DB, be it an webservice or a variable?

sigalsax: When retrieving an entry from DB do we also get there annotations? Or is that a separate table?

PR #1292, opened 2020-01-31
Adding OSS Suite Release design doc

izgeri:
Also updates design/README to be more clear about what's included in the design
folder

### DIFF
--- a/design/oss_suite_release.md
+++ b/design/oss_suite_release.md
@@ -1,0 +1,429 @@
+# Conjur OSS Suite Release
+
+## Motivation
+In this effort, we intend to build a formal release process for the _Conjur OSS Suite_,
+which we define as the set of components including the Conjur server, deployment utilities,
+command line interface (CLI), client libraries, and integrations with platforms and DevOps tools.
+
+At current, we have no formal process for releasing the full suite of tools in
+the Conjur OSS ecosystem so that:
+- It is clear to end users which version of Conjur OSS they should use, and what
+  integrations and tooling are compatible with that version.
+- Key use cases that involve more than two components working together are automatically
+  tested to ensure that the end-to-end experience in those use cases works as expected.
+- Development can continue on the master branch of individual components while
+  end users have a reliable way to pull "stable" releases for all components and
+  avoid unexpectedly running "edge" releases .
+- It is clear what features are officially released and which are still in progress -
+  at current, there is no way to easily track unreleased features in our broad suite
+  of tools and integrations, and formally creating a system for tracking unreleased
+  features will help us to do a better job of comprehensively testing and releasing
+  completed features.
+
+In this effort, we will build a formal Conjur OSS Suite Release process to begin
+addressing these concerns.
+
+## Desired Outcome
+There is an initial Conjur OSS Suite Release that validates compatibility of the
+included components. In addition, the infrastructure and tooling exists for creating
+future OSS Suite Releases that may include additional automated tests and components.
+
+## Acceptance Criteria
+- All key components of the Conjur OSS Suite will be grouped into a single
+  Conjur OSS Suite Release.
+- There is a special [release repository](https://github.com/cyberark/conjur-oss-suite-release)
+  that is used to:
+  - Pin the versions of the components included in the suite release.
+  - Add test case for end-to-end flows involving components of the OSS suite.
+  - Automatically run test suite each day against component versions pinned in
+    the suite release.
+  - When the pinned component versions are changed and the suite version is
+    incremented:
+    - Automatically generate GitHub releases that include release notes with links
+      to artifacts.
+    - Automatically generate a documentation-friendly release page that includes
+      a table of contents, release notes for each component, and links to artifacts.
+  - Document how end users should utilize the Conjur OSS Suite Release.
+
+In addition, as part of building this release, we'll also be creating:
+- A [GitHub landing page](https://cyberark.github.io/conjur) that links to the
+  repos included in the Conjur OSS Suite and describes what each one does.
+- A template to be added to the READMEs of repos in the Conjur OSS Suite to clarify
+  about the state of the code on master ("edge") and to encourage end users to
+  visit the Conjur OSS Suite Release to determine which versions of each product
+  to use.
+- A tool for identifying the unreleased changes in the components of the Conjur OSS
+  Suite Release, to help us in understanding whether a new OSS Suite Release is
+  needed.
+
+## Repositories Included in the Conjur OSS Suite Release
+- Conjur OSS Core
+  - [Conjur OSS](https://github.com/cyberark/conjur)
+    - [Conjur OSS Helm Chart](https://github.com/cyberark/conjur-oss-helm-chart)
+    - [Conjur OSS Cloudformation Templates](https://github.com/cyberark/conjur-aws)
+  - [Conjur CLI](https://github.com/cyberark/conjur-cli)
+  - Client Libraries
+    - [Golang](https://github.com/cyberark/conjur-api-go)
+    - [Java](https://github.com/cyberark/conjur-api-java)
+    - [.Net](https://github.com/cyberark/conjur-api-dotnet)
+    - [Python 3](https://github.com/cyberark/conjur-api-python3)
+    - [Ruby](https://github.com/cyberark/conjur-api-ruby)
+- Integrations
+  - [Ansible](https://github.com/cyberark/ansible-conjur-host-identity)
+  - Cloud Foundry
+    - [Service broker](https://github.com/cyberark/conjur-service-broker)
+    - [Buildpack](https://github.com/cyberark/cloudfoundry-conjur-buildpack)
+  - [Jenkins](https://github.com/cyberark/conjur-credentials-plugin)
+  - Kubernetes / OpenShift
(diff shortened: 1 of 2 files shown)


design/oss_suite_release.md
```diff
@@ -0,0 +1,300 @@
...
+  in the GitHub READMEs, which by custom in open source contains basic usage info).
+- Published contributor guidelines are available in a central place in GitHub.
+
+## User Stories
+
+### Open Source Users
+
+|||
+|---|---|
+|AS A|Conjur Open Source user|
```
alexkalish: I think that the ALL CAPS for template words is a little too screamy.  Maybe use italics?

design/oss_suite_release.md
```diff
@@ -0,0 +1,300 @@
+# Conjur OSS Suite Release
+
+## Feature Overview
```
alexkalish: I like calling the first section "Motivation" or maybe "Outcome".  Providing some type of overview (as you do below) is needed, but I feel like it's often more important to answer the "why" about a project.

design/oss_suite_release.md
```diff
@@ -0,0 +1,300 @@
...
+- It is clear what features are officially released and which are still in progress -
+  at current, there is no way to easily track unreleased features in our broad suite
+  of tools and integrations, and formally creating a system for tracking unreleased
+  features will help us to do a better job of comprehensively testing and releasing
+  completed features.
+
+In this effort, we will build a formal Conjur OSS Suite Release process to begin
+addressing these concerns.
+
+## Key Characteristics of Conjur OSS Suite Release
```
alexkalish: As we discussed, this could probably be changed to "Acceptance Criteria".  Also, we might want to remove some of the implementation details and move them down to the design section so that this focuses on "what" and not "how".

design/oss_suite_release.md
```diff
@@ -0,0 +1,300 @@
...
+- A [GitHub landing page](https://cyberark.github.io/conjur) that links to the
+  repos included in the Conjur OSS Suite and describes what each one does.
+- A template to be added to the READMEs of repos in the Conjur OSS Suite to clarify
+  about the state of the code on master ("edge") and to encourage end users to
+  visit the Conjur OSS Suite Release to determine which versions of each product
+  to use.
+- A tool for visualizing the unreleased changes in the components of the Conjur OSS
+  Suite Release, to help us in understanding whether a new OSS Suite Release is
+  needed
+
```
alexkalish: Can you add a "Definition of Done" header, probably here, just so that we don't forget to add this content?

design/oss_suite_release.md
```diff
@@ -0,0 +1,300 @@
...
+    - [Summon Chef Databags Provider](https://github.com/cyberark/summon-chefapi)
+    - [Summon Conjur Provider](https://github.com/cyberark/summon-conjur)
+    - [Summon Keyring Provider](https://github.com/cyberark/summon-keyring)
+    - [Summon S3 Provider](https://github.com/cyberark/summon-s3)
+
+## Considerations
+
+### How CyberArk Dynamic Access Provider (DAP) Will Use the OSS Release
+At current, when a new DAP appliance is built it is based on a manually pinned
+version of Conjur OSS. We propose implementing automation as part of this release
```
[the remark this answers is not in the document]
izgeri: The proposed solution (and the solution we go with is still TBD) is to have a bot update the version of Conjur pinned in DAP and submit a PR (that would include automatically running the test suite, presumably).

design/oss_suite_release.md
izgeri: as discussed, I'll leave this off the doc and add it to the epic later - I'd like to define the DoD once this doc is approved.

design/oss_suite_release.md
izgeri: we didn't talk about the implementation details here that you're suggesting we move - it would be good to touch base on this directly.

design/oss_suite_release.md
```diff
@@ -0,0 +1,300 @@
...
+    to its database via Secretless using credentials pulled from Conjur.
+  - Individual repos are responsible for running their own integration tests against
+    Conjur. The suite release test suite will enable integration testing multiple
+    components together (not just Component X + Conjur), and as part of the initial
+    effort we'll implement one such use case and ensure we're set up to add other
+    such use cases as we identify them in the future.
+  - Should the scope of this project include evaluating the component repositories
+    to determine the state of the automated tests and whether additional integration
+    tests need to be added?
+  - When a new suite release is being prepared, is there a way to run the automated
```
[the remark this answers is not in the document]
izgeri: I left this open for now - this is worth discussing, and it's TBD how we would implement this

design/oss_suite_release.md
```diff
@@ -0,0 +1,313 @@
...
+
+## High-Level Technical Design
+
+The [OSS Suite Release Repo](https://github.com/cyberark/conjur-oss-suite-release)
+will include:
+- A user-facing README with a high level overview of how Conjur OSS end users
+  should leverage the repo, with links to appropriate pages in the official documentation
+- A VERSION file that includes the current version of the OSS suite release
+- A machine-readable file with the pinned versions included in the current release
+- Version tags for each OSS suite release version that correspond to GitHub releases
```
andytinkham: Is an OSS suite release a new thing starting with this, or will version tags be put in for past releases for customers who are still using older components, perhaps?

izgeri: it's new! we'll be deciding what version to call it and which component versions to include in the course of implementation. we hope that current users will find the suite release a helpful way to determine which component versions to upgrade to, if they're considering an upgrade. definitely something I'd be glad to talk about on [discourse](https://discuss.cyberarkcommons.org) with end users anytime, too.

alexkalish:
Some thoughts/questions following our signoff meeting:

1. Regarding upgrades, can changelog standard be to put upgrade needs in [Changed] entries?
2. Can we re-version components that go into the suite with the suite release number?
3. Can we identify canonical download links for each components so that users do not need to link through to each component’s README?
4. Can we make it clear how the GitHub UX is different from the Docs UX?

design/oss_suite_release.md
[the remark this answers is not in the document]
diverdane: I think this is what you're suggesting... but I would envision (as a future goal) a process where the "candidate release" testing for each component is done on a regular basis, and maybe a test of a candidate release of the entire suite is done e.g. weekly, based on latest candidate releases of each component, so that we're not surprised when it comes time for a release.

izgeri:
@alexkalish -

1. We follow a standard for changelog formatting - https://keepachangelog.com/en/1.0.0/. In that
   standard, it defines `Changed` as
   > for changes in existing functionality. 
   It might be that we just need a special section for custom upgrade instructions. I've started to
   think that the release repo should have its own custom CHANGELOG.md that includes such
   things as custom sections on:
   - notable new features / release highlights
   - special upgrade instructions
   - known issues
   - ?
   where all of these items are included in the autogenerated release notes
1. I can see why this would be especially simple from an end user perspective, but I'm not sure
   it's possible. Whatever version we pick for the OSS suite release may conflict with an existing
   version of any component. Do you have any other ideas about standard naming conventions we
   might use to help with this?
1. For many things this is possible. For example, any artifact hosted in an external registry that
   is not a GitHub release, we can link to the artifacts external registry page for the specific
   version. For homebrew and our helm charts, we can make it as simple as running an `upgrade`
   command to get the latest. There are some components that will be trickier, however, and
   we'll have to evaluate on a case-by-case basis.
1. I will take a stab at working this into the requested UX improvements to the design doc today.

alexkalish:
@izgeri: 
1. Custom sections could be good.  However, even just a policy that special upgrade instructions need to be called out in `Changed` might be enough to start.
2. For container images, we can have multiple tags and namespace the ones for the suite, right?  I'm just not sure about all the other different package types.  I'm uncertain if it makes sense to do this for some components but not others?
3. Great.  Can this be added?  And I totally understand it won't always be easy/possible to provide a direct download link.
4. Thanks!

design/oss_suite_release.md
```diff
@@ -0,0 +1,429 @@
...
+  - Document how end users should utilize the Conjur OSS Suite Release.
+
+In addition, as part of building this release, we'll also be creating:
+- A [GitHub landing page](https://cyberark.github.io/conjur) that links to the
+  repos included in the Conjur OSS Suite and describes what each one does.
+- A template to be added to the READMEs of repos in the Conjur OSS Suite to clarify
+  about the state of the code on master ("edge") and to encourage end users to
+  visit the Conjur OSS Suite Release to determine which versions of each product
+  to use.
+- A tool for identifying the unreleased changes in the components of the Conjur OSS
```
hughsaunders: If possible I'd like to see a tool that covers the general diff case, allowing any two versions of the OSS suite release can be diffed. 

There are a few use cases for this:
* Generating release notes (showing only changes between the old and new releases specififed)
* Determining which changes an upgrade will bring, especially when skipping versions. 
* Determining unreleased changes (diff latest release against master)

Prior art: https://github.com/rcbops/osa_differ which generates a mark down report showing the differences between two openstack-ansible releases, including changes in all the sub projects.

izgeri: that sounds like a "nice to have", but based on the spike we're working on (cyberark/conjur-oss-suite-release#2) I think the team is considering this aspect!

PR #1294, opened 2020-02-04
Test `authentication-container-name` annotations is defaulted to `authenticator`

orenbm:
#### What does this PR do?

Adds a test where we create a host that doesn't have the annotation `authentication-container-name` and send an authentication request from a container named `authenticator`. This test validates that in case the annotation is not present then the container name is defaulted to `authenticator`.

### DIFF
--- a/ci/authn-k8s/dev/policies/policy.template.yml
+++ b/ci/authn-k8s/dev/policies/policy.template.yml
@@ -189,6 +189,11 @@
         authn-k8s/service-account: other-service-account
         authn-k8s/authentication-container-name: authenticator
 
+    - !host
+      id: test-app-no-container-annotation
+      annotations:
+        authn-k8s/namespace: {{CONJUR_AUTHN_K8S_TEST_NAMESPACE}}
+
   - !grant
     role: !layer
     members: *some-policy-hosts
--- a/ci/authn-k8s/dev/policies/policy.template.yml
+++ b/ci/authn-k8s/dev/policies/policy.template.yml
@@ -203,6 +203,29 @@
   role: !group conjur/authn-k8s/minikube/clients
   member: !layer some-policy
 
+# This policy is for testing that in case a host doesn't have any annotations defined
+# then the "authentication-container-name" annotation is defaulted to "authenticator"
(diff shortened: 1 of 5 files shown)


ci/authn-k8s/dev/policies/policy.template.yml
```diff
@@ -198,6 +203,31 @@
...
+# then the "authentication-container-name" annotation is defaulted to "authenticator"
+# and the host is authenticated.
+# We have a separate policy for this host so we can use the id '{{CONJUR_AUTHN_K8S_TEST_NAMESPACE}}/*/*'
+# without duplication.
+- !policy
+  id: host-without-container-name
+  body:
+    - !layer
+
+    - &hosts
```
jtuttle: Build is currently failing with:

```
[2020-02-04T18:35:44.294Z] {"error":{"code":"validation_failed","message":"policy_text Error at line 221, column 5 in  : Duplicate anchor hosts","details":[{"code":"validation_failed","target":"policy_text","message":"Error at line 221, column 5 in  : Duplicate anchor hosts"}]}}
```

I think you have to rename this `&hosts` anchor. Maybe change it to `&host-without-container-name-hosts`. If you look at the policy above, it does a similar thing.

ci/authn-k8s/dev/policies/policy.template.yml
```diff
@@ -198,6 +203,31 @@
...
+      - !host
+        id: {{CONJUR_AUTHN_K8S_TEST_NAMESPACE}}/*/*
+        annotations:
+          kubernetes/authentication-container-name: authenticator
```
jtuttle: This host shouldn't have the annotation defined, right?

ci/authn-k8s/dev/policies/policy.template.yml
orenbm: correct.

ci/authn-k8s/dev/policies/policy.template.yml
orenbm: correct again

Issue #1306, opened 2020-02-09
Authn Azure - retrieve provider-uri value in authn-azure class


orenbm:
In order to validate the Azure Access Token we need the provider URI that was stored in a variable under the authenticator policy.

We should use the mechanism developed in OIDC for retrieving secrets.

DOD
- [x] Implementation of how to retrieve provider-uri from DB has been completed
- [ ] PR has been reviewed and approved

PR #1314, opened 2020-02-09
Retrieve provider-uri value in authn-azure class

orenbm:
#### What does this PR do?

Retrieves the `provider-uri` variable value into the Azure authenticator class so it is available for token validation.

Connected to #1306

### DIFF
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -10,10 +10,12 @@ module Authentication
 
     Authenticator = CommandClass.new(
       dependencies: {
-
+        fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
+        logger:                      Rails.logger
       },
       inputs: [:authenticator_input]
     ) do
+
       extend Forwardable
       def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :credentials
 
@@ -30,6 +32,23 @@ module Authentication
 
       def validate_application_identity
 
+      end
+
+      def provider_uri
+        azure_authenticator_secrets["provider-uri"]
+      end
+
+      def azure_authenticator_secrets
+        @azure_authenticator_secrets ||= @fetch_authenticator_secrets.(
+          service_id: service_id,
+          conjur_account: account,
+          authenticator_name: authenticator_name,
+          required_variable_names: required_variable_names
+        )
+      end
+
+      def required_variable_names
+        @required_variable_names ||= %w(provider-uri)
       end
     end
 
--- a/app/domain/authentication/authn_oidc/authenticate.rb
+++ b/app/domain/authentication/authn_oidc/authenticate.rb
@@ -23,6 +23,9 @@ module Authentication
       inputs:       %i(authenticator_input)
     ) do
 
+      extend Forwardable
+      def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :webservice
+
       def call
         validate_account_exists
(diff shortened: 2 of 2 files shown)


app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -31,6 +32,23 @@ def validate_azure_token
...
+        @azure_authenticator_secrets ||= @fetch_authenticator_secrets.(
+          service_id: service_id,
+          conjur_account: account,
+          authenticator_name: authenticator_name,
+          required_variable_names: required_variable_names
+        )
+      end
+
+      def required_variable_names
+        @required_variable_names ||= %w(provider-uri)
```
sigalsax: I see `||=` in many places in our codebase. If I understand correctly, this statement reads
 `required_variable_names || required_variable_names = %w(provider-uri)`, where `required_variable_names` is only set if `required_variable_names` is logically false (i.e. if it's nil or false). 

Would `required_variable_names` ever be given a value elsewhere? Why do we do this in place of just `required_variable_names = %w(provider-uri)` b/c our goal is to update the `required_variable_names` to be the `provider-uri`

orenbm: This is actually memoization. It sets the value if it is not set already. We use it in our code so calculations will run only once.

sigalsax: Also why `%w` for `%w(provider-uri)` and not `%s`

To try to answer my own question, a couple of possible reasons:
1. Unity across our authenticators (this is how do it authn-oidc)
2. Set ourselves up for if we need to add another variable in the future.
3. b/c in `fetch_authenticator_secrets.rb`, we expect `required_variable_names` variable to be an array

orenbm: 3 is correct. We need an array of strings. `%s` will create symbols. Read more here: https://simpleror.wordpress.com/2009/03/15/q-q-w-w-x-r-s/

sigalsax: ^ changed it for `%q`

app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -57,14 +60,14 @@ def add_username_to_input
       end
```
sigalsax: I think you missed adding `update` in `def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :webservice`

orenbm: I didn't add it on purpose. The reason we use `def_delegators` is for readability. Someone who reads this class will know easily what are the inputs for this CommandClass, without the need to dig into `authenticator_input`. adding a delegator for `update` will not improve readability as a reader will not understand the usage of this method:
```
def add_username_to_input
      @authenticator_input = @authenticator_input.update(username: conjur_username)
end
```
 if it will be written as:
```
def add_username_to_input
      @authenticator_input = update(username: conjur_username)
end
```

So it's better to leave this method non-delegated and call ` @authenticator_input` explicitly.

app/domain/authentication/authn_azure/authenticator.rb
orenbm: Still the same answer - we need an array of strings in `fetch_authenticator_secrets`.

orenbm:
> let's follow the Boy Scout rule and "leave the campground cleaner than we found it" and also add
def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :credentials

Love the approach! I added this in another commit.

PR #1317, opened 2020-02-11
Add test script for authn-azure

orenbm:
This script will deploy Conjur in an AWS machine, load policies into
Conjur (authenticator policy, azure-app & a test var). It will then
send authn-azure requests from the Azure VM to Conjur to get an access token
and use it to retrieve a secret.

This commit adds the part for Conjur (deploy & load policies).

### DIFF
--- a/ci/authn-azure/setup-conjur.sh
+++ b/ci/authn-azure/setup-conjur.sh
@@ -1,0 +1,50 @@
+#!/bin/bash -ex
+
+sudo docker load -i conjur-image.tar
+
+# Remove old containers if they are up
+if [[ -d "conjur-quickstart" ]]
+then
+  pushd conjur-quickstart
+  sudo docker-compose down -v
+  popd
+fi
+
+rm -rf conjur-quickstart
+git clone --single-branch --branch local-conjur-image https://github.com/cyberark/conjur-quickstart.git
+cd conjur-quickstart
+sudo docker-compose down -v
+
+sed "s#{{ CONJUR_IMAGE_VERSION }}#$CONJUR_IMAGE_VERSION#g" ./Dockerfile.template > ./Dockerfile
+sudo docker-compose pull
+sudo docker-compose build
+sudo docker-compose run --no-deps --rm conjur data-key generate > data_key
+
+export CONJUR_DATA_KEY="$(< data_key)"
+echo "Generated CONJUR_DATA_KEY $CONJUR_DATA_KEY"
+
+# TODO: Make it work without sed. docker-compose should take the host's env var if it is not set but it doesn't work
+sed "s#{{ CONJUR_DATA_KEY }}#${CONJUR_DATA_KEY}#" ./docker-compose.yml.template > ./docker-compose.yml
+
+sudo docker-compose up -d
+
+# Wait for all components to be up before we create the account
+sleep 5
+sudo docker-compose exec -T conjur conjurctl account create cucumber
+
+sudo docker-compose exec -T client conjur init -u conjur -a cucumber
+ADMIN_API_KEY=$(sudo docker-compose exec -T conjur conjurctl role retrieve-key cucumber:user:admin | tr -d '\r')
+sudo docker-compose exec -T client conjur authn login -u admin -p $ADMIN_API_KEY
+
+# Copy the policy files into the directory that is volumed into the client
+cp ../policies/* conf/policy/
+
+sudo docker-compose exec -T client conjur policy load root policy/authn-azure.yml
+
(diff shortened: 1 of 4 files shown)


ci/authn-azure/setup-conjur.sh
```diff
@@ -0,0 +1,50 @@
...
+if [[ -d "conjur-quickstart" ]]
+then
+  pushd conjur-quickstart
+  sudo docker-compose down -v
+  popd
+fi
+
+rm -rf conjur-quickstart
+git clone --single-branch --branch local-conjur-image https://github.com/cyberark/conjur-quickstart.git
+cd conjur-quickstart
```
sigalsax: For uniformity, let's choose "cd" or pushd"

ci/authn-azure/setup-conjur.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "Generated CONJUR_DATA_KEY $CONJUR_DATA_KEY"
+
+# TODO: Make it work without sed. docker-compose should take the host's env var if it is not set but it doesn't work
+sed "s#{{ CONJUR_DATA_KEY }}#${CONJUR_DATA_KEY}#" ./docker-compose.yml.template > ./docker-compose.yml
+
+sudo docker-compose up -d
+
+# Wait for all components to be up before we create the account
+sleep 5
+sudo docker-compose exec -T conjur conjurctl account create cucumber
```
sigalsax: Why are we disabling pseudo-tty allocation with `-T` here? Is there a need for this here and in the next couple of commands?

ci/authn-azure/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -ex
+
+AWS_PEM_FILE_LOCATION="<location of pem file to login to machine>"
+USERNAME="<username of AWS machine>"
+AWS_MACHINE="<AWS machine>"
+
+pushd ../..
+  . version_utils.sh
```
sigalsax: why does this clock need to be indented

ci/authn-azure/test.sh
```diff
@@ -0,0 +1,24 @@
...
+  CONJUR_IMAGE_VERSION="$(version_tag)"
```
sigalsax: I would make this one statement `  export CONJUR_IMAGE_VERSION="$(version_tag)"` like you did this in the previous file on line 23 (`export CONJUR_DATA_KEY="$(< data_key)"`)

ci/authn-azure/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -ex
+
+AWS_PEM_FILE_LOCATION="<location of pem file to login to machine>"
+USERNAME="<username of AWS machine>"
```
sigalsax: I think better naming would be `AWS_USERNAME`

ci/authn-azure/test.sh
```diff
@@ -0,0 +1,24 @@
...
+AWS_MACHINE="<AWS machine>"
```
sigalsax: "DNS of AWS machine"

ci/authn-azure/test.sh
sigalsax: or `AWS_INSTANCE_USERNAME`

ci/authn-azure/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -ex
```
sigalsax: `-ex` are a couple of options that gives a sort of granularity in terms of failures and logs. I know there are `o, a`, etc but what does `-ex` specifically represent?

ci/authn-azure/setup-conjur.sh
orenbm: I `cd`'d because this is the only place that will take action. I use `pushd` when i do some actions in another place than the main area where the script is running. So I don't really need to `popd` after all the actions running in `conjur-quickstart`

ci/authn-azure/setup-conjur.sh
orenbm: Yes. This is the only way it works.

ci/authn-azure/test.sh
orenbm: we indent blocks inside `pushd` so it's clear it runs in a different place.

ci/authn-azure/test.sh
orenbm: correct

orenbm:
Merged this PR and created [this issue](https://github.com/cyberark/conjur/issues/1334)

PR #1318, opened 2020-02-11
Revise helm / AMI dependency in OSS suite release

izgeri:
- Remove helm / AMI repos from suite, since they are mechanisms for distributing product releases and are not products themselves
- Add items to high level technical design on updating the helm chart / re-publishing an AMI (TBD if Conjur core does this automatically already)
- Add K8s secrets provider to the suite

### DIFF
--- a/design/oss_suite_release.md
+++ b/design/oss_suite_release.md
@@ -58,9 +58,7 @@ In addition, as part of building this release, we'll also be creating:
 
 ## Repositories Included in the Conjur OSS Suite Release
 - Conjur OSS Core
-  - [Conjur OSS](https://github.com/cyberark/conjur)
-    - [Conjur OSS Helm Chart](https://github.com/cyberark/conjur-oss-helm-chart)
-    - [Conjur OSS Cloudformation Templates](https://github.com/cyberark/conjur-aws)
+  - [Conjur Core](https://github.com/cyberark/conjur)
   - [Conjur CLI](https://github.com/cyberark/conjur-cli)
   - Client Libraries
     - [Golang](https://github.com/cyberark/conjur-api-go)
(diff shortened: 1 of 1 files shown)


design/oss_suite_release.md
```diff
@@ -230,14 +229,16 @@ will include:
     links to artifacts for each component version included in the release
   - Note: when possible, links to artifacts should link directly to the canonical
     source, rather than linking to the component README (eg link directly to Dockerhub, etc)
+  - Bumping the version of Conjur pinned in the [Conjur helm chart](https://github.com/cyberark/conjur-oss-helm-chart)
```
alexkalish: I thought we decided this was out of scope.  Why the change to bring it in?

design/oss_suite_release.md
```diff
@@ -76,6 +74,7 @@ In addition, as part of building this release, we'll also be creating:
   - [Jenkins](https://github.com/cyberark/conjur-credentials-plugin)
   - Kubernetes / OpenShift
     - [Kubernetes authenticator client](https://github.com/cyberark/conjur-authn-k8s-client)
+    - [Conjur Kubernetes Secrets Provider](https://github.com/cyberark/secrets-provider-for-k8s)
```
alexkalish: Probably worth a discussion with IL, just so that understand this will be in the suite and what the repo requirements will be.

design/oss_suite_release.md
izgeri: we don't list this as out of scope in the doc, and I don't expect much effort needed to submit a PR to the helm chart repo with a conjur version bump. I think this is doable, and I'm fine including it based on my current understanding of it

PR #1324, opened 2020-02-13
Verify and decode azure token

orenbm:
#### What does this PR do?

Verifies and decodes the Azure token before it can be used to extract the host's application identity.

### DIFF
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -34,10 +34,9 @@ module Authentication
       private
 
       def validate_credentials_include_azure_token
-        # check that id token field exists and has some value
         unless decoded_credentials.include?(JWT_REQUEST_BODY_FIELD_NAME) &&
           !decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].empty?
-          raise Errors::Authentication::RequestBody::MissingRequestParam
+          raise Errors::Authentication::RequestBody::MissingRequestParam, JWT_REQUEST_BODY_FIELD_NAME
         end
       end
 
@@ -89,14 +88,16 @@ module Authentication
       end
 
       def xms_mirid
-        decoded_token[XMS_MIRID_TOKEN_FIELD_NAME].tap do |token_field_value|
-          @logger.debug(Log::ExtractedFieldFromAzureToken.new(XMS_MIRID_TOKEN_FIELD_NAME, token_field_value))
-        end
+        token_field_value(XMS_MIRID_TOKEN_FIELD_NAME)
       end
 
       def oid
-        decoded_token[OID_TOKEN_FIELD_NAME].tap do |token_field_value|
-          @logger.debug(Log::ExtractedFieldFromAzureToken.new(OID_TOKEN_FIELD_NAME, token_field_value))
+        token_field_value(OID_TOKEN_FIELD_NAME)
+      end
+
+      def token_field_value field_name
+        decoded_token[field_name].tap do |token_field_value|
+          @logger.debug(Log::ExtractedFieldFromAzureToken.new(field_name, token_field_value))
         end
       end
     end
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -10,9 +10,9 @@ module Authentication
 
     Authenticator = CommandClass.new(
       dependencies: {
-        fetch_authenticator_secrets:   Authentication::Util::FetchAuthenticatorSecrets.new,
-        verify_and_decode_token:       Authentication::AuthnOidc::VerifyAndDecodeToken.new,
-        logger:                        Rails.logger
+        fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
+        verify_and_decode_token:     Authentication::OAuth::VerifyAndDecodeToken.new,
+        logger:                      Rails.logger
       },
       inputs:       [:authenticator_input]
     ) do
@@ -21,8 +21,8 @@ module Authentication
       def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :credentials
 
       JWT_REQUEST_BODY_FIELD_NAME = "jwt"
-      XMS_MIRID_TOKEN_FIELD_NAME = "xms_mirid"
-      OID_TOKEN_FIELD_NAME = "oid"
+      XMS_MIRID_TOKEN_FIELD_NAME  = "xms_mirid"
+      OID_TOKEN_FIELD_NAME        = "oid"
 
       def call
         validate_credentials_include_azure_token
@@ -35,8 +35,10 @@ module Authentication
 
       def validate_credentials_include_azure_token
         # check that id token field exists and has some value
-        raise Errors::Authentication::RequestBody::MissingRequestParam, JWT_REQUEST_BODY_FIELD_NAME unless decoded_credentials.include?(JWT_REQUEST_BODY_FIELD_NAME) &&
+        unless decoded_credentials.include?(JWT_REQUEST_BODY_FIELD_NAME) &&
           !decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].empty?
+          raise Errors::Authentication::RequestBody::MissingRequestParam
+        end
       end
 
       # The credentials are in a URL encoded form data in the request body
@@ -66,9 +68,9 @@ module Authentication
       def azure_authenticator_secrets
         @azure_authenticator_secrets ||= @fetch_authenticator_secrets.(
           service_id: service_id,
-            conjur_account: account,
-            authenticator_name: authenticator_name,
-            required_variable_names: required_variable_names
+          conjur_account: account,
+          authenticator_name: authenticator_name,
+          required_variable_names: required_variable_names
         )
       end
 
@@ -87,13 +89,15 @@ module Authentication
       end
 
       def xms_mirid
-        @logger.debug(Log::ExtractedFieldFromAzureToken.new(XMS_MIRID_TOKEN_FIELD_NAME, decoded_token[XMS_MIRID_TOKEN_FIELD_NAME]))
-        decoded_token[XMS_MIRID_TOKEN_FIELD_NAME]
+        decoded_token[XMS_MIRID_TOKEN_FIELD_NAME].tap do |token_field_value|
+          @logger.debug(Log::ExtractedFieldFromAzureToken.new(XMS_MIRID_TOKEN_FIELD_NAME, token_field_value))
+        end
       end
 
       def oid
-        @logger.debug(Log::ExtractedFieldFromAzureToken.new(OID_TOKEN_FIELD_NAME, decoded_token[OID_TOKEN_FIELD_NAME]))
-        decoded_token[OID_TOKEN_FIELD_NAME]
+        decoded_token[OID_TOKEN_FIELD_NAME].tap do |token_field_value|
+          @logger.debug(Log::ExtractedFieldFromAzureToken.new(OID_TOKEN_FIELD_NAME, token_field_value))
+        end
       end
     end
 
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -6,48 +6,94 @@ module Authentication
     Err = Errors::Authentication::AuthnAzure
     Log = LogMessages::Authentication::AuthnAzure
     # Possible Errors Raised:
-    # TODO: Add errors
+    # TokenFieldNotFoundOrEmpty, MissingRequestParam
 
     Authenticator = CommandClass.new(
       dependencies: {
-        fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
-        logger:                      Rails.logger
+        fetch_authenticator_secrets:   Authentication::Util::FetchAuthenticatorSecrets.new,
+        verify_and_decode_token:       Authentication::AuthnOidc::VerifyAndDecodeToken.new,
+        logger:                        Rails.logger
       },
-      inputs: [:authenticator_input]
+      inputs:       [:authenticator_input]
     ) do
 
       extend Forwardable
-      def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :credentials
+      def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :credentials
+
+      JWT_REQUEST_BODY_FIELD_NAME = "jwt"
+      XMS_MIRID_TOKEN_FIELD_NAME = "xms_mirid"
+      OID_TOKEN_FIELD_NAME = "oid"
 
       def call
+        validate_credentials_include_azure_token
+        validate_azure_token
+        validate_required_token_fields_exist
         true
       end
 
       private
 
-      #def validate_azure_token
-      #
-      #end
-      #
-      #def validate_application_identity
-      #
-      #end
+      def validate_credentials_include_azure_token
+        # check that id token field exists and has some value
+        raise Errors::Authentication::RequestBody::MissingRequestParam, JWT_REQUEST_BODY_FIELD_NAME unless decoded_credentials.include?(JWT_REQUEST_BODY_FIELD_NAME) &&
+          !decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].empty?
+      end
+
+      # The credentials are in a URL encoded form data in the request body
+      def decoded_credentials
+        @decoded_credentials ||= Hash[URI.decode_www_form(credentials)]
+      end
+
+      def validate_azure_token
+        decoded_token
+      end
+
+      def decoded_token
+        @decoded_token ||= @verify_and_decode_token.call(
+          provider_uri:     provider_uri,
+          token_jwt:        decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME],
+          claims_to_verify: {
+            verify_iss: true,
+            iss:        provider_uri
+          }
+        )
+      end
 
       def provider_uri
-        azure_authenticator_secrets["provider-uri"]
+        @provider_uri ||= azure_authenticator_secrets["provider-uri"]
       end
 
       def azure_authenticator_secrets
         @azure_authenticator_secrets ||= @fetch_authenticator_secrets.(
           service_id: service_id,
-          conjur_account: account,
-          authenticator_name: authenticator_name,
-          required_variable_names: required_variable_names
+            conjur_account: account,
+            authenticator_name: authenticator_name,
+            required_variable_names: required_variable_names
         )
       end
 
       def required_variable_names
         @required_variable_names ||= %w(provider-uri)
+      end
+
+      def validate_required_token_fields_exist
+        validate_token_field_exists(XMS_MIRID_TOKEN_FIELD_NAME)
+        validate_token_field_exists(OID_TOKEN_FIELD_NAME)
+      end
+
+      def validate_token_field_exists field_name
+        @logger.debug(Log::ValidatingTokenFieldExists.new(field_name))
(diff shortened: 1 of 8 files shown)


app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
```diff
@@ -0,0 +1,88 @@
+require 'jwt'
+
+module Authentication
+  module AuthnAzure
```
Tovli: shouldn't the name of the module be like the name of the file?

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    Err = Errors::Authentication::AuthnAzure
+    Log = LogMessages::Authentication::AuthnAzure
+
+    # We use the OIDC log messages for Identity Provider related code
```
Tovli: what does it mean?
and why an object named OIDCLog is in AuthnAzure module?

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
```diff
@@ -0,0 +1,88 @@
...
+    OIDCLog = LogMessages::Authentication::AuthnAzure
+
+    VerifyAndDecodeAzureToken = CommandClass.new(
+      dependencies: {
+        # We have a ConcurrencyLimitedCache which wraps a RateLimitedCache which wraps a FetchProviderCertificate class
+        fetch_provider_certificate: ::Util::ConcurrencyLimitedCache.new(
```
Tovli: As you wrote a comment you are probably understand that this code is not readable...
Can you break it into readable parts?

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
```diff
@@ -0,0 +1,88 @@
...
+        @certs = @fetch_provider_certificate.(
+          provider_uri: @provider_uri,
+          refresh: force_read
+        )
+        @logger.debug(OIDCLog::IdentityProviderCertificateFetchedFromCache.new)
+        @certs
+      end
+
+      def verify_and_decode_azure_token
+        # We ensure that the certs are fresh just before we decode the token
```
Tovli: why is the comment needed in here?
the first method called ensure_certs_are_fresh. the second called decoded_token... seems pretty straight forward

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
```diff
@@ -0,0 +1,88 @@
...
+        ensure_certs_are_fresh
+
+        # `decoded_token` verifies the token before decoding it
```
Tovli: if decoded_token verifies, why not having it called verify_token ?
generally I would expect a function to be a verb and express an action

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
```diff
@@ -0,0 +1,88 @@
...
+          provider_uri: @provider_uri,
+          refresh: force_read
+        )
+        @logger.debug(OIDCLog::IdentityProviderCertificateFetchedFromCache.new)
+        @certs
+      end
+
+      def verify_and_decode_azure_token
+        # We ensure that the certs are fresh just before we decode the token
+        ensure_certs_are_fresh
```
Tovli: ensure_certs_are_fresh call decoded_token... why do you need to call it again afterwards?

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
```diff
@@ -0,0 +1,88 @@
...
+        return @decoded_token unless @decoded_token.nil?
+
+        # TODO: decide which claims we should verify
+        # specify the claims we need to verify
+        options = { verify_iss: true,
+                    iss: @provider_uri,
+                    verify_aud: true
+        }
+
+        @decoded_token = JWT.decode(
```
Tovli: as we use JWT in multiple places it might be good to have a dedicated object to deal with JWT related actions. then we can have one place to have defaults and change implementation if we would want in the future

Tovli: test

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
orenbm: No. The name of the command class is the name of the file.

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
orenbm: This file is not in the PR anymore

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
orenbm: This file is not in the PR anymore.

orenbm: And not really.

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
orenbm: Not in ruby. Whenever a method returns an object, the method name is the returned object.

orenbm: I changed the method name to `verified_and_decoded_token` and removed the comment.

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
orenbm: Look at the `rescue` clause. We first try to decode the token. Then, in case there is an error we fetch the certificates again to make sure we don't fail on an old cert error. After that we call `decoded_token` again to decode the token with fresh certs.

app/domain/authentication/authn_azure/verify_and_decode_azure_token.rb
orenbm: Where exactly do we have `JWT` other than this place?

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -6,35 +6,77 @@ module AuthnAzure
...
+      def oid
+        decoded_token["oid"]
+      end
 
       def provider_uri
-        azure_authenticator_secrets["provider-uri"]
+        @provider_uri ||= azure_authenticator_secrets["provider-uri"]
+      end
+
+      def decoded_credentials
```
sigalsax: in terms of flow, please move this up under the calling function

sigalsax: Because it is calling to a 3rd party, I suggest you comment on what `URI.decode_www_form` means and what we expect to happen

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -6,35 +6,77 @@ module AuthnAzure
...
+        raise Err::TokenFieldNotFoundOrEmpty, field_name if decoded_token[field_name].to_s.empty?
+      end
+
+      def xms_mirid
+        @logger.debug(Log::ExtractedXmsMiridFromAzureToked.new(decoded_token["xms_mirid"]))
+        decoded_token["xms_mirid"]
+      end
+
+      def oid
+        decoded_token["oid"]
```
sigalsax: why are we not able logging extracting oid? I don't believe this is sensitive info

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -6,35 +6,77 @@ module AuthnAzure
...
     # Possible Errors Raised:
-    # TODO: Add errors
+    # TokenFieldNotFoundOrEmpty, MissingRequestParam
 
     Authenticator = CommandClass.new(
       dependencies: {
-        fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
-        logger:                      Rails.logger
+        fetch_authenticator_secrets:   Authentication::Util::FetchAuthenticatorSecrets.new,
+        verify_and_decode_token:       Authentication::AuthnOidc::VerifyAndDecodeToken.new,
```
sigalsax: I understand that AuthnOidc and Azure are similar in the way the decoding process is handled. 
IMO involving the logic of another authenticator is confusing because this might not be clear to the next person touching this part of the code. 
I suggest we extract the overlaping logic we need to decode and verify the token for both authenticators and place it in another CommandClass and call this new class both here and in OIDC. That way, the next authenticator that is oauthn2/jwt token based, we will have a more flexible way to perform this action

app/domain/logs.rb
```diff
@@ -133,6 +133,17 @@ module AuthnK8s
 
       module AuthnAzure
 
+        # TODO: get security approval for this message
+        ExtractedXmsMiridFromAzureToked = ::Util::TrackableLogMessageClass.new(
+          msg: "Extracted xms_mirid '{0}' from Token",
```
sigalsax: "Token" here and "token" in the next log message are not inconsistent. Let's be consistent in terms of T or t when referencing "token"

app/domain/errors.rb
```diff
@@ -269,6 +269,11 @@ module AuthnK8s
 
       module AuthnAzure
 
+        TokenFieldNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
+          msg: "Field '{0-field-name}' not found or empty in Token",
```
sigalsax: same applies here

ci/authn-azure/run-authn-azure.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -ex
+
+# Get an Azure access token
+azure_access_token=$(curl \
```
sigalsax: The scripts have already been merged no?

app/domain/authentication/authn_azure/authenticator.rb
sigalsax: even if we don't want to print it out, I still think it is important for logging purposes 
`@logger.debug(Log::ExtractedXmsMiridFromAzureToked.new("oid"))`

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -6,35 +6,77 @@ module AuthnAzure
...
+      end
+
+      def validate_required_token_fields_exist
+        validate_token_field_exists( "xms_mirid")
+        validate_token_field_exists("oid")
+      end
+
+      def validate_token_field_exists field_name
+        @logger.debug(Log::ValidatingTokenFieldExists.new(field_name))
+        raise Err::TokenFieldNotFoundOrEmpty, field_name if decoded_token[field_name].to_s.empty?
```
sigalsax: I don't think it is enough to validate that the decoded token is not empty. In the test plan meeting, we also talked about validating that:
1. The token is readable and not corrupted (this might require further research on how a token can be corrupted)
2. xms_mirid is in the format we expect. We could check this by counting slashes if consistent or at least ensuring all parms in xms_mirid we expect to be there are (resource_group/subscription_id)

spec/app/domain/authentication/authn-azure/authenticator_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  ####################################
+  # request mock
+  ####################################
+
+  let (:authenticate_azure_token_request) do
+    request_body = StringIO.new
+    request_body.puts "jwt={\"xms_mirid\": \"some_xms_mirid_value\", \"oid\": \"some_oid_value\"}"
+    request_body.rewind
```
sigalsax: My understanding is that .rewind resets the index to zero? What is the purpose of this here?

spec/app/domain/authentication/authn-azure/authenticator_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+          )
+
+          ::Authentication::AuthnAzure::Authenticator.new(
+            fetch_authenticator_secrets: mock_fetch_secrets(
+                                           is_successful: true,
+                                           fetched_secrets: azure_authenticator_secrets
+                                         ),
+            verify_and_decode_token: mocked_verify_and_decode_token,
+            ).call(
+            authenticator_input: input_
```
sigalsax: I understand that `input_` references the authenticator_input but why `_` at the end of `input_`? Is this a stylistic thing?

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -6,35 +6,77 @@ module AuthnAzure
...
+        decoded_token["xms_mirid"]
+      end
+
+      def oid
+        decoded_token["oid"]
+      end
 
       def provider_uri
-        azure_authenticator_secrets["provider-uri"]
+        @provider_uri ||= azure_authenticator_secrets["provider-uri"]
```
sigalsax: We have a core test (#5 provider-uri variable loaded without value, `RequiredResourceMissing`) and I don't believe that was addressed.

app/domain/authentication/authn_azure/authenticator.rb
sigalsax: using OIDC as a dependency will also enforce us to use the OIDC specific errors and logs
Ex: `OIDCProviderCertificateFetchedFromCache` , etc

orenbm: Would you have it in the `Authentication` module? Would you create a new module named `OAuth`?

ci/authn-azure/run-authn-azure.sh
orenbm: Yes but as part of the deletion yesterday they were removed. This PR will re-add them.

app/domain/authentication/authn_azure/authenticator.rb
orenbm: 1. If the token is not a valid one then we will fail in the decoding so we're in the clear on this.
2. I think it will be easier to do this in a different PR. This one is already pretty big and your suggestion requires some UTs and more logic. I prefer we merge this and get to the xms_mirid validation after that.

spec/app/domain/authentication/authn-azure/authenticator_spec.rb
orenbm: Honestly I don't remember. It's from authn-oidc.

spec/app/domain/authentication/authn-azure/authenticator_spec.rb
orenbm: Ummm this was taken from the authn-oidc spec. Not sure why we did it this way then.

orenbm:
thanks for the review @sigalsax . Fixed some of your comments and commented on others.

ci/authn-azure/run-authn-azure.sh
```diff
@@ -0,0 +1,17 @@
...
+  'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fmanagement.azure.com%2F' \
```
nessiLahav: 1. why http and not https ? 
2. why the IP address is not in parameter like CONJUR_SERVER_DNS?

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -6,49 +6,95 @@ module AuthnAzure
...
+          claims_to_verify: {
+            verify_iss: true,
+            iss:        provider_uri
+          }
+        )
+      end
 
       def provider_uri
-        azure_authenticator_secrets["provider-uri"]
+        @provider_uri ||= azure_authenticator_secrets["provider-uri"]
```
nessiLahav: are you printing the provider uri to log as we did in OIDC?

orenbm: Yes, it happens inside `DiscoverOIDCProvider` which is called later in this stack

ci/authn-azure/run-authn-azure.sh
orenbm: This is actually the way to get a managed identity in Azure - https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http.

This URL is accessible to any client application running on the VM that can make HTTP REST calls

orenbm: it is a saved IP in azure, not our env.

app/domain/authentication/authn_azure/authenticator.rb
sigalsax: okay, adding this to our running list of things to add

app/domain/authentication/authn_azure/authenticator.rb
sigalsax: spoken about in person and decided on Oauth

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -6,49 +6,95 @@ module AuthnAzure
...
       },
-      inputs: [:authenticator_input]
+      inputs:       [:authenticator_input]
     ) do
 
       extend Forwardable
-      def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :credentials
+      def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :credentials
+
+      JWT_REQUEST_BODY_FIELD_NAME = "jwt"
```
sigalsax: I like this separation better (:

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -6,49 +6,95 @@ module AuthnAzure
...
         )
       end
 
       def required_variable_names
         @required_variable_names ||= %w(provider-uri)
       end
+
+      def validate_required_token_fields_exist
+        validate_token_field_exists(XMS_MIRID_TOKEN_FIELD_NAME)
+        validate_token_field_exists("oid")
```
sigalsax: `oid` -> `OID_TOKEN_FIELD_NAME`

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -6,49 +6,95 @@ module AuthnAzure
...
+      end
+
+      def validate_token_field_exists field_name
+        @logger.debug(Log::ValidatingTokenFieldExists.new(field_name))
+        raise Err::TokenFieldNotFoundOrEmpty, field_name if decoded_token[field_name].to_s.empty?
+      end
+
+      def xms_mirid
+        @logger.debug(Log::ExtractedFieldFromAzureToked.new(XMS_MIRID_TOKEN_FIELD_NAME, decoded_token[XMS_MIRID_TOKEN_FIELD_NAME]))
```
sigalsax: `ExtractedFieldFromAzureToked` -> `ExtractedFieldFromAzureToken`?

d -> n?

orenbm: nice

app/domain/authentication/authn_oidc/authenticate.rb
```diff
@@ -69,12 +69,9 @@ def add_username_to_input
         @authenticator_input = @authenticator_input.update(username: conjur_username)
       end
 
+      # The credentials are in a URL encoded form data in the request body
```
sigalsax: We talked about how credentials may not be too descriptive so what would credentials for us? I would add what we expect here in a comment

orenbm: the credentials differ from authenticator to authenticator. For example, in OIDC the credentials are an ID Token. In authn-azure it is an Azure access token. In authn it is an API Key.

ci/authn-azure/run-authn-azure.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -ex
```
moticless: Write to all your bash scripts the following header:
#!/bin/bash
set -euo pipefail

for explanation: https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

PR #1332, opened 2020-02-17
Downgrade pg version to fix the build

orenbm:
#### What does this PR do?
Downgrades the version of postgres to 9.3 in our test env.

The build is failing with an issue of Conjur with postrges. The `pg` container is being shut down immediately after we run `docker-compose up -d`

Until this is properly fix, this PR downgrades the `postgres` version to 9.3. [This issue](https://github.com/cyberark/conjur/issues/1333) was created to properly fix this.


orenbm:
Closing this as it was fixed - https://github.com/cyberark/conjur/pull/1336

PR #1343, opened 2020-02-19
Extract to oauth module

orenbm:
#### What does this PR do?

Extracts OAuthn related code into its own module and JWT related code into its own module.
This is done so other authenticators can use these parts of code independently from the OIDC authenticator, as it can be confusing to the reader which parts are OIDC related, which are OAuth related and which are JWT related.

### DIFF
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -151,7 +151,7 @@ class AuthenticateController < ApplicationController
       Errors::Authentication::Security::RoleNotFound,
       Errors::Authentication::Security::AccountNotDefined,
       Errors::Authentication::AuthnOidc::IdTokenFieldNotFoundOrEmpty,
-      Errors::Authentication::Jwt::TokenVerifyFailed,
+      Errors::Authentication::Jwt::TokenVerificationFailed,
       Errors::Authentication::Jwt::TokenDecodeFailed,
       Errors::Conjur::RequiredSecretMissing,
       Errors::Conjur::RequiredResourceMissing
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -151,8 +151,8 @@ class AuthenticateController < ApplicationController
       Errors::Authentication::Security::RoleNotFound,
       Errors::Authentication::Security::AccountNotDefined,
       Errors::Authentication::AuthnOidc::IdTokenFieldNotFoundOrEmpty,
-      Errors::Authentication::OAuth::TokenVerifyFailed,
-      Errors::Authentication::OAuth::TokenDecodeFailed,
+      Errors::Authentication::Jwt::TokenVerifyFailed,
+      Errors::Authentication::Jwt::TokenDecodeFailed,
       Errors::Conjur::RequiredSecretMissing,
       Errors::Conjur::RequiredResourceMissing
       raise Unauthorized
@@ -163,7 +163,7 @@ class AuthenticateController < ApplicationController
     when Errors::Authentication::RequestBody::MissingRequestParam
       raise BadRequest
 
-    when Errors::Authentication::OAuth::TokenExpired
+    when Errors::Authentication::Jwt::TokenExpired
       raise Unauthorized.new(err.message, true)
 
     when Errors::Authentication::OAuth::ProviderDiscoveryTimeout
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -173,7 +173,7 @@ class AuthenticateController < ApplicationController
       raise ServiceUnavailable
 
     when Errors::Authentication::OAuth::ProviderDiscoveryFailed,
-      Errors::Authentication::OAuth::ProviderFetchCertificateFailed
+      Errors::Authentication::OAuth::FetchProviderKeysFailed
       raise BadGateway
 
     else
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -151,8 +151,8 @@ class AuthenticateController < ApplicationController
       Errors::Authentication::Security::RoleNotFound,
       Errors::Authentication::Security::AccountNotDefined,
       Errors::Authentication::AuthnOidc::IdTokenFieldNotFoundOrEmpty,
-      Errors::Authentication::AuthnOidc::TokenVerifyFailed,
-      Errors::Authentication::AuthnOidc::TokenDecodeFailed,
+      Errors::Authentication::OAuth::TokenVerifyFailed,
+      Errors::Authentication::OAuth::TokenDecodeFailed,
       Errors::Conjur::RequiredSecretMissing,
       Errors::Conjur::RequiredResourceMissing
       raise Unauthorized
@@ -163,17 +163,17 @@ class AuthenticateController < ApplicationController
     when Errors::Authentication::RequestBody::MissingRequestParam
       raise BadRequest
 
-    when Errors::Authentication::AuthnOidc::IdTokenExpired
+    when Errors::Authentication::OAuth::TokenExpired
       raise Unauthorized.new(err.message, true)
 
-    when Errors::Authentication::AuthnOidc::ProviderDiscoveryTimeout
+    when Errors::Authentication::OAuth::ProviderDiscoveryTimeout
       raise GatewayTimeout
 
     when Errors::Util::ConcurrencyLimitReachedBeforeCacheInitialization
       raise ServiceUnavailable
 
-    when Errors::Authentication::AuthnOidc::ProviderDiscoveryFailed,
-      Errors::Authentication::AuthnOidc::ProviderFetchCertificateFailed
+    when Errors::Authentication::OAuth::ProviderDiscoveryFailed,
+      Errors::Authentication::OAuth::ProviderFetchCertificateFailed
       raise BadGateway
 
     else
--- a/app/domain/authentication/authn_oidc/validate_status.rb
+++ b/app/domain/authentication/authn_oidc/validate_status.rb
@@ -9,7 +9,7 @@ module Authentication
     ValidateStatus = CommandClass.new(
       dependencies: {
         fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
-        discover_oidc_provider: Authentication::OAuth::DiscoverIdentityProvider.new
+        discover_identity_provider: Authentication::OAuth::DiscoverIdentityProvider.new
       },
       inputs: %i(account service_id)
     ) do
@@ -39,7 +39,7 @@ module Authentication
       end
 
       def validate_provider_is_responsive
-        @discover_oidc_provider.(
+        @discover_identity_provider.(
           provider_uri: provider_uri
         )
       end
--- a/app/domain/authentication/authn_oidc/validate_status.rb
+++ b/app/domain/authentication/authn_oidc/validate_status.rb
@@ -9,7 +9,7 @@ module Authentication
     ValidateStatus = CommandClass.new(
       dependencies: {
         fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
-        discover_oidc_provider: Authentication::AuthnOidc::DiscoverOIDCProvider.new
+        discover_oidc_provider: Authentication::OAuth::DiscoverIdentityProvider.new
       },
       inputs: %i(account service_id)
     ) do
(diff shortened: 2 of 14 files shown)


app/controllers/authenticate_controller.rb
```diff
@@ -163,17 +163,17 @@ def handle_authentication_error(err)
...
-    when Errors::Authentication::AuthnOidc::ProviderDiscoveryTimeout
+    when Errors::Authentication::OAuth::ProviderDiscoveryTimeout
       raise GatewayTimeout
 
     when Errors::Util::ConcurrencyLimitReachedBeforeCacheInitialization
       raise ServiceUnavailable
 
-    when Errors::Authentication::AuthnOidc::ProviderDiscoveryFailed,
-      Errors::Authentication::AuthnOidc::ProviderFetchCertificateFailed
+    when Errors::Authentication::OAuth::ProviderDiscoveryFailed,
```
sigalsax: As spoken about let's think if OAuth is the correct naming b/c not all uses for JWT are Oauth

app/domain/logs.rb
```diff
@@ -29,50 +29,58 @@ module Security
...
 
+      module Jwt
+
+        TokenDecodeSuccess = ::Util::TrackableLogMessageClass.new(
+          msg: "Token decode succeeded",
+          code: "CONJ00005D"
+        )
+
+        TokenDecodeFailed = ::Util::TrackableLogMessageClass.new(
+          msg: "Failed to decode the token with the error '{0-exception}'",
```
sigalsax: Suggesting to remove "the"

app/domain/logs.rb
```diff
@@ -29,50 +29,58 @@ module Security
...
+          code: "CONJ00018D"
+        )
+
+      end
+
+      module AuthnOidc
+
+        ExtractedUsernameFromIDToked = ::Util::TrackableLogMessageClass.new(
+          msg: "Extracted username '{0}' from ID Token field '{1-id-token-username-field}'",
```
sigalsax: Let's have Shuli go through all these log/error messages so that they are uniform

app/controllers/authenticate_controller.rb
orenbm: This is not a JWT thing. It will always be an OAuth provider.

app/domain/logs.rb
orenbm: I'll pass them by shuli later on in one PR so it will be easier.

app/domain/authentication/jwt/verify_and_decode_token.rb
```diff
@@ -0,0 +1,62 @@
...
+
+module Authentication
+  module Jwt
+
+    Err = Errors::Authentication::Jwt
+    Log = LogMessages::Authentication::Jwt
+    # Possible Errors Raised:
+    # TokenExpired, TokenDecodeFailed, TokenVerifyFailed
+
+    # This class Verifies and decodes a JWT token. It doesn't connect with the issuer
```
sigalsax: verifies and decodes or Verifies and Decodes

app/domain/authentication/jwt/verify_and_decode_token.rb
```diff
@@ -0,0 +1,62 @@
...
+      private
+
+      def verified_and_decoded_token
+        # @jwt_decoder.decode returns an array with one decoded token so we take the first object
+        @decoded_token = @jwt_decoder.decode(
+          @token_jwt,
+          nil, # the key will be taken from options[:jwks] if present
+          should_verify,
+          @verification_options
+        ).first.tap do
```
sigalsax: `.first.tap` will perform the decode first and then return the object for logging?

app/domain/authentication/o_auth/fetch_provider_certificates.rb
```diff
@@ -13,7 +13,7 @@ module AuthnOidc
     FetchProviderCertificates = CommandClass.new(
       dependencies: {
         logger:                 Rails.logger,
-        discover_oidc_provider: Authentication::AuthnOidc::DiscoverOIDCProvider.new
+        discover_oidc_provider: ::Authentication::OAuth::DiscoverIdentityProvider.new
```
sigalsax: Would you be able to also reference a file from the same level?

Ex: `DiscoverIdentityProvider.new` instead of `::Authentication::OAuth::DiscoverIdentityProvider.new`

app/domain/errors.rb
```diff
@@ -126,24 +126,32 @@ module AuthnOidc
...
 
         TokenDecodeFailed = ::Util::TrackableErrorClass.new(
           msg: "Token decode failed (3rdPartyError ='{0}')",
           code: "CONJ00035E"
         )
 
-        IdTokenExpired = ::Util::TrackableErrorClass.new(
-          msg: "Token expired",
-          code: "CONJ00016E"
+        TokenVerifyFailed = ::Util::TrackableErrorClass.new(
```
sigalsax: TokenVerificationFailed

app/domain/errors.rb
```diff
@@ -126,24 +126,32 @@ module AuthnOidc
...
+          msg: "Token verification failed (3rdPartyError ='{0}')",
+          code: "CONJ00015E"
+        )
+
+      end
+
+      module AuthnOidc
+
+        IdTokenFieldNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
+          msg: "Field '{0-field-name}' not found or empty in ID Token. This field is defined in the id-token-user-property variable.",
```
sigalsax: `id-token-user-property` might be changed so I don't necessary like us hardcoding it here just in case this might change in the future

app/domain/logs.rb
```diff
@@ -29,50 +29,58 @@ module Security
...
         )
 
-        OIDCProviderDiscoverySuccess = ::Util::TrackableLogMessageClass.new(
+        IdentityProviderDiscoverySuccess = ::Util::TrackableLogMessageClass.new(
           msg: "Identity Provider discovery succeeded",
           code: "CONJ00008D"
         )
 
+        IdentityProviderCertificateFetchedFromCache = ::Util::TrackableLogMessageClass.new(
+          msg: "Identity Provider certificates fetched successfully from cache",
```
sigalsax: In OIDC it is certificates but in Azure they are keys. Let's think of a more general word instead of ` certificates `

app/domain/logs.rb
```diff
@@ -29,50 +29,58 @@ module Security
...
         ValidateProviderCertificateIsUpdated = ::Util::TrackableLogMessageClass.new(
           msg: "Validating Identity Provider certificates are up to date",
           code: "CONJ00019D"
         )
 
       end
 
+      module Jwt
+
+        TokenDecodeSuccess = ::Util::TrackableLogMessageClass.new(
```
sigalsax: In the error/log file we have different ways to express similar outcomes
Success, successful, succeeded and fail, failure, failed and verify, verification, etc

Let's aim to be more aligned here. If you don't want to tackle this in this PR, I will add it to the backlog of things TODO and will address this when we send these files to TW. What do you think?

app/controllers/authenticate_controller.rb
```diff
@@ -163,17 +163,17 @@ def handle_authentication_error(err)
...
+      Errors::Authentication::OAuth::ProviderFetchCertificateFailed
```
sigalsax: Also here, we have certificate but in Azure they are keys

app/domain/authentication/authn_oidc/validate_status.rb
```diff
@@ -9,7 +9,7 @@ module AuthnOidc
     ValidateStatus = CommandClass.new(
       dependencies: {
         fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
-        discover_oidc_provider: Authentication::AuthnOidc::DiscoverOIDCProvider.new
+        discover_oidc_provider: Authentication::OAuth::DiscoverIdentityProvider.new
```
sigalsax: discover_oidc_provider-> @discover_identity_provider

app/domain/authentication/jwt/verify_and_decode_token.rb
orenbm: No. As I mention in the comment - `# @jwt_decoder.decode returns an array with one decoded token so we take the first object`

`first` is a method to get the first object in an array.
`tap` is something else - https://medium.com/aviabird/ruby-tap-that-method-90c8a801fd6a

app/domain/authentication/o_auth/fetch_provider_certificates.rb
orenbm: Yes.

app/domain/errors.rb
orenbm: It is hard-coded in other places too. Let's not go to this rabbit hole in this PR

app/domain/logs.rb
orenbm: I'll go with keys

app/domain/authentication/o_auth/verify_and_decode_token.rb
```diff
@@ -41,7 +46,7 @@ def fetch_certs(force_read: false)
 
         @jwks = provider_certificates.jwks
         @algs = provider_certificates.algorithms
-        @logger.debug(Log::OIDCProviderCertificateFetchedFromCache.new)
+        @logger.debug(Log::IdentityProviderCertificateFetchedFromCache.new)
       end
 
       def verify_and_decode_token
```
sigalsax: We have
```
    def verify_and_decode_token
        ensure_certs_are_fresh
        verified_and_decoded_token
      end

      def ensure_certs_are_fresh
        verified_and_decoded_token
```
Why are we calling `verified_and_decoded_token` twice here? From `verify_and_decode_token` and `ensure_certs_are_fresh`

app/domain/authentication/o_auth/verify_and_decode_token.rb
```diff
@@ -1,26 +1,31 @@
...
-            ::Authentication::AuthnOidc::FetchProviderCertificates.new,
+            ::Authentication::OAuth::FetchProviderCertificates.new,
             refreshes_per_interval: 10,
             rate_limit_interval:    300, # 300 seconds (every 5 mins)
             logger: Rails.logger
           ),
           max_concurrent_requests: 3, # TODO: Should be dynamic calculation
           logger: Rails.logger
         ),
+        verify_and_decode_token: ::Authentication::Jwt::VerifyAndDecodeToken.new,
```
sigalsax: What is the difference between jwt and o_auth/verify_decode_token. It is confusing that both have the same naming

sigalsax: >  It first retrieves the JWKs from the identity provider and sets them as verification options for Authentication::Jwt::VerifyAndDecodeToken. That class does the offline verification and decode, using the JWKs retrieved from the OAuthn 2.0 Identity Provider.

Would like to discuss this comment because the last sentence isnt too clear to me

app/domain/logs.rb
orenbm: let's do that

orenbm: (add to TODO)

app/domain/authentication/o_auth/verify_and_decode_token.rb
orenbm: Yes. This is how `ensure_certs_are_fresh` works. I'll add a comment in the code:
```
      # ensure_keys_are_fresh will try to verify and decode the token and if it 
      # fails we fetch the provider keys again. We then verify and decode again
      # so we definitely don't fail because the keys are too old. If ensure_keys_are_fresh
      # will succeed to decode the token then the call to verified_and_decoded_token 
      # will not do a thing because the object is memoized
```

app/domain/authentication/o_auth/verify_and_decode_token.rb
orenbm: That's why I added a comment on each class. One is OAuth scoped and one is JWT scoped and they both verify and decode a token. If you have a better suggestion for naming tell me.

orenbm: what's unclear?

app/domain/authentication/o_auth/fetch_provider_certificates.rb
sigalsax: so i think `DiscoverIdentityProvider.new` would be cleaner

app/domain/logs.rb
sigalsax: added

app/domain/authentication/authn_oidc/validate_status.rb
sigalsax: As discussed, make this change in both places

PR #1344, opened 2020-02-19
Extract security and secrets specs from oidc

orenbm:
We would like to use these helpers in other specs than oidc.
For this we need to extract them into their own helpers.

### DIFF
--- a/spec/support/fetch_secrets_helper.rb
+++ b/spec/support/fetch_secrets_helper.rb
@@ -2,7 +2,25 @@
 
 shared_context "fetch secrets" do
 
-  let (:test_fetch_secrets_error) { "test-fetch-secrets-error" }
+  let(:test_fetch_secrets_error) { "test-fetch-secrets-error" }
+
+  let(:mocked_secret) do
+    double('Secret').tap do |secret|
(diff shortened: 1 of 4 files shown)


spec/support/fetch_secrets_helper.rb
```diff
@@ -16,3 +34,21 @@ def mock_fetch_secrets(is_successful:, fetched_secrets:)
...
+  it "fails when variable is missing" do
+    allow(Resource).to receive(:[])
+                         .with(/#{account}:variable:conjur\/#{authenticator_name}\/#{service}\/#{variable_name}/)
+                         .and_return(nil)
+
+    expect { subject }.to raise_error(Errors::Conjur::RequiredResourceMissing)
+  end
+
+  it "fails when variable has no value" do
+    allow(Resource).to receive(:[])
```
sigalsax: The test here is when the "variable has no value", the `resource_without_value` mock you created expects/can receive a resource so instead of `(:[])`, I would send a value to accurately mock what is happening in reality and it will just return nil anyway

orenbm: This is actually what happens here :)

PR #1345, opened 2020-02-20
Additional Logging for Code Coverage Debugging

hughsaunders:
#### What does this PR do?
Add detail to logs for debugging codeclimate submission failures.

This is an extension of existing commit https://github.com/cyberark/conjur/commit/15ee5b9ed4fc912bfd68c1565ef466b4bf80c24b

conjurinc/ops#516
See the failed jenkins build linked from the above issue, and discussion in the issue. 
Tricky, but you could create some simple json files and test merging them, check that empty files are picked up correctly.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -77,10 +77,15 @@ pipeline {
 
     stage('Submit Coverage Report'){
       steps{
-        archiveArtifacts artifacts: "coverage/.resultset.json", fingerprint: false
-        archiveArtifacts artifacts: "ci/authn-k8s/output/simplecov-resultset-authnk8s-gke.json", fingerprint: false
-        sh 'ci/submit-coverage'
-        publishHTML([reportDir: 'coverage', reportFiles: 'index.html', reportName: 'Coverage Report', reportTitles: '', allowMissing: false, alwaysLinkToLastBuild: true, keepAll: true])
+        script {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -77,9 +77,9 @@ pipeline {
 
     stage('Submit Coverage Report'){
       steps{
-        archiveArtifacts artifacts: "coverage/.resultset.json", fingerprint: false
-        archiveArtifacts artifacts: "ci/authn-k8s/output/simplecov-resultset-authnk8s-gke.json", fingerprint: false
         sh 'ci/submit-coverage'
+        archiveArtifacts artifacts: "coverage/.resultset*.json", fingerprint: false
+        archiveArtifacts artifacts: "ci/authn-k8s/output/simplecov-resultset-authnk8s-gke.json", fingerprint: false
```
jtuttle: We put these lines before `ci/submit-coverage` previously so that we could download and examine the artifacts from Jenkins in the event of an error. Do we no longer want that option?

hughsaunders: I moved it so that pre and post merge arrtifacts could be downloaded. I failed to realise that if there was an error in the submit-coverage script, the archiveArtifacts step wouldn't be reached. I'll add an exception handler.

PR #1346, opened 2020-02-20
Application identity validation implementation

sigalsax:
#### What does this PR do?
This PR adds the second validation in for the Azure authenticator, validating the application identity of the resource sending the request

Object received: Hash object from decode/valid token

DOD
- [x] Check integrity of annotations relevant to authn-azure (prefixed correctly, in proper format, has the 2 required annotations, and that there is not a combo of user and system assigned identity
- [x] Validate xms_mirid/oid based on annotations
   - [x] Grab the `key -> value` list from output of previous check and perform validation check
   - [x] Ensure that there user/system assigned flows are accounted for

TODO:
- [ ] validate xms_mirid format (length or # of slashes)

### DIFF
--- a/app/domain/authentication/authn_azure/application_identity.rb
+++ b/app/domain/authentication/authn_azure/application_identity.rb
@@ -1,0 +1,131 @@
+module Authentication
+  module AuthnAzure
+
+    Log = LogMessages::Authentication
+    Err = Errors::Authentication
+    # Possible Errors Raised: ConstraintNotSupported, MissingConstraint, IllegalConstraintCombinations
+
+    # This class defines an Azure application identity of a given Conjur resource.
+    # The constructor initializes an ApplicationIdentity object and validates that
+    # it is configured correctly.
+    # The difference between the validation in this constructor and
+    # the validation in ValidateApplicationIdentity is that here we validate that
+    # the application identity is configured correctly, and thus is a valid application
+    # identity. In ValidateApplicationIdentity we validate that the defined application
+    # identity is actually the one we expect from Azure. This validation is done through
+    # a combination of comparisons that are performed based on the type of assigned
+    # identity provided.
+    # For example, an application identity `authn-azure/some-value` is not
+    # valid and will fail the validation here. For example, application identity like
+    # `authn-azure/subscription-id` or `authn-azure/<service-id>/subscription-id`
+    # defined in the Conjur resource's annotations is a valid application identity
+    # because `subscription_id` is one of the defined, accepted constraints and will pass
+    # the validation here.
+    # If the Azure-related annotations in the resource defined in Conjur has Azure-specific
+    # annotations that do not match the supplied JWT token, then it will fail the
+    # validation of ValidateApplicationIdentity.
+    class ApplicationIdentity
+
+      def initialize(role_annotations:, service_id:)
+        @role_annotations = role_annotations
+        @service_id       = service_id
+
+        validate
+      end
+
+      def constraints
+        @constraints ||= {
+            subscription_id:          constraint_value("subscription-id"),
+            resource_group:           constraint_value("resource-group"),
+            user_assigned_identity:   constraint_value("user-assigned-identity"),
+            system_assigned_identity: constraint_value("system-assigned-identity"),
+        }.compact
+      end
+
+      private
+
+      # validate that the application identity is defined correctly
+      def validate
+        validate_azure_annotations_are_permitted
+        validate_required_constraints_exist
+        validate_constraint_combinations
+      end
+
+      # validating that the annotations listed for the Conjur resource align with the permitted Azure constraints
+      def validate_azure_annotations_are_permitted
+        validate_prefixed_permitted_annotations("authn-azure/")
+        validate_prefixed_permitted_annotations("authn-azure/#{@service_id}/")
+      end
+
+      # check if annotations with the given prefix is part of the permitted list
+      def validate_prefixed_permitted_annotations prefix
+        Rails.logger.debug(LogMessages::Authentication::ValidatingAnnotationsWithPrefix.new(prefix))
+
+        prefixed_annotations(prefix).each do |annotation|
+          annotation_name = annotation[:name]
+          next if prefixed_permitted_constraints(prefix).include?(annotation_name)
+          raise Err::AuthnAzure::ConstraintNotSupported.new(annotation_name.gsub(prefix, ""), permitted_constraints)
+        end
+      end
+
+      def prefixed_annotations prefix
+        @role_annotations.select do |a|
+          annotation_name = a.values[:name]
+
+          annotation_name.start_with?(prefix) &&
+              # verify we take only annotations from the same level
+              annotation_name.split('/').length == prefix.split('/').length + 1
+        end
+      end
+
+      # add prefix to all permitted constraints
+      def prefixed_permitted_constraints prefix
+        permitted_constraints.map { |k| "#{prefix}#{k}" }
+      end
+
+      def validate_required_constraints_exist
+        validate_constraint_exists :subscription_id
+        validate_constraint_exists :resource_group
+      end
+
+      def validate_constraint_exists constraint
+        raise Err::AuthnAzure::MissingConstraint.new(constraint) unless constraints[constraint]
+      end
+
+      # check the `service-id` specific constraint first to be more granular
+      def constraint_from_annotation constraint_name
+        annotation_value("authn-azure/#{@service_id}/#{constraint_name}") ||
+            annotation_value("authn-azure/#{constraint_name}")
+      end
+
+      def annotation_value name
+        annotation = @role_annotations.find { |a| a.values[:name] == name }
+
+        # return the value of the annotation if it exists, nil otherwise
+        if annotation
+          Rails.logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
+          annotation[:value]
+        end
+      end
+
+      def permitted_constraints
+        @permitted_constraints ||= %w(
+          subscription-id resource-group user-assigned-identity system-assigned-identity
+        )
+      end
+
+      # validates that the application identity doesn't include logical constraint
+      # combinations (e.g user_assigned_identity & system_assigned_identity)
+      def validate_constraint_combinations
+        identifiers = %i(user_assigned_identity system_assigned_identity)
+
+        identifiers_constraints = constraints.keys & identifiers
+        raise Err::IllegalConstraintCombinations, identifiers_constraints unless identifiers_constraints.length <= 1
+      end
+
+      def constraint_value constraint_name
+        constraint_from_annotation(constraint_name)
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -34,10 +34,9 @@ module Authentication
       private
 
       def validate_credentials_include_azure_token
-        # check that id token field exists and has some value
         unless decoded_credentials.include?(JWT_REQUEST_BODY_FIELD_NAME) &&
           !decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].empty?
-          raise Errors::Authentication::RequestBody::MissingRequestParam
+          raise Errors::Authentication::RequestBody::MissingRequestParam, JWT_REQUEST_BODY_FIELD_NAME
         end
       end
 
@@ -89,14 +88,16 @@ module Authentication
       end
 
       def xms_mirid
-        decoded_token[XMS_MIRID_TOKEN_FIELD_NAME].tap do |token_field_value|
-          @logger.debug(Log::ExtractedFieldFromAzureToken.new(XMS_MIRID_TOKEN_FIELD_NAME, token_field_value))
-        end
+        token_field_value(XMS_MIRID_TOKEN_FIELD_NAME)
       end
 
       def oid
-        decoded_token[OID_TOKEN_FIELD_NAME].tap do |token_field_value|
-          @logger.debug(Log::ExtractedFieldFromAzureToken.new(OID_TOKEN_FIELD_NAME, token_field_value))
+        token_field_value(OID_TOKEN_FIELD_NAME)
+      end
+
+      def token_field_value field_name
+        decoded_token[field_name].tap do |token_field_value|
+          @logger.debug(Log::ExtractedFieldFromAzureToken.new(field_name, token_field_value))
         end
       end
     end
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -4,6 +4,7 @@ module Authentication
   module AuthnAzure
 
     Err = Errors::Authentication::AuthnAzure
+    Log = LogMessages::Authentication::AuthnAzure
     # Possible Errors Raised:
     # TODO: Add errors
 
@@ -14,7 +15,7 @@ module Authentication
       inputs: [:authenticator_input]
     ) do
       extend Forwardable
-      def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :request_body
+      def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :credentials
 
       def call
         validate_azure_token
@@ -28,14 +29,6 @@ module Authentication
       end
 
       def validate_application_identity
-
-      end
-
-      def host
-
-      end
-
-      def host_annotations
 
       end
     end
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -10,10 +10,12 @@ module Authentication
 
     Authenticator = CommandClass.new(
       dependencies: {
-
+        fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
+        logger:                      Rails.logger
       },
       inputs: [:authenticator_input]
     ) do
+
       extend Forwardable
       def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :credentials
 
@@ -30,6 +32,23 @@ module Authentication
 
       def validate_application_identity
 
+      end
+
+      def provider_uri
+        azure_authenticator_secrets["provider-uri"]
+      end
+
+      def azure_authenticator_secrets
+        @azure_authenticator_secrets ||= @fetch_authenticator_secrets.(
+          service_id: service_id,
+          conjur_account: account,
+          authenticator_name: authenticator_name,
+          required_variable_names: required_variable_names
+        )
+      end
+
+      def required_variable_names
+        @required_variable_names ||= %w(provider-uri)
       end
     end
 
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -10,9 +10,9 @@ module Authentication
 
     Authenticator = CommandClass.new(
       dependencies: {
-        fetch_authenticator_secrets:   Authentication::Util::FetchAuthenticatorSecrets.new,
-        verify_and_decode_token:       Authentication::AuthnOidc::VerifyAndDecodeToken.new,
-        logger:                        Rails.logger
+        fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
+        verify_and_decode_token:     Authentication::OAuth::VerifyAndDecodeToken.new,
+        logger:                      Rails.logger
       },
       inputs:       [:authenticator_input]
     ) do
@@ -21,8 +21,8 @@ module Authentication
       def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :credentials
 
       JWT_REQUEST_BODY_FIELD_NAME = "jwt"
-      XMS_MIRID_TOKEN_FIELD_NAME = "xms_mirid"
-      OID_TOKEN_FIELD_NAME = "oid"
+      XMS_MIRID_TOKEN_FIELD_NAME  = "xms_mirid"
+      OID_TOKEN_FIELD_NAME        = "oid"
 
       def call
         validate_credentials_include_azure_token
@@ -35,8 +35,10 @@ module Authentication
 
       def validate_credentials_include_azure_token
         # check that id token field exists and has some value
-        raise Errors::Authentication::RequestBody::MissingRequestParam, JWT_REQUEST_BODY_FIELD_NAME unless decoded_credentials.include?(JWT_REQUEST_BODY_FIELD_NAME) &&
+        unless decoded_credentials.include?(JWT_REQUEST_BODY_FIELD_NAME) &&
           !decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].empty?
+          raise Errors::Authentication::RequestBody::MissingRequestParam
+        end
       end
 
       # The credentials are in a URL encoded form data in the request body
@@ -66,9 +68,9 @@ module Authentication
       def azure_authenticator_secrets
         @azure_authenticator_secrets ||= @fetch_authenticator_secrets.(
           service_id: service_id,
-            conjur_account: account,
-            authenticator_name: authenticator_name,
-            required_variable_names: required_variable_names
+          conjur_account: account,
+          authenticator_name: authenticator_name,
+          required_variable_names: required_variable_names
         )
       end
 
@@ -87,13 +89,15 @@ module Authentication
       end
 
       def xms_mirid
-        @logger.debug(Log::ExtractedFieldFromAzureToken.new(XMS_MIRID_TOKEN_FIELD_NAME, decoded_token[XMS_MIRID_TOKEN_FIELD_NAME]))
-        decoded_token[XMS_MIRID_TOKEN_FIELD_NAME]
+        decoded_token[XMS_MIRID_TOKEN_FIELD_NAME].tap do |token_field_value|
+          @logger.debug(Log::ExtractedFieldFromAzureToken.new(XMS_MIRID_TOKEN_FIELD_NAME, token_field_value))
+        end
       end
 
       def oid
-        @logger.debug(Log::ExtractedFieldFromAzureToken.new(OID_TOKEN_FIELD_NAME, decoded_token[OID_TOKEN_FIELD_NAME]))
-        decoded_token[OID_TOKEN_FIELD_NAME]
+        decoded_token[OID_TOKEN_FIELD_NAME].tap do |token_field_value|
+          @logger.debug(Log::ExtractedFieldFromAzureToken.new(OID_TOKEN_FIELD_NAME, token_field_value))
+        end
       end
     end
 
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -6,48 +6,94 @@ module Authentication
     Err = Errors::Authentication::AuthnAzure
     Log = LogMessages::Authentication::AuthnAzure
     # Possible Errors Raised:
-    # TODO: Add errors
+    # TokenFieldNotFoundOrEmpty, MissingRequestParam
 
     Authenticator = CommandClass.new(
       dependencies: {
-        fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
-        logger:                      Rails.logger
+        fetch_authenticator_secrets:   Authentication::Util::FetchAuthenticatorSecrets.new,
+        verify_and_decode_token:       Authentication::AuthnOidc::VerifyAndDecodeToken.new,
+        logger:                        Rails.logger
       },
-      inputs: [:authenticator_input]
+      inputs:       [:authenticator_input]
     ) do
 
       extend Forwardable
-      def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :username, :request, :credentials
+      def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :credentials
+
+      JWT_REQUEST_BODY_FIELD_NAME = "jwt"
+      XMS_MIRID_TOKEN_FIELD_NAME = "xms_mirid"
+      OID_TOKEN_FIELD_NAME = "oid"
 
       def call
+        validate_credentials_include_azure_token
+        validate_azure_token
+        validate_required_token_fields_exist
         true
       end
 
       private
 
-      #def validate_azure_token
-      #
-      #end
-      #
-      #def validate_application_identity
-      #
-      #end
+      def validate_credentials_include_azure_token
+        # check that id token field exists and has some value
+        raise Errors::Authentication::RequestBody::MissingRequestParam, JWT_REQUEST_BODY_FIELD_NAME unless decoded_credentials.include?(JWT_REQUEST_BODY_FIELD_NAME) &&
+          !decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].empty?
+      end
+
+      # The credentials are in a URL encoded form data in the request body
+      def decoded_credentials
+        @decoded_credentials ||= Hash[URI.decode_www_form(credentials)]
+      end
+
+      def validate_azure_token
+        decoded_token
+      end
+
+      def decoded_token
+        @decoded_token ||= @verify_and_decode_token.call(
+          provider_uri:     provider_uri,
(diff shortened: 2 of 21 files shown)


app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
+module Authentication
+  module AuthnAzure
+
+    Log = LogMessages::Authentication
+    Err = Errors::Authentication::AuthnAzure
+    # Possible Errors Raised: ConstraintNotSupported, MissingAnnotationConstraint, IllegalConstraintCombinations
+
+    # This class defines an application identity of a given Conjur host.
```
orenbm: an Azure application identity

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+    # The constructor initializes an ApplicationIdentity object and validates that
+    # it is configured correctly.
+    # The difference between the validation in this constructor and
+    # the validation in ValidateApplicationIdentity is that here we validate that
+    # the application identity is configured correctly, and thus is a valid application
+    # identity. In ValidateApplicationIdentity we validate that the defined application
+    # identity is actually the one we expect from Azure. This validation is done through
+    # a combination of comparisons that are performed based on the type of assigned
+    # identity provided.
+    # For example, an application identity `some-namespace/blah/some-value` is not
```
orenbm: some-namespace? :)

Please give an example with authn-azure annotations

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+    # valid and will fail the validation here. However, an application identity
+    # `some-namespace/service-account/some-value` is a valid application identity
```
orenbm: same here. Please also give context to _why_ it is a valid application and may fail in the `validate_application_identity` validations. It will help to understand the differences between the 2 validations

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+      end
+
+      private
+
+      # validate that the application identity is defined correctly
+      def validate
+        # check if annotations present are part of the permitted annotations for authn-azure
+        validate_permitted_annotations
+
+        # validate that a constraint exists on the required annotations
```
orenbm: pretty clear from the method name

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+            system_assigned_identity: constraint_value("system-assigned-identity"),
+        }.compact
+      end
+
+      private
+
+      # validate that the application identity is defined correctly
+      def validate
+        # check if annotations present are part of the permitted annotations for authn-azure
+        validate_permitted_annotations
```
orenbm: If it's not clear from the method name, please change it. `validate_annotations_are_permitted` will be ok imo.

orenbm: And remove the comment of course.

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+
+        # validate that a constraint exists on the required annotations
+        validate_required_constraints_exist
+        validate_constraint_combinations
+      end
+
+      def validate_permitted_annotations
```
orenbm: I think that we should add a method description

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+        validate_prefixed_permitted_annotations("authn-azure/")
+        validate_prefixed_permitted_annotations("authn-azure/#{@service_id}/")
+      end
+
+      # check if annotations with prefix is part of the permitted list
```
orenbm: with the given prefix

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+      # add prefix to all permitted constraints
+      def prefixed_permitted_constraints prefix
+        permitted_constraints.map { |k| "#{prefix}#{k}" }
+      end
+
+      def validate_required_constraints_exist
+        validate_constraint_exists :subscription_id
+        validate_constraint_exists :resource_group
+      end
+
```
orenbm: remove this line

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      def validate_constraint_exists constraint
+        raise Err::MissingAnnotationConstraint.new(constraint) unless constraints[constraint]
```
orenbm: Does the error name need to include the word `annotation`?

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+      end
+
+      # validates that the required annotations (subscription_id and resource_group) exist in annotations
```
orenbm: this method doesn't validate a thing. It seems that the description is for another method.

Also - please add a description for this method. why do we retrieve first `"authn-azure/#{@service_id}/#{constraint_name}"` and only then `"authn-azure/#{constraint_name}"`?

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+          subscription-id resource-group user-assigned-identity system-assigned-identity
+        )
+      end
+
+      # validates that the application identity doesn't include logical constraint
+      # combinations (e.g user_assigned_identity & system_assigned_identity)
+      def validate_constraint_combinations
+        identifiers = %i(user_assigned_identity system_assigned_identity)
+
+        identifiers_constraints = constraints.keys & identifiers
```
orenbm: Love it! Super elegant solution!

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -3,15 +3,14 @@
 module Authentication
   module AuthnAzure
 
+    Log = LogMessages::Authentication
```
orenbm: where did the `AuthnAzure` go?

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -20,18 +19,21 @@ module AuthnAzure
...
       private
 
-      #def validate_azure_token
-      #
-      #end
-      #
-      #def validate_application_identity
-      #
-      #end
+      # expecting to receive xms_mirid and oid as inputs (either global instance or function). This will change to hash map.
```
orenbm: Add a todo to remove this comment

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -20,18 +19,21 @@ module AuthnAzure
...
+      def validate_application_identity
+        @validate_application_identity.(
+            service_id: service_id,
+            account: account,
+            username: username,
+            xms_mirid: decoded_credentials["xms_mirid"],
+            oid: decoded_credentials["oid"]
```
orenbm: I added methods to get these variables so we also log the output to the log. Please change to:
```
xms_mirid: xms_mirid,
oid: oid
```

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
+require 'command_class'
+
+module Authentication
+  module AuthnAzure
+
+    Log = LogMessages::Authentication
```
orenbm: Where did the `AuthnAzure` module go?

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+    Err = Errors::Authentication::AuthnAzure
+    # Possible Errors Raised: RoleNotFound, InvalidApplicationIdentity
+
+    ValidateApplicationIdentity = CommandClass.new(
+        dependencies: {
+            role_class:                 Role,
+            resource_class:             Resource,
```
orenbm: Please add rooting (e.g `::Role`). It looks like `Role`, `Resource` & `ApplicationIdentity` are in the same level

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      def call
+        parse_xms_mirid
+        # compare xms_mirid with what is defined in annotations
+        validate_application_identity
+      end
+
+      private
+
+      # validate integrity of annotations
```
orenbm: What are we validating in this method? The comment seems wrong.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+      def application_identity
+        @application_identity ||= @application_identity_class.new(
+            role_annotations: role_annotations,
+            service_id: @service_id
+        )
+      end
+
+      def role_annotations
+        role.annotations
+      end
```
orenbm: seems redundant. You can call `role.annotations` in line 31 instead.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      # convert host ID to full host ID, cucumber:host:somepolicy
+      # TODO For a Conjur user, an incorrect full id will be returned
```
orenbm: You can remove this TODO but please add a bug for this issue as it's more general than our use-case. I can take it if you prefer.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+            role_annotations: role_annotations,
+            service_id: @service_id
+        )
+      end
+
+      def role_annotations
+        role.annotations
+      end
+
+      # convert host ID to full host ID, cucumber:host:somepolicy
```
orenbm: I don't like this comment, it's confusing (and the example is incorrect)

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+      # TODO For a Conjur user, an incorrect full id will be returned
+      def role_id
+        @role_id ||= @role_class.roleid_from_username(@account, @username)
+      end
+
+      def role
+        @role ||= @resource_class[role_id]
+        raise SecurityErr::RoleNotFound(role_id) if @username.nil?
+        return @role
```
orenbm: Please change to:
```
raise SecurityErr::RoleNotFound(role_id) if @username.nil?
@role ||= @resource_class[role_id]
```

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      # TODO add parsing format to check xms mirid is in proper format
```
orenbm: This is quite a big todo. I suggest that you create a method for `validate_xms_mirid_format` and leave it empty so it's more visible that it's missing.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+      # xms_mirid is a term in Azure to define a claim that describes the resource that holds the encoding of the instance's
```
orenbm: That holds among other details

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+      # subscription, resource group, and provider identity needed for authorization. xms_mirid is one of the fields in
```
orenbm: subscription id

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+      # the JWT token. This function will extract the relevant information from xms_mirid claim and populate a representative
+      # hash with the appropriate fields.
+      def parse_xms_mirid
+        # xms_mirid claim starts with an extra '/' so we will be ignoring it for parsing purposes
+        # ex: /subscription/<subscription_id> -> subscription/<subscription_id>
+        _, *field_split = @xms_mirid.split('/')
+        @xms_mirid_hash = Hash[field_split.each_slice(2).to_a]
```
orenbm: extract to method `xms_mirid_hash`?

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+        raise SecurityErr::RoleNotFound(role_id) if @username.nil?
+        return @role
+      end
+
+      # TODO add parsing format to check xms mirid is in proper format
+      # xms_mirid is a term in Azure to define a claim that describes the resource that holds the encoding of the instance's
+      # subscription, resource group, and provider identity needed for authorization. xms_mirid is one of the fields in
+      # the JWT token. This function will extract the relevant information from xms_mirid claim and populate a representative
+      # hash with the appropriate fields.
+      def parse_xms_mirid
```
orenbm: we usually call methods in the name of the returned object. I think we should have a method named `token_identity_from_xms_mirid` (or something) which will use memoization to create an object.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+        #   - validates that the correct provider (Microsoft.ManagedIdentity) for a user identity is defined in the
+        #     claim. If so, the 'user_assigned_identity' attribute will be added to the hash with the corresponding
+        #     value from xms_mirid claim
+        # system-assigned-identity:
+        #   - validates that the correct provider (Microsoft.Compute) for a system identity is defined in the
+        #     claim and its resource is one that we support. At current, we only support a virtualMachines resource.
+        #     If these conditions are met, a 'system_assigned_identity' attribute will be added to the hash with its
+        #     value being the oid field in the JWT token.
+        if @xms_mirid_hash["providers"] == "Microsoft.ManagedIdentity"
+          @logger.debug(Log::AuthnAzure::ExtractingIdentityForAuthentication.new(@xms_mirid_hash["providers"] << "/" << @xms_mirid_hash.keys[-1]))
```
orenbm: `@xms_mirid_hash["providers"] << "/" << @xms_mirid_hash.keys[-1]` -> `"#{@xms_mirid_hash["providers"]}/#{@xms_mirid_hash.keys[-1]}

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+          @token_identity[:user_assigned_identity] = @xms_mirid_hash["userAssignedIdentities"]
+        else
+          @logger.debug(Log::AuthnAzure::ExtractingIdentityForAuthentication.new(@xms_mirid_hash["providers"] << "/" << @xms_mirid_hash.keys[-1]))
```
orenbm: this looks like the same string as above. Please DRY your code.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+          @token_identity[:system_assigned_identity] = @oid
+        end
+      end
+
+      # validate the integrity of host annotations against the xms_mirid object representation
```
orenbm: please don't use the word `host`. verify also in other places in your code as I may have missed it.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,102 @@
...
+      def validate_application_identity
+        @logger.debug(Log::AuthnAzure::ValidatingApplicationIdentity.new(@xms_mirid_hash[@xms_mirid_hash.keys[-1]]))
```
orenbm: seems pretty weird to still use `@xms_mirid_hash` after you transformed it to `token_identity`. Please see how you can use `token_identity` here as well. It's not clear what `@xms_mirid_hash[@xms_mirid_hash.keys[-1]]` represents.

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -1,7 +1,7 @@
 module Authentication
   module AuthnK8s
 
-    Log = LogMessages::Authentication::AuthnK8s
+    Log = LogMessages::Authentication
```
orenbm: what?

app/domain/authentication/authn_k8s/validate_application_identity.rb
```diff
@@ -10,7 +10,7 @@ module AuthnK8s
 
     ValidateApplicationIdentity = CommandClass.new(
       dependencies: {
-        resource_class:              Resource,
+        resource_class:             Resource,
```
orenbm: root please

app/domain/errors.rb
```diff
@@ -277,6 +277,21 @@ module AuthnK8s
 
       module AuthnAzure
 
+        MissingAnnotationConstraint = ::Util::TrackableErrorClass.new(
+            msg: "Host does not have a {0} constraint defined in its annotations",
```
orenbm: please don't use `host`.

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+        raise Err::IllegalConstraintCombinations, identifiers_constraints unless identifiers_constraints.length <= 1
+      end
+
+      def constraint_value constraint_name
+        constraint_from_annotation(constraint_name)
+      end
```
jonahx: This method is superfluous and can be deleted -- essentially you just have to names for the same method.

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+
+      def validate_permitted_annotations
+        validate_prefixed_permitted_annotations("authn-azure/")
+        validate_prefixed_permitted_annotations("authn-azure/#{@service_id}/")
+      end
+
+      # check if annotations with prefix is part of the permitted list
+      def validate_prefixed_permitted_annotations prefix
+        Rails.logger.debug(Log::ValidatingAnnotationsWithPrefix.new(prefix))
```
jonahx: `Rails.logger` is an external dependency and should be injected.  One of the main benefits of having all of this auth code outside of rails in app/domain is that it has no dependency on rails.  This breaks that.

orenbm: So you would change this class to a command class called `init_application_identity`? or something like that.

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,132 @@
...
+    Log = LogMessages::Authentication
+    Err = Errors::Authentication::AuthnAzure
+    # Possible Errors Raised: ConstraintNotSupported, MissingAnnotationConstraint, IllegalConstraintCombinations
+
+    # This class defines an application identity of a given Conjur host.
+    # The constructor initializes an ApplicationIdentity object and validates that
+    # it is configured correctly.
+    # The difference between the validation in this constructor and
+    # the validation in ValidateApplicationIdentity is that here we validate that
+    # the application identity is configured correctly, and thus is a valid application
```
jonahx: It's good that you adding comments to clarify, but that this explanation is required is telling you that your names aren't clear enough yet.  Two different things should have different names. 

 It may be a sign that the logic isn't broken down in the best way too.

Also formatting nit: you should setup a hotkey in your editor so you can instantly square up things like this and keep clean margins.

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: I don't think this is a big TODO, b/c as discussed we assuming that the xms_mirid claim has a consistent length and are verifying it is in the proper format in terms of number of  `/`

Something along the lines of :
```
def validate_xms_mirid_format
   raise Err:Error unless xms_mirid_hash.length == 4`
end
```

orenbm: Don't you want to verify that it consists of all the expected fields too?

app/domain/authentication/authn_azure/application_identity.rb
sigalsax: Good catch 🥇 
Adding the following examples for application identities
Not valid - `authn-azure/some-value`
Valid - `authn-azure/subscription-id` or `authn-azure/subscription-id` or `authn-azure/<service-id>/subscription-id`

app/domain/authentication/authn_azure/application_identity.rb
sigalsax: After defining and giving examples of invalid and valid application identities, adding explanation that 
"`authn-azure/<service-id>/subscription-id` is a valid application identity because `subscription_id` is one of the defined, accepted constraints and will pass the validation here."

app/domain/authentication/authn_azure/application_identity.rb
sigalsax: updated to `validate_azure_annotations_are_permitted` b/c we are validating here if the annotations featured are ones we permit

app/domain/authentication/authn_azure/application_identity.rb
sigalsax: Changed this function to `validate_azure_annotations_are_permitted` with the comment,
"validating that the annotations listed for the host align with the defined, permitted constraints"

app/domain/authentication/authn_azure/application_identity.rb
orenbm: please note that the example `some-namespace/blah/some-value` if for an application identity that is in the id. I suggest writing down a whole host definition with annotations.

app/domain/authentication/authn_azure/application_identity.rb
sigalsax: I think it makes it more clear that we are checking for constraints for Azure in the annotations. 
The msg is reads: "Host does not have {0} constraint defined in its annotations"

If we want to reuse this for other authenticators, we can make it more general. WDYT?
Maybe `Host does not have the required constraint: {0-constraint} defined in policy`?

app/domain/authentication/authn_azure/application_identity.rb
orenbm: The example `some-namespace/service-account/some-value` refers to when the app id is in the host id and not in annotations. Please change your examples to hosts with app id defined in annotations

app/domain/authentication/authn_azure/application_identity.rb
orenbm: 1. Let's not use `Host`
2. What about `Role does not have the required constraint: {0-constraint}` without stating that it's in the annotations?

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: As explained, if Role was in Authentication > AuthnAzure, it would grab that module so you are correct it is not clear

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: why? In the `application_identity` class we are checking that the structure of the annotations is how we expect for the application-identity

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: @orenbm please open this one 🙏

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: This method calls the constructor of `application_identity_class` which creates an application identity object. The fact that the constructor does some validations is irrelevant in the current context.

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: What do you mean? My thought process is as follows:
My goal here is to locate a specific role (`test-policy/vm_sigal` for example). The endpoint is:`/:authenticator/test/cucumber/test-policy/vm_sigal/authenticate`

`role_id` would be `cucumber:test-policy/vm_host` and `@resource_class[cucumber:test-policy/vm_host]` returns the actual resource entry and its data in Conjur

But if `cucumber:test-policy/vm_host` or `@resource_class[cucumber:test-policy/vm_host]` returns nil, then the entity is NotFound

What is wrong with my thought process?

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: like so:
```
def xms_mirid_hash
   @xms_mirid_hash
end
```
?

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: just a small nit: for the request `/authenticator/test/cucumber/host/test-policy/vm_sigal/authenticate` the role id will be `cucumber:host:vm_sigal@test-policy`

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: not at all. more like this:
```
# xms_mirid claim starts with an extra '/' so we will be ignoring it for parsing purposes
# ex: /subscription/<subscription_id> -> subscription/<subscription_id>
def xms_mirid_hash
        _, *field_split = @xms_mirid.split('/')
        Hash[field_split.each_slice(2).to_a]
end
```

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: what about `token_identity_from_claims` b/c it is not only `xms_mirid` that is being used to build the `token_identity` object (also `oid` if system-assigned-identity). This might play into Jonah's comment about creating two separate objects...

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: intentional `"#` w.o a closing "?

sigalsax: updating to `"#{@xms_mirid_hash["providers"]}/#{@xms_mirid_hash.keys[-1]}"`

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: We discussed what to do about repeating this line and it was concluded we should write this twice. I do think we should make separate it in terms of transactions..

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: A couple of ideas as to why I am against this:
1. `xms_mirid_hash` is a direct copy of what is decoded and verified and `token_identity` object is my own object representation with manipulations on the `xms_mirid` claim. Therefore, I am wary to say `xms_mirid_hash` == `token_identity`... UNLESS we address the following:
2. The `token_identity` object doesn't have an entry for the resource name when Azure resource has a system-assigned-identity. We can definitely add it to the `token_identity` w.o harming the logic because we are comparing what are in the annotations against what is the `token_identity` and not the inverse
3. `@xms_mirid_hash[@xms_mirid_hash.keys[-1]]` I agree that this is a bit ugly and my goal is for this to be the last entry in the xms_mirid claim (the resource name). For example: `sigalVM`

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: cool.

app/domain/authentication/authn_k8s/application_identity.rb
sigalsax: made some logs more general during authn-azure development so updated them respectively in this file

app/domain/errors.rb
sigalsax: Updating to `Role`

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: But why is this line repeated? can't you log it before the `if`?

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: You log that you are validating `sigalVM` but you are actually validating the `oid`.

orenbm: I prefer adding in to the token identity rather than to continue referring `xms_mirid_hash`. `xms_mirid_hash` should be a temporary object that helps you move from the `xms_mirid` string to a token identity.

app/domain/authentication/authn_azure/application_identity.rb
sigalsax: Depends if your suggestion is to make it more general so we can reuse this for other authenticators?

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: `@xms_mirid_hash.length == 4 && @xms_mirid_hash.key?("subscriptions" && "resourcegroups" && "providers")`. Would I need to check more IYO?

Additionally, would providers be `Microsoft.Compute/virtualMachines/sigalVMUbuntu` ? b.c rn I am splitting at the `/` and therefore breaking apart the the above string

sigalsax: If we want to be `providers => Microsoft.Compute/virtualMachines/VMUbuntu` we need to change the logic a bit in terms of comparisons b/w user/system-assigned-identities

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: that is how it was before and it was mentioned that there is no value of this separated out without the function that holds this logic. Is this still the case?

app/domain/authentication/authn_azure/application_identity.rb
orenbm: What I want is to not use annotations in case the application identity will be implemented somewhere else in the future.

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: It's not critical IMO. `@xms_mirid_hash.length == 4 && @xms_mirid_hash.key?("subscriptions" && "resourcegroups" && "providers")` is fine.

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: I don't recall what was here before. Let's stick with extracting this to a method.

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: There is actually no bug here. We can go over it tomorrow.

app/domain/authentication/authn_azure/application_identity.rb
sigalsax: updated

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: deleteeeed

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: wasn't sure if oid was sensitive info. So you are suggesting that we have an extra attribute called `resource_name` where for user-identity it is the resource-name (the last parm in the xms_mirid claim) and for system-identity, oid is returned?

sigalsax: My thought process was the following:
- The last index, `sigalVMUbuntu` in the xms_mirid claim would be descriptive enough for the admin to understand which VM is being referred to without exposing too much info about the VM. The same for the user-assigned id

orenbm: let's not get into what is sensitive and what is not for now. I asked @shaharglazner about the `xms_mirid` and `oid`. anyway, you can't tell the user that your are validating `sigalVM` while you are actually validating the `oid`. any part of the `xms_mirid` that is useful for us you can put in the `TokenIdentity`

sigalsax:
closing why you may ask? b/c the branch got messy

PR #1373, opened 2020-02-26
Fix RoleNotAuthorizedOnWebservice message

orenbm:
The error message said that the role doesn't have `authenticate` privilege
even though sometimes we check for other privileges (e.g `update`). This PR
receives the privilege as input for the error message.

### DIFF
--- a/app/domain/authentication/authn_k8s/validate_pod_request.rb
+++ b/app/domain/authentication/authn_k8s/validate_pod_request.rb
@@ -38,7 +38,7 @@ module Authentication
 
       def validate_host_can_access_service
         return if host_can_access_service?
-        raise SecurityErr::RoleNotAuthorizedOnWebservice.new(host.role.id, service_id)
+        raise SecurityErr::RoleNotAuthorizedOnWebservice.new(host.role.id, "authenticate", service_id)
       end
 
       def validate_pod_exists
--- a/app/domain/authentication/validate_webservice_access.rb
+++ b/app/domain/authentication/validate_webservice_access.rb
@@ -60,11 +60,8 @@ module Authentication
       end
 
       def validate_user_has_access
-        # Ensure user has access to the service
         has_access = user_role.allowed_to?(@privilege, webservice_resource)
         unless has_access
-          @logger.debug(Log::UserNotAuthorized
-                          .new(@user_id, webservice_resource_id))
           raise Err::RoleNotAuthorizedOnWebservice.new(@user_id, @privilege, webservice_resource_id)
         end
       end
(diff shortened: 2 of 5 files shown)


eladkug:
LGTM. 
How to verify the fix:
Enable the authn-azure authenticator by curl command (

> curl -k -X PATCH  -H "Authorization: Token token=\"$admin_conjur_access_token\"    --data "enabled=true" https://<host_name_dns>/authn-azure/test/cucumber

if you did it with user which is not admin you will get 403 (which is o.k) but in the log you will see this:

> CONJ00006E 'alice' does not have 'authenticate' privilege on ...

And after the fix suppose to be :

> CONJ00006E 'alice' does not have 'update' privilege on ...

orenbm:
@eladkug - behaviour is validated. Please re-review and approve if it's ok.

I added some more code so I'd rather get a review from @sigalsax as well.

app/domain/authentication/validate_webservice_access.rb
```diff
@@ -60,12 +60,9 @@ def validate_user_is_defined
       end
 
       def validate_user_has_access
-        # Ensure user has access to the service
         has_access = user_role.allowed_to?(@privilege, webservice_resource)
         unless has_access
-          @logger.debug(Log::UserNotAuthorized
```
sigalsax: was this removed b.c of repetition for DEBUG/ERROR logs? In other words, logged in debug and logged in error?

app/domain/authentication/authn_k8s/validate_pod_request.rb
```diff
@@ -38,7 +38,7 @@ def validate_webservice_exists
 
       def validate_host_can_access_service
         return if host_can_access_service?
-        raise SecurityErr::RoleNotAuthorizedOnWebservice.new(host.role.id, service_id)
+        raise SecurityErr::RoleNotAuthorizedOnWebservice.new(host.role.id, "authenticate", service_id)
```
sigalsax: You might need to update specs too no? Ex: `validate_pod_request_spec` `validate_webservice_access_spec`

app/domain/authentication/validate_webservice_access.rb
orenbm: yes

app/domain/authentication/authn_k8s/validate_pod_request.rb
orenbm: No because the message didn't change.

orenbm:
@sigalsax please re-review

Issue #1377, opened 2020-02-28
Log authentication and login failures at the `info` level, rather than `debug`


micahlee:
The generic `401` error response from authenticators (specifically `authn-k8s`) can have a variety of root causes, some of them not directly related to the authentication credentials themselves, but rather issues with the authenticator configuration.

Currently the way to gather more information for a authentication failures is to increase the log level for Conjur to `debug`. However, this also turns on logging of all database queries, and because the rotator process executes a SQL query every second for pending rotations, this makes it very difficult to find the information you're looking for.

The proposal to improve this experience is to log authentication failures at the `info` log level, rather than `debug`.

micahlee:
Blocked by #1464

orenbm:
Unblocked by #1464 as we found no issue in logging these messages in INFO level

PR #1378, opened 2020-02-28
Log the initial line of authentication and login errors at `warn` level

micahlee:
#### Pending

- [ ] Security Review Sign-off

#### What does this PR do?

The generic 401 error response from authenticators (specifically authn-k8s) can have a variety of root causes, some of them not directly related to the authentication credentials themselves, but rather issues with the authenticator configuration.

This change improves the ability for a Conjur operator to quickly pinpoint the cause for an authentication failure, without changing log levels and restarting the Conjur process/container.

This is a follow-up issue to https://github.com/cyberark/conjur/pull/1219.

Connected to #1377

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Improved flows and rules around user creation
 - Kubernetes authenticator now returns 403 on unpermitted hosts instead of a 401
+- Authenticator will now log initial authentication failure reason at the `warn`
+  level rather than `debug`. Full stack trace is still available at `debug` level.
 
 ### Fixed
 - Updated broken links on server status page (#1341)
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -125,7 +125,7 @@ class AuthenticateController < ApplicationController
   private
 
   def handle_login_error(err)
-    logger.debug("Login Error: #{err.inspect}")
+    logger.warn("Login Error: #{err.inspect}")
     err.backtrace.each do |line|
       logger.debug(line)
     end
@@ -141,7 +141,7 @@ class AuthenticateController < ApplicationController
   end
 
   def handle_authentication_error(err)
-    logger.debug("Authentication Error: #{err.inspect}")
+    logger.warn("Authentication Error: #{err.inspect}")
     err.backtrace.each do |line|
       logger.debug(line)
     end


micahlee:
@andytinkham @shaharglazner 

I've created this draft PR to continue the discussion of raising the log level for the top line message from authentication failures (401) from `debug` to `warn`. This was initially included in a [previous PR](https://github.com/cyberark/conjur/pull/1219), but removed from that PR because we decided it was worth thinking more about any security implications of this change (See [this comment thread](https://github.com/cyberark/conjur/pull/1219#discussion_r358726636) for more context).

When you're able, would you please advise on the best way to continue, and what information you would like to continue progressing this forward?

Thanks so much!

CC: @orenbm @jtuttle @sgnn7 @alexkalish @jvanderhoof

micahlee:
My initial observations on the subject:

* The particular content for log messages comes from the Authenticator implementations.

*  The destination for these log messages is the normal Conjur server log (or DAP node log), typically accessed with `docker logs <container_name>`.

izgeri:
@micahlee can you provide some examples of the kinds of log messages you might see once this change is merged in? for example, in the authenticator troubleshooting we have identified a few really common problems - I'm curious whether there would be additional log output visible when people encounter these problems, which might make troubleshooting them (without changing the default log level to debug) easier

orenbm:
> The particular content for log messages comes from the Authenticator implementations.

That's not entirely correct. In the JWT decoder used in authn-oidc and authn-azure we have the following errors:
```
        TokenDecodeFailed = ::Util::TrackableErrorClass.new(
          msg: "Token decode failed (3rdPartyError ='{0}')",
          code: "CONJ00035E"
        )

        TokenVerificationFailed = ::Util::TrackableErrorClass.new(
          msg: "Token verification failed (3rdPartyError ='{0}')",
          code: "CONJ00015E"
        )
```

So we also log error messages that come from outside of our code.

orenbm:
> @micahlee I'm afraid that's not entirely correct. In the JWT decoder used in authn-oidc and authn-azure we have the following errors:

        TokenDecodeFailed = ::Util::TrackableErrorClass.new(
          msg: "Token decode failed (3rdPartyError ='{0}')",
          code: "CONJ00035E"
        )

        TokenVerificationFailed = ::Util::TrackableErrorClass.new(
          msg: "Token verification failed (3rdPartyError ='{0}')",
          code: "CONJ00015E"
        )
So we also log error messages that come from outside of our code.

However, I looked into the errors that are raised by the 3rd party and it doesn't include any sensitive data.

@micahlee @shaharglazner ^

andytinkham:
From a security standpoint, here's what I'd like to see us do for this change. We can make a list of the errors we know could be thrown. It's likely we won't get all of them, but we can get many, including all the ones we control. Review those errors individually and determine the sensitivity of the information in the error. This could be helped by making tests that address our handling of each error (which would be good to have anyhow), particularly if we made all our sensitive information conform to a pattern and then grepped the logs for that pattern after all our test runs.

More broadly, it would be really helpful generally to build out a tiered risk model of all our sensitive info - what do we have that is sensitive? Given a set of bad actor personas (e.g., disgruntled internal attacker who already has some access, external attacker, etc.), what could they do with each piece of sensitive data if they had access to it. I suspect we'd come up with something like secret values are the highest risk, then a tier that has less severe consequences and a tier with even less severe consquences and so on. This would give us a useful risk model that would be reusable any time we were touching data.

orenbm:
thanks @andytinkham ! 
I think this effort is a good opportunity to get all of the logs and errors that we write into the centralized files for [errors](https://github.com/cyberark/conjur/blob/master/app/domain/errors.rb) and [log messages](https://github.com/cyberark/conjur/blob/master/app/domain/logs.rb). this will make it easier for us to know what messages, and data, we write to our logs.

micahlee:
Thanks @andytinkham, I'm closing this PR for now until we can address your suggestions under a new issue: #1464

PR #1380, opened 2020-03-01
Decode JWT token fixes

orenbm:
This PR fixes 2 issues we had when decoding JWT tokens:
1. When we rescue a `JWT::DecodeError` we inspect the error to log it and to re-raise it but we forgot to mark the error in a variable.
2. We didn't rescue `Errno::ETIMEDOUT` when trying to discover the Oauth provider so we didn't raise a 504 Gateway Timeout error properly.

### DIFF
--- a/app/domain/authentication/o_auth/discover_identity_provider.rb
+++ b/app/domain/authentication/o_auth/discover_identity_provider.rb
@@ -34,7 +34,7 @@ module Authentication
         @discovered_provider = @open_id_discovery_service.discover!(@provider_uri).tap do
           @logger.debug(Log::IdentityProviderDiscoverySuccess.new)
         end
-      rescue HTTPClient::ConnectTimeoutError => e
+      rescue HTTPClient::ConnectTimeoutError, Errno::ETIMEDOUT => e
         raise_error(Err::ProviderDiscoveryTimeout, e)
       rescue => e
         raise_error(Err::ProviderDiscoveryFailed, e)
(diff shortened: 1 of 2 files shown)


app/domain/authentication/o_auth/discover_identity_provider.rb
```diff
@@ -34,7 +34,7 @@ def discover_provider
         @discovered_provider = @open_id_discovery_service.discover!(@provider_uri).tap do
           @logger.debug(Log::IdentityProviderDiscoverySuccess.new)
         end
-      rescue HTTPClient::ConnectTimeoutError => e
+      rescue HTTPClient::ConnectTimeoutError, Errno::ETIMEDOUT => e
```
sigalsax: notice that also here, https://github.com/lostisland/faraday/issues/718
they use ::HTTPClient::TimeoutError, Errno::ETIMEDOUT

Seems that this fixes the issue but I would like to understand the difference between these 2.

sigalsax: Do we have a test for these two?

orenbm: We don't have tests for 504 as we had problems with this in the past. I will create an issue for this.

sigalsax: As discussed in person we need to investigate this issue further and design a test to mock a 504 GatewayTimeout and not a 502 BadGateway response

PR #1382, opened 2020-03-01
Log more info when xms_mirid has invalid format

orenbm:


orenbm:
The codeclimate report has all these issues because this PR was opened against master in accident. These issues will be fixed in the PR from `authn-azure` to `master` once ready.

PR #1383, opened 2020-03-02
Raise ClaimInInvalidFormat properly

orenbm:
#### What does this PR do?
1. Raises a ClaimInInvalidFormat when the claim is actually invalid instead of only when the claim was valid. 
2. Parse the xms_mirid separately so it's clearer where the error came from.

### DIFF
--- a/app/domain/authentication/authn_azure/validate_application_identity.rb
+++ b/app/domain/authentication/authn_azure/validate_application_identity.rb
@@ -34,7 +34,7 @@ module Authentication
         begin
           required_keys = %w(subscriptions resourcegroups providers)
           missing_keys = required_keys - xms_mirid_hash.keys
-          raise Err::ClaimInInvalidFormat, "Required keys #{missing_keys} are missing" if missing_keys.empty?
+          raise Err::ClaimInInvalidFormat, "Required keys #{missing_keys} are missing" unless missing_keys.empty?
         rescue => e
           raise Err::ClaimInInvalidFormat, e.inspect
         end
--- a/app/domain/authentication/authn_azure/validate_application_identity.rb
+++ b/app/domain/authentication/authn_azure/validate_application_identity.rb
@@ -19,6 +19,7 @@ module Authentication
     ) do
 
       def call
+        parse_xms_mirid
         validate_xms_mirid_format
         token_identity_from_claims
         validate_azure_annotations_are_permitted
@@ -30,11 +31,9 @@ module Authentication
 
       private
 
-      def validate_xms_mirid_format
+      def parse_xms_mirid
         begin
-          required_keys = %w(subscriptions resourcegroups providers)
-          missing_keys = required_keys - xms_mirid_hash.keys
-          raise Err::ClaimInInvalidFormat, "Required keys #{missing_keys} are missing" unless missing_keys.empty?
+          xms_mirid_hash
         rescue => e
           raise Err::ClaimInInvalidFormat, e.inspect
         end
@@ -67,6 +66,12 @@ module Authentication
           end
           index += 1
         end
+      end
+
+      def validate_xms_mirid_format
+        required_keys = %w(subscriptions resourcegroups providers)
+        missing_keys = required_keys - xms_mirid_hash.keys
+        raise Err::ClaimInInvalidFormat, "Required keys #{missing_keys} are missing" unless missing_keys.empty?
       end
 
       # xms_mirid is a term in Azure to define a claim that describes the resource that holds the encoding of the instance's
--- a/app/domain/authentication/authn_azure/validate_application_identity.rb
+++ b/app/domain/authentication/authn_azure/validate_application_identity.rb
@@ -32,11 +32,7 @@ module Authentication
       private
 
       def parse_xms_mirid
-        begin
-          xms_mirid_hash
-        rescue => e
-          raise Err::ClaimInInvalidFormat, e.inspect
-        end
+        xms_mirid_hash
       end
 
       def xms_mirid_hash
@@ -48,30 +44,36 @@ module Authentication
       # according to fields we need to retrieve from the claim.
       # ultimately, transforming "/key1/value1/key2/value2" to {"key1" => "value1", "key2" => "value2"}
       def parsed_xms_mirid
-        split_xms_mirid = @xms_mirid_token_field.split('/')
+        begin
+          split_xms_mirid = @xms_mirid_token_field.split('/')
 
-        if split_xms_mirid.first == ''
-          split_xms_mirid = split_xms_mirid.drop(1)
-        end
-
-        index = 0
-        split_xms_mirid.each_with_object({}) do |property, xms_mirid_hash|
-          case property
-          when "subscriptions"
-            xms_mirid_hash["subscriptions"] = split_xms_mirid[index + 1]
-          when "resourcegroups"
-            xms_mirid_hash["resourcegroups"] = split_xms_mirid[index + 1]
-          when "providers"
-            xms_mirid_hash["providers"] = split_xms_mirid[index + 1, index + 3]
+          if split_xms_mirid.first == ''
+            split_xms_mirid = split_xms_mirid.drop(1)
           end
-          index += 1
+
+          index = 0
+          split_xms_mirid.each_with_object({}) do |property, xms_mirid_hash|
+            case property
+            when "subscriptions"
+              xms_mirid_hash["subscriptions"] = split_xms_mirid[index + 1]
+            when "resourcegroups"
+              xms_mirid_hash["resourcegroups"] = split_xms_mirid[index + 1]
+            when "providers"
+              xms_mirid_hash["providers"] = split_xms_mirid[index + 1, index + 3]
+            end
+            index += 1
+          end
+        rescue => e
+          raise Err::XmsMiridParseError.new(@xms_mirid_token_field, e.inspect)
         end
       end
 
       def validate_xms_mirid_format
         required_keys = %w(subscriptions resourcegroups providers)
         missing_keys = required_keys - xms_mirid_hash.keys
-        raise Err::ClaimInInvalidFormat, "Required keys #{missing_keys} are missing" unless missing_keys.empty?
+        unless missing_keys.empty?
+          raise Err::MissingRequiredKeysInXmsMirid.new(missing_keys, @xms_mirid_token_field)
+        end
       end
 
       # xms_mirid is a term in Azure to define a claim that describes the resource that holds the encoding of the instance's
--- a/app/domain/authentication/authn_azure/validate_application_identity.rb
+++ b/app/domain/authentication/authn_azure/validate_application_identity.rb
@@ -20,7 +20,7 @@ module Authentication
       def call
         validate_xms_mirid_format
         token_identity_from_claims
-        compare_token_identity_with_annotations
+        validate_token_identity_matches_annotations
       end
 
       private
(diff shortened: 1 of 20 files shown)


app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -30,11 +31,9 @@ def call
 
       private
 
-      def validate_xms_mirid_format
+      def parse_xms_mirid
         begin
-          required_keys = %w(subscriptions resourcegroups providers)
-          missing_keys = required_keys - xms_mirid_hash.keys
-          raise Err::ClaimInInvalidFormat, "Required keys #{missing_keys} are missing" if missing_keys.empty?
+          xms_mirid_hash
```
eladkug: In this function(xms_mirid_hash)  you dont raise exception . You prefer to handle it separately on the caller function? The problem is we will not know that the problem was with Hash creating.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -30,14 +31,8 @@ def call
...
 
-      def validate_xms_mirid_format
-        begin
-          required_keys = %w(subscriptions resourcegroups providers)
-          missing_keys = required_keys - xms_mirid_hash.keys
-          raise Err::ClaimInInvalidFormat, "Required keys #{missing_keys} are missing" if missing_keys.empty?
-        rescue => e
-          raise Err::ClaimInInvalidFormat, e.inspect
-        end
+      def parse_xms_mirid
```
eladkug: why we need three function for it:
1. parse_xms_mirid call to 
2. xms_mirid_hash which call to
3. parsed_xms_mirid

i think that we can remove at least one function here

app/domain/errors.rb
```diff
@@ -292,16 +292,21 @@ module AuthnAzure
           code: "CONJ00050E"
         )
 
-        ClaimInInvalidFormat = ::Util::TrackableErrorClass.new(
-          msg:  "xms_mirid claim has been received in an invalid format. Reason: {0}",
+        TokenFieldNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
+          msg:  "Field '{0-field-name}' not found or empty in token",
```
eladkug: not found or empty token

orenbm: what do you mean?

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: Each one has its own purpose. The Command Class's `call` method needs to tell the story of what is going on - thus it has a call to the `parse_xms_mirid`.

`xms_mirid_hash` needs to be memoized so we don't do the work of `parsed_xms_mirid` every time. that's why it calls the `parsed_xms_mirid` method that does all the work and memoizes its output. 

the `parsed_xms_mirid` method does the actual parsing and has that name as in ruby we give methods that return an object the name of the returned object.

app/domain/errors.rb
eladkug: need to remove the word "in" (empty **in** token)

orenbm: but it should be there :)

the field is missing inside the token. the token itself is not missing.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -49,23 +44,35 @@ def xms_mirid_hash
...
-            xms_mirid_hash["providers"] = split_xms_mirid[index + 1, index + 3]
+          index = 0
+          split_xms_mirid.each_with_object({}) do |property, xms_mirid_hash|
+            case property
+            when "subscriptions"
+              xms_mirid_hash["subscriptions"] = split_xms_mirid[index + 1]
+            when "resourcegroups"
+              xms_mirid_hash["resourcegroups"] = split_xms_mirid[index + 1]
+            when "providers"
+              xms_mirid_hash["providers"] = split_xms_mirid[index + 1, index + 3]
```
sgnn7: Drive-by comment: is a default case needed here?

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -49,23 +44,35 @@ def xms_mirid_hash
...
+            end
+            index += 1
           end
-          index += 1
+        rescue => e
+          raise Err::XmsMiridParseError.new(@xms_mirid_token_field, e.inspect)
+        end
+      end
+
+      def validate_xms_mirid_format
```
orenbm: @sigalsax do you think that we should validate also that the value of `providers` has 3 parts?

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: Thanks for taking a look @sgnn7 ! There is actually no need for a default case as any other properties are not needed in the hash. We added this so we don't assume too much on the structure of the `xms_mirid` field and we assume only what we have to assume.

However, I do see a bug here as is needs to be `split_xms_mirid[index + 1, 3]`. The current code would have taken `index + 3` parts of `split_xms_mirid` and put them in the value for `xms_mirid_hash["providers"]`, as can be seen here - https://apidock.com/ruby/v1_9_3_125/Array/slice

@sigalsax  ^^

sgnn7: @orenbm Fair enough - I just wanted to make sure that we wouldn't have exceptions when we got some unexpected case values :)

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -49,23 +44,35 @@ def xms_mirid_hash
...
+        required_keys = %w(subscriptions resourcegroups providers)
+        missing_keys = required_keys - xms_mirid_hash.keys
+        unless missing_keys.empty?
+          raise Err::MissingRequiredKeysInXmsMirid.new(missing_keys, @xms_mirid_token_field)
```
sigalsax: `MissingRequiredKeysInXmsMirid` please add this to the log issue for TWs
https://github.com/cyberark/conjur-docs/issues/742
Create a new section "Added logs" to the end so Shuli knows which new ones to review

app/domain/errors.rb
```diff
@@ -292,16 +292,21 @@ module AuthnAzure
           code: "CONJ00050E"
         )
 
-        ClaimInInvalidFormat = ::Util::TrackableErrorClass.new(
```
sigalsax: Please update here cyberark/conjur-docs#742 that you are removing this log

app/domain/errors.rb
```diff
@@ -292,16 +292,21 @@ module AuthnAzure
...
-          msg:  "xms_mirid claim has been received in an invalid format. Reason: {0}",
+        TokenFieldNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
+          msg:  "Field '{0-field-name}' not found or empty in token",
           code: "CONJ00051E"
         )
 
-        TokenFieldNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
-          msg:  "Field '{0-field-name}' not found or empty in token",
+        XmsMiridParseError = ::Util::TrackableErrorClass.new(
```
sigalsax: `XmsMiridParseError ` please add this to the log issue for TWs
cyberark/conjur-docs#742
Create a new section "Added logs" to the end so Shuli knows which new ones to review

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: good catch!

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: I don't think we should limit ourselves to 3 parts... regardless, we specify `index + 1, 3` so we will take the 3 elements and otherwise an exception will be caught in try/catch clause above

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: Although the refactor is cleaner, was there a problem with the logic that this was refactored?

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -30,14 +31,8 @@ def call
 
       private
 
-      def validate_xms_mirid_format
-        begin
-          required_keys = %w(subscriptions resourcegroups providers)
-          missing_keys = required_keys - xms_mirid_hash.keys
-          raise Err::ClaimInInvalidFormat, "Required keys #{missing_keys} are missing" if missing_keys.empty?
-        rescue => e
-          raise Err::ClaimInInvalidFormat, e.inspect
```
sigalsax: why was this removed in refactor? Don't we want to catch any errors that are raised?

app/domain/errors.rb
```diff
@@ -292,16 +292,21 @@ module AuthnAzure
...
+          msg:  "Failed to parse xms_mirid {0}. Reason: {1}",
           code: "CONJ00052E"
         )
 
+        MissingRequiredKeysInXmsMirid = ::Util::TrackableErrorClass.new(
+          msg:  "Required keys {0} are missing in xms_mirid {1}",
```
sigalsax: Why are we naming this "keys"? Maybe Fields? Properties?

sigalsax: `MissingRequiredKeysInXmsMirid ` please add this to the log issue for TWs
cyberark/conjur-docs#742
Create a new section "Added logs" to the end so Shuli knows which new ones to review

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: Yes. we raised an error and rescued it straight after that so the original error was never raised

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: That's not entirely correct. In case of the following xms_mirid:
```
/subscriptions/<subscription-id>/resourcegroups/<some-value>/providers/<just_one_value>
```
then `split_xms_mirid[index + 1, 3]` will be `[<just_one_value>]`. So if the providers part is less than 3 parts then `split_xms_mirid[index + 1, 3]` will return less than 3 parts and the `providers` field will be invalid.

orenbm: More importantly - Why don't you think we should limit ourselves to 3 parts? We already do it in the code.

app/domain/errors.rb
orenbm: I added these 2 as a PR comment

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: I rescue errors in the new method for the parse. This way it's clearer which logic raised the error.

app/domain/errors.rb
orenbm: Changed to "fields"

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: Okay, I was under the assumption that `split_xms_mirid[index + 1, 3]` but throw an index out of bounds exception.

sigalsax: So let's check for the length as well so that if Azure changes in the future, the user will know where the failure occured

app/domain/errors.rb
```diff
@@ -292,16 +292,26 @@ module AuthnAzure
...
         )
 
+        MissingRequiredFieldsInXmsMirid = ::Util::TrackableErrorClass.new(
+          msg:  "Required fields {0} are missing in xms_mirid {1}",
+          code: "CONJ00053E"
+        )
+
+        MissingProviderFieldsInXmsMirid = ::Util::TrackableErrorClass.new(
+          msg:  "Provider fields are missing in xms_mirid {1}",
+          code: "CONJ00053E"
```
sigalsax: update the error code to `CONJ00053E `

sigalsax: Do we want to print the providers field that were extracted instead of the whole xms_mirid string?

orenbm: No, I prefer that the user will get the whole field

orenbm:
Closing this due to conflicts that raised when we rebased the `authn-azure` branch onto master. Created PR https://github.com/cyberark/conjur/pull/1384 instead.

PR #1389, opened 2020-03-05
Generalize cucumber steps & helpers

orenbm:
#### What does this PR do?
Generalize some steps so they can be used in new tests.

As part of the authn-azure cucumber tests we came upon some steps that can be used if they will be generalized.

### DIFF
--- a/cucumber/api/features/step_definitions/authz_steps.rb
+++ b/cucumber/api/features/step_definitions/authz_steps.rb
@@ -67,7 +67,7 @@ Given(/^I permit user "([^"]*)" to "([^"]*)" user "([^"]*)"$/) do |grantee, priv
 end
 
 Given(/^I set annotation "([^"]*)" to "([^"]*)"$/) do |name, value|
-  @current_resource.add_annotation name: name, value: value
+  set_annotation_to_resource(name, value)
 end
 
 Given(/^I create (\d+) secret values?$/) do |n|
@@ -85,8 +85,9 @@ Given(/^I add the secret value(?: "([^"]*)")? to the resource(?: "([^"]*)")?$/) 
   Secret.create resource_id: resource_id, value: value
 end
 
-Given(/^I permit user "([^"]*)" to "([^"]*)" it$/) do |grantee, privilege|
-  grantee = lookup_user(grantee)
+Given(/^I permit (user|host) "([^"]*)" to "([^"]*)" it$/) do |role_type, grantee, privilege|
+  grantee = role_type == "user" ? lookup_user(grantee) : lookup_host(grantee)
+
   target = @current_resource
   unless grantee.allowed_to?(privilege, target)
     target.permit privilege, grantee
@@ -102,4 +103,11 @@ Given(/^I grant user "([^"]*)" to user "([^"]*)"$/) do |grantor, grantee|
   grantor = lookup_user(grantor)
   grantee = lookup_user(grantee)
   grantor.grant_to grantee
+end
+
+Given(/^I grant group "([^"]*)" to (user|host) "([^"]*)"$/) do |group_id, role_type, grantee|
+  group = lookup_group(group_id)
+
+  grantee = role_type == "user" ? lookup_user(grantee) : lookup_host(grantee)
+  group.grant_to grantee
 end
--- a/cucumber/api/features/step_definitions/user_steps.rb
+++ b/cucumber/api/features/step_definitions/user_steps.rb
@@ -18,6 +18,12 @@ Given(/^I have user "([^"]*)"$/) do |login|
   end
 end
 
+Given(/^I have host "([^"]*)"$/) do |login|
+  unless host_exists?(login)
+    create_host login, @current_user || admin_user
+  end
+end
+
 Given(/^I create a new admin-owned user "(.*?)"$/) do |login|
   create_user login, admin_user
 end
--- a/cucumber/api/features/support/authz_helpers.rb
+++ b/cucumber/api/features/support/authz_helpers.rb
@@ -33,6 +33,12 @@ module AuthzHelpers
 
     @resources[resource_id] = @current_resource
   end
+
+  def set_annotation_to_resource(name, value, resource = @current_resource)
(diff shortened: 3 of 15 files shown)


cucumber/api/features/support/machine_helpers.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+#
+# Utility methods for performing actions in remote machines
+#
```
sgnn7: I kind of feel like `machine**` is maybe too generic. What about `ssh_**` instead or something similar?

orenbm: @sgnn7 i was debating between `ssh_helpers` and `machine_helpers` and went with the latter for 2 reasons:
1. this is what we have in appliance: https://github.com/conjurinc/appliance/blob/master/features/step_definitions/machine_steps.rb
2. we may want to add in future more abilities such as to scp to a machine.

what do you think? go with `ssh_helpers` and add `scp_helpers` in the future or stay with machine?

sgnn7: Hmm... I see your reasoning and it makes sense though "machine" is pretty ambiguous even though we have it in the appliance. Since `scp` is done over `ssh`, I think `ssh_helpers` would still be valid but even `remote_machine_helpers` or something similar would be fine as long as it's not just plain `machine_*`.

orenbm: renamed to `ssh_`, thanks!

cucumber/api/features/step_definitions/authz_steps.rb
```diff
@@ -67,7 +67,7 @@
 end
 
 Given(/^I set annotation "([^"]*)" to "([^"]*)"$/) do |name, value|
-  @current_resource.add_annotation name: name, value: value
+  set_annotation_to_resource(name, value)
```
sigalsax: Missing resource param

cucumber/api/features/step_definitions/authz_steps.rb
```diff
@@ -103,3 +104,10 @@
   grantee = lookup_user(grantee)
   grantor.grant_to grantee
 end
+
+Given(/^I grant group "([^"]*)" to (user|host) "([^"]*)"$/) do |group_id, role_type, grantee|
```
sigalsax: Are you anticipating this will be used in the future? B.c rn this is not used in codebase

orenbm: It is used in another PR for authn-azure tests

cucumber/api/features/step_definitions/user_steps.rb
```diff
@@ -18,6 +18,12 @@
   end
 end
 
+Given(/^I have host "([^"]*)"$/) do |login|
```
sigalsax: This also isn't used in code, I assume we will need this for later?

orenbm: Yes, i need it in the azure tests i'm writing

cucumber/api/features/support/authz_helpers.rb
```diff
@@ -33,6 +33,12 @@ def i_have_a_new_resource(kind, identifier)
 
     @resources[resource_id] = @current_resource
   end
+
+  def set_annotation_to_resource(name, value, resource = @current_resource)
+    unless resource.annotations.find { |a| a.values[:name] == name }
+      resource.add_annotation name: name, value: value
```
sigalsax: add_annotations is a function that does not yet exist so wouldn't I be getting method not found error?

orenbm: but the function does exist.

orenbm: it was used in `authz_steps.rb`. i just moved the usage to the helper

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -173,13 +187,27 @@ def create_user login, owner
...
-
     roleid = "cucumber:user:#{login}"
-    Role.create(role_id: roleid).tap do |user|
-      Credentials[role: user] || Credentials.new(role: user).save(raise_on_save_failure: true)
+    create_role(roleid, owner)
+  end
+
+  # Create a regular host, owned by the admin user
+  def create_host login, owner
+    unless login
```
sigalsax: doesn't look like login will ever not be supplied so let's look into why we need the `unless login`

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -173,13 +187,27 @@ def create_user login, owner
...
+      login = USER_NAMES[@host_index]
```
sigalsax: why do we need this array? RestHelpers defines this as 
```
USER_NAMES = %w[auto-larry auto-mike auto-norbert auto-otto].freeze
```
so I am trying to understand why we have this and a host_index?

cucumber/api/features/support/ssh_helpers.rb
```diff
@@ -0,0 +1,16 @@
...
+require 'net/ssh'
+
+# Utility methods for performing actions in remote machines
+module SshHelpers
+
+  def run_command_in_machine(machine_ip, machine_username, machine_password, command)
+    ssh = Net::SSH.start(machine_ip, machine_username, :password => machine_password)
+    res = ssh.exec!(command)
+    ssh.close
+    res
```
sigalsax: I imagine we can use .tap here instead of just returning `res`

cucumber/api/features/support/rest_helpers.rb
orenbm: thanks, removed this logic.

cucumber/api/features/support/rest_helpers.rb
orenbm: we had it for the default username. I removed it.

cucumber/api/features/support/ssh_helpers.rb
orenbm: Done

PR #1392, opened 2020-03-08
Run each authenticator's cucumber tests separately in Jenkins

orenbm:
#### What does this PR do?

We would like to run each authenticator in its separate Jenkins step
so we can easily pinpoint a failure to know which authenticator failed.
This will also let us run smaller tests locally while developing an authenticator,
and will make adding tests for a new authenticator easier.


ci/test
```diff
@@ -30,12 +30,33 @@ all() {
   done
 }
 
-cucumber_authenticators() {
-  : DOC - Runs Cucumber Authenticator features
+cucumber_authenticators_config() {
+  : DOC - Runs Cucumber Authenticator configuration features
```
jonahx: I've noticed this `: DOC` thing a few places... what's it being used for?   Why not just use a comment?  Is automatic documentation being generated someplace?

Jenkinsfile
```diff
@@ -49,8 +49,17 @@ pipeline {
         stage('RSpec') {
           steps { sh 'ci/test rspec' }
         }
-        stage('Authenticators') {
-          steps { sh 'ci/test cucumber_authenticators' }
+        stage('Authenticators Config') {
+          steps { sh 'ci/test cucumber_authenticators_config' }
+        }
+        stage('Authenticators Status') {
+          steps { sh 'ci/test cucumber_authenticators_status' }
```
eladkug: Both related to OIDC i would change the name of the state for OIDC Authenticators..., and the feature file name as well..

cucumber/authenticators_status/features/authn_status.feature
```diff
@@ -100,7 +45,6 @@ Feature: Authenticator status check
       member: !user alice
     """
     And I am the super-user
-    And I successfully set OIDC variables
```
eladkug: if we dont need it can we remove the implementation as well?

Jenkinsfile
orenbm: Why is `Authenticators Config` OIDC related? this test suite include `available_authn.feature` & `authn_config.feature` which have nothing to do with OIDC.

`Authenticators Status` _is_ related to OIDC but is actually separate. As you can see it doesn't assume that an OIDC provider is in tact. This test checks all the non-specific status checks (e.g webservice is loaded) so it runs separately. Please take a look at `authn_status.feature` to see this.

Let me know if you still think it should run under OIDC.

cucumber/authenticators_status/features/authn_status.feature
orenbm: But we need the implementation in the OIDC tests. I only removed it from here so there is no dependency on OIDC as this test checks all the non-specific authenticator checks.

orenbm:
@jonahx @micahlee I noticed there are CC issues. I'm not fixing them at the moment so your review will be easier.

Jenkinsfile
eladkug: agree for `Authenticators Config`. 
In `authn_status.feature `i see the last 2 tests contain:
When I GET "/**authn-oidc/keycloak/**non-existing-account/status" in several tests so this one should be oidc.
so lets change the Status to OIDC status (or moving the specific tests to different place)

orenbm: Again - the separation here is that these tests checks the general checks. We use here the OIDC authenticator just because it is the only authenticator that implements this check. The tests in `authn_status_oidc.feature` tests the oidc specific tests. The tests in `authn_status.feature` tests the general tests. 

Here's a rule of thumb - If once `authn-azure` is implemented we can replace every occurrence of `authn-oidc` to `authn-azure` and the test is still the same then that test belongs in `authn_status.feature`. If the replacements changed the test then it is oidc specific.

orenbm: Take a look here - https://github.com/cyberark/conjur/blob/master/app/domain/authentication/validate_status.rb

all these checks are not authenticator-specific. `authn_status.feature` tests this file. The only reason that feature file has the word OIDC is because if we wouldn't have had it then the check will fail [here](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/validate_status.rb#L29) . If the authenticator doesn't exist or doesn't implement the status check then we can't test the case that the webservice doesn't exist

orenbm: And we will need to duplicate these tests in every `authn_status_something.feature` instead of having it in one place that doesn't assume anything about the authenticator. 

You can see that `authn_Status.feature` runs without even having a `keycloak` container up.

orenbm:
@eladkug I added a description to `authn_status.feature` to improve readability. wdyt?

PR #1393, opened 2020-03-09
Fix RoleNotFound initialized constant SecurityErr

sigalsax:
#### What does this PR do?
In the code we have
```
@role ||= @resource_class[role_id].tap do |role|
  raise SecurityErr::RoleNotFound(role_id) unless role
end
```
but we don’t ever have the following:
```
SecurityErr = Errors::Authentication::Security
```

so we get an error: initialized constant SecurityErr

This PR fixes this bug

### DIFF
--- a/app/domain/authentication/authn_azure/validate_application_identity.rb
+++ b/app/domain/authentication/authn_azure/validate_application_identity.rb
@@ -5,7 +5,11 @@ module Authentication
 
     Log = LogMessages::Authentication::AuthnAzure
     Err = Errors::Authentication::AuthnAzure
-    # Possible Errors Raised: RoleNotFound, InvalidApplicationIdentity
+    SecurityErr = Errors::Authentication::Security
+
+    # Possible Errors Raised: RoleNotFound, InvalidApplicationIdentity, XmsMiridParseError,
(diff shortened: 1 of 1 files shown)


orenbm:
@sigalsax it seems that your local branch is not up to date with `authn-azure`. This PR should have only 1 changed file.

orenbm:
Also - if this is the only occurence of `SecurityErr` I think it is better to have `Errors::Authentication::Security` in the code itslef:

```
@role ||= @resource_class[role_id].tap do |role|
  raise Errors::Authentication::Security::RoleNotFound(role_id) unless role
end
```

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -161,7 +160,7 @@ def validate_token_identity_matches_annotations
 
       def role
         @role ||= @resource_class[role_id].tap do |role|
-          raise SecurityErr::RoleNotFound(role_id) unless role
+          raise raise Errors::Authentication::Security::RoleNotFound(role_id) unless role
```
orenbm: is the word `raise` written twice?

PR #1395, opened 2020-03-09
Refactor authenticators_common to _authenticators_common

orenbm:
The directory '_authenticators_common' needs to start with an underscore
so every authenticator directory is below it in the hierarchy. If the authenticator
directory is above it (e.g authenticators_azure is above authenticators_common in an
alphabetical order) then we will not be able to load into it the methods defined in authenticators_common.

A comment about this was written in `cucumber.yml` where we load the directory. It is not
possible to add a note in the load lines themselves as it mucks up the yml file.


sigalsax:
Please take a look at codeclimate if there are some helpful tips that we could take

PR #1404, opened 2020-03-10
Test improvements

orenbm:
#### What does this PR do?

1. Fix ssh helper for running command in remote machine - Fixes the way we close the connection. Also sends the password in the way suggested in CC.
2. Allow running cucumber tests with env_args without services - In the current implementation it was not possible to send env_args
without also sending services. In authn-azure we need to send env args
but we don't have any additional services so we need this ability.

### DIFF
--- a/cucumber/api/features/support/ssh_helpers.rb
+++ b/cucumber/api/features/support/ssh_helpers.rb
@@ -6,9 +6,9 @@ require 'net/ssh'
 module SshHelpers
 
   def run_command_in_machine(machine_ip, machine_username, machine_password, command)
-    ssh = Net::SSH.start(machine_ip, machine_username, :password => machine_password)
+    ssh = Net::SSH.start(machine_ip, machine_username, password: machine_password)
     ssh.exec!(command).tap do
-      close
+      ssh.close
     end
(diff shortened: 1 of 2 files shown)


cucumber/api/features/support/ssh_helpers.rb
```diff
@@ -6,9 +6,9 @@
 module SshHelpers
 
   def run_command_in_machine(machine_ip, machine_username, machine_password, command)
-    ssh = Net::SSH.start(machine_ip, machine_username, :password => machine_password)
```
sigalsax: why was this changed? `:something` is a property of a hash that targets the key. Is NetSSH not expecting a password property?

sigalsax:
NOTE: this is opened against master and not authn-azure

cucumber/api/features/support/ssh_helpers.rb
orenbm: It does expect `password` and still gets it. I just changed the syntax

orenbm:
> NOTE: this is opened against master and not authn-azure

@sigalsax it should be :)

PR #1409, opened 2020-03-11
Miscellaneous improvements found during authn-azure dev

orenbm:
During the development of authn-azure we came upon different places
where , as jolly good boy and girl scouts, we improved the code.
This PR centralizes all these improvements that are not related
to authn-azure.

### DIFF
--- a/app/domain/authentication/authn_k8s/application_identity.rb
+++ b/app/domain/authentication/authn_k8s/application_identity.rb
@@ -143,13 +143,14 @@ module Authentication
           annotation_name = a.values[:name]
 
           # Calculate the granularity level of the annotation.
-          # For example, the annotation "authn-k8s/namespace" is in the general level,
-          # and applies to every host that tries to authenticate with the k8s authenticator, regardless
-          # of the service id.
-          # The annotation "authn-k8s/#{@service_id}/namespace" is on the service-id level,
-          # and applies only to hosts trying to authenticate with the authenticator "authn-k8s/#{@service_id}".
+          # For example, the annotation "authn-k8s/namespace" is in the general
+          # level, and applies to every host that tries to authenticate with the
+          # k8s authenticator, regardless of the service id.
+          # The annotation "authn-k8s/#{@service_id}/namespace" is on the
+          # service-id level, and applies only to hosts trying to authenticate
+          # with the authenticator "authn-k8s/#{@service_id}".
           annotation_granularity_level = annotation_name.split('/').length
-          prefix_granularity_level = prefix.split('/').length
+          prefix_granularity_level     = prefix.split('/').length
 
           annotation_name.start_with?(prefix) &&
             # Verify we take only annotations from the same level.
--- a/app/domain/authentication/authn_k8s/application_identity.rb
+++ b/app/domain/authentication/authn_k8s/application_identity.rb
@@ -85,7 +85,9 @@ module Authentication
         controllers = %i(deployment deployment_config stateful_set)
 
         controller_constraints = constraints.keys & controllers
-        raise Err::IllegalConstraintCombinations, controller_constraints unless controller_constraints.length <= 1
+        unless controller_constraints.length <= 1
+          raise Errors::Authentication::IllegalConstraintCombinations, controller_constraints
+        end
       end
 
       def constraint_value constraint_name
@@ -102,7 +104,7 @@ module Authentication
 
         # return the value of the annotation if it exists, nil otherwise
         if annotation
-          Rails.logger.debug(Log::RetrievedAnnotationValue.new(name))
+          Rails.logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
           annotation[:value]
         end
       end
@@ -127,7 +129,7 @@ module Authentication
       end
 
       def validate_prefixed_permitted_annotations prefix
-        Rails.logger.debug(Log::ValidatingAnnotationsWithPrefix.new(prefix))
+        Rails.logger.debug(LogMessages::Authentication::ValidatingAnnotationsWithPrefix.new(prefix))
 
         prefixed_k8s_annotations(prefix).each do |annotation|
           annotation_name = annotation[:name]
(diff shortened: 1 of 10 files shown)


spec/support/fetch_secrets_helper.rb
```diff
@@ -18,7 +18,7 @@
 
   let(:mocked_secret) do
     double('Secret').tap do |secret|
-      allow(secret).to receive(:value).and_return("mocked-secret")
+      allow(secret).to receive(:value).and_return(+"mocked-secret") # unfreezing the secret for authn-azure tests
```
jonahx: 1. Put comment above so line doesn't exceed 80 chars.
2. Can you explain more how this will be used, because my first instinct is that there should never be a use-case where you need to mutate strings.

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -85,7 +85,7 @@ def validate_constraint_combinations
         controllers = %i(deployment deployment_config stateful_set)
 
         controller_constraints = constraints.keys & controllers
-        raise Err::IllegalConstraintCombinations, controller_constraints unless controller_constraints.length <= 1
+        raise Errors::Authentication::IllegalConstraintCombinations, controller_constraints unless controller_constraints.length <= 1
```
jonahx: just a formatting nit, but for line length reasons i'd prefer:

```
        unless controller_constraints.length <= 1
          raise Errors::Authentication::IllegalConstraintCombinations, controller_constraints
        end
```

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -141,7 +141,7 @@ def prefixed_k8s_annotations prefix
           annotation_name = a.values[:name]
 
           annotation_name.start_with?(prefix) &&
-            # Verify we take only annotations from the same level
+            # verify we take only annotations from the same level
```
jonahx: I'd add a code example to the comments here showing an actual name and prefix so the reader can see what this logic is doing.  eg, why does "same level" mean that the number of parts of one is one more than the number of parts of the other?

also consider breaking into multiple statements for clarity.

spec/app/domain/authentication/authn_k8s/validate_pod_request_spec.rb
```diff
@@ -42,13 +42,13 @@
...
   let(:pod) { double("Pod", :spec => pod_spec) }
-  
+
   let(:k8s_object_lookup_class) { double("K8sObjectLookup") }
 
   let(:validate_application_identity) { double("ValidateApplicationIdentity") }
 
-  let(:dependencies) { { resource_class:             double(),
-                         k8s_object_lookup_class:   k8s_object_lookup_class,
+  let(:dependencies) { { resource_class:                double(),
```
jonahx: Formatting nits:

For multiline statements ruby style guide prescribes `do.. end` over `{...}`.

Separately, I'd opt for the opening / closing braces of hash on separate lines, though I'm unsure what the guidelines are on that, just stating my taste on that one...
```
let(:dependencies) do 
  { 
    resource_class:                double(),
    k8s_object_lookup_class:       k8s_object_lookup_class,
    validate_application_identity: validate_application_identity
  }
end
```
I really want to automate all this formatting stuff... it's a time waster.  I have a possible solution I'll be posting about today.

spec/app/domain/authentication/authn_k8s/validate_pod_request_spec.rb
```diff
@@ -80,7 +80,7 @@
       .to receive(:new)
             .and_return(validate_application_identity)
     allow(validate_application_identity).to receive(:call)
-                                          .and_return(true)
+                                              .and_return(true)
```
jonahx: More formatting.  In my recent work I've been doing:
```
allow(Authentication::AuthnK8s::ValidateApplicationIdentity).to(
  receive(:new).and_return(validate_application_identity)
)
allow(validate_application_identity).to(
  receive(:call).and_return(true)
)
```
Wdyt?

spec/support/fetch_secrets_helper.rb
orenbm: 1. Done.
2. Please see here - https://github.com/cyberark/conjur/pull/1406

spec/app/domain/authentication/authn_k8s/validate_pod_request_spec.rb
orenbm: It does look better but we have the current style 100+ places. Let's address this out of the scope of this PR.

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -140,9 +142,18 @@ def prefixed_k8s_annotations prefix
...
 
+          # Calculate the granularity level of the annotation.
+          # For example, the annotation "authn-k8s/namespace" is in the general level,
+          # and applies to every host that tries to authenticate with the k8s authenticator, regardless
+          # of the service id.
+          # The annotation "authn-k8s/#{@service_id}/namespace" is on the service-id level,
+          # and applies only to hosts trying to authenticate with the authenticator "authn-k8s/#{@service_id}".
+          annotation_granularity_level = annotation_name.split('/').length
+          prefix_granularity_level = prefix.split('/').length
+
```
jonahx: Much better.

I have two comments, one important and one cosmetic, neither of them requiring you to make any changes atm.

## Important comment

What is going on here is indicating that we are missing an `AnnotationName` value object (or something similar) which would house the logic and allow you to write something like:
```ruby
annotation_name.granularity == prefix.granularity
```
To be clear, the details of the correct interface are something I'd need more context to advise on, I'm just trying to indicate the general idea that you don't want code like the above responsible for having to deal with this kind of thing.  More importantly, without a value object to house it, the same logic will often spread out in multiple places in your code base.

This has been a huge problem with the details around constructing conjur variables, eg, you'll see colon-joining code scattered in tens of places throughout this code base and even others.  We've made this mistake in other places too.

This change is probably out of scope for this PR, I'm just bringing it up as something to be aware of when you see it and in future design work.

## Cosmetic comment

Turn on a paragraph hard-wrapping hotkey in your editor so your comments will have clean 80-char margins :)

app/domain/errors.rb
```diff
@@ -173,104 +178,99 @@ module AuthnIam
...
 
         UntrustedClientCertificate = ::Util::TrackableErrorClass.new(
-          msg: "Client certificate cannot be verified by certification authority",
+          msg:  "Client certificate cannot be verified by certification authority",
           code: "CONJ00030E"
         )
 
         CommonNameDoesntMatchHost = ::Util::TrackableErrorClass.new(
-          msg: "Client certificate CN must match host name. Cert CN: {0}. " \
+          msg:  "Client certificate CN must match host name. Cert CN: {0}. " \
```
jonahx: This is cool, I was unaware you could do this.

We should do this everywhere we can to avoid the really long lines.

app/domain/logs.rb
```diff
@@ -138,27 +133,27 @@ module AuthnK8s
...
 
       RateLimitedCacheUpdated = ::Util::TrackableLogMessageClass.new(
-        msg: "Rate limited cache updated successfully",
+        msg:  "Rate limited cache updated successfully",
         code: "CONJ00016D"
       )
 
       RateLimitedCacheLimitReached = ::Util::TrackableLogMessageClass.new(
-        msg: "Rate limited cache reached the '{0-limit}' limit and will not call target for the next '{1-seconds}' seconds",
+        msg:  "Rate limited cache reached the '{0-limit}' limit and will not call target for the next '{1-seconds}' seconds",
```
jonahx: Use your line splitting trick here too.

app/domain/authentication/authn_k8s/application_identity.rb
orenbm: Thanks, we'll take this as part of the refactor we'll do for the application identity.

PR #1414, opened 2020-03-12
Add integration tests for authn-azure

orenbm:
#### What does this PR do?
Adds integration tests for authn-azure and runs them in the Jenkins pipeline.

We use a Jenkins plugin that creates an Azure VM for the lifetime of the pipeline and terminates it once authn-azure tests are done.


Jenkinsfile
```diff
@@ -69,6 +69,39 @@ pipeline {
...
+                env.KEEP_AZURE_AUTHN_INSTANCE = "true"
+                while(env.KEEP_AZURE_AUTHN_INSTANCE == "true"){
+                  sleep(time: 1, unit: "SECONDS")
+                }
+              }
+            }
+          }
+        }
+        stage('Test Azure Authenticator'){
+          steps{
```
sigalsax: add space after `{`

ci/authn-azure/check_dependencies.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -eo pipefail
+
+check_env_var() {
+  if [[ -z "${!1+x}" ]]; then
+# where ${var+x} is a parameter expansion which evaluates to nothing if var is unset, and substitutes the string x otherwise.
```
sigalsax: indentation

ci/test
```diff
@@ -199,10 +199,10 @@ _get_azure_env_args() {
...
-  azure_env_args="$azure_env_args -e AZURE_AUTHN_INSTANCE_IP=$AZURE_MACHINE_IP"
-  azure_env_args="$azure_env_args -e AZURE_AUTHN_INSTANCE_USERNAME=$AZURE_MACHINE_USERNAME"
-  azure_env_args="$azure_env_args -e AZURE_AUTHN_INSTANCE_PASSWORD=$AZURE_MACHINE_PASSWORD"
+  azure_env_args="$azure_env_args -e AZURE_RESOURCE_GROUP=$AZURE_RESOURCE_GROUP"
+  azure_env_args="$azure_env_args -e AZURE_AUTHN_INSTANCE_IP=$AZURE_AUTHN_INSTANCE_IP"
+  azure_env_args="$azure_env_args -e AZURE_AUTHN_INSTANCE_USERNAME=$AZURE_AUTHN_INSTANCE_USERNAME"
+  azure_env_args="$azure_env_args -e AZURE_AUTHN_INSTANCE_PASSWORD=$AZURE_AUTHN_INSTANCE_PASSWORD"
   azure_env_args="$azure_env_args -e USER_ASSIGNED_IDENTITY=$USER_ASSIGNED_IDENTITY"
 
 # TODO: add this once available
```
sigalsax: We want to wait until you add systen-assigned-identity? b/c you have some TODOs around this

orenbm: No. The PR is big as it is. Let's merge this and add the rest of the tests later.

cucumber/authenticators_azure/features/authn_status_azure.feature
```diff
@@ -0,0 +1,161 @@
...
+        annotations:
+          description: Authentication service for prod, based on Open ID Connect.
+
+      - !webservice
+        id: status
+        annotations:
+          description: Status service to verify the authenticator is configured correctly
+
+      - !variable
+        id: id-token-user-property
```
sigalsax: why do we need this for authn-azure?

sigalsax: shouldnt this be `provider-uri`?

sigalsax: to mock some dummy variable that is not the `provider uri`? If so, I would just not add any value b/c that is confusing because we also have OIDC that actually uses this value

cucumber/authenticators_azure/features/authn_azure.feature
```diff
@@ -0,0 +1,49 @@
...
+        resource: !webservice
+    """
+    And I am the super-user
+    And I successfully set Azure variables
+    And I have host "no-assigned-identity-app"
+    And I set Azure annotations to host "no-assigned-identity-app"
+    And I grant group "conjur/authn-azure/prod/apps" to host "no-assigned-identity-app"
+
+  Scenario: A valid Azure token to get Conjur access token
+    # We want to verify the returned access token is valid for retrieving a secret
```
sigalsax: Instead of adding a comment here, I suggest you move it to the feature
`Feature: Azure Authenticator - Users can authenticate with Azure authenticator` ->
`Feature: Azure Authenticator - Users can authenticate with Azure authenticator and fetch secret`

cucumber/authenticators_azure/features/authn_azure.feature
```diff
@@ -0,0 +1,49 @@
...
+    And I permit host "no-assigned-identity-app" to "execute" it
+    And I add the secret value "test-secret" to the resource "cucumber:variable:test-variable"
+
+    And I fetch an Azure access token inside machine
+    When I authenticate via Azure with token as host "no-assigned-identity-app"
+    Then host "no-assigned-identity-app" is authorized
+    And I successfully GET "/secrets/cucumber/variable/test-variable" with authorized user
+
+  Scenario: A valid provider-uri without trailing slash works
+    # We want to verify the returned access token is valid for retrieving a secret
```
sigalsax: remove this comment and I suggest to add to Feature title

cucumber/authenticators_azure/features/authn_azure.feature
```diff
@@ -0,0 +1,49 @@
...
+    And I set Azure annotations to host "no-assigned-identity-app"
+    And I grant group "conjur/authn-azure/prod/apps" to host "no-assigned-identity-app"
+
+  Scenario: A valid Azure token to get Conjur access token
+    # We want to verify the returned access token is valid for retrieving a secret
+    Given I have a "variable" resource called "test-variable"
+    And I permit host "no-assigned-identity-app" to "execute" it
+    And I add the secret value "test-secret" to the resource "cucumber:variable:test-variable"
+
+    And I fetch an Azure access token inside machine
```
sigalsax: "from inside machine"

sigalsax: or "while inside machine"

cucumber/authenticators_azure/features/authn_azure.feature
```diff
@@ -0,0 +1,49 @@
...
+    When I authenticate via Azure with token as host "no-assigned-identity-app"
+    Then host "no-assigned-identity-app" is authorized
```
sigalsax: authorized to do what? authorized to authenticate?

sigalsax: "has been authorized by Conjur" or just "has been authorized". By this message, it isn't clear that this means it has been given an access token by Conjur

cucumber/authenticators_azure/features/authn_azure.feature
```diff
@@ -0,0 +1,49 @@
...
+    And I successfully GET "/secrets/cucumber/variable/test-variable" with authorized user
+
+  Scenario: A valid provider-uri without trailing slash works
+    # We want to verify the returned access token is valid for retrieving a secret
+    Given I have a "variable" resource called "test-variable"
+    And I permit host "no-assigned-identity-app" to "execute" it
+    And I add the secret value "test-secret" to the resource "cucumber:variable:test-variable"
+
+    And I fetch an Azure access token inside machine
```
sigalsax: "from inside" or "while inside"

cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb
```diff
@@ -0,0 +1,31 @@
+Given(/^I successfully set Azure variables$/) do
+  set_azure_provider_uri_variable
```
sigalsax: so if I am understanding here, you are either using the default value for `set_azure_provider_uri_variable` which is `azure_provider_uri` or setting it. If you use `azure_provider_uri` then you are using the `AZURE_TENANT_ID`. What did we gain from this? The ability to take from the env or be dynamic?

cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb
```diff
@@ -0,0 +1,31 @@
...
+
+Given(/^I successfully set Azure provider-uri variable to value "([^"]*)"$/) do |provider_uri|
+  set_azure_provider_uri_variable(provider_uri)
+end
+
+Given(/^I successfully set Azure provider-uri variable without trailing slash$/) do
+  set_azure_provider_uri_variable(azure_provider_uri.chop)
+end
+
+Given(/I fetch an Azure access token inside machine/) do
```
sigalsax: this is only for a system-assigned-id, otherwise for user-assigned-id, you need the client id

sigalsax: I would create two separate functions 
```
retrieve_azure_access_token_user_assigned_id (client_id)
       command = "curl 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&client_id=?resource=https%3A%2F%2Fmanagement.azure.com%2F' -H Metadata:true -s | jq -r '.access_token'"

retrieve_azure_access_token_system_assigned_id
```

cucumber/authenticators_azure/features/support/authn_azure_helper.rb
```diff
@@ -0,0 +1,76 @@
...
+  def authenticate_azure_token(service_id:, account:, username:, azure_token: parsed_azure_token)
+    username = username.gsub("/", "%2F")
+    path = "#{conjur_hostname}/authn-azure/#{service_id}/#{account}/#{username}/authenticate"
+
+    payload = {}
+    unless azure_token.nil?
+      payload["jwt"] = azure_token
+    end
+
+    post(path, payload)
```
sigalsax: this triggers a POST to conjur with path `/authenticate` the jwt in the body (payload)? This mocks the actual request?

cucumber/authenticators_azure/features/support/authn_azure_helper.rb
```diff
@@ -0,0 +1,76 @@
...
+  # TODO: add this once available
+  #def system_assigned_identity
+  #  @system_assigned_identity ||= validated_env_var('SYSTEM_ASSIGNED_IDENTITY')
+  #end
+
+  def user_assigned_identity
+    @user_assigned_identity ||= validated_env_var('USER_ASSIGNED_IDENTITY')
+  end
+
+  def parsed_azure_token
```
sigalsax: From what I can tell, what is returned here isn't a parsed_azure_token, it is just the azure token as a string.

dev/cli
```diff
@@ -46,6 +47,24 @@ function enable_oidc() {
   env_args="$env_args $keycloak_env_args"
 }
 
+function enable_azure() {
+
+  . ../ci/authn-azure/check_dependencies.sh
+
+  echo "Setting Azure details as env variables"
```
sigalsax: Let's keep the echos uniform across all  enable_authenticators

For oidc it is `echo "Extracting keycloak variables & setting as env variables"`

For azure it is `echo "Setting Azure details as env variables"`

dev/cli
```diff
@@ -46,6 +47,24 @@ function enable_oidc() {
...
+  azure_env_args="$azure_env_args -e AZURE_AUTHN_INSTANCE_USERNAME=$AZURE_AUTHN_INSTANCE_USERNAME"
+  azure_env_args="$azure_env_args -e AZURE_AUTHN_INSTANCE_PASSWORD=$AZURE_AUTHN_INSTANCE_PASSWORD"
+  azure_env_args="$azure_env_args -e AZURE_SUBSCRIPTION_ID=$AZURE_SUBSCRIPTION_ID"
+  azure_env_args="$azure_env_args -e AZURE_RESOURCE_GROUP=$AZURE_RESOURCE_GROUP"
+  azure_env_args="$azure_env_args -e USER_ASSIGNED_IDENTITY=$USER_ASSIGNED_IDENTITY"
+#  azure_env_args="$azure_env_args -e SYSTEM_ASSIGNED_IDENTITY=$SYSTEM_ASSIGNED_IDENTITY" # TODO: add this once available
+  env_args="$env_args $azure_env_args"
+}
+
 unset COMPOSE_PROJECT_NAME
```
sigalsax: curious what this does?

dev/start
```diff
@@ -129,6 +133,21 @@ if [[ $ENABLE_AUTHN_LDAP = true ]]; then
   docker-compose exec conjur conjurctl policy load cucumber /src/conjur-server/dev/files/authn-ldap/policy.yml
 fi
 
+if [[ $ENABLE_AUTHN_AZURE = true ]]; then
+  enabled_authenticators="$enabled_authenticators,authn-azure/prod"
+
+  . ../ci/authn-azure/check_dependencies.sh
```
sigalsax: We are checking this twice; once here and once in cli. Do we need these both?

cucumber/authenticators_azure/features/authn_status_azure.feature
orenbm: nice catch. missed it while copying this policy from OIDC.

cucumber/authenticators_azure/features/authn_azure.feature
orenbm: Renamed the scenario to `Users can authenticate with Azure authenticator and fetch secret`

ci/authn-azure/check_dependencies.sh
```diff
@@ -0,0 +1,23 @@
...
+check_env_var "AZURE_TENANT_ID"
+check_env_var "AZURE_SUBSCRIPTION_ID"
+check_env_var "AZURE_RESOURCE_GROUP"
+check_env_var "AZURE_AUTHN_INSTANCE_USERNAME"
+check_env_var "AZURE_AUTHN_INSTANCE_PASSWORD"
+check_env_var "USER_ASSIGNED_IDENTITY"
+
+# These variables should come from Jenkins
+check_env_var "AZURE_AUTHN_INSTANCE_IP"
+#check_env_var "SYSTEM_ASSIGNED_IDENTITY" # TODO: add this once available
```
sigalsax: why is `SYSTEM_ASSIGNED_IDENTITY` coming from Jenkins and not also summon?
Also do we want to set both of these in env at the same time due to incompatibility/invalid combination?

cucumber/authenticators_azure/features/authn_azure.feature
orenbm: went with "from"

cucumber/authenticators_azure/features/authn_azure.feature
orenbm: authorized by Conjur - got a 200 ok response. This is not a new step so if you think it should be changed, let's do it in a separate PR to master

Jenkinsfile
```diff
@@ -69,6 +69,39 @@ pipeline {
         stage('OIDC Authenticator') {
           steps { sh 'ci/test cucumber_authenticators_oidc' }
         }
+        stage('Allocate Azure Authenticator Instance'){
+          steps {
+            script {
+              node('azure-authn'){
```
sigalsax: not 💯 on what this is doing- is this setting up the environment?

cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb
orenbm: In future tests we will give custom inputs for the provider uri, such that will end with 502 & 504 response codes.

Jenkinsfile
sigalsax: this just runs separate code in the context of a workspace?

cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb
orenbm: I started with the user assigned id which is the basic one and plan to get the system assigned id code once i test it. I don't want to add helpers that are not used in this PR to make it less noisy.

cucumber/authenticators_azure/features/support/authn_azure_helper.rb
orenbm: 1. We don't mock anything in integration tests.
2. Yes, it does exactly what you said in the first sentence.

cucumber/authenticators_azure/features/support/authn_azure_helper.rb
orenbm: renamed to `def azure_token`

dev/cli
orenbm: not sure tbh

dev/start
orenbm: yes because these are 2 scripts that need the variables.

ci/authn-azure/check_dependencies.sh
orenbm: > why is SYSTEM_ASSIGNED_IDENTITY coming from Jenkins and not also summon?

Because it is created in each build as each pipeline has its own Azure VM. Same as with the IP.

> do we want to set both of these in env at the same time due to incompatibility/invalid combination?

We will have tests for both use-cases in the same pipeline, then they are both needed.

Jenkinsfile
orenbm: I added a comment on this in the commit. Still confused?

dev/cli
orenbm: Google it and tell me? :)

hughsaunders: `COMPOSE_PROJECT_NAME` is the prefix used to name containers when created via docker-compose

Jenkinsfile
```diff
@@ -69,6 +69,39 @@ pipeline {
         stage('OIDC Authenticator') {
           steps { sh 'ci/test cucumber_authenticators_oidc' }
         }
+        stage('Azure Authenticator preparation - Allocate Azure Authenticator Instance') {
```
hughsaunders: This pattern is unfamiliar, so probably worth adding a comment about why the two stages are necessary.

dev/cli
orenbm: Oh right, it sat somewhere in the back of my head. thanks!

Jenkinsfile
orenbm: Thanks, added a comment

cucumber/authenticators_azure/features/authn_azure.feature
sigalsax: I think it is unclear. Only I looked at the code, it became clear. IMO worth the refactor

cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb
sigalsax: But that's the thing it shouldn't work if you use user-assigned-id b/c your endpoint is:
```
http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fmanagement.azure.com%2F
```
So you are missing the client-id in your request. For user-assigned-identity, it must have the client-id. Like so:
```
http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&client_id=?resource=https%3A%2F%2Fmanagement.azure.com%2F
```

Jenkinsfile
sigalsax: In the comment `one` should have an uppercase O (:

sigalsax: What do you mean when you say "Allocating an Azure VM"? As in getting the VM up and configuring it to be ready for running tests against?

Jenkinsfile
sigalsax: I would mention why this is a special case. Like why don't we have this for OIDC for example? That way it is more clear that this is pattern circumstantial

cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb
orenbm: it is not a user-assigned identity token. It is a system-assigend one, only i am not using the identity itself (the oid). Once we will add the user-assigned id test I will add another method for that.

Jenkinsfile
orenbm: without the word "again". it is a new machine.

Jenkinsfile
orenbm:
```
// We have 2 stages for the Azure Authenticator tests. One is
// responsible for allocating an Azure VM, from which we will get an
// Azure access token for Conjur authentication. It sets the Azure VM IP
// in the env and waits until the authn-azure tests are finished. We use
// this mechanism as we need a live Azure VM for getting the Azure token
// and we don't want to have a long running-machine deployed in Azure.
```

@sigalsax what do you think? I believe the last sentence is redundant. I can't say why we don't have it for OIDC because we just don't need it there.

cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb
sigalsax: So let's not use "Users" in the Scenario for now?
```
  Scenario: Users can authenticate with Azure authenticator and fetch secret
```
Maybe, Machines/Azure instances

orenbm: I will change to hosts but it will always be host. We will not authenticate users, regardless of the token type.

cucumber/authenticators_azure/features/authn_azure.feature
orenbm: Created PR - https://github.com/cyberark/conjur/pull/1425

PR #1415, opened 2020-03-12
Update Authenticator Test Suite for Rails 5 Upgrade

micahlee:
*NOTE:* This PR merges into the `rails-5-again` branch, not `master`.

This PR updates the authenticator test suite to make the necessary fixes for the Rails 5 upgrade.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,9 +49,9 @@ pipeline {
         stage('RSpec') {
           steps { sh 'ci/test rspec' }
         }
-        // stage('Authenticators') {
-        //   steps { sh 'ci/test cucumber_authenticators' }
-        // }
+        stage('Authenticators') {
+          steps { sh 'ci/test cucumber_authenticators' }
+        }
         stage('Policy') {
           steps { sh 'ci/test cucumber_policy' }
         }
--- a/cucumber/authenticators/features/authn_status.feature
+++ b/cucumber/authenticators/features/authn_status.feature
@@ -107,13 +107,21 @@ Feature: Authenticator status check
     And the authenticator status check fails with error "UserNotAuthorizedInConjur: CONJ00006E"
 
   Scenario: An authenticator without an implemented status check returns 503
-    Given I login as "alice"
+    Given a policy:
+    """
+    - !user alice
+    """
+    And I login as "alice"
     When I GET "/authn-ldap/test/cucumber/status"
     Then the HTTP response status code is 501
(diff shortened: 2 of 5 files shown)


micahlee:
Yeah, that seems to also be the case for the `rails-5-again` branch as well: https://jenkins.conjur.net/job/cyberark--conjur/job/rails-5-again/

The particular error is because the GKE coverage results are missing.

Jenkinsfile
```diff
@@ -49,9 +49,9 @@ pipeline {
         stage('RSpec') {
           steps { sh 'ci/test rspec' }
         }
-        // stage('Authenticators') {
-        //   steps { sh 'ci/test cucumber_authenticators' }
-        // }
+        stage('Authenticators') {
+          steps { sh 'ci/test cucumber_authenticators' }
+        }
```
orenbm: @micahlee please note that this has changed so you'll have some conflicts. See latest jenkinsfile -https://github.com/cyberark/conjur/blob/master/Jenkinsfile#L60

cucumber/authenticators/features/authn_status.feature
```diff
@@ -107,13 +107,21 @@ Feature: Authenticator status check
     And the authenticator status check fails with error "UserNotAuthorizedInConjur: CONJ00006E"
 
   Scenario: An authenticator without an implemented status check returns 503
-    Given I login as "alice"
+    Given a policy:
+    """
+    - !user alice
+    """
+    And I login as "alice"
```
orenbm: @micahlee can you explain why this is part of the PR? is this change related to the Rails 5 upgrade?

Jenkinsfile
micahlee: Yeah, we are aware that were going to have some heavy rebasing to do. A lot has been going on in the testing framework lately! 🙂

cucumber/authenticators/features/authn_status.feature
micahlee: Yes. When we started debugging test failures we discovered that we had quite a few scenarios that assumed a deterministic order, and that there was database state from prior runs.

We have started correcting this by doing more to ensure that each scenario "starts from scratch" and defines the entire scenario context itself.

These two scenarios are examples that assumed the user `alice` was already created in the database. So they now just need to specify that explicitly in the scenario by defining the user in policy.

orenbm: Nice, thanks!

When you create a PR to master please write this in a commit so it's clear what the change is . thanks for this!

PR #1417, opened 2020-03-12
Move development instructions out of README

izgeri:
Moves development instructions out of README
Also update release instructions in CONTRIBUTING

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,18 +1,340 @@
 # Contributing to Conjur
 
-Thanks for your interest in Conjur. Before contributing, please take a moment to read and sign our <a href="https://github.com/cyberark/conjur/blob/master/Contributing_OSS/CyberArk_Open_Source_Contributor_Agreement.pdf" download="conjur_contributor_agreement">Contributor Agreement</a>. This provides patent protection for all Conjur users and allows CyberArk to enforce its license terms. Please email a signed copy to <a href="oss@cyberark.com">oss@cyberark.com</a>.
+Thanks for your interest in Conjur. Before contributing, please take a moment to
+read and sign our <a href="https://github.com/cyberark/conjur/blob/master/Contributing_OSS/CyberArk_Open_Source_Contributor_Agreement.pdf" download="conjur_contributor_agreement">Contributor Agreement</a>.
+This provides patent protection for all Conjur users and allows CyberArk to enforce
+its license terms. Please email a signed copy to <a href="oss@cyberark.com">oss@cyberark.com</a>.
 
 For general contribution and community guidelines, please see the [community repo](https://github.com/cyberark/community).
 
+- [Prerequisites](#prerequisites)
+- [Build Conjur as a Docker image](#build-conjur-as-a-docker-image)
+- [Set up a development environment](#set-up-a-development-environment)
+    + [LDAP Authentication](#ldap-authentication)
+    + [RubyMine IDE Debugging](#rubymine-ide-debugging)
+    + [Visual Studio Code IDE Debugging](#visual-studio-code-ide-debugging)
+  * [Development CLI](#development-cli)
+    + [Step into the running Conjur container](#step-into-the-running-conjur-container)
+    + [View the admin user's API key](#view-the-admin-user-s-api-key)
+    + [Load a policy](#load-a-policy)
+- [Testing](#testing)
+  * [RSpec](#rspec)
+  * [Cucumber](#cucumber)
+    + [Run all the cukes:](#run-all-the-cukes-)
+    + [Run just one feature:](#run-just-one-feature-)
+- [Pull Request Workflow](#pull-request-workflow)
+- [Style guide](#style-guide)
+- [Changelog maintenance](#changelog-maintenance)
+- [Releasing](#releasing)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,18 +1,312 @@
...
+[get-git]: https://git-scm.com/downloads
+[get-docker-compose]: https://docs.docker.com/compose/install
+
+## Build Conjur as a Docker image
+
+It's easy to get started with Conjur and Docker:
+
+1. install dependencies (as above)
+1. clone this repository
+1. run the build script in your terminal:
```
sgnn7: These should all start capitalized

README.md
```diff
@@ -428,61 +150,19 @@ Permission to use this service is controlled by the built-in resource
...
-- Create a pull request to have the release acked and merged.
-  - https://github.com/cyberark/conjur/pull/new/vx.y.z
+We welcome contributions of all kinds to Conjur. See our [contributing
+guide][contrib].
 
-Deleting and changing tags should be avoided. If in any doubt if the release will
-be accepted, before creating a tag push the (VERSION and CHANGELOG) changes in
-a branch and ask for approval. Then create and push a tag on `master` once it's
-been merged.
+[contrib]: https://github.com/cyberark/conjur/blob/master/CONTRIBUTING.md
```
sgnn7: This feels superfluous as it's only used once and that usage is 2 lines above it

README.md
```diff
@@ -42,323 +59,28 @@ Migrating data from Conjur Open Source to Conjur EE is simple using our
 
 [migration]: design/MIGRATION.md
 
-# Development
```
sgnn7: We should probably leave this section in and link it to contributing header. Not everyone that wants to work on the code wants to do it to contribute the code back.

CONTRIBUTING.md
```diff
@@ -21,3 +315,72 @@ Use [this guide][style] to maintain consistent style across the Conjur project.
...
+   has changed since the last tag - make sure the NOTICES file is up-to-date with
+   the new versions
+
+### Update the version and changelog
+1. Examine the changelog and decide on the version bump rank (major, minor, patch).
+1. Change the title of _Unreleased_ section of the changelog to the target
+version.
+  - Be sure to add the date (ISO 8601 format) to the section header.
+1. Add a new, empty _Unreleased_ section to the changelog.
+  - Remember to update the references at the bottom of the document.
```
jvanderhoof: Adding another two spaces to these bullets will set them under the numbered list item when rendered. They are currently showing up as indented with the numbered list items.

CONTRIBUTING.md
```diff
@@ -21,3 +315,72 @@ Use [this guide][style] to maintain consistent style across the Conjur project.
...
+  - git will ask you to enter the tag message. These will become the release notes.
+  Format should be like this (note the subject line and message):
+
+        Version x.y.z
+
+        This is a human-readable overview of the changes in x.y.z. It should be a
+        consise, at-a-glance summary. It certainly isn't a direct copy-and-paste
+        from the changelog.
+
+1. Push the tag: `git push vx.y.z` (or `git push origin vx.y.z` if you are working
```
jvanderhoof: What do you think about using `git push v1.2.3` or something like that.  putting the `v` (which follows our convention) next to the `x` reads a bit weird to me.

README.md
izgeri: the CONTRIBUTING file is still linked to, it's just lower and in a section called `Contributing` consistent with the rest of our repos

take a look there and see if it could be clearer about telling people where to look for development instructions, I'm happy to edit it if you think it could be better.

README.md
izgeri: you're not wrong - but I copied it from what we're putting in other project READMEs, and it's not really hurting anything by being there so 🤷‍♀

CONTRIBUTING.md
izgeri: fixed - even though I didn't write this, you're right, and might as well fix it now

README.md
sgnn7: I think from my perspective, as long as there's some indication that there's "development" info in the contributing doc it's fine with me.

CONTRIBUTING.md
izgeri: yeah, just noticed there was a v0.1.1 in the example above it too - updated to make them all consistent

README.md
izgeri: updated the comment so that it indicates it includes "general development instructions"

README.md
```diff
@@ -30,6 +30,23 @@ Conjur provides secrets management and application identity for modern infrastru
...
+  * [Database](#database)
+    + [DATABASE_URL environment variable](#database-url-environment-variable)
+    + [Database initialization](#database-initialization)
+  * [Authenticators](#authenticators)
+  * [Rotators](#rotators)
+  * [Secrets and keys](#secrets-and-keys)
+    + [Important: avoid data loss](#important--avoid-data-loss)
+  * [Account management](#account-management)
+- [Versioning](#versioning)
+- [License](#license)
```
sgnn7: Contributing link is not in this ToC. I think there may be other missing headers too.

izgeri: good catch!! fixed

PR #1423, opened 2020-03-13
Remove low-impact possum references

jtuttle:
This PR removes all of the low-impact references to the old "possum" project name. The remaining references will be more difficult to replace and should be handled as a separate effort, if at all.


app/models/schemata.rb
```diff
@@ -8,10 +8,10 @@
 # doesn't exist, it is ignored. As a result, SQL and DDL operations normally apply 
 # to the "public" schema unless a schema is indicated specifically (e.g. "myschema.mytable").
 #
-# Possum can be configure with an alternative search path, such as "possum, public". In this case,
-# DDL and SQL will be applied to the primary schema (in this example, "possum").
+# Conjur can be configure with an alternative search path, such as "conjur, public". In this case,
```
micahlee:
```suggestion
# Conjur can be configured with an alternative search path, such as "conjur, public". In this case,
```

db/migrate/20170710163523_create_resources_textsearch.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # This migration has been ported over from Conjur v3's /authz project. I had to 
-# make a few changes for possum that have been documented with comments.
+# make a few changes for Conjur v5 that have been documented with comments.
```
micahlee: WDYT about calling this Conjur OSS instead? It's confusing here because the Conjur version here is currently `1.4.7`, so it's not obvious where 5 comes from.

jtuttle: Yeah sounds good. You're right v5 is not really meaningful these days.

PR #1426, opened 2020-03-15
Hosts can authenticate with Azure Authenticator

orenbm:
#### What does this PR do?
We would like to let hosts authenticate from Azure VMs using an Azure access token. This PR adds such and authenticator. The authenticator verifies the Azure access token against the Azure Identity Provider, and if the token is valid it also validates the Conjur host has the correct application identity stored in its
annotations, by comparing it to the data stored in the Azure access token.

We also added a Status Check for this authenticator and of course UTs and integration tests.

More integration tests will be added to this PR but other than that it is ready for review.

### DIFF
--- a/app/domain/authentication/authn_azure/application_identity.rb
+++ b/app/domain/authentication/authn_azure/application_identity.rb
@@ -1,0 +1,41 @@
+module Authentication
+  module AuthnAzure
+
+    Log = LogMessages::Authentication::AuthnAzure
+
+    class ApplicationIdentity
+
+      def initialize(role_annotations:, service_id:)
+        @role_annotations = role_annotations
+        @service_id       = service_id
+      end
+
+      def constraints
+        @constraints ||= {
+          subscription_id:          constraint_value("subscription-id"),
+          resource_group:           constraint_value("resource-group"),
+          user_assigned_identity:   constraint_value("user-assigned-identity"),
+          system_assigned_identity: constraint_value("system-assigned-identity")
+        }.compact
+      end
+
+      private
+
+      # check the `service-id` specific constraint first to be more granular
+      def constraint_value constraint_name
+        annotation_value("authn-azure/#{@service_id}/#{constraint_name}") ||
+          annotation_value("authn-azure/#{constraint_name}")
+      end
+
+      def annotation_value name
+        annotation = @role_annotations.find { |a| a.values[:name] == name }
+
+        # return the value of the annotation if it exists, nil otherwise
+        if annotation
+          Rails.logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
+          annotation[:value]
+        end
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_azure/application_identity.rb
+++ b/app/domain/authentication/authn_azure/application_identity.rb
@@ -5,9 +5,10 @@ module Authentication
 
     class ApplicationIdentity
 
-      def initialize(role_annotations:, service_id:)
+      def initialize(role_annotations:, service_id:, logger:)
         @role_annotations = role_annotations
         @service_id       = service_id
+        @logger           = logger
       end
 
       def constraints
@@ -32,7 +33,7 @@ module Authentication
 
         # return the value of the annotation if it exists, nil otherwise
         if annotation
-          Rails.logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
+          @logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
           annotation[:value]
         end
       end
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -21,12 +21,8 @@ module Authentication
       extend Forwardable
       def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :credentials, :username
 
-      XMS_MIRID_TOKEN_FIELD_NAME = "xms_mirid".freeze
-      OID_TOKEN_FIELD_NAME       = "oid".freeze
-
       def call
         validate_azure_token
-        validate_required_token_fields_exist
         validate_application_identity
       end
 
@@ -37,13 +33,16 @@ module Authentication
       end
 
       def decoded_token
-        @decoded_token ||= @verify_and_decode_token.call(
-          provider_uri:     provider_uri,
-          token_jwt:        decoded_credentials.jwt,
-          claims_to_verify: {
-            verify_iss: true,
-            iss:        provider_uri
-          }
+        @decoded_token ||= DecodedToken.new(
+          decoded_token_hash: @verify_and_decode_token.call(
+            provider_uri:     provider_uri,
+            token_jwt:        decoded_credentials.jwt,
+            claims_to_verify: {
+              verify_iss: true,
+              iss:        provider_uri
+            }
+          ),
+          logger:             @logger
         )
       end
 
@@ -56,8 +55,8 @@ module Authentication
           service_id: service_id,
           account: account,
           username: username,
-          xms_mirid_token_field: xms_mirid,
-          oid_token_field: oid
+          xms_mirid_token_field: decoded_token.xms_mirid,
+          oid_token_field: decoded_token.oid
         )
       end
 
@@ -78,30 +77,6 @@ module Authentication
 
       def required_variable_names
         @required_variable_names ||= %w(provider-uri)
-      end
-
-      def validate_required_token_fields_exist
-        validate_token_field_exists(XMS_MIRID_TOKEN_FIELD_NAME)
-        validate_token_field_exists(OID_TOKEN_FIELD_NAME)
-      end
-
-      def validate_token_field_exists field_name
-        @logger.debug(Log::ValidatingTokenFieldExists.new(field_name))
-        raise Err::TokenFieldNotFoundOrEmpty, field_name if decoded_token[field_name].to_s.empty?
-      end
-
-      def xms_mirid
-        token_field_value(XMS_MIRID_TOKEN_FIELD_NAME)
-      end
-
-      def oid
-        token_field_value(OID_TOKEN_FIELD_NAME)
-      end
-
-      def token_field_value field_name
-        decoded_token[field_name].tap do |token_field_value|
-          @logger.debug(Log::ExtractedFieldFromAzureToken.new(field_name, token_field_value))
-        end
       end
     end
 
--- a/app/domain/authentication/authn_azure/authenticator.rb
+++ b/app/domain/authentication/authn_azure/authenticator.rb
@@ -21,30 +21,16 @@ module Authentication
       extend Forwardable
       def_delegators :@authenticator_input, :service_id, :authenticator_name, :account, :credentials, :username
 
-      JWT_REQUEST_BODY_FIELD_NAME = "jwt".freeze
-      XMS_MIRID_TOKEN_FIELD_NAME  = "xms_mirid".freeze
-      OID_TOKEN_FIELD_NAME        = "oid".freeze
+      XMS_MIRID_TOKEN_FIELD_NAME = "xms_mirid".freeze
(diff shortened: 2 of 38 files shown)


cucumber/authenticators_azure/features/authn_status_azure.feature
```diff
@@ -0,0 +1,151 @@
...
+
+    - !grant
+      role: !group conjur/authn-azure/prod/operators
+      member: !user alice
+    """
+    And I am the super-user
+    And I login as "alice"
+    When I GET "/authn-azure/prod/cucumber/status"
+    Then the HTTP response status code is 500
+    And the authenticator status check fails with error "RequiredResourceMissing: CONJ00036E"
```
jvanderhoof: Where are these error codes documented?

cucumber/authenticators_azure/features/support/authn_azure_helper.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    payload = {}
+    payload["jwt"] = azure_token
+
+    post(path, payload)
+  end
+
+  private
+
+  def set_azure_variable(variable_name, value)
```
jvanderhoof: What do you think about renaming this to `create_and_set_azure_variable`.  Using `set` makes me think of setters, which this method is not 😄.

cucumber/authenticators_azure/features/authn_status_azure.feature
orenbm: in error.rb

jvanderhoof: Is there a ticket to get these into our docs?  They would really help troubleshooting.

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      private
+
+      # check the `service-id` specific constraint first to be more granular
+      def constraint_value constraint_name
+        annotation_value("authn-azure/#{@service_id}/#{constraint_name}") ||
+          annotation_value("authn-azure/#{constraint_name}")
+      end
+
+      def annotation_value name
```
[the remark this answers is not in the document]
jonahx: Use parens: ruby style guide.  This should be getting flagged by CC.

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,41 @@
...
+        annotation = @role_annotations.find { |a| a.values[:name] == name }
+
+        # return the value of the annotation if it exists, nil otherwise
+        if annotation
+          Rails.logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
```
jonahx: This should be passed into the initializer

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -0,0 +1,41 @@
...
+          system_assigned_identity: constraint_value("system-assigned-identity")
+        }.compact
+      end
+
+      private
+
+      # check the `service-id` specific constraint first to be more granular
+      def constraint_value constraint_name
+        annotation_value("authn-azure/#{@service_id}/#{constraint_name}") ||
+          annotation_value("authn-azure/#{constraint_name}")
```
jonahx: Without looking at this more deeply, something strikes me as off about having two ways to store the annotation.  It might be ok if this logic only exists in this one place.  Otherwise, it should likely be moved to an object of its own.

app/domain/authentication/authn_azure/authenticator.rb
```diff
@@ -0,0 +1,139 @@
...
+
+      def status(authenticator_status_input:)
+        Authentication::AuthnAzure::ValidateStatus.new.(
+          account: authenticator_status_input.account,
+          service_id: authenticator_status_input.service_id
+        )
+      end
+    end
+  end
+end
```
jonahx: 👍 pleasure to read

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,185 @@
...
+      end
+
+      # we expect the xms_mirid claim to be in the format of /subscriptions/<subscription-id>/...
+      # therefore, we ignore the first slash of the xms_mirid claim and group the entries in key-value pairs
+      # according to fields we need to retrieve from the claim.
+      # ultimately, transforming "/key1/value1/key2/value2" to {"key1" => "value1", "key2" => "value2"}
+      def parsed_xms_mirid
+        split_xms_mirid = @xms_mirid_token_field.split('/')
+
+        if split_xms_mirid.first == ''
```
jonahx: Add comment explaining what this is for.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -0,0 +1,185 @@
...
+          case property_name
+          when "subscriptions"
+            xms_mirid_hash["subscriptions"] = split_xms_mirid[property_value_index]
+          when "resourcegroups"
+            xms_mirid_hash["resourcegroups"] = split_xms_mirid[property_value_index]
+          when "providers"
+            xms_mirid_hash["providers"] = split_xms_mirid[property_value_index, 3]
+          end
+          index += 1
+        end
```
jonahx: This section here could be made much more idiomatic.  if i'm understanding the code the `split_xms_mirid` is an array that looks something like this:
```ruby
arr = ['subscriptions', 'foo', 'resourcegroups', 'bar', 'providers', 'one', 'two', 'three']
```
with the hash keys and their values in adjacent elements, with providers containing 3 elements after it.    if so, here's a more functional alternative to the above logic:

```ruby
# Slice points:
#        v                      v                        v
# arr = ["subscriptions", "foo", "resourcegroups", "bar", "providers", "one", "two", "three"]

arr = ["subscriptions", "foo", "resourcegroups", "bar", "providers", "one", "two", "three"]

xms_hash = arr
  .slice_before(/subscriptions|resourcegroups|providers/)
  .each_with_object({}) do |part, hash|
    key, val = part.first, part.drop(1)
    val = val.first if val.length == 1 
    hash[key] = val
end

p xms_hash
#=> {"subscriptions"=>"foo", "resourcegroups"=>"bar", "providers"=>["one", "two", "three"]}
```
Additionally, since there's enough logic here and with the validations you do, I'd strongly consider making `XmsMirid` into  its own value object.  You could keep it in this same file if you want, but it would take a string in its constructor, do all the parsing and validation internally, and expose methods `subscriptions`, `resourcegroups`, and `providers`.

app/domain/authentication/authn_azure/application_identity.rb
orenbm: Thanks, let's tackle this in the application identity refactor

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: Beautiful!

cucumber/authenticators_azure/features/authn_status_azure.feature
orenbm: Nothing like that is planned as far as I know. @alexkalish @InbalZilberman can you prioritize this?

app/domain/authentication/authn_azure/xms_mirid.rb
```diff
@@ -0,0 +1,66 @@
...
+    Err = Errors::Authentication::AuthnAzure
+
+    class XmsMirid
+
+      REQUIRED_KEYS = %w(subscriptions resourcegroups providers).freeze
+
+      def initialize(xms_mirid_token_field)
+        @xms_mirid_token_field = xms_mirid_token_field
+
+        @mirid_parts_hash = mirid_parts_hash
```
sigalsax: was there a reason the xms prefixed was chopped off?

ci/authn-azure/check_dependencies.sh
```diff
@@ -0,0 +1,29 @@
...
+check_env_var "AZURE_SUBSCRIPTION_ID"
+check_env_var "AZURE_RESOURCE_GROUP"
+check_env_var "AZURE_AUTHN_INSTANCE_USERNAME"
+check_env_var "AZURE_AUTHN_INSTANCE_PASSWORD"
+check_env_var "USER_ASSIGNED_IDENTITY"
+check_env_var "USER_ASSIGNED_IDENTITY_CLIENT_ID"
+
+# These variables should come from Jenkins
+check_env_var "AZURE_AUTHN_INSTANCE_IP"
+#check_env_var "SYSTEM_ASSIGNED_IDENTITY" # TODO: add this once available
```
sigalsax: system assigned identity will be tackled in a separate PR?

orenbm: Yes, once we can do it

app/domain/authentication/authn_azure/xms_mirid.rb
orenbm: Please see @jonahx 's comment: https://github.com/cyberark/conjur/pull/1446#discussion_r396012256

wdyt?

cucumber/authenticators_azure/features/authn_status_azure.feature
```diff
@@ -0,0 +1,151 @@
...
+
+  # TODO: add this tes when issue #1085 is done
```
sigalsax: "test"

cucumber/authenticators_azure/features/support/authn_azure_helper.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def retrieve_system_assigned_azure_access_token
+    retrieve_azure_access_token(retrieve_azure_token_command(is_user_assigned: false))
+  end
+
+  def retrieve_azure_token_command is_user_assigned:
+    base_command = "curl 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01" \
+      "&resource=https%3A%2F%2Fmanagement.azure.com%2F"
+
+    base_command = "#{base_command}&client_id=#{user_assigned_identity_client_id}" if is_user_assigned
```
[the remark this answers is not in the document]
sigalsax: really nice

cucumber/authenticators_azure/features/support/authn_azure_helper.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def retrieve_user_assigned_azure_access_token
+    retrieve_azure_access_token(retrieve_azure_token_command(is_user_assigned: true))
+  end
+
+  def retrieve_system_assigned_azure_access_token
+    retrieve_azure_access_token(retrieve_azure_token_command(is_user_assigned: false))
+  end
+
+  def retrieve_azure_token_command is_user_assigned:
```
sigalsax: couldn't you just send true or false? `is_user_assigned: true/false` for readability?

app/domain/authentication/authn_azure/xms_mirid.rb
```diff
@@ -0,0 +1,66 @@
+module Authentication
+  module AuthnAzure
+
+    Err = Errors::Authentication::AuthnAzure
+
+    class XmsMirid
```
sigalsax: I like how to put the xms_mirid in its own class. A great idea.

cucumber/authenticators_azure/features/authn_status_azure.feature
orenbm: will be removed with the TODO so i won't bother

cucumber/authenticators_azure/features/support/authn_azure_helper.rb
orenbm: i don't understand your question. can you write down your suggestion?

sigalsax:
```
calling function:
retrieve_azure_access_token(retrieve_azure_token_command (false))
...

function:
def retrieve_azure_token_command is_user_assigned
```

orenbm: ah yes, it is possible. but i want it to be more explicit so the reader doesn't wonder what `false` is.

PR #1432, opened 2020-03-17
Add UTs for Jwt::VerifyAndDecodeToken

orenbm:
Add UTs that validate that:
- If the decoder succeeds then it returns the decoded token
  as one object in a list
- That we raise Jwt::TokenDecodeFailed in case the decoding failed
- That we raise Jwt::TokenExpired if the token is expired
- That we raise Jwt::TokenVerificationFailed in case the decoded failed
  to verify the token.

As we don't have a way to simulate an invalid or expired token in an
integration test, this is our way to verify our behaviour.


orenbm:
> What about TokenFieldNotFoundOrEmpty ? it wasnt tested manually this is the only way to verify this.

This is already tested here: https://github.com/cyberark/conjur/blob/authn-azure/spec/app/domain/authentication/authn-azure/authenticator_spec.rb#L180

PR #1433, opened 2020-03-17
Add tests for the Azure Authenticator

orenbm:
#### What does this PR do?
Adds some tests for the Azure Authenticator, according to the [test plan](https://github.com/cyberark/conjur/blob/master/design/authn_azure/authn_azure_solution_design.md#test-plan)

The only test that still needs to be added is the System-Assigned Identity scenario. This test will be added in another PR as we don't have the infra for it yet.

We also have some tests from the test plan that will not be tested in OSS:
image


cucumber/authenticators_azure/features/authn_azure_basic_host.feature
```diff
@@ -0,0 +1,94 @@
...
+    And I fetch an Azure access token from inside machine
+    And I authenticate via Azure with token as host "test-app"
+    Then it is bad gateway
+    # Check recovery to a valid provider uri
+    When I successfully set Azure variables
+    And I fetch an Azure access token from inside machine
+    And I authenticate via Azure with token as host "test-app"
+    And host "test-app" has been authorized by Conjur
+
+  Scenario: Bad Gateway is raised in case of an invalid ID Provider hostname
```
eladkug: What is the different from previous test ? 99% the same

cucumber/authenticators_azure/features/authn_azure_hosts.feature
```diff
@@ -0,0 +1,138 @@
...
+  Scenario: Host with user-assigned-identity annotation is authorized
+    And I have host "user-assigned-identity-app"
+    And I set subscription-id annotation to host "user-assigned-identity-app"
+    And I set resource-group annotation to host "user-assigned-identity-app"
+    And I set user-assigned-identity annotation to host "user-assigned-identity-app"
+    And I grant group "conjur/authn-azure/prod/apps" to host "user-assigned-identity-app"
+    And I fetch a user-assigned Azure access token from inside machine
+    When I authenticate via Azure with token as host "user-assigned-identity-app"
+    Then host "user-assigned-identity-app" has been authorized by Conjur
+
```
eladkug: Why in vanilla flow you dont verify something from log. for example : `Application identity validated`

cucumber/authenticators_azure/features/authn_azure_hosts.feature
```diff
@@ -0,0 +1,138 @@
...
+    And I have host "no-subscription-id-app"
+    And I set resource-group annotation to host "no-subscription-id-app"
+    And I grant group "conjur/authn-azure/prod/apps" to host "no-subscription-id-app"
+    And I fetch an Azure access token from inside machine
+    And I save my place in the log file
+    When I authenticate via Azure with token as host "no-subscription-id-app"
+    Then it is unauthorized
+    And The following appears in the log after my savepoint:
+    """
+    Errors::Authentication::AuthnAzure::RoleMissingConstraint
```
eladkug: I think you need to add to log verification 
`Role does not have the required constraint: subscription_id
` because you are testing specific missing.

cucumber/authenticators_azure/features/authn_azure_performance.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Azure Authenticator performance
```
eladkug: same here with Azure Authenticator -

cucumber/authenticators_azure/features/authn_azure_performance.feature
```diff
@@ -0,0 +1,42 @@
...
+
```
eladkug: If you can describe the feature like you did before it will be great. It was very clear.

Jenkinsfile
```diff
@@ -92,11 +92,10 @@ pipeline {
           steps {
             script {
               while (!env.AZURE_AUTHN_INSTANCE_IP?.trim()) {
-                sleep(time: 1, unit: "SECONDS")
+                sleep(time: 15, unit: "SECONDS")
```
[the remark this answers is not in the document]
orenbm: they have different purposes. the `Azure Authenticator preparation - Allocate Azure Authenticator Instance` stage sets `AZURE_AUTHN_INSTANCE_IP` which is required for the tests that run in `Azure Authenticator` so the loop in `Azure Authenticator` is waiting for `AZURE_AUTHN_INSTANCE_IP` to be set before it starts.

Then `Azure Authenticator preparation - Allocate Azure Authenticator Instance` waits until `Azure Authenticator` finishes its work so the Azure machine stays alive.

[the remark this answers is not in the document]
orenbm: Note that they are looping on different variables

cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
```diff
@@ -0,0 +1,120 @@
+Feature: Hosts can authenticate with Azure authenticator
```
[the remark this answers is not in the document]
orenbm: Did I?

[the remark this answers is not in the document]
orenbm: Ok, i'll do that

cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
```diff
@@ -0,0 +1,120 @@
...
+    And I fetch an Azure access token from inside machine
+    And I save my place in the log file
+    When I authenticate via Azure with token as host "test-app"
+    Then it is unauthorized
+    And The following appears in the log after my savepoint:
+    """
+    Errors::Conjur::RequiredResourceMissing
+    """
+
+  # TODO: add this test when issue #1085 is done
```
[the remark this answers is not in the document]
orenbm: We don't to assume the test order. It may run after a test in another scenario.

cucumber/authenticators_azure/features/authn_azure_basic_host.feature
```diff
@@ -0,0 +1,94 @@
...
+    Given I fetch an Azure access token from inside machine
+    And I authenticate via Azure with token as host "test-app"
+    And host "test-app" has been authorized by Conjur
+    # Update provider uri to a different hostname and verify `provider-uri` has changed
+    When I add the secret value "https://different-provider:8443" to the resource "cucumber:variable:conjur/authn-azure/prod/provider-uri"
+    And I fetch an Azure access token from inside machine
+    And I authenticate via Azure with token as host "test-app"
+    Then it is bad gateway
+    # Check recovery to a valid provider uri
+    When I successfully set Azure variables
```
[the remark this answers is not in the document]
orenbm: I can change the step to `I successfully set Azure variables with the correct values` if you think it's better.

cucumber/authenticators_azure/features/authn_azure_basic_host.feature
```diff
@@ -0,0 +1,94 @@
...
+    And I fetch an Azure access token from inside machine
+    And I authenticate via Azure with token as host "test-app"
+    And host "test-app" has been authorized by Conjur
+
```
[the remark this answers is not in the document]
orenbm: We test that in the vanilla test. I don't want to add the variable and value in every test. In the vanilla flow we verify that the Conjur access token is good for retrieving secrets (i must admit that i don't like that either but i had more important battles when we had that conversation)

cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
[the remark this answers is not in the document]
eladkug: 👍 i agree, i read the issue again and understand better

cucumber/authenticators_azure/features/authn_azure_basic_host.feature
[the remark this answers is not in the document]
eladkug: yes.. its better

cucumber/authenticators_azure/features/authn_azure_basic_host.feature
orenbm: This test is for testing that we return a Bad Gateway response code in case of an invalid Provider URI.

The test above it is for testing that modifying the `provider-uri` variable reflects on the ID Provider endpoint.

More info here: https://github.com/cyberark/conjur/commit/5907a0c5de9faa4a472965ccb4fad4262f3c693f

cucumber/authenticators_azure/features/authn_azure_hosts.feature
orenbm: that's just a random message. i don't want to verify each message.

eladkug: maybe verify there is no error on logs? we dont know what happens in logs

orenbm: I don't think we need - or easily can - check that the logs don't contain error message. We have UTs for checking that if an error is raised then we fail the request so verifying a 200 response code is enough for that.

cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
```diff
@@ -0,0 +1,120 @@
+Feature: Azure Authenticator - Hosts can authenticate with Azure authenticator
+
```
eladkug: add description

cucumber/authenticators_azure/features/authn_azure_basic_host.feature
eladkug: This case is not related to bad_cofiguration ?

orenbm:
> This case is not related to bad_cofiguration ?

@eladkug I moved it

cucumber/authenticators_azure/features/authn_azure_basic_host.feature
```diff
@@ -0,0 +1,94 @@
...
+    And I authenticate via Azure with token as host "test-app"
+    Then host "test-app" has been authorized by Conjur
+    And I successfully GET "/secrets/cucumber/variable/test-variable" with authorized user
+
+  Scenario: Changing provider-uri dynamically reflects on the ID Provider endpoint
+    Given I fetch an Azure access token from inside machine
+    And I authenticate via Azure with token as host "test-app"
+    And host "test-app" has been authorized by Conjur
+    # Update provider uri to a different hostname and verify `provider-uri` has changed
+    When I add the secret value "https://different-provider:8443" to the resource "cucumber:variable:conjur/authn-azure/prod/provider-uri"
```
nessiLahav: Do we want to validate also this change is written to the log?

cucumber/authenticators_azure/features/authn_azure_basic_host.feature
orenbm: Moved to bad_configuration feature

cucumber/authenticators_azure/features/authn_azure_basic_host.feature
orenbm: yeah that's a good point, thanks.

PR #1435, opened 2020-03-17
Log Azure constraint name in annotation type

orenbm:
We would like to log the constraints as an annotation (e.g `resource-group` instead of `resource_group`) so it is clearer to user what they need to fix.

### DIFF
--- a/app/domain/authentication/authn_azure/validate_application_identity.rb
+++ b/app/domain/authentication/authn_azure/validate_application_identity.rb
@@ -165,6 +165,8 @@ module Authentication
         constraints.map { |constraint| annotation_type_constraint(constraint) }
       end
 
+      # converts the constraint to be in annotation style (e.g resource-group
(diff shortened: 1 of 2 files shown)


app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -153,12 +155,20 @@ def validate_token_identity_matches_annotations
...
           end
         end
         @logger.debug(Log::ValidatedApplicationIdentity.new)
       end
 
+      def annotation_type_constraints constraints
+        constraints.map { |constraint| annotation_type_constraint(constraint) }
+      end
+
+      def annotation_type_constraint constraint
```
nessiLahav: I think we should add comment to the function that explain why we are doing it

orenbm: Thanks

PR #1439, opened 2020-03-19
Tmp/rails 5 again squash

jonahx:
BRANCH ONLY FOR TESTING -- WILL BE DELETED, NOT MERGED


jonahx:
Was always a temp branch.   No longer needed.

Issue #1440, opened 2020-03-19
K8s authenticator fails to authenticate using Deployment name for recent K8s versions


diverdane:
## Summary
The Kubernetes authenticator plugin is failing to authenticate and results in 401 errors when:
* A deployment name is being used as a Kubernetes authn identity
* Kubernetes version that the app is running on is recent (v1.16 or newer)
* (This may not be an absolute requirement) Follower is running externally to the cluster that the app is on

This failure was reported in this SalesForce case:
https://cyberark.lightning.force.com/lightning/r/Case/5002J00001JXbzfQAD/view
and a dap-support issue was also created:
https://github.com/conjurinc/dap-support/issues/19

When follower debugs are enabled with the configuration described above, the following error is seen in the follower logs:
```
<15>1 2020-03-16T14:58:27.000+00:00 dap-access-node-66f8df9946-qg2k4 conjur-possum 1277 - [meta sequenceId="12"] [origin=192.168.247.106] [request_id=27e5a551-2584-406d-bb70-ad1fae84cee6] [tid=1477] Authentication Error: #<Authentication::AuthnK8s::K8sObjectLookup::K8sForbiddenError: forbidden: User "system:serviceaccount:dap:conjur-cluster" cannot get path "/oapi/v1">
```
Full log is here: https://cyberark.lightning.force.com/lightning/r/ContentDocument/0692J000008icAbQAI/view

What's curious is that the issue was reported in a non-OpenShift environment, but the API endpoint being queried is an OpenShift specific endpoint.

Looking at the code where this error is happening, you can see some kubernetes clients being created in the K8s authenticator: https://github.com/cyberark/conjur/blob/1278b6630cdd5e8f924429b58245dda9a85a2ef9/app/domain/authentication/authn_k8s/k8s_object_lookup.rb#L150-L163
```
       # If more API versions appear, add them here.
      # List them in the order that you want them to be searched for methods.
      def k8s_clients
        @clients ||= [
          kubectl_client,
          KubeClientFactory.client(api: 'apis/apps', version: 'v1beta2', host_url: api_url, options: options),
          KubeClientFactory.client(api: 'apis/apps', version: 'v1beta1', host_url: api_url, options: options),
          KubeClientFactory.client(api: 'apis/extensions', version: 'v1beta1', host_url: api_url, options: options),
          # OpenShift 3.3 DeploymentConfig
          KubeClientFactory.client(api: 'oapi', version: 'v1', host_url: api_url, options: options),
[...]

PR #1445, opened 2020-03-20
Fixes support for deployment k8s authen for k8s 1.16 or newer

diverdane:
#### What does this PR do?
This change fixes support for using a deployment name as a Kubernetes
authenticator ID for Kubernetes server version 1.16 or newer.

This support had been broken for Kubernetes versions 1.16 or newer
because the server support for deployments (and some other Kubernetes
resources) using the deprecated api/v1beta1 endpoints was removed as
of Kubernetes 1.16, and the Kubernetes client that we use in the
Conjur Kubernetes authenticator was not updated.

The authenticator client had support for:
- Group: api (core), Version: v1

But was missing support for:
- Group: apis/apps, Version: v1
- Group: apis/extensions, Version: v1

The fix is to add explicit support for the missing apps/extensions
endpoints.

This change also adds some sorely needed debug level messages to
help with triage/troubleshooting of this client.

Fixes Issue #1440

Fixes Issue #1440

CHANGELOG.md was updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Postgres advisory lock used for rotation
 - Updated conjur-policy-parser to
   [v3.0.4](https://github.com/conjurinc/conjur-policy-parser/blob/master/CHANGELOG.md#v304).
+- Fix support for using deployment as K8s authentication resource type for Kubernetes >= 1.16 (#1440)
 
 ## [1.4.7] - 2020-03-12
 
--- a/app/domain/authentication/authn_k8s/k8s_object_lookup.rb
+++ b/app/domain/authentication/authn_k8s/k8s_object_lookup.rb
@@ -152,13 +152,36 @@ module Authentication
       def k8s_clients
         @clients ||= [
           kubectl_client,
-          KubeClientFactory.client(api: 'apis/apps', version: 'v1beta2', host_url: api_url, options: options),
-          KubeClientFactory.client(api: 'apis/apps', version: 'v1beta1', host_url: api_url, options: options),
-          KubeClientFactory.client(api: 'apis/extensions', version: 'v1beta1', host_url: api_url, options: options),
+          KubeClientFactory.client(
+            api: 'apis/apps', version: 'v1', host_url: api_url,
+            options: options
+          ),
+          KubeClientFactory.client(
+            api: 'apis/apps', version: 'v1beta2', host_url: api_url,
+            options: options
+          ),
+          KubeClientFactory.client(
+            api: 'apis/apps', version: 'v1beta1', host_url: api_url,
+            options: options
+          ),
+          KubeClientFactory.client(
+            api: 'apis/extensions', version: 'v1', host_url: api_url,
+            options: options
+          ),
+          KubeClientFactory.client(
+            api: 'apis/extensions', version: 'v1beta1', host_url: api_url,
+            options: options
+          ),
           # OpenShift 3.3 DeploymentConfig
-          KubeClientFactory.client(api: 'oapi', version: 'v1', host_url: api_url, options: options),
+          KubeClientFactory.client(
+            api: 'oapi', version: 'v1', host_url: api_url,
+            options: options
+          ),
           # OpenShift 3.7 DeploymentConfig
-          KubeClientFactory.client(api: 'apis/apps.openshift.io', version: 'v1', host_url: api_url, options: options)
+          KubeClientFactory.client(
+            api: 'apis/apps.openshift.io', version: 'v1', host_url: api_url,
+            options: options
+          )
         ]
       end
 


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Changed
 - Postgres advisory lock used for rotation
+- Fix support for using deployment as K8s authen ID for Kubernetes >= 1.16
```
izgeri:
```suggestion
- Fix support for using deployment as K8s authentication resource type for Kubernetes >= 1.16 (#1440)
```

app/domain/authentication/authn_k8s/k8s_object_lookup.rb
```diff
@@ -134,12 +134,29 @@ def invoke_k8s_method method_name, *arguments
         k8s_client_for_method(method_name).send(method_name, *arguments)
       end
 
+      def k8s_client_info client
+        group    = client.instance_variable_get(:@api_group)
+        version  = client.instance_variable_get(:@api_version)
+        entities = client.instance_variable_get(:@entities).to_yaml
+        "api_group: #{group}, api_version: #{version}, entities: #{entities}"
+      end
```
jonahx: I didn't realize you wanted to use these methods in production code.  They really aren't meant to be used like this.  If the client is making them private, they really should be considered private.

I can't do the research on figured out if there's a more proper way to do this, but just be aware that what we're doing here is very brittle and liable to break with the next client release.  If you've considered all that and still decided this is the only way, let me know.

app/domain/authentication/authn_k8s/k8s_object_lookup.rb
```diff
@@ -134,12 +134,29 @@ def invoke_k8s_method method_name, *arguments
...
+
       # Methods move around between API versions across releases, so search the
       # client API objects to find the method we are looking for.
       def k8s_client_for_method method_name
         k8s_clients.find do |client|
           begin
-            client.respond_to?(method_name)
+            responded   = client.respond_to?(method_name)
```
jonahx: `has_method` would be a more accurate name

app/domain/authentication/authn_k8s/k8s_object_lookup.rb
izgeri: if this part is questionable or merits further discussion, I don't think it's needed to get the client support for new API versions in (right, @diverdane?) and we can separate this out into another PR to discuss there

app/domain/authentication/authn_k8s/k8s_object_lookup.rb
```diff
@@ -134,12 +134,29 @@ def invoke_k8s_method method_name, *arguments
...
+            # The call to respond_to? has updated the list of entities
+            # supported by that client, so we can gather that now.
+            client_info = k8s_client_info(client)
+            result = if responded
+              "supported"
+            else
+              "NOT supported"
+            end
```
jonahx: ruby =/= Go :)

A more idiomatic way to write this would be:
```ruby
support_desc = responded ? "supported" : "NOT supported"
```

app/domain/authentication/authn_k8s/k8s_object_lookup.rb
```diff
@@ -134,12 +134,29 @@ def invoke_k8s_method method_name, *arguments
...
+            Rails.logger.debug("Method #{method_name} is #{result} for client: #{client_info}")
```
jonahx: Wrap:
```ruby
            Rails.logger.debug(
              "Method #{method_name} is #{result} for client: #{client_info}"
            )
```

app/domain/authentication/authn_k8s/k8s_object_lookup.rb
```diff
@@ -152,8 +169,10 @@ def k8s_client_for_method method_name
       def k8s_clients
         @clients ||= [
           kubectl_client,
+          KubeClientFactory.client(api: 'apis/apps', version: 'v1', host_url: api_url, options: options),
           KubeClientFactory.client(api: 'apis/apps', version: 'v1beta2', host_url: api_url, options: options),
           KubeClientFactory.client(api: 'apis/apps', version: 'v1beta1', host_url: api_url, options: options),
+          KubeClientFactory.client(api: 'apis/extensions', version: 'v1', host_url: api_url, options: options),
```
jonahx: Break for line length: 

```ruby
      def k8s_clients
        @clients ||= [
          kubectl_client,
          KubeClientFactory.client(
            api: 'apis/apps', version: 'v1', host_url: api_url, options: options
          ),
          KubeClientFactory.client(
            api: 'apis/apps', version: 'v1beta2', host_url: api_url,
            options: options
          ),
          KubeClientFactory.client(
            api: 'apis/apps', version: 'v1beta1', host_url: api_url,
            options: options
          ),
          KubeClientFactory.client(
            api: 'apis/extensions', version: 'v1', host_url: api_url,
            options: options
          ),
          KubeClientFactory.client(
            api: 'apis/extensions', version: 'v1beta1', host_url: api_url,
            options: options
          ),
          # OpenShift 3.3 DeploymentConfig
          KubeClientFactory.client(
            api: 'oapi', version: 'v1', host_url: api_url, options: options
          ),
          # OpenShift 3.7 DeploymentConfig
          KubeClientFactory.client(
            api: 'apis/apps.openshift.io', version: 'v1', host_url: api_url,
            options: options
          )
        ]
      end
```

app/domain/authentication/authn_k8s/k8s_object_lookup.rb
diverdane: Yes, this isn't necessary for the fix. I'm going to bail on adding the debug level messages for now. I took a look at the kubeclient documentation, and I don't see any good way to expose the "entities" that are supported by each client. (There's an "all_entities" call documented, but it gives me an empty result. :( ).

app/domain/authentication/authn_k8s/k8s_object_lookup.rb
diverdane: This is what happens when a Go coder tries writing Ruby.

PR #1446, opened 2020-03-21
Split xms_mirid logic into its own class

orenbm:
### What does this PR do?

This PR has 2 commits. The larger one splits `xms_mirid` logic into its own class

We have too much xms_mirid logic for parsing and validations to have it
inside the general ValidateApplicationIdentity class. Moving this logic
into its own class makes it easier to read the code and test it.

The smaller commit passes the `Rails.logger` into `ApplicationIdentity` so it is easier to understand that it is a dependency and to mock it in case we want to.

### DIFF
--- a/app/domain/authentication/authn_azure/xms_mirid.rb
+++ b/app/domain/authentication/authn_azure/xms_mirid.rb
@@ -1,0 +1,64 @@
+module Authentication
+  module AuthnAzure
+
+    Err = Errors::Authentication::AuthnAzure
+
+    class XmsMirid
+
+      REQUIRED_KEYS = %w(subscriptions resourcegroups providers)
+
+      def initialize(xms_mirid_token_field)
+        @xms_mirid_token_field = xms_mirid_token_field
+
+        @mirid_parts_hash = mirid_parts_hash
+        validate
+      end
+
+      def subscriptions
+        @mirid_parts_hash["subscriptions"].first
+      end
+
+      def resource_groups
+        @mirid_parts_hash["resourcegroups"].first
+      end
+
+      def providers
+        @mirid_parts_hash["providers"]
+      end
+
+      private
+
+      # we expect the xms_mirid claim to be in the format of /subscriptions/<subscription-id>/...
+      # therefore, we ignore the first slash of the xms_mirid claim and group the entries in key-value pairs
+      # according to fields we need to retrieve from the claim.
+      # ultimately, transforming "/key1/value1/key2/value2" to {"key1" => "value1", "key2" => "value2"}
+      def mirid_parts_hash
+        raw_mirid_parts = @xms_mirid_token_field.split('/')
+
+        # accept also an xms_mirid field that doesn't start with a slash
+        mirid_parts = raw_mirid_parts.first == '' ?
+                        raw_mirid_parts.drop(1) :
+                        raw_mirid_parts
+
+        # transform ["subscriptions", "a", "resourcegroups", "b", "providers", "c", "d", "e"]
+        # to {"subscriptions"=>["a"], "resourcegroups"=>["b"], "providers"=>["c", "d", "e"]}
+        mirid_parts.slice_before(Regexp.new(REQUIRED_KEYS.join('|')))
+          .map{|x| [x.first, x.drop(1)] }.to_h
+      rescue => e
+        # this should never occur, it's here to enhance supportability in case it does
+        raise Err::XmsMiridParseError.new(@xms_mirid_token_field, e.inspect)
+      end
+
+      def validate
+        missing_keys = REQUIRED_KEYS - @mirid_parts_hash.keys
+        unless missing_keys.empty?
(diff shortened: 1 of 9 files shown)


app/domain/authentication/authn_azure/xms_mirid.rb
```diff
@@ -0,0 +1,65 @@
...
+        # accept also an xms_mirid field that doesn't start with a slash
+        if split_xms_mirid.first == ''
+          split_xms_mirid = split_xms_mirid.drop(1)
+        end
+
+        split_xms_mirid
+          .slice_before(/subscriptions|resourcegroups|providers/)
+          .each_with_object({}) do |part, hash|
+          key, val  = part.first, part.drop(1)
+          val       = val.first if val.length == 1
```
jonahx: I'm about to make a subtle point I didn't bother with earlier, but now that you've done the extraction of this object you might enjoy.  Notice the:
```ruby
val       = val.first if val.length == 1
```
violates the "avoid mutation" heuristic I made above.  Tbf, in this case I don't really mind it, but I point it out as something that will disappear automatically when we make another change...

Notice there is something funny about what we're doing here, in that we return a _hash with a non-uniform data type_.  You're allowed to do this in loosy-goosy ruby, but you shouldn't.  That is, the _natural_ data type being returned here is `Hash[string]: [string]` -- a hash whose values are arrays of strings.  That's what we get if we just delete line 45.  So we've done something funny by making it into "a hash whose values are sometimes strings and sometimes arrays of strings".

Now that we have a nice little value object, we don't need to commit this minor sin anymore.  Instead, we can move the knowledge that we want "subscriptions" and "resourcegroups" to be simple strings into their accessor methods.  And memoization allows us to do this without no performance hit either:

```ruby
      def subscriptions
        @subscriptions ||= @xms_mirid_hash["subscriptions"].first
      end

      def resource_groups
        @resource_groups ||= @xms_mirid_hash["resource_groups"].first
      end
```

spec/app/domain/authentication/authn-azure/xms_mirid_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  let(:xms_mirid_token_missing_subscription_id_field) { "/resourcegroups/some-resource-group-value/providers/Microsoft.ManagedIdentity/userAssignedIdentities/some-system-assigned-identity-value" }
+
+  let(:xms_mirid_token_missing_resource_groups_field) { "/subscriptions/some-subscription-id-value/providers/Microsoft.ManagedIdentity/some-user-assigned-identity-value" }
+
+  let(:xms_mirid_token_missing_providers_field) { "/subscriptions/some-subscription-id-value/resourcegroups/some-resource-group-value/" }
+
+  let(:xms_mirid_token_missing_initial_slash) { "subscriptions/some-subscription-id-value/resourcegroups/some-resource-group-value/providers/Microsoft.ManagedIdentity/userAssignedIdentities/some-system-assigned-identity-value" }
+
+  let(:invalid_xms_mirid_token_providers_field) { "/subscriptions/some-subscription-id-value/resourcegroups/some-resource-group-value/providers/Microsoft.ManagedIdentity/some-user-assigned-identity-value" }
```
jonahx: line break these

spec/app/domain/authentication/authn-azure/xms_mirid_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+  context "An invalid xms_mirid claim in the Azure token" do
+    context "that is missing subscriptions field in xms_mirid" do
+      subject do
+        Authentication::AuthnAzure::XmsMirid.new(
+          xms_mirid_token_missing_subscription_id_field
+        )
+      end
+      it "raises an error" do
+        expect { subject }.to raise_error(::Errors::Authentication::AuthnAzure::MissingRequiredFieldsInXmsMirid)
+      end
```
jonahx: line break nit:
```ruby
expect { subject }.to raise_error(
  ::Errors::Authentication::AuthnAzure::MissingRequiredFieldsInXmsMirid
)
```

spec/app/domain/authentication/authn-azure/xms_mirid_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        Authentication::AuthnAzure::XmsMirid.new(
+          invalid_xms_mirid_token_providers_field
+        )
+      end
+      it "raises an error" do
+        expect { subject }.to raise_error(::Errors::Authentication::AuthnAzure::MissingProviderFieldsInXmsMirid)
+      end
+    end
+  end
+end
```
jonahx: Really nice, readable, thorough unit tests.

app/domain/authentication/authn_azure/xms_mirid.rb
orenbm: 🤤indeed!

PR #1451, opened 2020-03-24
Specify and Default to Ruby 2.5 in Gemfile and Dockerfile

micahlee:
### What does this PR do?

This PR updates the Gemfile to specify that Conjur expects Ruby 2.5. This provides a more explicit error when attempting to use a different version of Ruby.

The PR also updates the Dockerfile to update the `ruby` and `gem` symlinks to explicitly point to Ruby 2.5. This is because the Ubuntu 20.04 base images defaults to Ruby 2.7 now, with no clear, better way to change the default.

#### Change log

#### Test coverage

#### Follow-on issues

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,9 @@ RUN apt-get update -y && \
                        tzdata \
                        # needed to build some gem native extensions:
                        libz-dev \
-    && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


sgnn7:
@micahlee It might be good for us to also figure out how we can decouple OS ruby version from required ruby version (rvm/rubyenv/etc) at some point - I feel like this has bit us in the rear before with debify too.

PR #1459, opened 2020-03-30
Remove API docs folder

jtuttle:
The docs team has their own copy of the API docs that they're now using to publish our public docs.


sgnn7:
@jtuttle The branch needs a rebase I think

Issue #1434, opened 2020-03-17
Test hosts with `system-assigned-identity` annotation can authenticate with `authn-azure`


orenbm:
In [this PR](https://github.com/cyberark/conjur/pull/1433) we added all the tests for the Azure Authenticator except for the scenario where hosts can (or cannot) authenticate with `authn-azure` with the `system-assigned-identity` annotation. 

This is not possible at the moment as we have an issue with our Jenkins plugin to enable both system assigned and user assigned identities. Once we overcome that we should add these tests.

DoD:
- [ ] Test for successful authentication with the `system-assigned-identity` annotation is added and passes
- [ ] Test for unsuccessful authentication with the `system-assigned-identity` annotation is added and passes

PR #1463, opened 2020-03-31
Conjur is tested with Azure system assigned identities

orenbm:
Connected to #1434 

We would like to test scenarios where the host that authenticates
with authn-azure has a system-assigned-identity annotation.

The scenarios include:
  - host with correct value succeeds
  - host with incorrect value fails
  - host with both system-assigned-identity and -assigned-identity annotations fails

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -78,8 +78,11 @@ pipeline {
         stage('Azure Authenticator preparation - Allocate Azure Authenticator Instance') {
           steps {
             script {
-              node('azure-authn') {
-                env.AZURE_AUTHN_INSTANCE_IP = sh(script: 'curl icanhazip.com', returnStdout: true)
+              node('azure-linux') {
+                // get `ci/authn-azure/get_system_assigned_identity.sh` from scm
+                checkout scm
+                env.AZURE_AUTHN_INSTANCE_IP = sh(script: 'curl icanhazip.com', returnStdout: true).trim()
+                env.SYSTEM_ASSIGNED_IDENTITY = sh(script: 'ci/authn-azure/get_system_assigned_identity.sh', returnStdout: true).trim()
                 env.KEEP_AZURE_AUTHN_INSTANCE = "true"
                 while(env.KEEP_AZURE_AUTHN_INSTANCE == "true") {
                   sleep(time: 15, unit: "SECONDS")
@@ -91,7 +94,7 @@ pipeline {
         stage('Azure Authenticator') {
           steps {
             script {
-              while (!env.AZURE_AUTHN_INSTANCE_IP?.trim()) {
+              while (!env.AZURE_AUTHN_INSTANCE_IP?.trim() || !env.SYSTEM_ASSIGNED_IDENTITY?.trim()) {
                 sleep(time: 15, unit: "SECONDS")
               }
               sh(
--- a/ci/authn-azure/get_system_assigned_identity.sh
+++ b/ci/authn-azure/get_system_assigned_identity.sh
@@ -1,0 +1,49 @@
+#!/bin/bash -ex
+
+### Retrieves a System Assigned Identity access token from Azure metadata service and
+### returns the System Assigned Identity Object ID
+
+main() {
+  azure_access_token="$(curl 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fmanagement.azure.com%2F' -H Metadata:true | jq -r '.access_token')"
+
+  decoded_token_payload=$(decode_jwt_payload "$azure_access_token")
+
+  object_id="$(echo "$decoded_token_payload" | jq -r '.oid')"
+
+  echo "$object_id"
+}
+
+# Decodes the payload of a given JWT
+decode_jwt_payload(){
+  decode_jwt_part "$1" 2
+}
+
+# Decodes a JWT and returns either the header or the payload
+# $1 - the encoded JWT
+# $2 - the required part: 1 for header, 2 for payload, 3 for signature
+decode_jwt_part(){
+  jwt_part_to_decode="$(get_jwt_part "$1" "$2")"
+
+  decoded_jwt_part="$(decode_base64 "$jwt_part_to_decode")"
+
+  echo "$decoded_jwt_part"
+}
+
+# Splits a given JWT and returns the required part
+# $1 - the encoded JWT
+# $2 - the required part: 1 for header, 2 for payload, 3 for signature
+get_jwt_part(){
+  echo -n $1 | cut -d "." -f $2
+}
+
+# Decodes a given data that is base64 encoded
+decode_base64() {
+  local len=$((${#1} % 4))
+  local result="$1"
+  if [ $len -eq 2 ]; then result="$1"'=='
+  elif [ $len -eq 3 ]; then result="$1"'='
+  fi
+  echo "$result" | tr '_-' '/+' | openssl enc -d -base64
+}
+
+main
--- a/ci/authn-azure/secrets.yml
+++ b/ci/authn-azure/secrets.yml
@@ -1,3 +1,8 @@
+# Variables listed in this file are required in order to run the authn-azure tests.
+#
+# NOTE: we also need the variables AZURE_AUTHN_INSTANCE_IP & SYSTEM_ASSIGNED_IDENTITY
+#       but they will come from the Jenkins stage that allocates the Azure VM
+
 AZURE_TENANT_ID: !var ci/azure/tenant-id
 AZURE_SUBSCRIPTION_ID: !var ci/azure/subscription-id
 AZURE_RESOURCE_GROUP: !var ci/azure/authn-test/resource-group
@@ -5,5 +10,3 @@ USER_ASSIGNED_IDENTITY: !var ci/azure/authn-test/user-assigned-id
 USER_ASSIGNED_IDENTITY_CLIENT_ID: !var ci/azure/authn-test/user-assigned-id-client-id
 AZURE_AUTHN_INSTANCE_USERNAME: !var ci/azure/authn-test/instance-user
 AZURE_AUTHN_INSTANCE_PASSWORD: !var ci/azure/authn-test/instance-password
-
-# TODO: add system-assigned identity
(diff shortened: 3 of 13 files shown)


cucumber/authenticators_azure/features/authn_azure_hosts.feature
```diff
@@ -117,8 +141,23 @@ Feature: Azure Authenticator - Different Hosts can authenticate with Azure authe
     Errors::Authentication::AuthnAzure::InvalidApplicationIdentity
     """
 
+  Scenario: Host with incorrect system-assigned-identity annotation is authorized
```
eladkug: this scenario is denied/unauthorized

Jenkinsfile
```diff
@@ -91,7 +94,7 @@ pipeline {
         stage('Azure Authenticator') {
           steps {
             script {
-              while (!env.AZURE_AUTHN_INSTANCE_IP?.trim()) {
+              while (!env.AZURE_AUTHN_INSTANCE_IP?.trim() || !env.SYSTEM_ASSIGNED_IDENTITY?.trim()) {
```
eladkug: why we need !env.SYSTEM_ASSIGNED_IDENTITY?.trim() here?

Jenkinsfile
```diff
@@ -78,8 +78,11 @@ pipeline {
         stage('Azure Authenticator preparation - Allocate Azure Authenticator Instance') {
```
eladkug: I think "Allocate Azure Authenticator Instance" is enough.

ci/authn-azure/get_system_assigned_identity.sh
```diff
@@ -0,0 +1,25 @@
...
+  elif [ $len -eq 3 ]; then result="$1"'='
+  fi
+  echo "$result" | tr '_-' '/+' | openssl enc -d -base64
+}
+
+decode_jwt(){
+  decode_base64_url "$(echo -n $2 | cut -d "." -f $1)" | jq .
+}
+
+function bl_decode_jwt_payload(){
```
eladkug: why here you use function and other functions doesnt contain this?

ci/authn-azure/secrets.yml
```diff
@@ -5,5 +5,3 @@ USER_ASSIGNED_IDENTITY: !var ci/azure/authn-test/user-assigned-id
 USER_ASSIGNED_IDENTITY_CLIENT_ID: !var ci/azure/authn-test/user-assigned-id-client-id
 AZURE_AUTHN_INSTANCE_USERNAME: !var ci/azure/authn-test/instance-user
 AZURE_AUTHN_INSTANCE_PASSWORD: !var ci/azure/authn-test/instance-password
-
```
eladkug: Just to make this clear i would add a commend that SYSTEM_ASSIGNED_IDENTITY is specific variable that taking during Allocate Azure Authenticator Instance.

ci/authn-azure/get_system_assigned_identity.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -ex
+
+decode_base64_url() {
+  local len=$((${#1} % 4))
```
sigalsax, line 4: there is a lot of technical, syntax in this function that I am not familiar with and when looking at the decode_base64_url function, I think it would benefit from an explanation as to what you are trying to accomplish and why you are moding how the length of the result will give the a result an extra `=` or not
For my knowledge:
`#` sign. After looking it up, basically means prefix removal and I assume this translates to prefix removal of the access token?

Why does the length of the result, determine if the result has an extra `=` or not?

Am I understanding correctly that you are translating/replacing all `_-` patterns with  `/+`

ci/authn-azure/secrets.yml
sigalsax: ➕ ^

cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
```diff
@@ -57,7 +57,7 @@ Feature: Azure Authenticator - Bad authenticator configuration leads to an error
     And I have host "test-app"
     And I set Azure annotations to host "test-app"
     And I grant group "conjur/authn-azure/prod/apps" to host "test-app"
-    And I fetch an Azure access token from inside machine
+    And I fetch a non-assigned Azure access token from inside machine
```
sigalsax, line 60: missing the `identity` part IMO, `non-assigned identity`

cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
```diff
@@ -114,8 +114,8 @@ Feature: Azure Authenticator - Bad authenticator configuration leads to an error
     And I have host "test-app"
     And I set Azure annotations to host "test-app"
     And I grant group "conjur/authn-azure/prod/apps" to host "test-app"
-    And I successfully set Azure variables with the correct values
-    And I fetch an Azure access token from inside machine
+    And I successfully set Azure provider-uri variable with the correct values
```
sigalsax, line 117: Do you think that in the future we will need other variables for authn-azure configuration besides `provider-uri`? 
If so, I would avoid being specific "I successfully set Azure provider-uri variable with the correct values" -> "I successfully set Azure "provider-uri" variable with the correct values" so that we can replace that with any `provider_uri` with any variable and it will functionally be the same

cucumber/authenticators_azure/features/authn_azure_hosts.feature
```diff
@@ -34,11 +34,21 @@ Feature: Azure Authenticator - Different Hosts can authenticate with Azure authe
     When I authenticate via Azure with token as host "user-assigned-identity-app"
     Then host "user-assigned-identity-app" has been authorized by Conjur
 
+  Scenario: Host with system-assigned-identity annotation is authorized
+    And I have host "system-assigned-identity-app"
+    And I set subscription-id annotation to host "system-assigned-identity-app"
+    And I set resource-group annotation to host "system-assigned-identity-app"
+    And I set system-assigned-identity annotation to host "system-assigned-identity-app"
+    And I grant group "conjur/authn-azure/prod/apps" to host "system-assigned-identity-app"
+    And I fetch a system-assigned Azure access token from inside machine
```
sigalsax, line 43: I really like the refactor in the code! (:

```
Given(/I fetch a (non|system|user)-assigned Azure access token from inside machine/) do |identity_type|
  case identity_type
  when "non", "system"
    retrieve_system_assigned_azure_access_token
  when "user"
    retrieve_user_assigned_azure_access_token
  end
end
```
very nice

cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
eladkug: I think this is a  better way, because here the user will understand which variable the automation set and it will be easy to debug in case of failure.
In my vision if you see the results via jenkins/report you will understand exactly the test and the parameters.

Jenkinsfile
```diff
@@ -78,8 +78,11 @@ pipeline {
...
           steps {
             script {
-              node('azure-authn') {
-                env.AZURE_AUTHN_INSTANCE_IP = sh(script: 'curl icanhazip.com', returnStdout: true)
+              node('azure-linux') {
+                // get `ci/authn-azure/get_system_assigned_identity.sh` from scm
+                checkout scm
```
sigalsax, line 83: `checkout scm` allows whatever is below it to run inside the Azure VM?

Jenkinsfile
orenbm: these 2 variables are required for authn-azure tests and are retrieved from the parallel step where we allocate the azure vm. for this reason we wait for the tests to run until we have those variables in the env

Jenkinsfile
orenbm: because this stage runs in parallel with all the other tests (and not only azure) i wanted to make it obvious that it's related to the azure authenticator tests

ci/authn-azure/get_system_assigned_identity.sh
orenbm: made the whole script clearer and easier to understand.

ci/authn-azure/secrets.yml
orenbm: Added a note

ci/authn-azure/get_system_assigned_identity.sh
orenbm: i didn't go into detail of how this method works, and i don't think it is very important. 

Please take a look again at the script. I re-arranged it to be more readable.

cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
orenbm: i don't think the azure authenticator will have additional variables in the future. If it will have, it won't be hard to replace it. I agree with elad that the value of explicitly saying what the step does is high.

Jenkinsfile
orenbm: it gets the whole source code, including the script that we need to run in the azure vm

cucumber/authenticators_azure/features/authn_azure_hosts.feature
eladkug: need to change here too..(you changed for user-assigned-identity only ...

orenbm: oh yeah missed it. thanks!

ci/authn-azure/get_system_assigned_identity.sh
```diff
@@ -0,0 +1,49 @@
...
+# $2 - the required part: 1 for header, 2 for payload, 3 for signature
+decode_jwt_part(){
+  jwt_part_to_decode="$(get_jwt_part "$1" "$2")"
+
+  decoded_jwt_part="$(decode_base64 "$jwt_part_to_decode")"
+
+  echo "$decoded_jwt_part"
+}
+
+# Splits a given JWT and returns the required part
```
sigalsax, line 32: thanks!

ci/authn-azure/get_system_assigned_identity.sh
sigalsax: 👍 on the explanation

PR #1469, opened 2020-04-02
Fix message for invalid providers field in xms_mirid

orenbm:
After we check that the length of the xms_mirid's `providers`
field equals 3 we raise the error `MissingProviderFieldsInXmsMirid`.
The field is not missing as if it were we would raise `MissingRequiredFieldsInXmsMirid`
instead.

### DIFF
--- a/app/domain/authentication/authn_azure/xms_mirid.rb
+++ b/app/domain/authentication/authn_azure/xms_mirid.rb
@@ -58,7 +58,7 @@ module Authentication
         end
 
         unless @mirid_parts_hash["providers"].length == 3
-          raise Err::MissingProviderFieldsInXmsMirid, @xms_mirid_token_field
+          raise Err::InvalidProviderFieldsInXmsMirid, @xms_mirid_token_field
         end
       end
     end
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -307,8 +307,10 @@ unless defined? Errors::Authentication::AuthenticatorNotFound
           code: "CONJ00053E"
         )
 
-        MissingProviderFieldsInXmsMirid = ::Util::TrackableErrorClass.new(
-          msg:  "Provider fields are missing in xms_mirid {1}",
+        InvalidProviderFieldsInXmsMirid = ::Util::TrackableErrorClass.new(
+          msg:  "Provider fields are in invalid format in xms_mirid {1}." \
+                "xms_mirid must contain the resource provider namespace, the " \
+                "resource type, and the resource name",
           code: "CONJ00054E"
         )
       end
--- a/spec/app/domain/authentication/authn-azure/xms_mirid_spec.rb
+++ b/spec/app/domain/authentication/authn-azure/xms_mirid_spec.rb
@@ -116,7 +116,7 @@ RSpec.describe 'Authentication::AuthnAzure::XmsMirid' do
         )
       end
       it "raises an error" do
-        expect { subject }.to raise_error(::Errors::Authentication::AuthnAzure::MissingProviderFieldsInXmsMirid)
+        expect { subject }.to raise_error(::Errors::Authentication::AuthnAzure::InvalidProviderFieldsInXmsMirid)
       end
     end
   end


orenbm:
@shuligh can you please review the following error: `Provider fields are in invalid format in xms_mirid {1}. It should have 3 parts`.

For example, in case the xms_mirid field is `/subscriptions/some-subscription-id-value/resourcegroups/some-resource-group-value/providers/Microsoft.ManagedIdentity/some-user-assigned-identity-value` then the `providers` fields are `Microsoft.ManagedIdentity/some-user-assigned-identity-value` which are only 2 parts (separated by `/`). I am not sure how to phrase the error message.

Thanks!

app/domain/authentication/authn_azure/xms_mirid.rb
```diff
@@ -58,7 +58,7 @@ def validate
         end
 
         unless @mirid_parts_hash["providers"].length == 3
-          raise Err::MissingProviderFieldsInXmsMirid, @xms_mirid_token_field
+          raise Err::InvalidProviderFieldsInXmsMirid, @xms_mirid_token_field
```
sigalsax, line 61: Like this better

app/domain/errors.rb
```diff
@@ -307,8 +307,9 @@ module AuthnAzure
           code: "CONJ00053E"
         )
 
-        MissingProviderFieldsInXmsMirid = ::Util::TrackableErrorClass.new(
-          msg:  "Provider fields are missing in xms_mirid {1}",
+        InvalidProviderFieldsInXmsMirid = ::Util::TrackableErrorClass.new(
+          msg:  "Provider fields are in invalid format in xms_mirid {1}." \
```
sigalsax, line 311: Do we need a `.` here?

orenbm: there is another sentence below so the message that users will get is `Provider fields are in invalid format in xms_mirid {1}. It should have 3 parts`

app/domain/errors.rb
```diff
@@ -307,8 +307,9 @@ module AuthnAzure
...
+                "It should have 3 parts",
```
sigalsax, line 312: Do we need this hint here?

orenbm: I think that without it the message is very ambiguous and doesn't tell the customer much. Without it the customer will think "well, what _is_ a valid format?"

shuligh:
> @shuligh can you please review the following error: `Provider fields are in invalid format in xms_mirid {1}. It should have 3 parts`.
> 
> For example, in case the xms_mirid field is `/subscriptions/some-subscription-id-value/resourcegroups/some-resource-group-value/providers/Microsoft.ManagedIdentity/some-user-assigned-identity-value` then the `providers` fields are `Microsoft.ManagedIdentity/some-user-assigned-identity-value` which are only 2 parts (separated by `/`). I am not sure how to phrase the error message.
> 
> Thanks!

I agree we deifnitely need to tell the use what is invalid about it. What are these "parts" made up?

orenbm:
> What are these "parts" made up?

They are a 3-parted resource-identifier: `{resource-provider-namespace}/{resource-type}/{resource-name}`.

For exmaple:
`Microsoft.Compute/virtualMachines/myVM`

shuligh:
> > What are these "parts" made up?
> 
> They are a 3-parted resource-identifier: `{resource-provider-namespace}/{resource-type}/{resource-name}`.
> 
> For exmaple:
> `Microsoft.Compute/virtualMachines/myVM`

So from your slack message :
>for example:
Microsoft.Compute/virtualMachines/some-system-assigned-identity-value"
14:43
in general it can be {rg?esource-provider-namespace}/{resource-type}/{resource-name}

so I would say 
Provider fields are in invalid format in xms_mirid {1}.  xms_mirid must contain the resource provider namespace, the resource type, and the resource name.


Is that too long?

shuligh:
BTW, what version is this for?

orenbm:
> Is that too long?
I think it's fine, thanks :)

orenbm:
> BTW, what version is this for?

Next one :)

app/domain/errors.rb
```diff
@@ -307,8 +307,10 @@ module AuthnAzure
           code: "CONJ00053E"
         )
 
-        MissingProviderFieldsInXmsMirid = ::Util::TrackableErrorClass.new(
-          msg:  "Provider fields are missing in xms_mirid {1}",
+        InvalidProviderFieldsInXmsMirid = ::Util::TrackableErrorClass.new(
+          msg:  "Provider fields are in invalid format in xms_mirid {1}." \
+                "xms_mirid must contain the resource provider namespace, the " \
```
sigalsax, line 312: Question: now that I think about it, is it proper “resource provider namespace”?

orenbm: I saw that somewhere but looking [here](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-providers-and-types) makes me want to change it to "resource provider"

orenbm: in any case, searching google for "azure resource provider namespace" will lead the customer to the answer. I'll leave it up to you if you want to create a new PR and change it.

sigalsax: I am okay with this then. Seems like we could go both ways

PR #1475, opened 2020-04-03
Add rake task, which gets next available error code

h-artzi:
### What does this PR do?
makes it easier to get next available error number available for use

N/A

### How to test?
1. ./build.sh
2. cd dev
3. ./start
4. rake error_code:next and it should print out the number 58

#### Change log

#### Test coverage

#### Follow-on issues
- [X] No follow-up issues are required

### DIFF
--- a/lib/tasks/next_available_error.rake
+++ b/lib/tasks/next_available_error.rake
@@ -1,0 +1,8 @@
+require 'util/error_code'
+
+namespace :error_code do
+  task :next do
+    error_code = Error::ConjurCode.new('./app/domain/errors.rb')
+    error_code.print_next_available
+  end
+end
--- a/lib/util/error_code.rb
+++ b/lib/util/error_code.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module Error
+  # helps get information regarding current error codes.
+  class ConjurCode
+    def initialize(path)
+      @path = path
+      validate
+    end
+
+    def print_next_available
+      id = next_code_id
+      unless id
+        $stderr.puts "The path doesn't contain any files with Conjur error codes"
+        return
+      end
+      puts format("The next available error number is %d ( CONJ%05dE )", id, id)
+    end
+
+    # separate data-gathering from printing
+    def next_code_id
+      max_code = existing_codes.max
+      max_code ? max_code + 1 : nil
+    end
+
+    def existing_codes
+      codes = File.foreach(@path).map do |line|
+        match = /code: \"CONJ(?<num>[0-9]+)E\"/.match(line)
+        match ? match[:num].to_i : nil
+      end
+      codes.compact
+    end
+
+    def validate
+      return if File.file?(@path)
+      raise format("The following path:%s was not found", @path)
+    end
+  end
+end
--- a/spec/app/domain/util/trackable_error_spec.rb
+++ b/spec/app/domain/util/trackable_error_spec.rb
@@ -21,3 +21,35 @@ RSpec.describe 'Util::TrackableErrorClass' do
     end
   end
 end
+
+require 'util/error_code'
+
+describe Error::ConjurCode do
+  let(:error_code_valid) do
+    Error::ConjurCode.new('./spec/app/domain/util/errors/valid')
+  end
+  let(:error_code_invalid) { Error::ConjurCode.new('invalid_file_path') }
+  let(:error_code_empty) do
+   Error::ConjurCode.new('./spec/app/domain/util/errors/empty')
+  end
+  
+  it 'raises error when file path is invalid' do
+    expect { error_code_invalid }.to raise_error(RuntimeError)
+  end
+
+  it 'outputs the next available error code' do 
+    expect do
+      error_code_valid.print_next_available
+    end.to output(
+      "The next available error number is 5 ( CONJ00005E )\n"
+    ).to_stdout
+  end
+
+  it 'outputs that the file lacks error codes' do 
+    expect do
+      error_code_empty.print_next_available
+    end.to output(
+      "The path doesn't contain any files with Conjur error codes\n"
+    ).to_stderr
+  end
+end
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -323,6 +323,19 @@ root@9feae5e5e001:/src/conjur-server# cucumber --profile authenticators    # run
 root@9feae5e5e001:/src/conjur-server# cucumber --profile api cucumber/api/features/resource_list.feature
 ```
 
+### Rake Tasks
+
+Rake tasks are easy to run from within the `conjur` server container:
+
+- Get the next available error code from [errors](./app/domain/errors.rb)
+  ```sh-session
+  root@aa8bc35ba7f4:/src/conjur-server# rake error_code:next
+  ```
+    The output will be similar to
+  ```sh-session
+  The next available error number is 63 ( CONJ00063E )
+  ```
+
 ## Pull Request Workflow
 
 1. [Fork the project](https://help.github.com/en/github/getting-started-with-github/fork-a-repo)
--- a/spec/app/domain/util/errors/empty
+++ b/spec/app/domain/util/errors/empty
@@ -1,0 +1,1 @@
+There are no error codes in here!!
--- a/spec/app/domain/util/errors/valid
+++ b/spec/app/domain/util/errors/valid
@@ -1,0 +1,24 @@
+
+# frozen_string_literal: true
+
+require 'util/trackable_error_class'
+
+module Errors
+  module Conjur
+
+    RequiredResourceMissing = Util::TrackableErrorClass.new(
+      msg:  "Missing required resource: {0-resource-name}",
+      code: "CONJ00001E"
+    )
+
+    RequiredSecretMissing = Util::TrackableErrorClass.new(
+      msg:  "Missing value for resource: {0-resource-name}",
+      code: "CONJ00002E"
+    )
+
+    RequiredSecretMissing = Util::TrackableErrorClass.new(
+      msg:  "Missing value for resource: {0-resource-name}",
+      code: "CONJ00004E"
+    )
+  end
+end


lib/tasks/next_available_error.rake
```diff
@@ -0,0 +1,12 @@
+namespace :err do
```
micahlee, line 1: What do you think about changing the command you run to `error_code:next` instead?

h-artzi: Makes sense

jvanderhoof:
Rake tests are non-trivial to write.  Let's move the functionality into a utility class which will simplify testing.

jvanderhoof:
A good post on testing Rake: https://thoughtbot.com/blog/test-rake-tasks-like-a-boss

lib/tasks/next_available_error.rake
```diff
@@ -0,0 +1,12 @@
+namespace :error_code do
+  task :next do
+    err_num = 0
+    File.foreach('./app/domain/errors.rb') do |line|
+      match = line.match(/code: \"CONJ(?<num>[0-9]+)E\"/)
+      if match
+        err_num = [err_num, match[:num].to_i].max
+      end
+    end
+    puts "The next available error number is #{err_num+1}"
```
micahlee, line 10: What do you think about adding the full code in parens for easy copy/paste? For example:
```
$ rake error_code:next
The next available error number is 58 ( CONJ00058E )
```

micahlee:
Hey @h-artzi, any updates to this PR?

jonahx:
Bump.

h-artzi:
What are your thoughts on the following CC suggestion `Provide an exception class and message as arguments to raise`?

lib/util/error_code.rb
```diff
@@ -0,0 +1,43 @@
...
+          err_num = [err_num, match[:num].to_i].max
+        end
+      end
+
+      puts(error_code(err_num+1))
+    end
+
+    def validate
+      msg = format("The following path:%s was not found", @path)
+      raise FileNotFound.new(msg) unless File.file?(@path)
```
micahlee, line 25: Yeah, this is just telling you the idiomatic way to write this would be:

```suggestion
      raise FileNotFound, msg unless File.file?(@path)
```

micahlee:
Hey @h-artzi, I'm running through this again locally now and will then review the test additions. Would you mind rebasing this in the meantime?

lib/util/error_code.rb
```diff
@@ -0,0 +1,40 @@
...
+      raise FileNotFound.new(format("The following path:%s was not found", @path)) unless File.file?(@path)
+    end
+
+    def error_code(num)
+      format("The next available error number is %d ( CONJ%05dE )", num, num)
+    end
+  end
+end
+
+class FileNotFound < StandardError
```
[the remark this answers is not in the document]
jonahx: No need for a custom error here, just ditch this class and `raise "Path not found"` in our validate method.

lib/util/error_code.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    def next_available
+      err_num = -1
+      File.foreach(@path) do |line|
+        match = line.match(/code: \"CONJ(?<num>[0-9]+)E\"/)
+        if match
+          err_num = [err_num, match[:num].to_i].max
+        end
+      end
```
jonahx, line 18: Two things here: 

If you were going to use this method, embrace the ruby way -- use the functional methods from the Enumerable class.  `select` all lines you want, `map` to pull the error numbers, then call `max`.  No procedural loops.

However, parsing the line text is brittle (eg, imagine someone refers to one of these error codes in comments, and has made a typo there).  Ruby allows you to introspect.  

`Errors::Conjur.constants` should give you all the Error class names, possibly others you don't care about.  Filter through them, seeing which respond to the `code` method.  Get all your codes that way.  Then apply the above logic to those strings.

h-artzi: I chose the more brittle options due to the fact that I ran into issues loading the errors module. It would result in an error stating that it could not load `util/trackable_error_class` found in errors.rb.

h-artzi:
@micahlee the updated issue is here #1557

lib/util/error_code.rb
```diff
@@ -0,0 +1,38 @@
...
+
+module Util
+  # helps get information regarding current error codes.
+  class ErrorCode
+    def initialize(path)
+      @path = path
+      validate
+    end
+
+    def next_available
```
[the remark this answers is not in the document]
h-artzi: I currently think this method is easy to follow as is, however, please share your thoughts

lib/util/error_code.rb
```diff
@@ -0,0 +1,38 @@
...
+      code_lines = File.foreach(@path).select {|line| /code: \"CONJ[0-9]+E\"/.match(line)}
+      codes = code_lines&.map { |match| /[0-9]+/.match(match)[0].to_i }
```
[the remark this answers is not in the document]
h-artzi: Is there an alternative to safe navigators?

lib/util/error_code.rb
jonahx: You should be able to just require it explicitly.

One of the most pernicious qualities of Rails is it's "convenient" auto-loading of classes.  This means you don't have to specify the things you're using with `require` statements at the top.  Which makes all your dependencies implicit, as well as screwing you when you use the files outside Rails.  I'm just guessing that's what's happening here (could be wrong).

lib/util/error_code.rb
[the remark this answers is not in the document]
jonahx: Are you sure it give you nil when there are no results?  I would have expected it to give you an empty array, in which case you can safely use `map` with the safe navigation.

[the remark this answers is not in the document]
jonahx: Also, I think the logic with setting the empty case to -1, then incrementing, then testing for 0 in _another_ function  is too tricky.  It introduces unnecessary coupling between the two functions as well.

Instead, I'd suggest:

```ruby
    def next_available
      code_lines = File.foreach(@path).select {|line| /code: \"CONJ[0-9]+E\"/.match(line)}
      codes = code_lines&.map { |match| /[0-9]+/.match(match)[0].to_i }
      puts  "The file doesn't contain any error messages" and return if codes.empty?
      code = codes.max + 1
      puts format("The next available error number is %d ( CONJ%05dE )", code, code)
    end
```
and then you can delete the `error_code` function (which, btw, was misleadingly named :) .... `print_error_code` would have been more accurate)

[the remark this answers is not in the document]
h-artzi: It does give an empty array, so not sure why it thinks it a nil check. I will update `next_available`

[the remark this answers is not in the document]
jonahx:
```ruby
    def print_next_available
      id = next_code_id
      puts "The file doesn't contain any error messages" and return unless id
      puts format("The next available error number is %d ( CONJ%05dE )", id, id)
    end

    # separate data-gathering from printing
    def next_code_id
      max_code = existing_codes.max
      max_code ? max_code + 1 : nil
    end

    def existing_codes
      File.foreach(@path).select do |line|
        /code: \"CONJ([0-9])+E\"/.match(line) do |m|
          m[1].to_i
        end
      end
    end
```

lib/tasks/next_available_error.rake
```diff
@@ -0,0 +1,8 @@
+require 'util/error_code'
+
+namespace :error_code do
+  task :next do
+    error_code = Util::ErrorCode.new('./app/domain/errors.rb')
```
jonahx, line 5: Something I missed: we should avoid `Util`.  Do we already have an `Error` module or `Rake` module?  I'd put it in there instead.

lib/util/error_code.rb
```diff
@@ -0,0 +1,46 @@
...
+  class ErrorCode
+    def initialize(path)
+      @path = path
+      validate
+    end
+
+    def print_next_available
+      id = next_code_id
+      unless id
+        puts "The file doesn't contain any error messages"
```
jonahx, line 14: Accuracy: "The path doesn't contain any files with Conjur error codes"

spec/app/domain/util/trackable_error_spec.rb
```diff
@@ -21,3 +21,33 @@
...
+    expect do
+      error_code_valid.print_next_available
+    end.to output("The next available error number is 5 ( CONJ00005E )\n").to_stdout
+  end
+
+  it 'outputs next available error code but forgets to append leading zeros' do 
+    expect do
+      error_code_valid.print_next_available
+    end.to_not output("The next available error number is 58 ( CONJ5E )\n").to_stdout
+  end
```
jonahx, line 46: This result of this test is logically implied by the result of the test above, so it can be removed.

spec/app/domain/util/trackable_error_spec.rb
```diff
@@ -21,3 +21,33 @@
...
+
+  it 'outputs that the file lacks error codes' do 
+    expect do
+      error_code_empty.print_next_available
+    end.to output("The path doesn't contain any files with Conjur error codes\n").to_stdout
```
jonahx, line 51: A nitpick (don't bother changing unless super easy): This one should probably be printed to stderr.

lib/util/error_code.rb
```diff
@@ -0,0 +1,38 @@
...
+    def next_code_id
+      max_code = existing_codes.max
+      max_code ? max_code + 1 : nil
+    end
+
+    def existing_codes
+      codes = File.foreach(@path).map do |line|
+        match = /code: \"CONJ(?<num>[0-9]+)E\"/.match(line)
+        match ? match[:num].to_i : nil
+      end.compact
```
jonahx, line 30: Forgot to delete this guy.

lib/util/error_code.rb
```diff
@@ -0,0 +1,38 @@
...
+      codes.compact
+    end
+
+    def validate
+      raise format("The following path:%s was not found", @path) unless File.file?(@path)
+    end
```
jonahx, line 36: To preserve margin:
```ruby
def validate
  unless File.file?(@path)
    raise format("The following path:%s was not found", @path)
  end
end
```
Generally, think: "If it fits easily on one line use the "unless/if at the end" syntax.  Once it overflows, just give up and go for traditional block syntax.

spec/app/domain/util/trackable_error_spec.rb
```diff
@@ -21,3 +21,27 @@
     end
   end
 end
+
+require 'util/error_code'
+
+describe Error::ConjurCode do
+  let(:error_code_valid) { Error::ConjurCode.new('./spec/app/domain/util/errors/valid') }
```
jonahx, line 28: Margin overflow again:
```ruby
let(:error_code_valid) do 
  Error::ConjurCode.new('./spec/app/domain/util/errors/valid')
end
```

spec/app/domain/util/trackable_error_spec.rb
```diff
@@ -21,3 +21,27 @@
...
+  let(:error_code_invalid) { Error::ConjurCode.new('invalid_file_path') }
+  let(:error_code_empty) { Error::ConjurCode.new('./spec/app/domain/util/errors/empty') }
+
```
jonahx, line 31: Here too.

spec/app/domain/util/trackable_error_spec.rb
```diff
@@ -21,3 +21,27 @@
...
+  
+  it 'raises error when file path is invalid' do
+    expect { error_code_invalid }.to raise_error(RuntimeError)
+  end
+
+  it 'outputs the next available error code' do 
+    expect do
+      error_code_valid.print_next_available
+    end.to output("The next available error number is 5 ( CONJ00005E )\n").to_stdout
+  end
```
jonahx, line 40: Suggested 80 line formatting.  No official guidelines here, just my opinion:

```ruby
  it 'outputs the next available error code' do 
    expect do
      error_code_valid.print_next_available
    end.to(
      output("The next available error number is 5 ( CONJ00005E )\n"
    ).to_stdout
  end
```

PR #1479, opened 2020-04-07
Add design doc for Application Identity validation

orenbm:
When we add a new authenticator to Conjur, we will need to add the mechanism
of validating the host's Application Identity. Currently we have no generic
way to do that and because of that it is less straight-forward to add a new
authenticator.

This doc tackles this issue and suggests a new approach.

### DIFF
--- a/design/authenticators/application-identity/application-identity.md
+++ b/design/authenticators/application-identity/application-identity.md
@@ -1,0 +1,468 @@
+# Application Identity
+
+##### Note
+
+In this doc, I will cover authenticating hosts with Conjur. Everything about 
+authenticating hosts is also relevant to users. For the scope of this document, 
+host, user or role are interchangeable.
+  
+## Introduction
+
+In several authenticators, such as `authn-k8s` & `authn-azure`, we require hosts 
+to contain some data which define their identity. This data is referred to as Application Identity. 
+The Application Identity gives the authenticator another layer of
+authentication, where it can authenticate applications using not only their
+Conjur identity and permissions (e.g host can `authenticate` with the
+authenticator `webservice`) but also their native identity (e.g Azure
+resource characteristics)
+ 
+For example, in the Azure Authenticator the Application Identity includes:
+- Subscription ID
+- Resource Group
+- System-Assigned Identity (optional)
+- User-Assigned Identity (optional)
+ 
+A host that will authenticate using the Azure authenticator will be defined
+ in its annotations as follows:
+ ```
+- !host 
+  id: test-app 
+  annotations: 
+    authn-azure/subscription-id: test-subscription 
+    authn-azure/resource-group: test-group 
+``` 
+
+While in the Azure Authenticator the Application Identity is always stored in the host annotations,
+in the Kubernetes Authenticator the Application Identity can be stored either
+in the host's ID or in its annotations. 
+For example:
+ ```
+- !host
+  id: test-app-service-account
+  annotations:
+    authn-k8s/namespace: some-namespace
+    authn-k8s/service-account: some-service-account
+
+- !host
+  id: some-namespace/service_account/some-service-account
+```
+
+In the example above, both hosts have the same Application Identity. One stores 
+it in its annotation and the other in its ID. However, their Conjur identity is not
+the same.
+
+## Issue Description
+
+When we add a new authenticator to Conjur, we need to add the mechanism
+of validating the host's Application Identity. Currently we have no generic
+way to do it and because of that it is less straight-forward to add a new
+authenticator. An example of what a generic implementation is can be seen in
+the [Authenticator Status Check](../authenticators-status). As you can see
+in the [Implementing Authenticator Status API](../authenticators-status/authn-status-new-implem.md) 
+doc, it is very simple to add a new Status Check to an Authenticator. This happens because the flow
+starts with general checks that are relevant to all authenticators, and all
+that needs to be done in order to add a new Status Check is to implement a
+method named `status` in the relevant `Authenticator` class, and create a
+Command Class that will perform the actual validations. 
+ 
+However, when validating the Application Identity we currently have no
+generic mechanism, as can be seen in the implementations for the [Azure
+Authenticator](/app/domain/authentication/authn_azure/validate_application_identity.rb) and 
+[Kubernetes Authenticator](/app/domain/authentication/authn_k8s/validate_application_identity.rb), 
+where a lot of code is duplicated with minor changes. 
+
+## Solution
+
+Before we can find the solution, we need to understand how Application
+Identities are related across authenticators, and how they are different.
+ 
+For the sake of simplicity, from here and throughout the document, Application 
+Identities will be defined only in the host's annotations, and not in the ID. 
+In future authenticators we will not develop an option to define it in the ID and it
+is available in the K8s Authenticator only for backwards compatibility.
+
+### Defining Application Identity Validation
+
+The Application Identity validation can be roughly split into 3 parts:
+ - Extract the Application Identity from the host
+ - Validate the Application Identity is defined correctly
+ - Validate the Application Identity matches the authentication request
+
+Let's break down each part to better understand them.
+
+#### Extract the Application Identity from the host
+
+Let's look into the Azure Authenticator's Application Identity class:
+```
+class ApplicationIdentity
+
+    def initialize(role_annotations:, service_id:, logger:)
+        @role_annotations = role_annotations
+        @service_id       = service_id
+        @logger           = logger
+    end
+    
+    def constraints
+        @constraints ||= {
+          subscription_id:          constraint_value("subscription-id"),
+          resource_group:           constraint_value("resource-group"),
+          user_assigned_identity:   constraint_value("user-assigned-identity"),
+          system_assigned_identity: constraint_value("system-assigned-identity")
+        }.compact
+    end
+    
+    private
+    
+    # check the `service-id` specific constraint first to be more granular
+    def constraint_value constraint_name
+        annotation_value("authn-azure/#{@service_id}/#{constraint_name}") ||
+          annotation_value("authn-azure/#{constraint_name}")
+    end
+    
+    def annotation_value name
+        annotation = @role_annotations.find { |a| a.values[:name] == name }
+        
+        # return the value of the annotation if it exists, nil otherwise
+        if annotation
+          @logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
+          annotation[:value]
+    end
+end
+```
+
+The class gets the host's annotations and the authenticator's service-id, 
+and then builds a `constraints` hash from the annotations. 
+
+This is the case also in [`authn-k8s`](/app/domain/authentication/authn_k8s/application_identity.rb), 
+with only one difference as the K8s Authenticator also looks in the host's ID. 
+
+Putting that aside we can see that the behavior is exactly the same. 
+In this case the general class will be defined as follows:
+- input
+  - authenticator name
+  - authenticator service-id
+  - annotations
+  - List of relevant annotations (e.g `[subscription-id, resource-group
+  , system-assigned-identity, user-assigned-identity]
+  ` for `authn
+  -azure`)
+- output
+  - `constraints` hash.
+   
+#### Validate the Application Identity is defined correctly
+
+In the Azure Authenticator and in the K8s Authenticator we have requirements on 
+how the annotations are defined. As can be seen 
+[here](/app/domain/authentication/authn_azure/validate_application_identity.rb), 
+we have the following validations:
+  - Validate annotations are permitted
+    - Here we validate that the `authn-azure` annotations are known. For
+      example, the annotation `authn-azure/subscription-id` is valid, whereas
+      the annotation `authn-azure/blah` is invalid. We do that to minimize
+      errors from the user, where they will define, for example, the
+      annotation `authn-azure/subscription-ir` (note the typo). In this
+      case we'd rather fail the authentication and point them to their
+      mistake rather than ignore it, and authenticate the request in a weaker
+       granularity than what is expected by the user.
+  - Validate required constraints exist
+    - For example, in the Azure Authenticator we require hosts to have the
+     `subscription-id` and the `resource-group` annotations. In the K8s
+      Authenticator we require the `namespace` annotation.
+  - Validate constraint combinations
+    - For example, in the Azure Authenticator we do not allow hosts to have
+     both the `user-assigned-identity` and the `system-assigned-identity`
+     annotations. In the K8s we do not allow hosts to have any combination of
+      the annotations `deployment`, `deployment-config` & `stateful-set`.
+      
+Also here we can see some generalization. 
+In this case the general class will be defined as follows:
+- input
+  - The following lists of annotation names
+    - permitted
+    - required
+    - non-permitted combinations
+  - The `constraints` hash built in the previous step
+- output
+  - Nothing if validation succeeded, relevant error if it failed
+
+#### Validate the Application Identity matches the authentication request
+
+Once we extracted the Application Identity from the host's annotations, and
+verified it is defined correctly, we can validate that the authentication
+request is valid by validating the Application Identity. For example, if
+the k8s host is permitted to authenticate from namespace `namespace-1`, and
+the authentication request was sent from a pod that is in namespace
+`namespace-2` we will fail the request.
+
+In a way, we can define this step as comparing 2 objects of Application
+Identity - one that was extracted from the host and one that was extracted
+from the authentication request. In the example above we will have the
+following Application Identity objects:
+ - Host Application Identity
+   - constraints
+     - namespace: `namespace-1`
+ - Authentication request Application Identity
+   - constraints
+     - namespace: `namespace-2`
+     
+All we need to do is iterate over the Host Application Identity `constraints` hash 
+and verify that all of its fields have the same values as the 
+corresponding `constraints` hash fields in the Authentication request Application Identity object.
+
+In this case the general class will be defined as follows:
+- input
+  - Host Application Identity object
+  - Authentication request Application Identity object
+- output
+  - Nothing if validation succeeded, InvalidApplicationIdentity error if it
+   failed
+   
+### Implementation
+
+In conclusion, our general ValidateApplication class will need the following
+inputs:
+  - Authenticator name (e.g 'authn-azure')
+  - Authenticator's service-id (e.g 'prod')
+  - host's annotations
+  - Different lists of annotation names
+    - permitted
+    - required
+    - non-permitted combinations
+  - Authentication request Application Identity object
+  
+It will perform the following steps:
+  1. Build the Host Application Identity object using the:
+     - authenticator name
+     - authenticator service-id
+     - host's annotations
+     - list of permitted annotations
+  2. Verify the Host Application Identity is configured correctly using the:
+     - Host Application Identity object built in the previous step
+     - list of required constraints
+     - list of non-permitted constraints combinations
+  3. Verify the Host Application Identity matches the authentication request
+   using the:
+     - Host Application Identity object built in the first step
+     - Authentication request Application Identity object given in the input
+      to this class
+      
+Let's see how the classes will look like, at a high level. After we go over all the
+classes, we'll see a usage example for the Azure Authenticator.
+
+#### Application Identity class
+
+We will have 2 different Application Identity classes. Although their
+interface is the same, they are not entirely identical. 
+
+At first, the `constraints` method will return a simple key-value hash. We
+should challenge that and find a better way to do it in Ruby.
+
+##### Host Application Identity
+
+```
+class HostApplicationIdentity
+
+  def initialize(authenticator_name:, service_id:, role_annotations:, permitted_constraints:)
+    @authenticator_name = authenticator_name
+    @service_id       = service_id
+    @role_annotations = role_annotations
+    @permitted_constraints = permitted_constraints
+  end
+
+  def constraints
+    @constraints ||= @permitted_constraints.each_with_object({}) do |annotation_name, constraints|
+      constraints[annotation_name] = constraint_value(annotation_name)
+    end
+  end
+
+  private
+
+  # check the `service-id` specific constraint first to be more granular
+  def constraint_value constraint_name
+    annotation_value("#{@authenticator_name}/#{@service_id}/#{constraint_name}") ||
+      annotation_value("#{@authenticator_name}/#{constraint_name}")
+  end
+
+  def annotation_value name
+    annotation = @role_annotations.find { |a| a.values[:name] == name }
+
+    # return the value of the annotation if it exists, nil otherwise
+    if annotation
+      @logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
+      annotation[:value]
+    end
+  end
+end
+```
+
+##### Authentication request Application Identity
+
+The input for this class will be built in each authenticator as it differs
+between them. For simplicity, the class will receive the `constraints` hash
+as an input. We should challenge that to find a better way.
+```
+class AuthenticationRequestApplicationIdentity
+  attr_reader :constraints
+
+  def initialize(constraints:)
+    @constraints = constraints
+  end
+end
+```
+
+#### ValidateApplicationIdentity class
+
+This class will use the classes defined above to validate the Application Identity.
+
+```
+ValidateApplicationIdentity ||= CommandClass.new(
+  dependencies: {
+    host_application_identity_class: HostApplicationIdentity
+  },
+  inputs:       %i(
+    authenticator_name
+    service_id
+    host_annotations
+    permitted_constraints
+    required_constraints
+    non_permitted_combinations
+    authentication_request_application_identity
+  )
+) do
+
+  def call
+    extract_application_identity_from_host
+    verify_application_identity_configuration
+    verify_application_identity_matches_authentication_request
+  end
+
+  def extract_application_identity_from_host
+    @host_application_identity_class.new(
+      authenticator_name: @authenticator_name, 
+      service_id: @service_id, 
+      role_annotations: @host_annotations, 
+      permitted_constraints: @permitted_constraints,
+    )
+  end
+
+  def verify_application_identity_configuration
+    # verify that `@host_application_identity_class.constraints.keys` contains
+    # all members of `@required_constraints` and do not have more than one
+    # member of `@non_permitted_combinations`
+    #
+    # Note: this can be extracted to another CommandClass 
+  end
+
+  def verify_application_identity_matches_authentication_request
+    # iterate over `@host_application_identity_class.constraints` 
+    # and verify each value matches the corresponding one in 
+    # `@authentication_request_application_identity.constraints`
+    #
+    # Note: this can be extracted to another CommandClass 
+  end
+end
+```
+
+A call to this class will be made in the relevant Authenticator class. For
+example, the Azure Authenticator class will look roughly like this:
+ 
+```
+module Authentication
+  module AuthnAzure
+
+    Authenticator = CommandClass.new(
+      dependencies: {
+        <other dependencies>,
+        resource_class:             ::Resource
+        authentication_request_application_identity_class: AuthenticationRequestApplicationIdentity
+        validate_application_identity: ValidateApplicationIdentity.new,
+      },
+      inputs:       [:authenticator_input]
+    ) do
+
+      def call
+        <perform other authentication steps>
+        validate_application_identity
+      end
+
+      private
+      
+      def validate_application_identity
+        @validate_application_identity.call(
+          authenticator_name: @authenticator_input.authenticator_name
+          service_id: @authenticator_input.service_id
+          host_annotations: role.annotations
+          permitted_constraints: %w(
+                                     subscription-id
+                                     resource-group
+                                     user-assigned-identity
+                                     system-assigned-identity
+                                 )
+          required_constraints: %w(
+                                    subscription-id
+                                    resource-group
+                                 )
+          non_permitted_combinations: %w(
+                                         user-assigned-identity
+                                         system-assigned-identity
+                                       )
+          authentication_request_application_identity: @authentication_request_application_identity_class.new(
+            constraints: constraints_from_authentication_request
+          )
+        )
+      end
+
+      def constraints_from_authentication_request
+        return @constraints_from_authentication_request if @constraints_from_authentication_request
+
+        @constraints_from_authentication_request = {
+          subscription_id: xms_mirid.subscriptions,
+          resource_group:  xms_mirid.resource_groups
+        }
+
+        if xms_mirid.providers.include? "Microsoft.ManagedIdentity"
+          @constraints_from_authentication_request[:user_assigned_identity] = xms_mirid.providers.last
+        else
+          @constraints_from_authentication_request[:system_assigned_identity] = @oid_token_field
+        end
+
+        @constraints_from_authentication_request
+      end
+
+      # @xms_mirid_token_field is extracted from the request body
+      def xms_mirid
+        @xms_mirid ||= XmsMirid.new(@xms_mirid_token_field)
+      end
+
+      def role
+        @role ||= @resource_class[role_id]
+      end
+  end
+end
+```
+ 
+## Application Identity stored in the host's ID
+
+As mentioned in the beginning of the doc, we made an assumption that
+Application Identities can be defined only in the host's annotations. 
+However, in the K8s Authenticator it can be done also in the host's ID, to maintain
+backwards-compatibility. For now it is just one authenticator out of two that
+implements the Application Identity method but once we grow and have more
+authenticators (e.g AWS Authenticator) then it will make less sense to
+handle Application Identities defined in host IDs in the general classes.
+
+To solve that, we can perform some manipulation inside the K8s Authenticator.
+Before we call `ValidateApplicationIdentity` and pass into it the host's
+annotations, we will check where the Application Identity is defined (as we
+do [here](/app/domain/authentication/authn_k8s/application_identity.rb) in
+the method `application_identity_in_annotations?`). If it's in the
+annotations then we have no problem and we can pass the annotations to 
+`ValidateApplicationIdentity`. If not, we can take the host ID and build a host
+annotations object from it. For example, if the host ID is 
+`namespace-1/service_account/service-account-1` then we will add to the host
+annotations map 2 fields:
+  - `authn-k8s/namespace: namespace-1`
+  - `authn-k8s/service-account: service-account-1`
+  
+As a bonus, we may even add these annotations permanently to the host so next
+time it tries to authenticate we won't need to do this manipulation. 


design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
+# Application Identity
+
+##### Note
+
+In this doc I talk about authenticating hosts with Conjur. Everything I say
```
sigalsax, line 5: In this doc, I will cover authenticating hosts with Conjur. Everything about authenticating hosts is also relevant to users. For the scope of this document, `host`, `user` or `role` are interchangeable"

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+ about authenticating hosts is relevant also for authenticating users
+ . Whenever I say one of `host`, `user` or `role` it has the same meaning.
```
sigalsax, line 7: This preamble is very useful. Like that you added it

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+  
+## Introduction
+
+In several authenticators we require hosts to contain some data which define
```
sigalsax, line 11: in several or in all? I would list them in (). Example: (authn-k8s, authn-azure, ...)

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+resource characteristics)
+ 
+ For example, in the Azure Authenticator the Application Identity
+  includes:
+   - Subscription ID
+   - Resource Group
+   - System-Assigned Identity (optional)
+   - User-Assigned Identity (optional)
+ 
+A host that will authenticate using the Azure authenticator will be defined
```
sigalsax, line 26: will be defined in its annotations as follows:

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+ as follows:
+ ```
+- !host 
+  id: test-app 
+  annotations: 
+    authn-azure/subscription-id: test-subscription 
+    authn-azure/resource-group: test-group 
+``` 
+
+In the example above, the Application Identity is stored as annotations.
```
sigalsax, line 36: remove this sentence if you implement the previous comment

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+starts with general checks that are relevant to all authenticators, and all
+that needs to be done in order to add a new Status Check is to implement a
+method named `status` in the relevant `Authenticator` class, and create a
+ Command Class that will perform the actual validations. 
+ 
+However, when validating the Application Identity we currently have no
+generic mechanism, as can be seen in the implementations for the [Azure
+Authenticator](/app/domain/authentication/authn_azure/validate_application_identity.rb) and 
+[Kubernetes Authenticator](/app/domain/authentication/authn_k8s/validate_application_identity.rb), where a lot of code is duplicated with
+ minor changes.
```
sigalsax, line 73: why is this on a new line?

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+[Kubernetes Authenticator](/app/domain/authentication/authn_k8s/validate_application_identity.rb), where a lot of code is duplicated with
+ minor changes. 
+
+## Solution
+
+Before we can find the solution, we need to understand how Application
+ Identities are related across authenticators, and how they are different.
+ 
+At this point, to ease the problem, I will act as if Application Identities
+can be defined only in the host's annotations, and not in the ID. In future
```
sigalsax, line 81: For the sake of simplicity, from here and throughout the doc, Application Identities will be defined only in the host's annotations, and not in the ID.

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+          annotation[:value]
+    end
+end
+```
+
+The class gets the host's annotations and the authenticator's service-id, 
+and then builds a `constraints` hash from the annotations. 
+
+This is the case also in [`authn-k8s`](/app/domain/authentication/authn_k8s/application_identity.rb), with only one difference as the K8s Authenticator
+also looks in the host's ID.
```
sigalsax, line 139: why new line?

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+
+The class gets the host's annotations and the authenticator's service-id, 
+and then builds a `constraints` hash from the annotations. 
+
+This is the case also in [`authn-k8s`](/app/domain/authentication/authn_k8s/application_identity.rb), with only one difference as the K8s Authenticator
+also looks in the host's ID. 
+
+Putting that aside we can see that the behavior is exactly the same. 
+In this case the general class will be defined as follows:
+- input
```
sigalsax, line 143: align this line with rest of points

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+  - List of relevant annotations (e.g `[subscription-id, resource-group
+  , system-assigned-identity, user-assigned-identity]
+  ` for `authn
+  -azure`)
+- output
+  - `constraints` hash.
+   
+#### Validate the Application Identity is defined correctly
+
+In the Azure Authenticator and in the K8s Authenticator we have some
```
sigalsax, line 156: remove "some". fluff

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+  Application Identity object.
+  
+In this case the general class will be defined as follows:
+- input
+  - Host Application Identity object
+  - Authentication request Application Identity object
+- output
+  - Nothing if validation succeeded, InvalidApplicationIdentity error if it
+   failed
+
```
sigalsax, line 221: the previous sections are all very very comprehensive. Great job explaining!

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+   failed
+   
+### Implementation
+
+In conclusion, our general ValidateApplication class will need the following
+inputs:
+  - Authenticator name (e.g 'authn-azure')
+  - Authenticator's service-id (e.g 'prod')
+  - host's annotations
+  - Different lists of annotation names
```
sigalsax, line 229: A list of annotation names that can be a combination of any of the following:

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+    - permitted
+    - required
+    - non-permitted combinations
+  - Authentication request Application Identity object
```
sigalsax, line 233: this is the request uri right?

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+     - Host Application Identity object built in the previous step
+     - list of required constraints
+     - list of non-permitted constraints combinations
+  3. Verify the Host Application Identity matches the authentication request
+   using the:
+     - Host Application Identity object built in the first step
+     - Authentication request Application Identity object given in the input
+      to this class
+      
+Let's see how the classes will look, in high level.
```
sigalsax, line 251: "at a high level"

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+  - Authenticator's service-id (e.g 'prod')
+  - host's annotations
+  - Different lists of annotation names
+    - permitted
+    - required
+    - non-permitted combinations
+  - Authentication request Application Identity object
+  
+It will perform the following steps:
+  1. Build the Host Application Identity object using the:
```
sigalsax, line 236: Here do you mean a completely, separate object that will have the attributes listed below?
In other words an object that has an authenticator name, service-id, host-annotations attributes will be dynamically filled based on the incoming request and the list of permitted annotations predefined?
So before runtime, the Azure Authentication Identity object would look like:
```
    authenticator name: null
    authenticator service-id: null
    host's annotations: null
    list of permitted annotation: subscription id, resource group, etc.
```
and at runtime:
```
    authenticator name: authn-azure
    authenticator service-id: oren-azure
    host's annotations: subscription-id: my-great-subscription-id, resource-group: my-resource-group
    list of permitted annotation: subscription id, resource group, etc.
```

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+        end
+      end
+    end
+```
+
+##### Authentication request Application Identity
+
+The input for this class will be built in each authenticator as it differs
+ between them. For simplicity, the class will receive the `constraints` hash
+  as an input. We should challenge that to find a better way.
```
sigalsax, line 305: what other options do we have? a separate object?

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+                                       )
+          authentication_request_application_identity: @authentication_request_application_identity_class.new(
+            constraints: @token_identity
+          )
+        )
+      end
+
+      def token_identity_from_claims
+        @token_identity = {
+          subscription_id: xms_mirid.subscriptions,
```
sigalsax, line 421: where does `xms_mirid` come from if not as input?

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+          required_constraints: %w(
+                                    subscription-id
+                                    resource-group
+                                 )
+          non_permitted_combinations: %w(
+                                         user-assigned-identity
+                                         system-assigned-identity
+                                       )
+          authentication_request_application_identity: @authentication_request_application_identity_class.new(
+            constraints: @token_identity
```
sigalsax, line 414: should this be
`constraints: token_identity_from_claims`? b/c on first run, token_identity will be null and nothing calls `token_identity_from_claims` so why not trigger the loading of the `token_identity` by replacing with what i suggested unless you intend for it to be different?

sigalsax: or will that be filled as part of the `<perform other authentication steps>` calls?

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,464 @@
...
+  )
+) do
+
+  def call
+    extract_application_identity_from_host
+    verify_configuration
+    verify_application_identity_matches_authentication_request
+  end
+
+  def extract_application_identity_from_host
```
sigalsax, line 340: Correct me if I am wrong but I feel like `extract_application_identity_from_host` is not really representative of what this function is doing. `.new` just initializes the object but doesn't extract anything. Only host_application_identity.constraints "extracts" and checks if the annotations present are permitted

Maybe `initialize_application_identity_from_host`?

design/authenticators/application-identity/application-identity.md
orenbm: yes. i'll add that.

design/authenticators/application-identity/application-identity.md
orenbm: auto-indent. i disabled it.

design/authenticators/application-identity/application-identity.md
orenbm: the input won't have one list that is a combination of the following. It will have 3 lists of annotations.

orenbm: They can also be empty if the authenticator doesn't have such a constraint. Because we iterate over the list then if the list is empty -> no error will be raised.

design/authenticators/application-identity/application-identity.md
orenbm: what do you mean? I am referring the the Application Identity that is extracted from the Authentication request. For example, in `authn-azure` it is the identity that is extracted from the azure access token

design/authenticators/application-identity/application-identity.md
orenbm: No. this class is described above like this:
```
- input
  - authenticator name
  - authenticator service-id
  - annotations
  - List of relevant annotations (e.g `[subscription-id, resource-group
  , system-assigned-identity, user-assigned-identity]
  ` for `authn
  -azure`)
- output
  - `constraints` hash.
```

It will have only one attribute - `constraints`

design/authenticators/application-identity/application-identity.md
orenbm: oh no, i thought you meant something else. I actually don't have yet a better idea but it seems that we can find one if we look for it.

design/authenticators/application-identity/application-identity.md
orenbm: it does but it's not very interesting. I'll add a comment to clarify

design/authenticators/application-identity/application-identity.md
orenbm: you are correct, i missed it. Changed it now.

design/authenticators/application-identity/application-identity.md
orenbm: i don't want to go into this implementation detail.

orenbm: but you are correct :)

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,466 @@
...
+- input
+  - authenticator name
+  - authenticator service-id
+  - annotations
+  - List of relevant annotations (e.g `[subscription-id, resource-group
+  , system-assigned-identity, user-assigned-identity]
+  ` for `authn
+  -azure`)
+- output
+  - `constraints` hash.
```
orenbm, line 150: @jonahx is there a better option than a hash? Please keep this question in mind while reading the rest of the doc and answer it once you're finished.

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,466 @@
...
+      
+Let's see how the classes will look, at a high level.
+
+#### Application Identity class
+
+We will have 2 different Application Identity classes. Although their
+interface is the same, they are not entirely identical. 
+
+At first, the `constraints` method will return a simple key-value hash. We
+should challenge that and find a better way to do it in Ruby.
```
orenbm, line 257: @jonahx any idea?

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,466 @@
...
+        end
+      end
+    end
+```
+
+##### Authentication request Application Identity
+
+The input for this class will be built in each authenticator as it differs
+between them. For simplicity, the class will receive the `constraints` hash
+as an input. We should challenge that to find a better way.
```
orenbm, line 303: @jonahx any idea?

design/authenticators/application-identity/application-identity.md
jvanderhoof: The dangerous way to do this is with `method_missing` (powerful, but treat with great care):
```
➜  dap-wiki git:(master) irb
2.5.1 :001 > class Foo
2.5.1 :002?>     def initialize()
2.5.1 :003?>         @attributes = {
2.5.1 :004 >               bing: 'bang',
2.5.1 :005 >               ping: 'pong'
2.5.1 :006?>           }
2.5.1 :007?>       end
2.5.1 :008?>
2.5.1 :009?>     def method_missing(m, *args, &blk)
2.5.1 :010?>         if @attributes.key?(m.to_sym)
2.5.1 :011?>             @attributes[m.to_sym]
2.5.1 :012?>           else
2.5.1 :013?>             super
2.5.1 :014?>           end
2.5.1 :015?>       end
2.5.1 :016?>   end
 => :method_missing
2.5.1 :017 > foo = Foo.new
 => #<Foo:0x00007f9e74186f40 @attributes={:bing=>"bang", :ping=>"pong"}>
2.5.1 :018 > foo.ping
 => "pong"
2.5.1 :019 > foo.baz
Traceback (most recent call last):
        3: from /Users/jasonvanderhoof/.rvm/rubies/ruby-2.5.1/bin/irb:11:in `<main>'
        2: from (irb):19
        1: from (irb):13:in `method_missing'
NoMethodError (undefined method `baz' for #<Foo:0x00007f9e74186f40>)
```

jvanderhoof:
@orenbm, this feels like a heavyweight solution to a problem.  What's the one-sentence outcome you're hoping to achieve? Is it something like "simple validation of authenticator attributes"?

design/authenticators/application-identity/application-identity.md
```diff
@@ -0,0 +1,468 @@
...
+In a way, we can define this step as comparing 2 objects of Application
+Identity - one that was extracted from the host and one that was extracted
+from the authentication request. In the example above we will have the
+following Application Identity objects:
+ - Host Application Identity
+   - constraints
+     - namespace: `namespace-1`
+ - Authentication request Application Identity
+   - constraints
+     - namespace: `namespace-2`
```
izgeri, line 206: where does the namespace info come from in authn-k8s when an authentication request is sent from an app with a host identity defined using annotations?

previously the host would have authenticated as `{namespace}/service_account/{service account}`; the namespace / service account would be sent with the request in the form of the host ID (`CONJUR_AUTHN_LOGIN`) and the service_id would be extracted from the request route

now when a host authenticates as `my-host` in the `CONJUR_AUTHN_LOGIN`, where the host was defined in policy using the `{service account}` and `{namespace}` values as host annotations, how does the server extract the "authentication request application identity" values for service account and namespace?

sorry if this is more clearly defined elsewhere, I've just never thought about this aspect of defining hosts by annotations before and your example is raising this question for me

izgeri:
just wanted to comment that this is a very well thought out and clear document, and it clearly will be valuable to future authenticators to have this centralized logic in place. nice work identifying the issue and drafting this proposal @orenbm!

jonahx:
Going to add one long comment with quotes from original instead of inline
GH comments.

## General Comments

This is thorough and attentive to detail.  I appreciate the effort that went
into this, and I'd very much like to see more features developed in this way:
With a thoughtful writeup, and lots of time for iterative feedback from other
devs.

My overall impression is that the design feels too complex what the feature,
essentially, is.  I spent a long time thinking about and examining existing
code to try to figure out why.

## Specifics

I'll start with a few comments about places where the writing of the document
itself wasn't clear enough to me.  This may seem unrelated to the technical
design, but I think they are actually connected.

So first, I want the "Introduction" to frame the general problem more clearly.
Both what it is, and how it relates to Conjur authentication.  Instead, we jump
right in to an assumed knowledge of speficic Conjur authenticators:

> In several authenticators, such as authn-k8s & authn-azure, we require hosts  

We're starting at 1K feet high instead of 10K feet high where we want to be.

> The Application Identity gives the authenticator another layer of
> authentication,

Let's be more precise here.  "Authentication" (generally speaking, not just in
Conjur) is how you verify an identity.  In the most basic case, a client enters
a username/password.  The client's knowledge of the password verifies that he
is the true owner of the username (ie, that's his true identity).

So it's confusing to talk about "identity adding a layer of authentication",
since "authentication" is "the process of verifying identity".

I think what you meant in the above quote is that it is essentially an "alias",
or like a secondary "username" you can use when authenticating.

I think we should explicitly define our terms.  Here are some suggestions:

- "identity" -- the actual thing that you are.  I might call you "Oren" or "Mr.
  Ben Meir" but I always refer to the same identity, the actual flesh and blood
  person who _is_ Oren.
- "identifiers" -- any way of referring to an identity.  "Oren" and "Mr. Ben
  Meir" are identifiers. In our current context, I think it the Conjur host
  would be the identity, and it can be referred to by multiple identifers.

Then you could say stuff like "The annotations are used as another way to
_identify_ a Conjur host" or "The annotations are identifiers of the Conjur
host."

> In the example above, both hosts have the same Application Identity. One
> stores it in its annotation and the other in its ID. However, their Conjur
> identity is not the same.

This was confusing to me.  How can the same identifier refer to two different
hosts?

> However, when validating the Application Identity we currently have no
> generic mechanism, as can be seen in the implementations for the Azure
> Authenticator and Kubernetes Authenticator, where a lot of code is duplicated
> with minor changes.

I looked at these files but wasn't able to immediately see the duplicated
code...  I'd like to know exactly what it is, since this will help me
understand the abstraction you're wanting to create here.

> The Application Identity validation can be roughly split into 3 parts:
> 
> - Extract the Application Identity from the host
> - Validate the Application Identity is defined correctly
> - Validate the Application Identity matches the authentication request

```
class ApplicationIdentity
  ...
end
```

Couple of issues here:

First, this class is basically an `AnnotationRepo` -- that is, it's a just a
wrapper over the database that (typically) returns domain obects (in this case,
hashes).  Here the wrapper is doing 2 things: logging, and building the
annotation path for you (by adding the full prefix with the `service_id`).

But an `AnnotationRepo` is not an abstraction you'd repeat for every 
type of application identity.  You'd just have one of them.

> Validate the Application Identity is defined correctly

`ValidateAzureAnnotations` shoudn't be a `CommandClass`.

Remember, `CommandClass` should be used to implement "use case objects"
(sometimes called "service" objects).  It's where the logic of multiple things
working together lives, so typically you'll only need one when you have 2 or
more dependencies.

Really `ValidateAzureAnnotations` has only pure functional logic.  It can be impleented as a pure value object or pure function: 

You give it some annotations, it returns if they're valid or not.

It can include validation messages, of course, but it won't actually raise any
errors or do any logging.  Just pure input/output.  The error and logging would
be done by the `CommandClass` that calls it.

Here's the big lesson: Pure functions / value objects are the easiest things to
understand and reason about.  Not to mention test (no mocks needed like  there
are when testing a `CommandClass`!).  Therefore we want to push as much of our
logic into them as possible.

The `CommandClass` pieces which wire things together, then, can remain lean and
just delegate to our pure functional code.

> The following lists of annotation names
> - permitted
> - required
> - non-permitted combinations

I wouldn't abstract on this quite yet.  Future authenticators may have
requirements that don't fit into this pattern.  Instead, per above: Make the
API just "take in annotation names, return a validation result".

> Validate the Application Identity matches the authentication request

Before moving on here, let's retrospect on what's happened so far.  Because
conceptually, it's interesting.  Really all we've done is pull a bunch of 
stuff out of our own database and validate it.

Now that's odd... you might wonder why do we have to validate our own data?
Don't you normally validate data _before_ you stick in a DB?

What's happening is here is that we're missing data types.  We're using
free-form annotations as a way to create ad-hoc types that might otherwise be
stored in their own tables.

What are those types?

They are different _kinds_ of hosts.  We don't just have _hosts_.  We now have
_Azure hosts_ and _OIDC hosts_ and so on.  But they are currently  _invisible
types_... they exist only implicitly.  

We very likely want a design that makes these explicit.

Something like this...

A generic `HostRepo` (see my comments above on `AnnotationRepo`), which houses 
the generic annotation logic (just fetching them), and then an `AzureHostRepo` and `OIDCHostRepo`, etc... whose only job is to query the database and return  
fully-formed, validated `AzureHost` and `OIDCHost` objects that work just like 
we want.  Or return an error, if the requesed host cannot be coerced into one of  these types.

Because that's what all this application identity validation is really about...
that's the underlying concept here.  Is the host id you're asking for actuall
an _Azure_ host, or are its annotations inconsistent with that type of host?

Now when we validate the request the implementation detail of looping over
constraints can go away...

What you're really asking is something like: `AzureHost.matches?(request)`

This is just the high-level idea, but think that abstraction would simplify the
code a lot and match how we're naturally thinking about things.

design/authenticators/application-identity/application-identity.md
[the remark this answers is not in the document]
orenbm: the duplicated code is (showing snippets from azure but they appear also in k8s):
```
def initialize(role_annotations:, service_id:, logger:)
        @role_annotations = role_annotations
        @service_id       = service_id
        @logger           = logger
      end

      def constraints
        @constraints ||= {
          subscription_id:          constraint_value("subscription-id"),
          resource_group:           constraint_value("resource-group"),
          user_assigned_identity:   constraint_value("user-assigned-identity"),
          system_assigned_identity: constraint_value("system-assigned-identity")
        }.compact
      end

# check the `service-id` specific constraint first to be more granular
      def constraint_value constraint_name
        annotation_value("authn-azure/#{@service_id}/#{constraint_name}") ||
          annotation_value("authn-azure/#{constraint_name}")
      end

      def annotation_value name
        annotation = @role_annotations.find { |a| a.values[:name] == name }

        # return the value of the annotation if it exists, nil otherwise
        if annotation
          @logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
          annotation[:value]
        end
      end

# validating that the annotations listed for the Conjur resource align with the permitted Azure constraints
      def validate_azure_annotations_are_permitted
        validate_prefixed_permitted_annotations("authn-azure/")
        validate_prefixed_permitted_annotations("authn-azure/#{@service_id}/")
      end

      # check if annotations with the given prefix is part of the permitted list
      def validate_prefixed_permitted_annotations prefix
        @logger.debug(LogMessages::Authentication::ValidatingAnnotationsWithPrefix.new(prefix))

        prefixed_annotations(prefix).each do |annotation|
          annotation_name = annotation[:name]
          next if prefixed_permitted_constraints(prefix).include?(annotation_name)
          raise Err::ConstraintNotSupported.new(annotation_name.gsub(prefix, ""), permitted_constraints)
        end
      end

      def prefixed_annotations prefix
        @role_annotations.select do |a|

            # verify we take only annotations from the same level
            annotation_name.split('/').length == prefix.split('/').length + 1
        end
      end

      # add prefix to all permitted constraints
      def prefixed_permitted_constraints prefix
        permitted_constraints.map { |k| "#{prefix}#{k}" }
      end

      def permitted_constraints
        @permitted_constraints ||= %w(
          subscription-id resource-group user-assigned-identity system-assigned-identity
        )
      end

 def validate_constraint_exists constraint
        unless application_identity.constraints[constraint]
          raise Err::RoleMissingConstraint, annotation_type_constraint(constraint)
        end
      end

 # validates that the application identity doesn't include logical constraint
      # combinations (e.g user_assigned_identity & system_assigned_identity)
        identifiers = %i(user_assigned_identity system_assigned_identity)

        identifiers_constraints = application_identity.constraints.keys & identifiers
        unless identifiers_constraints.length <= 1
          raise Errors::Authentication::IllegalConstraintCombinations, annotation_type_constraints(identifiers_constraints)
        end
      end

def validate_token_identity_matches_annotations
        application_identity.constraints.each do |constraint|
          annotation_type  = constraint[0].to_s
          annotation_value = constraint[1]
          unless annotation_value == @token_identity[annotation_type.to_sym]
            raise Err::InvalidApplicationIdentity, annotation_type_constraint(annotation_type)
          end
        end
        @logger.debug(Log::ValidatedApplicationIdentity.new)
      end

      def annotation_type_constraints constraints
        constraints.map { |constraint| annotation_type_constraint(constraint) }
      end

      # converts the constraint to be in annotation style (e.g resource-group
      # instead of resource_group) to enhance supportability
      def annotation_type_constraint constraint
        constraint.to_s.tr('_', '-')
      end
```

This is pretty much the whole code of authn-azure's application identity validation and it is implemented also in authn-k8s

orenbm:
> @orenbm, this feels like a heavyweight solution to a problem. What's the one-sentence outcome you're hoping to achieve? Is it something like "simple validation of authenticator attributes"?

@jvanderhoof We already have this validation in Conjur, but each authenticator has its on implementation for it although the code is identical in a huge part of it. 

What I'm hoping to achieve is a generic mechanism to validate the application identity so that future authenticators will have minimal implementation, and it can leverage from the general code.

This is the same as we have in authenticators, where there is a general flow (check the security, audit, create an access token) and a specific part that verifies the credentials.

design/authenticators/application-identity/application-identity.md
orenbm: As can be seen in [our docs](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/Latest/en/Content/Integrations/Kubernetes_AppIdentity.htm?Highlight=application%20identity#DefineKubernetesResourcesasDAPhosts) the namespace can be defined in the annotation `authn-k8s/namespace`. 

the host `{namespace}/service_account/{service account}` can be defined as follows:
```
- host 
  id: some-host
  annotations:
    - authn-k8s/namespace: {namespace}
    - authn-k8s/service-account: {service account}
```

The extraction of the values from the request stayed the same, we didn't touch that logic.

orenbm:
thanks @jonahx for the thorough response. I'll comment here regarding your comments:

> So first, I want the "Introduction" to frame the general problem more clearly.

Noted. I assumed some knowledge regarding our authenticators and jumped into introducing the Application Identity. @sigalsax  said that she had an easy time joining my train of thought but she is more familiar with this area.

> I think what you meant in the above quote is that it is essentially an "alias",
or like a secondary "username" you can use when authenticating.

That is actually incorrect. The Application Identity is a layer of authorization (as opposing to a layer of authentication i called it earlier). It limits the host's authentication ability. For example, the host:
```
   - !host
      id: some-host
      annotations:
        authn-k8s/namespace: some-namespace
```
can authenticate to Conjur using the authn-k8s, only from the namespace `some-namespace` and not from another.

The Application Identity is not an alias or a secondary username, but a layer of authorization. I think that it is clearer now how 2 Conjur hosts can have the same Application Identity:
 ```
- !host
  id: test-app-service-account
  annotations:
    authn-k8s/namespace: some-namespace
    authn-k8s/service-account: some-service-account

- !host
  id: some-namespace/service_account/some-service-account
```
One defines it in the ID and one defines it in the annotations but both hosts can authenticate with Conjur from a pod in the namespace `some-namespace` and that has the service-account `some-service-account`.

> I looked at these files but wasn't able to immediately see the duplicated
code

I addressed this question [here](https://github.com/cyberark/conjur/pull/1479#discussion_r426560275). As you can see it is most of the code. The only difference is that the k8s authenticator also handles an Application Identity defined in the ID.

> ValidateAzureAnnotations shoudn't be a CommandClass.

I had this debate with myself when we implemented this in authn-azure. It seemed better to split it out at the time but i can't disagree with you. I'll take this note if we implement the generic validator.

> I wouldn't abstract on this quite yet. Future authenticators may have
requirements that don't fit into this pattern. Instead, per above: Make the
API just "take in annotation names, return a validation result".

I don't really understand your suggestion here. Can you elaborate?

> Now that's odd... you might wonder why do we have to validate our own data?
Don't you normally validate data before you stick it in a DB?

What's happening here is that we're missing data types. We're using
free-form annotations as a way to create ad-hoc types that might otherwise be
stored in their own tables.

You are totally correct and we addressed this issue in the past. Ideally, these validations would occur upon policy load. However, we didn't get the time to redesign this and add a new data-type instead of using the current mechanism of annotations.

> They are different kinds of hosts. We don't just have hosts. We now have
Azure hosts and OIDC hosts and so on. But they are currently invisible
types... they exist only implicitly.

While this is conceptually correct, we also have a requirement to let hosts authenticate with multiple authenticators.

File: design/authenticators/application-identity/application-identity.md [hunk 24]
```diff
@@ -0,0 +1,468 @@
...
+Identities are related across authenticators, and how they are different.
+ 
+For the sake of simplicity, from here and throughout the document, Application 
+Identities will be defined only in the host's annotations, and not in the ID. 
+In future authenticators we will not develop an option to define it in the ID and it
+is available in the K8s Authenticator only for backwards compatibility.
+
+### Defining Application Identity Validation
+
+The Application Identity validation can be roughly split into 3 parts:
```
eranha, line 86: It'a actually three responsibilities.

jonahx:
I re-read the k8's `application_identity.rb` to get more context for my answer,
and I noticed a few things...

You had asked me a while back what you could do to improve:

```ruby 
def constraints
  @constraints ||= {
    pod:               constraint_value("pod"),
    service_account:   constraint_value("service_account"),
    deployment:        constraint_value("deployment"),
    deployment_config: constraint_value("deployment_config"),
    stateful_set:      constraint_value("stateful_set")
  }.compact
end
```

You'd complained that, since the attributes being returned could be different
depending on the host, you had to use a hash rather than a fixed object.  That
is, you recognized that it wasn't returning a _fixed type_ and that (correctly)
didn't sit right with you.

The answer to your question is found in how we're _using_ the above method.  I
did a "Find Usages" in RubyMine, and there were only 2 relevant non-spec
results.  Here it is being used in `ValidateApplicationIdentity`:

```ruby
def validate_pod_metadata
  application_identity.constraints.each do |constraint|
    resource_type   = constraint[0].to_s
    resource_name   = constraint[1]
    resource_object = k8s_resource_object(
      resource_type,
      resource_name,
      application_identity.namespace
    )
```

That is, conceptually, `constraints` really returns an a list of K8s resources.
That's your missing object.  If you add `K8sResource` value object,
`constraints` can have a fixed return type: `[<K8sResource>]` and the method
`constraints` can be renamed to reflect that.

Note: You'll need to create a new value type for `K8sResource` since the
existing `k8s_resource_object` is something else: It interacts with the K8s API
and returns results from that.

Additionally, I think you can probably rewrite that `ApplicationIdentity` class
and remove 1/3 or more of the code.

I would suggest doing this before contining the "Application Identity"
abstraction being proposed here.  As I mentioned above, "Application Identity"
is not really what this is about.

My suggestion is to return to first principles:

1. Write down in a plain English bullet list, at a high level, what validations
we're doing in `ApplicationIdentity`.
2. Make sure that code -- both the names, the number methods, and their
structure -- reflect what you wrote in step 1.

Remember the _goal_ is not to have many one-line methods.  The goal is to make
your high-level logic clear and suppress extraneous detail.  One line private
methods can often help with this. But if... 

1. They aren't named carefully or 
2. If they're not broken down at the right level of granularity or
3. If keeping the call chain in your head becomes harder than following a single, longer method body 

Then it's time to rethink.

Getting back to the main project now... My advice here is more speculative 
since I'm not 100% sure I understand all the requirements...

Given my current understanding, I think what all this is really about is
validating hosts/applications.  Different hosts have different  sets of
properties, which are stored in annotations (or paths) and we need to validate
them after we pull them out of the database.  The basic pattern (correct
me if I'm wrong here) is something like:

1. Query the database using a host_id (or other application identifer?)
2. Get back a bunch of annotations
3. Validate those annotations are valid since we don't do so when inserting
   them which we should but that's a separate discussion.
4. Turn that set of annotations into a semantically meaningful object like a
   k8s resource (this part we're not doing yet)
5. Do something with those objects (that's not part of the scope of this document)


```ruby
# Alternative name: ApplicationRepo?
class HostRepo
  def k8_application(host_id)
    # 1. Does db query for the annotations
    # 2. Delegates to the validation logic currently in 
    #    ApplicationIdentity
    # 3. Returns [K8sResource] per above
  end

  # Same for azure and others
end
```

Then `ApplictionIdentity` get renamed to something like `K8Application` or
`K8Host` or `K8sResources` (I'm a little hazy on the correct terminology here,
but I'm guessing a K8sApplication is made up of a set of K8s resources?).  That
is, we create our K8sApplication  from the annotations, and they get validated
as part of that creation:

```ruby
class K8sApplication
  def initialize(host_id:, host_annotations:, service_id:)
    # etc...
  end
end
```

orenbm:
thanks @jonahx ! please review [this PR](https://github.com/cyberark/conjur/pull/1580). I think that after we merge it, it will be much easier to generalize the validations.

orenbm:
this effort is at a halt at the moment and may be continues in the future. Until then, i am closing this PR

Issue #1487, opened 2020-04-09
BUG: Pipeline 'Package' steps succeeds when it doesn't produce a valid debian package


micahlee:
The upgrade to Rails 5 includes dependency updates that also need to be updated when we build
the debian package for the DAP appliance. The package step is currently not producing a valid
`.deb` package, but the pipeline step indicates that it is successful.

When the a valid debian package cannot be build, the CI pipeline step should indicate this by failing
with an error.

PR #1494, opened 2020-04-13
Fix Conjur Debian Package

micahlee:
### What does this PR do?

This PR fixes the debian package pipeline step that was silently broken after updating to Bundler 2
with the Rails 5 update.

The commits may be reviewed independently. The package step can now also be easily run on
a local developer workstation with:

```
$ ./package.sh
```

N/A

Connected to #1487

#### Change log

#### Test coverage

#### Follow-on issues

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -1,6 +1,14 @@
+<% require 'securerandom' %>
+
 development:
   secret_key_base: "boop"
 test:
   secret_key_base: "boop"
 production:
-  secret_key_base: "secret_key_base is intentionally not set for this application"
+  # Conjur doesn't use HTTP cookies or Rails key generation,
+  # so we set secret_key_base to a random value on application startup. 
+  secret_key_base: <%= SecureRandom.hex(64) %>
+appliance:
+  # Conjur doesn't use HTTP cookies or Rails key generation,
+  # so we set secret_key_base to a random value on application startup.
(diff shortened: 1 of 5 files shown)


config/secrets.yml
```diff
@@ -1,6 +1,10 @@
+<% require 'securerandom' %>
+
 development:
   secret_key_base: "boop"
 test:
   secret_key_base: "boop"
 production:
-  secret_key_base: "secret_key_base is intentionally not set for this application"
+  secret_key_base: <%= SecureRandom.hex(64) %>
```
jtuttle, line 8: Was the original value causing problems here? I would actually prefer it be set to something like an English sentence that explains WTF Is going on just in case someone ends up trying to use it. Setting it to an actual value makes it look like it's being used.

micahlee: How about a comment instead?

My concern with this is that this doesn't actually "unset" it, it just uses a throwaway value. The Rails 4 solution was acceptably because we could be confident that Rails was not using the value without it raising the error.

With the current solution, if in the future something else in Rails begins to use the `secret_key_base`, we won't know about it, and having it is as a well-known, hardcoded value leaves us at risk.

A comment that we don't intend to use the `secret_key_base`, so we set it to a random value on startup, should address both concerns, right?

jtuttle: Yep, comment sounds good.

Issue #1312, opened 2020-02-09
Challenge the way authenticators store non-sensitive information


orenbm:
At current, we store all information - sensitive and non-sensitive - as variables in the authenticator policy.
For example, in authn-azure the variable `provider-uri` is stored as a variable:
```
- !policy
  id: conjur/authn-azure/prod
  body:
  - !webservice

  - !variable
    id: provider-uri  
      
  - !group apps
  
  - !permit
    role: !group apps
    privilege: [ read, authenticate ]
    resource: !webservice
```

This information is not sensitive. We should consider storing this data as an annotation, as follows:
```
# policy id needs to match the convention `conjur/authn-azure/<service ID>`
- !policy
  id: conjur/authn-azure/prod
  body:
  - !webservice
     annotations:
       provider-uri: <some-value>
      
  - !group apps
  
  - !permit
    role: !group apps
    privilege: [ read, authenticate ]
    resource: !webservice
```

This can improve the UX because adding the provider-uri as a variable in the policy demands later that the admin user populate its value through the CLI. Further, adding it as annotation entry makes it easier to set it as a key-value pair and is therefore more readable.

We currently have it a variable for 2 reasons:
1. This is the way it is already implemented in authn-oidc 
2. In case the value is changed the user needs to load the policy again with a replace flag.

We should challenge if we can change it to annotations.

DOD
- [ ] Challenge leads to a decision
- [ ] This decision is updated in the solution-design with explanation

orenbm:
This suggestion adds complexity to the user rather than reducing it. More info can be found in the PR.

PR #1501, opened 2020-04-19
Add design doc for storing authenticators non-sensitive information

orenbm:
Connected to #1312 

It's best to read the design [here](https://github.com/cyberark/conjur/blob/design-authn-data/design/authenticators/storing_non_sensitive_info.md) and add comments in the [`Files changed` tab](https://github.com/cyberark/conjur/pull/1501/files).

In several authenticators we need to store some information for their
authentication process. For example, in `authn-oidc` we have 2 variables
that store this kind of data:
  - provider-uri
  - id-token-user-property
  
These 2 variables are defined in the authenticator policy and are loaded with
values before users can authenticate with the authenticator. However, this data
is not sensitive and was stored in variables just because we didn't think of a 
better solution. As the data is not sensitive, it makes sense to store them in
other methods than in variables, which can improve the UX of the authenticator
configuration and make another simplicity step.

### DIFF
--- a/design/authenticators/storing_non_sensitive_info.md
+++ b/design/authenticators/storing_non_sensitive_info.md
@@ -1,0 +1,331 @@
+# Store Authenticators non-sensitive information
+
+- [Store Authenticators non-sensitive information](#store-authenticators-non-sensitive-information)
+  * [Introduction](#introduction)
+  * [Solution](#solution)
+    + [Authenticator modification](#authenticator-modification)
+    + [Backwards compatibility](#backwards-compatibility)
+  * [Test plan](#test-plan)
+  * [Performance](#performance)
+    + [Configuration](#configuration)
+    + [Authentication](#authentication)
+  * [Security](#security)
+  * [Documentation](#documentation)
+  * [Version Update](#version-update)
+  * [Delivery Plan](#delivery-plan)
+  
+## Introduction
+
+In several authenticators we need to store some information for their
+authentication process. For example, in `authn-oidc` we have 2 variables
+that store this kind of data:
+  - provider-uri
+  - id-token-user-property
+  
+These 2 variables are defined in the authenticator policy and are loaded with
+values before users can authenticate with the authenticator. However, this data
+is not sensitive and was stored in variables just because we didn't think of a 
+better solution. As the data is not sensitive, it makes sense to store them in
+other methods than in variables, which can improve the UX of the authenticator
+configuration and make another simplicity step.
+
+## Solution
+
+The suggested solution is to store this data in the authenticator webservice's
+annotations. 
+
+For example, let's look at a policy that defines an `authn-oidc`
+authenticator:
+
+```
+- !policy
+  id: conjur/authn-oidc/keycloak
+  body:
+  - !webservice
+    annotations:
+      description: Authentication service for Keycloak, based on Open ID Connect.
+
+  - !variable
+    id: provider-uri
+
+  - !variable
+    id: id-token-user-property
+
+  - !group
+    id: users
+    annotations:
+      description: Group of users who can authenticate using the authn-oidc/keycloak authenticator
+
+  - !permit
+    role: !group users
+    privilege: [ read, authenticate ]
+    resource: !webservice
+```
+
+After loading the policy above, the operator will load the variables with values:
+```
+conjur variable values add conjur/authn-oidc/keycloak/provider-uri "some-provider-uri"
+conjur variable values add conjur/authn-oidc/keycloak/id-token-user-property "some-property"
+```
+
+In the suggested approach, the policy will not have variables, and will be defined
+as follows:
+```
+- !policy
+  id: conjur/authn-oidc/keycloak
+  body:
+  - !webservice
+    annotations:
+      description: Authentication service for Keycloak, based on Open ID Connect.
+      provider-uri: some-provider-uri
+      id-token-user-property: some-property
+
+  - !group
+    id: users
+    annotations:
+      description: Group of users who can authenticate using the authn-oidc/keycloak authenticator
+
+  - !permit
+    role: !group users
+    privilege: [ read, authenticate ]
+    resource: !webservice
+```
+
+This improves the UX of the authenticator configuration for the following reasons:
+  - The user doesn't need to load values into each variable so we have fewer steps
+    of authenticator configuration.
+  - The policy shows clearly all the data of the authenticator, comparing to
+    the current policy where the data of the variables is not visible.
+
+### Authenticator modification
+
+What happens if the user configured an authenticator, and then needs to replace
+one of the non-sensitive data objects? In this case, the user will need to reload
+the policy with the `--delete` flag. 
+
+For example, let's look at the following scenario. At first, the policy tree
+is empty:
+```
+root@daf912c19ac3:/# conjur list
+[
+  "myConjurAccount:policy:root"
+]
+```
+
+Then I load the policy:
+```
+- !policy
+  id: some-policy
+  body:
+  - !webservice
+    annotations:
+      annotation-key: annotation-value
+
+  - !variable some-variable
+```
+
+with `conjur policy load root some-policy.yml`. 
+
+When we run `conjur list` we will get:
+```
+root@daf912c19ac3:/# conjur list
+[
+  "myConjurAccount:policy:root",
+  "myConjurAccount:policy:some-policy",
+  "myConjurAccount:webservice:some-policy",
+  "myConjurAccount:variable:some-policy/some-variable"
+]
+```
+
+and when we run `conjur show myConjurAccount:webservice:some-policy` we will get:
+```
+root@daf912c19ac3:/# conjur show myConjurAccount:webservice:some-policy
+{
+  "created_at": "2020-04-19T12:16:50.866+00:00",
+  "id": "myConjurAccount:webservice:some-policy",
+  "owner": "myConjurAccount:policy:some-policy",
+  "policy": "myConjurAccount:policy:root",
+  "permissions": [
+
+  ],
+  "annotations": [
+    {
+      "name": "annotation-key",
+      "value": "annotation-value",
+      "policy": "myConjurAccount:policy:root"
+    }
+  ]
+}
+```
+
+We can see that the webservice has the annotation `annotation-key` with the value 
+`annotation-value`.
+
+Now, for some reason, I decide to replace the value of `annotation-key` to
+`some-other-annotation-value`.
+
+All I need to do is change the policy to:
+```
+- !policy
+  id: some-policy
+  body:
+  - !webservice
+    annotations:
+      annotation-key: some-other-annotation-value
+
+  - !variable some-variable
+```
+
+and load it with `conjur policy load --delete root some-policy.yml`.
+
+Now when we run `conjur list` we will get:
+```
+root@daf912c19ac3:/# conjur list
+[
+  "myConjurAccount:policy:root",
+  "myConjurAccount:policy:some-policy",
+  "myConjurAccount:webservice:some-policy",
+  "myConjurAccount:variable:some-policy/some-variable"
+]
+```
+
+and when we run `conjur show myConjurAccount:webservice:some-policy` we will get:
+```
+root@daf912c19ac3:/# conjur show myConjurAccount:webservice:some-policy
+{
+  "created_at": "2020-04-19T12:16:50.866+00:00",
+  "id": "myConjurAccount:webservice:some-policy",
+  "owner": "myConjurAccount:policy:some-policy",
+  "policy": "myConjurAccount:policy:root",
+  "permissions": [
+
+  ],
+  "annotations": [
+    {
+      "name": "annotation-key",
+      "value": "some-other-annotation-value",
+      "policy": "myConjurAccount:policy:root"
+    }
+  ]
+}
+```
+
+We can see that the annotation `annotation-key` now has the value 
+`some-other-annotation-value`.
+
+Furthermore, if we write the following policy:
+```
+- !policy
+  id: some-policy
+  body:
+  - !webservice
+    annotations:
+      annotation-key: yet-another-annotation-value
+```
+
+without the variable `some-variable`, and load it using the `--delete` flag,
+we can see that the variable is not deleted from the DB:
+```
+root@daf912c19ac3:/# conjur list
+[
+  "myConjurAccount:policy:root",
+  "myConjurAccount:policy:some-policy",
+  "myConjurAccount:webservice:some-policy",
+  "myConjurAccount:variable:some-policy/some-variable"
+]
+```
+
+In conclusion, we can see that we have a method to update authenticators' metadata
+when it is stored in annotations.
+
+### Backwards compatibility
+
+We will need to support authentication using variables as well, for `authn-oidc`
+and `authn-azure`, to maintain backwards-compatibility. This means that for these
+two authenticators we will first look for the values in the annotations and if 
+they are not present we will retrieve their secrets from the DB. 
+
+New authenticators can implement only storing non-sensitive data in the webservice's
+annotations. 
+
+## Test plan
+
+The following tests should be implemented for `authn-oidc` and `authn-azure`:
+
+| **Scenario**                                | **Given**                                                                                 | **When**                                  | **Then**                                                                                                         |
+|---------------------------------------------|-------------------------------------------------------------------------------------------|-------------------------------------------|------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


design/authenticators/storing_non_sensitive_info.md
```diff
@@ -0,0 +1,331 @@
...
+    annotations:
+      description: Group of users who can authenticate using the authn-oidc/keycloak authenticator
+
+  - !permit
+    role: !group users
+    privilege: [ read, authenticate ]
+    resource: !webservice
+```
+
+This improves the UX of the authenticator configuration for the following reasons:
```
rafis3, line 94: @orenbm I wonder if this suggestion solves the right problem. The first UX concern is related to needing a few steps for updating multiple variables. And the second UX concern can be done by bulk fetching all the variables, which Conjur supports today, just not in through the CLI.

If we would add bulk update of variables to Conjur, and added bulk update and retrieval to the Conjur CLI, wouldn't that solve the two authenticator UX concerns that were raised here? We could improve the UX even further with the suggestions I mentioned above.

design/authenticators/storing_non_sensitive_info.md
```diff
@@ -0,0 +1,331 @@
...
+In several authenticators we need to store some information for their
+authentication process. For example, in `authn-oidc` we have 2 variables
+that store this kind of data:
+  - provider-uri
+  - id-token-user-property
+  
+These 2 variables are defined in the authenticator policy and are loaded with
+values before users can authenticate with the authenticator. However, this data
+is not sensitive and was stored in variables just because we didn't think of a 
+better solution. As the data is not sensitive, it makes sense to store them in
```
rafis3, line 28: @orenbm why does it make sense to store them in annotations rather than variables? The variables can store any data, not just sensitive data. It provides a unified location where all the authenticator configuration is located - both sensitive and non-sensitive information would be stored in the same place. In your proposal, the user would have to understand which configuration should be in annotations and which in variables and use different APIs for each. In that aspect, I think the UX is less optimal.

What if we simplified the configuration of authenticators through new commands for the CLI?
For example, to define a new authenticator, call:
```bash
$ conjur config authenticator add authn-oidc/my-idp --provider-uri my-uri --id-token-user-property prop
```

Then to see the authenticator, a user could run:
```bash
$ conjur config authenticator show authn-oidc/my-idp
```
Update the authenticator with:
```bash
$ conjur config authenticator update authn-oidc/my-idp --provider-uri my-new-uri
```

And delete an authenticator with:
```bash
$ conjur config authenticator delete authn-oidc/mu-idp
```

Each command would perform the required policy manipulations behind the scenes. So one can either work with the policies directly as done today, or do it in simple commands through the CLI.

I think there are ways that we can improve the UX even further, without breaking the API and splitting the data into two different storage locations (annotations and variables).

What do you think?

rafis3:
@orenbm, @AndrewCopeland thanks a lot for the initiative of improving our authentications configuration user experience! Please see my comments

design/authenticators/storing_non_sensitive_info.md
orenbm: how is that a better UX? In your suggestion the user still needs to run a command to load the variables so it's still one more step (and a tedious one) in the configuration.

Furthermore, with the webservice annotation we get a visible and readable key-value map of metadata. in your suggestion we don't get that.

I see how your suggestion is better than our current process but not how it's better than the one suggested in this doc (and also - your suggestion will require more dev time).

design/authenticators/storing_non_sensitive_info.md
orenbm: 1. The user doesn't need to understand different APIs. He will load a policy with the non-sensitive data and will load sensitive data into variables. It also makes sense to store sensitive data in variables and non-sensitive data in another data-object, as variables hold secrets by definition.
2. As it's a nice feature to define authenticators via the CLI, it has nothing to do with the suggested approach. If we implement the suggested feature, then the implementation of the CLI command will be slightly different but it won't change anything from the user's perspective. We can do it in the future but I wouldn't treat these 2 features as related or dependant on each other. I also think that the implementation time, and testing time, of your suggestion is A LOT longer than the one suggested in this doc.

design/authenticators/storing_non_sensitive_info.md
```diff
@@ -0,0 +1,331 @@
...
+  - The user doesn't need to load values into each variable so we have fewer steps
+    of authenticator configuration.
+  - The policy shows clearly all the data of the authenticator, comparing to
+    the current policy where the data of the variables is not visible.
+
+### Authenticator modification
+
+What happens if the user configured an authenticator, and then needs to replace
+one of the non-sensitive data objects? In this case, the user will need to reload
+the policy with the `--delete` flag.
```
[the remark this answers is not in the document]
AndrewCopeland: `--delete` looks much more scary than it actually is. It will only explicitly delete conjur resources when you use the `!delete`action within the policy. Everything else will remain present and nothing will be deleted implicitly. `--replace` is what should be used very carefully.

Also a lot of customers typically only use replace (especially for authenticators) so I think `--replace` will be used in the wild rather than `--delete`

design/authenticators/storing_non_sensitive_info.md
AndrewCopeland: Having all of the CLI commands would be beneficial. However I still think that non-sensitive authenticator stored as annotations makes configuring authenticators easier, less room for errors and easier to troubleshoot (On the support side of things).

For OIDC we turn 3 steps into 1 step.
For Azure we turn 2 steps into 1 step.

rafis3: I think we need to find the root cause of a problem, to solve it properly and make our product better. And not be tempted to make shortcuts that create technical debt. If we want to put information as annotation, because currently, it's easier to view them, that is a concern to me. If we want to put information in annotation because we believe that this is where this information absolutely should to be in, that's a different story. But I feel that currently, the answer is the former.

I am concerned that config information of the same instance (an authenticator) would be scattered between variables and annotations. While this saves 1-2 steps of commands, it requires the end user to hold a deeper understanding of the product. The user would need to know which parameter goes where and the user would have to use two different approaches (`update variable` and `load policy`) to perform the same logical action (provide authenticator config data). The user currently loads a policy anyway, however the policy load is logically defining the authenticator, then updating the variables is how you logically set the authenticator configuration. With your offer, the authenticator definition the two logical steps are mixed into 1.5 of the process is done with policy load (definition + half of the authenticator config), then the 0.5 remaining is completed with a variable update of the sensitive config (if exists).

design/authenticators/storing_non_sensitive_info.md
rafis3: I think that my suggestion provides a better UX because it abstracts away from the user, how the authenticator is built and how it is saves in Conjur. The user would not need to understand the policy structure of the authenticators and would not need to understand if a value needs to go into an annotation or a variable. Instead the user would only need to be familiar with a single command that would ask from him/her all the required parameters. So it's a single command that would define and config the authenticator accordingly.

If we want a way to see variable values more easily, then maybe what we should really do, is add an API that gets the policy yaml and if a certain parameter is specified (and if you have permissions to read the variables) an API that would return both the policy and the variable values in it. We don't have to move data from one place to another, just because we like better how it looks, we can improve the UX of our existing tools and make them more convenient and friendly

design/authenticators/storing_non_sensitive_info.md
rafis3: @AndrewCopeland in your comment:

> For OIDC we turn 3 steps into 1 step.
> For Azure we turn 2 steps into 1 step.

You are are reducing the number of API calls, but you are not reducing the number of steps to one, because with the suggestion above, you are adding an `if` step for the user. The user needs to think and understand, where does each parameter go? Is it sensitive or non sensitive? I think that adds complexity and requires more knowledge and familiarity with the product to get right. Same with the retrieval. Instead of getting all the config values in the same way, the user would have to show half of the configuration values with one command and the rest of the configuration values with another.

Lastly, there's also the permissions consideration. Setting configuration in annotations, requires different permissions than updating variables. You would no longer be able to give a user limited permission to only modify an existing authenticator. Every user would have to have create/delete/modify permissions always.

rafis3:
I think what would really help, is a tradeoff analysis. There are pros and cons to every suggestion and both sides need to be reflected in order to create a shared understanding and to create an equal discussion for everyone that wants to be involved.

design/authenticators/storing_non_sensitive_info.md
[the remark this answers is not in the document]
orenbm: thanks Andrew, I'll add more content to the documentation section that explains what `--delete` does. It also replaces any entity that is present in the policy. 

can you elaborate on how customers use replace for authenticators? How does their policy tree look like? In the docs the policy id for an authenticator is `conjur/authn-something/service-id` so if they will load that into root with `--replace` they will remove everything else from `root`. The only way that will work is if they create a policy named `conjur`, a policy named `authn-something` and then a policy named `service-id` and load with `--replace` into it.

design/authenticators/storing_non_sensitive_info.md
orenbm: >  While this saves 1-2 steps of commands, it requires the end user to hold a deeper understanding of the product. The user would need to know which parameter goes where and the user would have to use two different approaches (update variable and load policy) to perform the same logical action (provide authenticator config data)

> with the suggestion above, you are adding an if step for the user. The user needs to think and understand, where does each parameter go? Is it sensitive or non sensitive? I think that adds complexity and requires more knowledge and familiarity with the product to get right.

Same as above. this complexity is not in the customer's side.

> Same with the retrieval. Instead of getting all the config values in the same way, the user would have to show half of the configuration values with one command and the rest of the configuration values with another.

That's incorrect. The customer will never retrieve these secrets. They are defined for the use of the authenticator, which retrieves them during the authentication process. So again, no added complexity for the customer.
That is not correct. _We_ will defined the authenticator and we will decide wether the data is sensitive or not. If a community member will create the authenticator then they will need to decide that. The end-user doesn't need to understand that as the docs will state which data should go in annotations and which data should go in variables. I actually think that it makes sense to tell the user to store metadata in annotations and secrets in variables. 

> Lastly, there's also the permissions consideration. Setting configuration in annotations, requires different permissions than updating variables. You would no longer be able to give a user limited permission to only modify an existing authenticator. Every user would have to have create/delete/modify permissions always.

What do you mean here? Currently for a user to update a configuration he will need `update` permissions on the variable. After this change he will need `update` permissions on the webservice. Do you see a scenario in which a user will get permissions on the variables of the authenticator and not on the webservice? it makes no sense.

rafis3: >The end-user doesn't need to understand that as the docs will state which data should go in annotations and which data should go in variables

Going to the docs is not a great user experience. But running a CLI that completes your required and optional parameters is. Because it relieves you from the need to rely on the documentation and prevents you from making any unnecessary human errors, such as parameter typos.

AndrewCopeland: > Going to the docs is not a great user experience. But running a CLI that completes your required and optional parameters is

To configure/enable any authenticator I expect the user to go to the docs, I do not expect a customer to understand how to configure authenticators without going to the docs. I understand your point of view for having the conjur-cli enable the authenticator but this brings up another concern which is policy as code. If we enable authenticators with the conjur-cli we do not have a record of which authenticators are configured within the policy.

> You are are reducing the number of API calls, but you are not reducing the number of steps to one, because with the suggestion above, you are adding an if step for the user. The user needs to think and understand, where does each parameter go?

I do not see any `if` step here because the docs **should** walk the user through which annotations need to be set and then which variables need to set.

> If we want to put information as annotation, because currently, it's easier to view them, that is a concern to me. If we want to put information in annotation because we believe that this is where this information absolutely should to be in, that's a different story. 

I think it's both, it's easier to view and update also it is where non-sensitive information should reside.

design/authenticators/storing_non_sensitive_info.md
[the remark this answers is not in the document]
AndrewCopeland: @orenbm  It really depends on the customers. Some customers do what you mentioned. They have a `conjur` sub branch and load all authenticators as `--replace` into that subranch. Something like:
`conjur policy load --replace conjur authenticators.yml`. While others have one large policy they load on the root level. That one large policy will have all authenticators and apps and whenever an authenticator needs to be changed/added they will update the root policy and load with `--replace`

sgnn7:
From my point of view, I do think we need an easier way to configure the authenticators (through CLI or similar) but using annotations seems like it hinges on a fact that we don't store sensitive data for our authenticator config there but that becomes a problem with things like the external K8s authenticator which has a mix of sensitive and non-sensitive configuration fields (exactly what @sashaCher was pointing out). Also, how do we decide what is "sensitive" or not? Having the data split between variables and annotations in this case also seems like it would risk user confusion (like @rafis3 was hinting) and I don't know for sure if the benefits would outweigh the costs. 

Anyways, I like that we're thinking more about UX and authenticator configuration definitely needs the attention!

alexkalish:
Wow, some great discussion here.  @orenbm: I have one primary question: is there a business motivation for this change (e.g. is it requested by PM or a customer based on a clear need) or are you looking to improve a poor UX?  Thanks.

orenbm:
> Wow, some great discussion here. @orenbm: I have one primary question: is there a business motivation for this change (e.g. is it requested by PM or a customer based on a clear need) or are you looking to improve a poor UX? Thanks.

It was raised by @AndrewCopeland when we design the Azure authenticator.

sharonr78:
Thanks @orenbm for this great initiative. 
I read this design and loved the idea, so it made me instantly pick your side.
Using annotations for the non-sensitive data does make sense to me.. It improves clarity when the user loads the policy and see the context right away, and the user doesn’t need to load values into variables.
Annotations can be the exact field for such non-sensitive data, but the user will need to learn it is there. Not a tough task to do.

However, after I read all the comments, I found some logical explanations, which made it hard for me to pick a side, especially when this is not a technology I have experience with.

All commenters contradicted the suggestion with some logic answers, like Srdjan, who indicated an issues with both sensitive and non-sensitive data (“Also, how do we decide what is “sensitive” or not? Having the data split between variables and annotations in this case also seems like it would risk user confusion”). Sasha mentioned an issue with coupled properties which might break the authenticator.

I learned that a good UX tries to lower the cognitive load from the user as much as possible and reduce friction, like in multiple options - The more options the user needs to choose from, the harder it will take him to do so. In our case, I’m not sure if this lowers the confusion or might raise new one.

Since I don’t have any experience with Conjur’s variable and policies, and both approaches have their pros and cons, I suggest doing what I’d be doing in such a dilemma - user testing, or A/B testing.
You can prepare a comparison table with both designs (A - using variables, B - using annotations) side by side, its pros and cons, so it will be easy to notice the differences.
Then you can run a poll among the SEs in Slack and get votes for each design. This will give you a quantitive value for your test.
In addition, we can ask our users in the OSS community if this design makes sense and can improve their UX.
Lastly, we can use the Technical Community portal which serves approx. 30K customers and partners and raise this solution.

Unfortunately, I can’t tip the scales on this manner, since there’s no single truth in UX and various approaches may fit. Eventually, we are not the users and only user-testing and feedback loops help creating a lovable and usable product.

orenbm:
Sadly, this proposal was not accepted and was claimed to be adding UX complexity instead of reducing it. Therefore, I'm closing this issue.

orenbm:
Trying to re-focus the conversation on this to get a go/no-go on this feature.

When you answer the following question, please don't refer to the other proposal suggested in this conversation, regarding the CLI addition. The question is this - if we move the authenticator's non-sensitive data from variables to annotations on the webservice, will it increase the UX or decrease it?

One point that I saw in the comments is that users will need to refer to the documentation to understand which data is sensitive and which is not. As @AndrewCopeland stated correctly, users already need to go to the docs to configure authenticators so it doesn't matter. They also don't need to understand if the data is sensitive or not because the docs will state which data should be stored in variables and which in annotations.

@rafis3 @AndrewCopeland @cyberark/conjur-core-team i'd like to get your thoughts on this. thanks :)

rafis3:
I vote against this feature. I think it makes the user experience more complicated.
But I would love to hear other opinions and well, to see where the majority stands.

jonahx:
We should consider very carefully the security implications of this.

Because from a usability perspective, by making this distinction, we introduce a type of user error which was not possible before: namely, that someone mistakenly stores sensitive information in a non-sensitive store.  And this error has the most serious possible consequences.

Whereas if everything stored is secure by default, there is no harm done by securing something which doesn't really need to be secure.

For example, one thing that happens (in both configuration and code) is that rather than understanding something fully, people just copy something else that they see.  Is it possible that an operator sees annotations being used and, assuming they're secure, puts something _else_ sensitive in an annotation.

We really want our product to be bulletproof, even in the face of careless users.

I haven't delved into how plausible this will be in this case, but I just wanted to raise this as something to think about and have a security review of.

orenbm:
Looks like this suggestion adds complexity for the user instead of reducing it. Closing the issue and the PR.

PR #1505, opened 2020-04-23
Add default content type for Policy load API endpoints

micahlee:
### What does this PR do?

If no content-type is specified in a POST request, Rack assumes the content
type is either `application/x-www-form-urlencoded` or
`multipart/form-data`.

The Conjur policy documents are YAML content, and some special
characters allowed in policy cause the Rack pody parsing to fail, leading to
a Rack error: `Invalid or incomplete POST parameters`.

This PR adds a Rack middleware that allow default content-types to be
configured on a per request path basis to better control the behavior when
no content-type is provided in the request headers.

Connected to conjurinc/dap-support#74

#### Change log

#### Test coverage

#### Follow-on issues

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -60,6 +60,12 @@ class ApplicationController < ActionController::API
 
   around_action :run_with_transaction
 
+  # sets the default content type header on incoming requests that match the
+  # path_match regex
+  def self.set_default_content_type_for_path(path_match, content_type)
+    ::Rack::DefaultContentType.content_type_by_path[path_match] = content_type
+  end
+
   private
 
   # Wrap the request in a transaction.
--- a/lib/rack/default_content_type.rb
+++ b/lib/rack/default_content_type.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+require 'forwardable'
+
+module Rack
+
+  # DefaultContentType is a Rack middleware that supports
+  # providing a default Content-Type header for requests
+  # to specific paths that do not include a Content-Type
+  # header in the request.
+  #
+  # Default content types are configured by adding entries
+  # to the Rack::DefaultContentType.default_content_types
+  # class member hash.
+  #
+  # Each hash key must be a regular expression to test
+  # request paths for a match. The hash value is the content
+  # type string to add to the request.
+  #
+  class DefaultContentType
+
+    # A Hash-like whose keys are Regex objects. It looks up values by matching
+    # keys given to its #[] method against those regexes, and returning the
+    # value of the first match.
+    class RegexHash
+      extend Forwardable
+
+      # Note: this will work with a Hash or keyword args
+      def initialize(**kwargs)
+        @vals_by_regex = kwargs
+      end
+
+      def [](key)
+        matching_re = @vals_by_regex.each_key.find { |re, _| re =~ key }
+        @vals_by_regex[matching_re] # Note: returns nil if no match
+      end
+
+      def_delegators :@vals_by_regex, :[]=
+    end
+
+    def self.content_type_by_path
+      @content_type_by_path ||= RegexHash.new
+    end
+
+    def initialize(app)
+      @app = app
+    end
+
+    # :reek:DuplicateMethodCall
+    def call(env)
+      # If a content type is already present on the request, we
+      # don't need to do anything
+      if content_type_missing?(env) && default_content_type(env)
+        env['CONTENT_TYPE'] = default_content_type(env)
+        Rack::Request.new(env).body.rewind
+      end
+
+      @app.call(env)
+    end
+
+    # :reek:UtilityFunction
+    # :reek:NilCheck
+    def content_type_missing?(env)
+      env['CONTENT_TYPE']&.empty?
+    end
+
+    # :reek:UtilityFunction
+    def default_content_type(env)
+      Rack::DefaultContentType.content_type_by_path[env['PATH_INFO']]
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Title of status page is now `Conjur Status` again, rather than only
   `Conjur` ([conjurinc/dap-support](https://github.com/conjurinc/dap-support/issues/75)).
+- Policy load API endpoints now default to the `application/x-yaml` content-type
+  if no content type is provided in the request
+  ([conjurinc/dap-support#74](https://github.com/conjurinc/dap-support/issues/74)).
 
 ### Changed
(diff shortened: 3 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -60,6 +60,23 @@ class RecordExists < Exceptions::RecordExists
 
   around_action :run_with_transaction
 
+  # path_default_content_type configures default content-types from within
```
jtuttle, line 63: This is very informative, but I think you can convey enough information to be useful with a much shorter comment. The params are named such that they're fairly obvious and if I need an example I can grep for one in the code or look up the gem easily enough. Comment could be something like:

```
# sets the default content type header on incoming requests for controller methods
# that match the path_match regex
```

app/controllers/application_controller.rb
```diff
@@ -60,6 +60,23 @@ class RecordExists < Exceptions::RecordExists
...
+  # - path_match: Regular expression to match an incoming request path against.
+  # - content_type: String containing the content-type to use for a request
+  #                 to the matching path that does not contain a content-type
+  #                 header.
+  #
+  # ==== Examples:
+  #
+  # path_default_content_type /^\/policies/, 'application/x-yaml'
+  #
+  def self.path_default_content_type(path_match, content_type)
```
jonahx, line 76: Since this is mutating the variable and not just returning, let's use a verb name:

```
set_default_content_type_for_path
```

lib/rack/default_content_type.rb
```diff
@@ -0,0 +1,59 @@
...
+  #
+  # Each hash key must be a regular expression to test
+  # request paths for a match. The hash value is the content
+  # type string to add to the request.
+  #
+  # :reek:ClassVariable
+  # rubocop:disable Style/ClassVars
+  class DefaultContentType
+    def self.default_content_types
+      @@default_content_types ||= {}
```
jonahx, line 22: You should always avoid `@@`.  And indeed it should be superfluous here.  The receiving object, because of the `self.` qualifier, is the class itself.  So `@default_content_types` would refer to a variable on the class object, not one of its instances.

lib/rack/default_content_type.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      # If a content type is already present on the request, we
+      # don't need to do anything.
+      env = update_env(env) unless env['CONTENT_TYPE'].present?
```
jonahx, line 40: Since you mutating the environment, it's clearer to write:
```ruby
update_env(env) unless env['CONTENT_TYPE'].present?
```

lib/rack/default_content_type.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      @app.call(env)
+    end
+
+    private
+
+    # :reek:FeatureEnvy
+    # :reek:UtilityFunction
```
jonahx, line 48: This reek comments look valid and related to my comment above about `@@`.  I'll try post a solution that avoid all this in a few minutes.

jonahx:
```ruby
=begin

Currently we're doing:

    config.middleware.use ::Rack::DefaultContentType
  end

Instead we'll do:

    config.middleware.use(
      ::Rack::DefaultContentType,
      { %r{^\/policies} => 'application/x-yaml'}
    )
  end

And you can delete:

  def self.path_default_content_type(path_match, content_type)
    ::Rack::DefaultContentType.default_content_types[path_match] = content_type
  end

And replace it with a comment that the configuration is done in
config.application.rb, just so readers know there is some magic
afoot.

=end

module Rack


  # DefaultContentType is a Rack middleware that supports
  # providing a default Content-Type header for requests
  # to specific paths that do not include a Content-Type
  # header in the request.
  #
  # Default content types are configured by adding entries
  # to the Rack::DefaultContentType.default_content_types
  # class member hash.
  #
  # Each hash key must be a regular expression to test
  # request paths for a match. The hash value is the content
  # type string to add to the request.
  class DefaultContentType

    # A Hash-like whose keys are Regex objects. It looks up values by matching
    # keys given to its #[] method against those regexes, and returning the
    # value of the first match.
    class RegexHash
      # Note: this will work with a Hash or keyword args
      def initialize(**kwargs)
        @vals_by_regex = kwargs
      end

      def [](key)
        matching_re = @vals_by_regex.each_key.find { |re, v| re =~ key }
        @vals_by_regex[matching_re] # Note: returns nil if no match
      end
    end

    def initialize(app, content_type_by_path)
      @app = app
      @content_type_by_path = RegexHash.new(content_type_by_path)
    end

    def call(env)
      # Note: Avoid Rails dependency -- don't use `.present?`
      content_type_missing = env['CONTENT_TYPE']&.empty?
      default_content_type = @content_type_by_path[env['PATH_INFO']]

      # If a content type is already present on the request, we
      # don't need to do anything.
      if content_type_missing && default_content_type
        env['CONTENT_TYPE'] = default_content_type
        Rack::Request.new(env).body.rewind
      end

      @app.call(env)
    end
  end

end
```

lib/rack/default_content_type.rb
micahlee: Interesting... I hadn't realized there was such a thing as a "class instance variable" distinct from a "class variable." Cool!

lib/rack/default_content_type.rb
micahlee: > Avoid Rails dependency -- don't use `.present?`

I fine with the change, but we're in a Rails app. Why are we avoiding using it?

micahlee: > And replace it with a comment that the configuration is done in
config.application.rb, just so readers know there is some magic
afoot.

I think I get where you're coming from with this suggestion. I still think, however, it's going to be easier to discover, understand, and maintain if the middleware only provides the application-wide capability to set default content types, with the path-specific settings configured closer to the path-specific logic (either the routes or the controller).

A global config with a comment is a weak connection with it comes to understanding what is happening when debugging later.

micahlee:
Thanks @jtuttle and @jonahx!

This is ready for another look when one of you has time again. Thanks!

lib/rack/default_content_type.rb
jonahx: > I fine with the change, but we're in a Rails app. Why are we avoiding using it?

Because this is middleware, not part of the Rails app.

jonahx: > A global config with a comment is a weak connection with it comes to understanding what is happening when debugging later.

I partially agree, but you're really choosing between lesser evils.  

The downside of the solution you merged is that you're using mutation, and allowing the middleware to be mutated at runtime, from anywhere, at any time.  This can give rise much trickier bugs than an immutable middleware that is statically configured one time.  So yes, the way I originally suggested has the downside that the comment is deleted, and then it's not clear by looking locally in the controller what's happening.  But I think this is unlikely to happen, especially if the comment says "Do not delete" :)

PR #1508, opened 2020-04-24
Prevent image from rebuilding in tests

jonahx:
Speeds up the CI by about 4 minutes by preventing unnecessary docker image reubilds in the tests

#1511

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -8,14 +8,14 @@
 . version_utils.sh
 
 TAG="$(version_tag)"
-RUN_DEV=true
+jenkins=false # Running on Jenkins (vs local dev machine)
 
 while [[ $# -gt 0 ]]
 do
 key="$1"
 case $key in
     --jenkins)
-    RUN_DEV=false
+    jenkins=true
     ;;
     *)
     ;;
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -44,12 +44,20 @@ function flatten() {
...
 }
 
-echo "Building image conjur:$TAG"
-docker build -t "conjur:$TAG" .
-flatten "conjur:$TAG"
+image_not_built() {
+  [[ "$(docker images -q "$1" 2> /dev/null)" == "" ]]
+}
+
+if image_not_built "conjur:$TAG"; then
```
micahlee, line 51: There's a risk here that attempting to rebuild the image locally after change the code won't work unless the TAG (namely git commit hash) has also changed.

We may want to explicitly log if we are skipping the build because the image is up to date to make this more obvious at a glance.

micahlee: Another option could be to incorporate the `--jenkins` flag into this condition, so this primarily affects the CI nodes, rather than local dev (where `build.sh` happens less frequently and more intentionally).

jonahx: Great point.  I like the latter suggestion.

PR #1509, opened 2020-04-27
Design proposal for a simplifying authenticators management in Conjur

rafis3:
Following the [proposal](https://github.com/cyberark/conjur/pull/1501) given by @orenbm and @AndrewCopeland, I wanted to propose another alternative for simplifying the user experience around authenticators management.

The proposal suggests creating new CLI commands that would support adding/updating/deleting authenticators, as well as enabling and disabling them. Each in a single step, instead of the multiple steps that are required today.

### DIFF
--- a/design/proposals/authenticator_management_improvement.md
+++ b/design/proposals/authenticator_management_improvement.md
@@ -1,0 +1,266 @@
+# Authenticators Management Improvement Proposal
+
+- [Authenticators Management Improvement Proposal](#authenticators-management-improvement-proposal)
+  - [Introduction](#introduction)
+  - [Proposed Solution](#proposed-solution)
+    - [Add Authenticator](#add-authenticator)
+      - [Functionality](#functionality)
+      - [Synopsis](#synopsis)
+      - [Usage Example](#usage-example)
+      - [API Specification](#api-specification)
+    - [Delete Authenticator](#delete-authenticator)
+      - [Functionality](#functionality-1)
+      - [Synopsis](#synopsis-1)
+      - [Usage Example](#usage-example-1)
+      - [API Specification](#api-specification-1)
+    - [Update Authenticator](#update-authenticator)
+      - [Functionality](#functionality-2)
+      - [Synopsis](#synopsis-2)
+      - [Usage Example](#usage-example-2)
+      - [API Specification](#api-specification-2)
+    - [Enable Authenticator](#enable-authenticator)
+      - [Functionality](#functionality-3)
+      - [Synopsis](#synopsis-3)
+      - [Usage Example](#usage-example-3)
+      - [API Specification](#api-specification-3)
+    - [Disable Authenticator](#disable-authenticator)
+      - [Functionality](#functionality-4)
+      - [Synopsis](#synopsis-4)
+      - [Usage Example](#usage-example-4)
+      - [API Specification](#api-specification-4)
+  - [Implementation Details](#implementation-details)
+  - [Backwards Compatibility](#backwards-compatibility)
+  - [Performance](#performance)
+  - [Security](#security)
+  - [Documentation](#documentation)
+  - [Version Update](#version-update)
+  - [Delivery Plan](#delivery-plan)
+  
+## Introduction
+
+Conjur authenticators are defined as sub-policies, under the `conjur` policy.  
+Currently to define an authenticator, the user needs to complete the following steps:
+
+- Read the documentation
+- Understand how the authenticator policy is structured
+- Create a policy yaml file that defines the authenticator
+- Load the policy using Conjur CLI
+- Fill the authenticator policy variables with their values
+- Enable the authenticator through an environment variable or REST API call
+
+To delete an authenticator, the user needs to complete these steps.  
+Either:
+
+- Create a policy with a `!delete` statement, on the authenticator policy
+- Load the policy using Conjur CLI
+
+Or:
+
+- Update the `conjur` policy yaml file and take out the authenticator
+- Reload the `conjur` policy using the `--replace` option  
+
+The current user experience requires a certain level of expertise. The user needs to understand the policy structure of each authenticator and the process is comprised of multiple steps. This proposal offers an easier way for managing authenticators.
+
+## Proposed Solution
+
+The proposed solution is to add a few simple Conjur CLI commands (and possibly APIs, see two alternatives below), that would hide the complexity and provide a single step to perform each action:
+
+### Add Authenticator
+
+#### Functionality
+
+Defines a new authenticator in Conjur. Behind the scenes, this command will creat the authenticator policy and fill it with all its required data. If the data is not required to be given by the user, like the `authn-k8s` CA cert and key, the information will be generated automatically.
+
+#### Synopsis
+
+conjur config authenticator add *authenticator-type/service-id* [*parameters*]
+
+#### Usage Example
+
+```shell script
+$ conjur config authenticator add authn-oidc/my-idp --provider-uri my-uri --id-token-user-property prop
+```
+
+#### API Specification
+
+- **URL**
+
+    `/:authenticator/:service_id/:account`
+
+    Example: `/authn-oidc/my-idp/my-company`
+
+- **Method:**
+
+    `POST`
+
+- **URL Parameters:**
+
+  - `:authenticator` - The authenticator type (e.g. authn-k8s, authn-oidc).
+  - `:service_id` - The name of the specific authenticator instance to be created.
+  - `:account` - The Conjur account in which to create the authenticator.
+
+- **Return Value:**
+  
+  The API will return the appropriate HTTP code. If the operation was successful, it will also return the policy yaml that was loaded for the newly created authenticator.
+
+### Delete Authenticator
+
+#### Functionality
+
+Deletes an existing authenticator in Conjur. Deletes the authenticator policy and deletes the authenticator from the enabled/disabled authenticators list.
+
+#### Synopsis
+
+conjur config authenticator delete *authenticator-type/service-id*
+
+#### Usage Example
+
+```shell script
+$ conjur config authenticator delete authn-oidc/my-idp
+```
+
+#### API Specification
+
+- **URL**
+
+    `/:authenticator/:service_id/:account`
+
+    Example: `/authn-oidc/my-idp/my-company`
+
+- **Method:**
+
+    `DELETE`
+
+- **URL Parameters:**
+
+  - `:authenticator` - The authenticator type (e.g. authn-k8s, authn-oidc).
+  - `:service_id` - The name of the specific authenticator instance to be deleted.
+  - `:account` - The Conjur account in which to delete the authenticator.
+
+- **Return Value:**
+  
+  The API will return the appropriate HTTP code.
+
+### Update Authenticator
+
+#### Functionality
+
+Updates an existing authenticator in Conjur. Updates the authenticator variables.
+
+#### Synopsis
+
+conjur config authenticator update *authenticator-type/service-id* [*parameters*]
+
+#### Usage Example
+
+```shell script
+$ conjur config authenticator update authn-oidc/my-idp --provider-uri new-uri
+```
+
+#### API Specification
+
+- **URL**
+
+    `/:authenticator/:service_id/:account`
+
+    Example: `/authn-oidc/my-idp/my-company`
+
+- **Method:**
+
+    `PATCH`
+
+- **URL Parameters:**
+
+  - `:authenticator` - The authenticator type (e.g. authn-k8s, authn-oidc).
+  - `:service_id` - The name of the specific authenticator instance to update.
+  - `:account` - The Conjur account in which to update the authenticator.
+
+- **Return Value:**
+  
+  The API will return the appropriate HTTP code. If the operation was successful, it will also return the update policy yaml of the modified authenticator.
+
+### Enable Authenticator
+
+#### Functionality
+
+Enables an authenticator in the Conjur database (can be overriden by environment variable).
+
+#### Synopsis
+
+conjur config authenticator enable *authenticator-type/service-id*
+
+#### Usage Example
+
+```shell script
+$ conjur config authenticator enable authn-oidc/my-idp
+```
+
+#### API Specification
+
+This already exists, please see the relevant documentation.
+
+### Disable Authenticator
+
+#### Functionality
+
+Disables an authenticator in the Conjur database (can be overriden by environment variable).
+
+#### Synopsis
+
+conjur config authenticator disable *authenticator-type/service-id*
+
+#### Usage Example
+
+```shell script
+$ conjur config authenticator disable authn-oidc/my-idp
+```
+
+#### API Specification
+
+This already exists, please see the relevant documentation.
+
+## Implementation Details
+
+The new functionality will be developed on the Conjur side. New APIs will be exposed for this new functionality.  
+The CLI will use these new APIs. In order for the CLI to remain up to date with the authenticator types, their required parameters and optional parameters, Conjur will expose an internal API that would provide this authenticator schema information. This API will be called by the CLI whenever this information will be needed. For example, when running `conjur config authenticator add/update {authenticator-type}/{service-id} --help`.
+
+## Backwards Compatibility
+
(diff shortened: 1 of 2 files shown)


AndrewCopeland:
We have a lot of users who do not even use the conjur cli when managing policy. 
What is being proposed is the customer is going to have to:
1. Log into the CLI
2. Enabled the authenticator via the CLI
3. Write policy that links a host to the authenticator
4. Load the policy so host has access to the authenticator.

It seems like the same amount or more effort required since the customer will always need to load policy at the end of the day. However if we were going to do this approach I think creating a API endpoint for the authenticator is needed rather than just creating a policy loading wrapper in the cli.

Also I have brought this up on Oren's original proposal, this goes directly against policy as code. If we needed to re-hydrate the policy from scratch we would then need to load all of the authenticators individually from the cli.

micahlee:
Thanks for sharing, @rafis3!

I also agree that the API capability should be a must for this to keep the information about particular authenticators and what config data they need together in the core Conjur repo, rather than split between Conjur and the CLI where they would need to be added and maintained in both places over time.

Not to mention the API would give greater automation flexibility for those who skip the CLI and use the API directly.

design/authenticators/authenticator_management_improvement.md
```diff
@@ -0,0 +1,247 @@
...
+  
+## Introduction
+
+Conjur authenticators are defined as sub-policies, under the `conjur` policy.  
+Currently to define an authenticator, the user needs to complete the following steps:
+
+- Read the documentation
+- Understand how the authenticator policy is structured
+- Create a policy yaml file that defines the authenticator
+- Load the policy using Conjur CLI
```
orenbm, line 49: It's worth to mention an additional step for most of the authenticators - load values into variables created in the policy

design/authenticators/authenticator_management_improvement.md
```diff
@@ -0,0 +1,247 @@
...
+Defines a new authenticator in Conjur. Behind the scenes, this command will creat the authenticator policy and fill it with all its required data. If the data is not required to be given by the user, like the `authn-k8s` CA cert and key, the information will be generated automatically.
+
+#### Synopsis
+
+conjur config authenticator add *authenticator-type/service-id* [*parameters*]
+
+#### Example
+
+```shell script
+$ conjur config authenticator add authn-oidc/my-idp --provider-uri my-uri --id-token-user-property prop
```
orenbm, line 82: I think that `--provider-uri` is quite error prone as you assume that the only addition that will be is for variables. I'd rather have something like `--metadata provider-uri=my-uri`.

This way you have the flexibility of adding other authenticator parameters in the future and you can encapsulate the step for populating the variable.

design/authenticators/authenticator_management_improvement.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+### Delete Authenticator
+
+#### Functionality
+
+Deletes an existing authenticator in Conjur. Deletes the authenticator policy.
```
orenbm, line 89: it should also disable the authenticator in the DB IMO to complete the task.

design/authenticators/authenticator_management_improvement.md
```diff
@@ -0,0 +1,247 @@
...
+Updates an existing authenticator in Conjur. Updates the authenticator variables.
+
+#### Synopsis
+
+conjur config authenticator update *authenticator-type/service-id* [*parameters*]
+
+#### Example
+
+```shell script
+$ conjur config authenticator update authn-oidc/my-idp --provider-uri new-uri
```
orenbm, line 114: same here. I would use a more generic flag than `--provider-uri`

design/authenticators/authenticator_management_improvement.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+### Enable Authenticator
+
+#### Functionality
+
+Enables an authenticator in Conjur.
```
orenbm, line 121: As we have 2 methods of enablement it's best to specify where this will enable the authenticator. I guess that the DB is the most suitable via the CLI

design/authenticators/authenticator_management_improvement.md
```diff
@@ -0,0 +1,247 @@
...
+#### Example
+
+```shell script
+$ conjur config authenticator disable authn-oidc/my-idp
+```
+
+## Backwards Compatibility
+
+- The authenticators keep their same structure in Conjur. Users could still manage them in the way it's done today.  
+- These CLI commands are uniting policy loads and variable updates into a single step, therefore running the commands requires permissions to both load a policy under the `conjur` policy and to updated the variables in that policy.
```
orenbm, line 152: ... to update ...

design/authenticators/authenticator_management_improvement.md
```diff
@@ -0,0 +1,247 @@
...
+
+No security implications. These new CLI commands perform the same action, with an easier experience.
+
+## Documentation
+
+We will need to update the docs of every authenticator, to specify how to use the new CLI commands. In addition, we will need to add documentation for enabling/disabling authenticators.
+
+## Drawbacks
+
+Every new authenticator that would be created for Conjur, should now have a corresponding Conjur CLI command support as
```
orenbm, line 168: why? Isn't this CLI command generic? All it needs to do is create a policy and add some variables to it.

orenbm:
As a side comment - I think that this proposal can live side-by-side with [my proposal](https://github.com/cyberark/conjur/pull/1501). The only difference is what happens in the background of your commands.

design/authenticators/authenticator_management_improvement.md
rafis3: I don't understand why it's error prone and how your suggestion solves it. Can you please elaborate?
The reason why I suggested authenticator specific flags, is to guide the user to provide the mandatory information, instead of requiring the users to check the docs and understand by themselves what information needs to be provided

design/authenticators/authenticator_management_improvement.md
rafis3: Right, I forgot to mention that step. I'll fix it. Thanks!

design/authenticators/authenticator_management_improvement.md
rafis3: Agree, I forgot that we don't have a cascading delete here. I'll add it.

design/authenticators/authenticator_management_improvement.md
rafis3: I asked you some questions about it. Let's discuss above and decide for both.

design/authenticators/authenticator_management_improvement.md
rafis3: Yes, that was my thought. I'll specify it, it's worth mentioning to create a shared understanding.

design/authenticators/authenticator_management_improvement.md
rafis3: Because if it's generic, it doesn't guide the users to input the right information and it prevents them from misconfiguring. I think there's big value in that user experience.

design/authenticators/authenticator_management_improvement.md
orenbm: what i meant is that if authenticators will have other members than variables then this won't work. However, if you use specific tags then it's ok because the implementation is on our side.

I wouldn't go in that direction though as it's can't be generic and will require implementation for each authenticator (with the need to fix it when we do changes)

orenbm: you suggest that if the `--provider-uri` flag is not present then we will fail the request?

design/authenticators/authenticator_management_improvement.md
orenbm: what will the non-generic implemntation include? Only required parameters?

rafis3:
Thanks for the comments and feedback everyone. I agree with all of you that the better alternative is making the change in Conjur, rather than in the CLI. So I added another recommendation section in the end.

@AndrewCopeland, the proposal doesn't take away all the steps, only the ones related to defining the authenticators. So logging in, defining access control, loading policies, are still required.
Regarding your second point, I don't think it goes against infrastructure as code because of two reasons:
1. You still automate the policy load with a tool. The tool would run API/CLI calls. So instead of running `conjur policy load` and `conjur variable` commands to create an authenticator, you'd run a `conjur config authenticator add`.
1. You can still use the old way, though less recommended, because you'd rather use the abstracted layer rather than updating the raw policy and variables data.

sgnn7:
👍 from me and 👍  on the serverside option. I would like to float an additional idea to this so that we don't need hard-coupling between CLI and server: a server-side "authenticator descriptor API".

This would follow something like this workflow:
- CLI contacts the API when you invoke `conjur config authenticator` (or when you log in) to fetch the latest authenticator descriptors
- Server provides the available authenticator types and parameter list(s) (types, names)
- CLI verifies the command/params and has no need for baked-in description of authenticators)before it sends them to server

This should allow us to have the CLI be generic and capable of working with newer server as we add/modify authenticators in the future.

rafis3:
@sgnn7 I love the idea! I added it to the description of alternative 1.

All, any additional comments or can I merge this proposal?
If we all agree that alternative 1 is preferred, I can also just remove the second alternative.

design/authenticators/authenticator_management_improvement.md
rafis3: Yes, the user experience would be that if you don't specify the required parameters it will fail.
Please have a look at my latest addition, based on the suggestion that Srdjan raised. I think this solves your concern as I understand it.

rafis3:
@cyberark/conjur-core-team, @AndrewCopeland, @orenbm, @sgnn7 would love to get a final thumbs up or down, if you think this should be merged or not, so I could close this PR.
Thanks!

micahlee:
@rafis3 , what does merging this PR mean? Is it just a"design direction the team likes" for further consideration? Or is there some further commitment that this is something we're going to do?

If the latter, shouldn't this include greater feedback than just engineers? This proposal is more than just technical implementation details, but covers the behavior and end-user interaction with Conjur.

rafis3:
Thanks for the response @micahlee. This isn't currently intended for any allocated effort. It's more more supposed to capture a thought, a direction for a certain area of our product, so when the time comes to work on it, we'll have a small piece of vision for it. As such, where is the appropriate place to capture it? Is it an issue with certain tags? is it in the `design` folder? Or maybe something else?

micahlee:
Yeah, I basically have the same questions... :)

What do you think about a subdirectory `design/proposals` just to help clarify that this isn't the implemented design yet?

The lifecycle for this is something I'm very interested in fleshing out and solidifying though. I *really* support the ability to propose improvements and enhancements, but am still concerned that this becomes another ice box...

jvanderhoof:
I'm strongly adding my support to the API approach.  It allows us to create three levels of interaction: UI, CLI, and API, which open this feature up for three different personas.

Thanks for the thoughtful proposal @orenbm & @rafis3!

design/authenticators/authenticator_management_improvement.md
```diff
@@ -0,0 +1,248 @@
...
+## Implementation Options
+
+### Alternative 1 - Server Side
+
+The new functionality will be developed on the Conjur side. New APIs will be exposed for this new functionality. For example, in order to add an authenticator, the client will call a `POST` request to `/{authenticator-type}/{service-id}/{account}`. Example:
+
+```shell script
+curl --request POST \
+  --data '{"providerUri": "my-uri", "id-token-user-property": "prop"}' \
+  https://conjur.mycompany.net/authn-oidc/my-authenticator/my-account
```
jvanderhoof, line 176: The above syntax raises the question if authenticators should be allowed to span Conjur accounts. Conjur accounts were originally intended to allow separation between groups (ex. staging/production, or marketing/infrastructure/etc.).

rafis3: Why does this syntax raise this question? Currently authenticators don't span Conjur accounts and in this example, the specific authenticator under the specific account are specified in the URL. To compare, to authenticate with an authenticator you would call `https://conjur.mycompany.net/authn-oidc/my-authenticator/my-account/authenticate`

design/proposals/authenticator_management_improvement.md
```diff
@@ -0,0 +1,248 @@
...
+Or:
+
+- Update the `conjur` policy yaml file and take out the authenticator
+- Reload the `conjur` policy using the `--replace` option  
+
+The current user experience requires a certain level of expertise. The user needs to understand the policy structure of each authenticator and the process is comprised of multiple steps. This proposal offers an easier way for managing authenticators.
+
+## Proposed Solution
+
+The proposed solution is to add a few simple Conjur CLI commands (and possibly APIs, see two alternatives below), that would hide the complexity and provide a single step to perform each action:
```
InbalZilberman, line 68: @rafis3 I think API is a must have in our type of customers that many types performs these kind of actions in their pipeline, scripts, automated way where connecting to the CLI is a complex option.

design/proposals/authenticator_management_improvement.md
```diff
@@ -0,0 +1,248 @@
...
+
+conjur config authenticator update *authenticator-type/service-id* [*parameters*]
+
+#### Example
+
+```shell script
+$ conjur config authenticator update authn-oidc/my-idp --provider-uri new-uri
+```
+
+### Enable Authenticator
```
InbalZilberman, line 118: @rafis3 Do we really need this functionality AT ALL?
How about just making all authenticators enabled by default? 
Alternatively, making an authenticator be enabled only if some role has permissions to it?
WDYT?
What is the need this functionality is coming to answer?

For backward compatibility I am thinking of a configuration and not improving the existing mechanism.

design/proposals/authenticator_management_improvement.md
InbalZilberman: By mistake, just now read all the above thread.. so this point is well discussed

design/proposals/authenticator_management_improvement.md
rafis3: In our current architecture, yes this is needed. You want in any instance to have only the minimal required authenticators. Maybe in the future we'll be able to remove this need.

InbalZilberman: @rafis3 If you agree that in the future we dont need it, i think it is not worth the effort of having a CLI providing this usability. what it means for enabling the authenticators by default?

design/proposals/authenticator_management_improvement.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Proposed Solution
+
+The proposed solution is to add a few simple Conjur CLI commands (and possibly APIs, see two alternatives below), that would hide the complexity and provide a single step to perform each action:
+
+### Add Authenticator
+
+#### Functionality
+
+Defines a new authenticator in Conjur. Behind the scenes, this command will creat the authenticator policy and fill it with all its required data. If the data is not required to be given by the user, like the `authn-k8s` CA cert and key, the information will be generated automatically.
```
jvanderhoof, line 72: @micahlee has a prototype/POC with this functionality floating around.  It came up with the OCP Follower deployment improvements spike.  It's pretty powerful stuff.

jvanderhoof:
@rafis3, a couple of higher-level questions:
1. How does a user know what values to pass into the POST to create a particular authenticator?
1. How does a user know if they've successfully created a functioning authenticator (configured correctly and able to connect to the authentication endpoint)?

rafis3:
@jvanderhoof please see below:

> 1. How does a user know what values to pass into the POST to create a particular authenticator?

There will be an API that specifies the schema of every authenticator. See my little explanation about it in this quote from the doc:
```
Conjur will expose an internal API that would provide this authenticator schema information. This API will be called by the CLI whenever this information will be needed.
```

> 2. How does a user know if they've successfully created a functioning authenticator (configured correctly and able to connect to the authentication endpoint)?

There will be two ways:
1. The user will get a successful response from the creation (or update) of the authenticator. This will give an indication that the information was given properly.
1. The authenticator status API can provide additional information about the state of the authenticator and the validity of its parameters. For more information, see [this](https://github.com/cyberark/conjur/blob/master/design/authenticators/authenticators-status/authn_status_general.md) page.

design/authenticators/authenticator_management_improvement.md
jvanderhoof: The above is missing the account scope in the URL.

PR #1510, opened 2020-04-27
add roles when using conjurctl policy load

h-artzi:
### What does this PR do?
When using conjurctl to load a policy, it failed to create role credentials. This PR allows the creation of roles. 

[dap-support/#80](https://github.com/conjurinc/dap-support/issues/80)

### How to test?
Check the ticket and run it locally with the appropriate image. It is expected to successfully run and output the created roles.

#### Change log

#### Test coverage

Note: I am not sure that this should be a supported feature in the future.

#### Follow-on issues
- [X] No follow-up issues are required

### DIFF
--- a/lib/root_loader.rb
+++ b/lib/root_loader.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'json'
+
 # BootstrapLoader is used to load an initial "root" policy when the database is completely empty.
 class RootLoader
   class << self
@@ -17,25 +19,60 @@ class RootLoader
         admin = ::Role[admin_id] || ::Role.create(role_id: admin_id)
         if admin_password = ENV['CONJUR_ADMIN_PASSWORD']
           $stderr.puts "Setting 'admin' password"
-          admin_credentials = Credentials[role: admin] || Credentials.create(role: admin)
-          admin_credentials.password = admin_password
-          admin_credentials.save
+          set_admin_password(admin, admin_password)
         end
 
         root_policy_resource = Loader::Types.find_or_create_root_policy(account)
+        policy_ver = policy_version(
+          role: admin, 
+          policy: root_policy_resource, 
+          filename: filename
+        )
 
-        policy_version = PolicyVersion.new role: admin, policy: root_policy_resource, policy_text: File.read(filename)
-        policy_version.policy_filename = filename
-        policy_version.perform_automatic_deletion = true
-        policy_version.delete_permitted = true
-        policy_version.update_permitted = true
-        policy_version.save
+        acc_roles = accepted_roles(policy_ver)
+        created_roles = create_roles(acc_roles)
 
-        loader = Loader::Orchestrate.new policy_version
-        loader.load
+        $stderr.puts(
+          JSON.pretty_generate(
+            created_roles: created_roles,
+            version: policy_ver.version
+          )
+        )
+
+        end_t = Time.now
+        $stderr.puts "Loaded policy in #{end_t - start_t} seconds"
       end
-      end_t = Time.now
-      $stderr.puts "Loaded policy in #{end_t - start_t} seconds"
+    end
+
+    def set_admin_password(admin_role, password)
+      admin_credentials = Credentials[role: admin_role] || Credentials.create(role: admin_role)
+      admin_credentials.password = password
+      admin_credentials.save
+    end
+
+    def create_roles accepted_roles
+      accepted_roles.each_with_object({}) do |role, memo|
+        credentials = Credentials[role: role] || Credentials.create(role: role)
+        role_id = role.id
+        memo[role_id] = { id: role_id, api_key: credentials.api_key }
+      end
+    end
+
+    def accepted_roles policy_version
+      loader = Loader::Orchestrate.new policy_version
+      loader.load
+      loader.new_roles.select do |role|
+        %w(user host).member?(role.kind)
+      end
+    end
+
+    def policy_version role:, policy:, filename:
+      policy_version = PolicyVersion.new role: role, policy: policy, policy_text: File.read(filename)
+      policy_version.policy_filename = filename
+      policy_version.perform_automatic_deletion = true
+      policy_version.delete_permitted = true
+      policy_version.update_permitted = true
+      policy_version.save
     end
   end
 end
--- a/.reek.yml
+++ b/.reek.yml
@@ -311,6 +311,7 @@ detectors:
     exclude:
     - initialize
     - RootLoader#load
+    - RootLoader#policy_version
     - PossumWorld#authn_local_request
     - PossumWorld#create_user
     - PossumWorld#denormalize


lib/root_loader.rb
```diff
@@ -33,6 +35,16 @@ def load account, filename
...
+
+        created_roles = loader.new_roles.select do |role|
+          %w(user host).member?(role.kind)
+        end.inject({}) do |memo, role|
+          credentials = Credentials[role: role] || Credentials.create(role: role)
+          memo[role.id] = { id: role.id, api_key: credentials.api_key }
+          memo
+        end
+
+        $stderr.puts JSON.pretty_generate(created_roles: created_roles, version: policy_version.version)
```
jtuttle, line 47: Can you break this line? It's pretty long.

lib/root_loader.rb
```diff
@@ -33,6 +35,17 @@ def load account, filename
...
         loader = Loader::Orchestrate.new policy_version
         loader.load
+
+        created_roles = loader.new_roles.select do |role|
+          %w(user host).member?(role.kind)
+        end.inject({}) do |memo, role|
+          credentials = Credentials[role: role] || Credentials.create(role: role)
+          memo[role.id] = { id: role.id, api_key: credentials.api_key }
+          memo
+        end
```
jonahx, line 45: I'd split this whole thing up into 2 statements, rather than chaining.

Also, see how the above has the awkward pattern 
```ruby
memo[blah] = stuff.....
memo
```
`each_with_object` is made for avoiding just this.

jonahx: So this is exactly what I meant.... buuuuuut

I don't get how it's working because `each_with_object` reverses the arguments relative to `reduce`, so that the memo is in the 2nd arg:

https://ruby-doc.org/core-2.7.1/Enumerable.html#method-i-each_with_object

lib/root_loader.rb
```diff
@@ -33,6 +35,17 @@ def load account, filename
...
+        acceppted_roles = loader.new_roles.select do |role|
+          %w(user host).member?(role.kind)
+        end
+
+        created_roles = acceppted_roles.each_with_object({}) do |memo, role|
+          memo[role.id] = { id: role.id, api_key: credentials.api_key }
+        end
+
+        $stderr.puts JSON.pretty_generate(created_roles: created_roles, 
+                                          version: policy_version.version)
```
jonahx, line 48: Formatting nitpicks:

No poetry mode per ruby style guide, ie, needs parens after `puts`.

Also, suggested line breaking:

```ruby
$stderr.puts(
  JSON.pretty_generate(
    created_roles: created_roles,
    version: policy_version.version
  )
)
```

lib/root_loader.rb
```diff
@@ -33,6 +35,17 @@ def load account, filename
 
         loader = Loader::Orchestrate.new policy_version
         loader.load
+
+        acceppted_roles = loader.new_roles.select do |role|
```
micahlee, line 39:
```suggestion
        accepted_roles = loader.new_roles.select do |role|
```

lib/root_loader.rb
```diff
@@ -33,6 +35,22 @@ def load account, filename
 
         loader = Loader::Orchestrate.new policy_version
         loader.load
+
+        accepted_roles = loader.new_roles.select do |role|
+          %w(user host).member?(role.kind)
+        end
+
+        created_roles = acceppted_roles.each_with_object({}) do |role, memo|
```
jtuttle, line 43: Typo here: `acceppted`. I would have thought this would break when testing, have you tried this out?

h-artzi: Good catch! It broke while testing.

lib/root_loader.rb
```diff
@@ -31,11 +33,31 @@ def load account, filename
...
-        loader = Loader::Orchestrate.new policy_version
-        loader.load
+        create_roles policy_version
+
+        end_t = Time.now
+        $stderr.puts "Loaded policy in #{end_t - start_t} seconds"
+      end
+    end
+
+    def create_roles policy_version
```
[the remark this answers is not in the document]
jonahx: I just noticed this after approving, but per CC you should extract this:

```ruby
      loader = Loader::Orchestrate.new policy_version
      loader.load
      accepted_roles = loader.new_roles.select do |role|
        %w(user host).member?(role.kind)
      end
```
into an `accepted_roles(policy_version)` method.

lib/root_loader.rb
```diff
@@ -23,19 +25,45 @@ def load account, filename
         end
 
         root_policy_resource = Loader::Types.find_or_create_root_policy(account)
+        policy_version = policy_version(role: admin, policy: root_policy_resource, filename: filename)
+
+        accepted_roles = accepted_roles(policy_version)
+        create_roles(accepted_roles)
```
jonahx, line 31: To avoid confusion with the method name, we should use something diff for the local var name:
```ruby
        acc_roles = accepted_roles(policy_version)
        create_roles(acc_roles)
```

lib/root_loader.rb
```diff
@@ -23,19 +25,45 @@ def load account, filename
         end
 
         root_policy_resource = Loader::Types.find_or_create_root_policy(account)
+        policy_version = policy_version(role: admin, policy: root_policy_resource, filename: filename)
```
jonahx, line 28: split long line:
```ruby
policy_version = policy_version(
  role: admin,
  policy: root_policy_resource,
  filename: filename  
)
```
and per other comment, use something different from local var name than the method name, `policy_ver` or something.

lib/root_loader.rb
```diff
@@ -12,30 +14,66 @@ class << self
...
-      $stderr.puts "Loaded policy in #{end_t - start_t} seconds"
+    end
+
+    def policy_version role:, policy:, filename:
+      policy_version = PolicyVersion.new role: role, policy: policy, policy_text: File.read(filename)
+      policy_version.policy_filename = filename
+      policy_version.perform_automatic_deletion =
+        policy_version.delete_permitted =
+        policy_version.update_permitted = true
+      policy_version.save
```
jonahx, line 76: What's happening here?  It looks like something went wrong...

jtuttle: Just being naughty and seeing if reek would complain. 🤭 

Was trying a multiple assignment to get around the statement limit.

jtuttle: And I would have gotten away with it too if it wasn't for that darn indentation warning! 🐕

jonahx: hahahahaha

This is a clear case where reek may be ignored.  BUT.... the deeper reason it's throwing the warning is because the API for `PolicyVersion.new`  uses the "create and configure" pattern, which is poor and should be avoided.  Instead the constructor should take a hash and return an immutable object.

Also, what does the dog emoji mean?

jonahx:
```
[true] * 3
```
😆 😆 

I think it's time to admit defeat and just ignore the CC error.

jtuttle: Hah yeah I jokingly suggested `[true] * 3` to Hadar too. 😂 

The dog emoji was a reference to Scooby Doo's "and I would have gotten away with it too if it wasn't for you darn kids!" 😄

PR #1512, opened 2020-04-28
Test that authenticators write audit message

orenbm:
When Conjur users perform authentication request we write audit messages for success and failures.
We write the messages to the audit log, which is actually the Rails log (which is derived from the  value of RAILS_ENV).

This PR verifies that the LDAP, Azure and OIDC authenticators write audit messages.

I didn't add tests for other authenticators for the following reasons:
- `authn-k8s`: we currently don't have an ability to read the Conjur log
- `authn-iam`: we don't have integration tests for this authenticator
- `authn`: we don't have integration tests for this authenticator

### DIFF
--- a/cucumber/authenticators_azure/features/authn_azure_hosts.feature
+++ b/cucumber/authenticators_azure/features/authn_azure_hosts.feature
@@ -177,3 +177,19 @@ Feature: Azure Authenticator - Different Hosts can authenticate with Azure authe
     """
     Errors::Authentication::Security::RoleNotAuthorizedOnWebservice
     """
+
+  # This test runs a failing authentication request that is already
+  # tested in another scenario (Host without resource-group annotation is denied).
+  # We run it again here to verify that we write a message to the audit log
+  Scenario: Authentication failure is written to the audit log
+    And I have host "no-resource-group-app"
+    And I set subscription-id annotation to host "no-resource-group-app"
+    And I grant group "conjur/authn-azure/prod/apps" to host "no-resource-group-app"
+    And I fetch a non-assigned-identity Azure access token from inside machine
+    And I save my place in the audit log file
+    When I authenticate via Azure with token as host "no-resource-group-app"
+    Then it is unauthorized
+    And The following appears in the audit log after my savepoint:
+    """
+    cucumber:host:no-resource-group-app failed to authenticate with authenticator authn-azure service cucumber:webservice:conjur/authn-azure/prod
+    """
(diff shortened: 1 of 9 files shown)


cucumber/authenticators_ldap/features/authn_ldap.feature
```diff
@@ -109,3 +114,15 @@ Feature: Users can login with LDAP credentials from an authorized LDAP server
...
+  # tested in another scenario (An LDAP user authorized in Conjur can't login with a bad password).
+  # We run it again here to verify that we write a message to the audit log
+  Scenario: Authentication failure is written to the audit log
+    Given I save my place in the log file
+    When my LDAP password is wrong for authorized user "alice"
+    Then it is unauthorized
+    And The following appears in the log after my savepoint:
+    """
+    cucumber:user:alice failed to authenticate with authenticator authn-ldap service cucumber:webservice:conjur/authn-ldap/test
+    """
```
eladkug, line 128: new line is missing

cucumber/authenticators_azure/features/authn_azure_hosts.feature
```diff
@@ -177,3 +177,19 @@ Feature: Azure Authenticator - Different Hosts can authenticate with Azure authe
...
+
+  # This test runs a failing authentication request that is already
+  # tested in another scenario (Host without resource-group annotation is denied).
+  # We run it again here to verify that we write a message to the audit log
+  Scenario: Authentication failure is written to the audit log
+    And I have host "no-resource-group-app"
+    And I set subscription-id annotation to host "no-resource-group-app"
+    And I grant group "conjur/authn-azure/prod/apps" to host "no-resource-group-app"
+    And I fetch a non-assigned-identity Azure access token from inside machine
+    And I save my place in the log file
```
eladkug, line 189: it audit log here, right? this test looks the same as "Host without resource-group annotation is denied"  beside the last line in logs, but the user need to understand it different log. I would write " And I save my place in the audit_log file".. all over the places related to audit_logs

cucumber/authenticators_ldap/features/authn_ldap.feature
eladkug, line 128: New line is missing

cucumber/authenticators_oidc/features/authn_oidc_performance.feature
```diff
@@ -0,0 +1,38 @@
+Feature: Users can authneticate with OIDC authenticator
```
eladkug, line 1: if this is already new file , i would change the feature name to explain it for performance

cucumber/authenticators_azure/features/authn_azure_hosts.feature
orenbm: that is correct and i wrote a comment about it just before the scenario.

eladkug: I saw your comment and its great, but writing audit_logs instead of log can help(during reading test results if we have some failures)

orenbm: i changed the code a bit to make it clearer that we look for audit messages (although they are in the main log file)

PR #1521, opened 2020-05-04
Openssl 1.0.2final

nahumcohen:
### What does this PR do?
- set openssl version to 1.0.2u
- enable FIPS
- override 3rd parties non FIPS complaint encryption methods

https://github.com/cyberark/conjur/issues/1527

#### Change log
- [v] The CHANGELOG has been updated

#### Test coverage
- [v] The changes in this PR do not require tests

#### Follow-on issues
- [v] This merge is pending slosilo merge and then fixing the Gemfile and Gemfile.lock

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Raise proper error of an authn request with a non-existing user to the `authn`
   authenticator ([cyberark/conjur#1591](https://github.com/cyberark/conjur/pull/1591))
 
+### Changed
+- Uses OpenSSL 1.0.2u to support FIPS compliance 
+- Conjur can be configured to run in FIPS compliant or Non-FIPS compliant mode depending on requirements.
+  FIPS Compliant mode is slightly slower then non-FIPS compliant
+
 ## [1.7.1] - 2020-06-03
 
 ### Added
@@ -59,9 +64,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Removed
 - Images are no longer published to Quay.io.
-
-### Security
-- Upgraded Rails to `v5.2.4.3` to resolve [CVE-2020-8164](https://groups.google.com/forum/#!topic/rubyonrails-security/f6ioe4sdpbY).
 
 ## [1.6.0] - 2020-04-14
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM cyberark/ubuntu-ruby-fips:20.04-latest
 
 ENV DEBIAN_FRONTEND=noninteractive \
     PORT=80 \
@@ -10,19 +10,14 @@ EXPOSE 80
 
 RUN apt-get update -y && \
     apt-get -y dist-upgrade && \
-    apt-get install -y build-essential \
+    apt-get install -y libz-dev
+
+RUN apt-get install -y build-essential \
                        curl \
                        git \
-                       libpq-dev \
                        ldap-utils \
-                       postgresql-client \
-                       ruby2.5 ruby2.5-dev \
                        tzdata \
-                       # needed to build some gem native extensions:
-                       libz-dev \
     && rm -rf /var/lib/apt/lists/*
-
-RUN gem install --no-document --version 2.1.4 bundler
 
 WORKDIR /opt/conjur-server
 
--- a/Gemfile
+++ b/Gemfile
@@ -31,10 +31,10 @@ gem 'sequel-rails'
 
 gem 'activesupport'
 gem 'base32-crockford'
-gem 'bcrypt-ruby', '~> 3.0.0'
+gem 'bcrypt', '~> 3.1.2'
 gem 'gli', require: false
 gem 'listen'
-gem 'slosilo', '~> 2.1'
+gem 'slosilo', '~> 2.2'
 
 # Explicitly required as there are vulnerabilities in older versions
 gem "ffi", ">= 1.9.24"
--- a/app/domain/util/open_ssl/x509/quick_csr.rb
+++ b/app/domain/util/open_ssl/x509/quick_csr.rb
@@ -23,7 +23,7 @@ module Util
         #
         def initialize(
           common_name:,
-          rsa_key: OpenSSL::PKey::RSA.new(1048),
+          rsa_key: OpenSSL::PKey::RSA.new(2048),
           alt_names: []
         )
           @cn = common_name
--- a/ci/authn-oidc/keycloak/fetchCertificate
+++ b/ci/authn-oidc/keycloak/fetchCertificate
@@ -3,7 +3,7 @@
 # This script retrieves a certificate from the keycloak OIDC provider.
 # It is needed to communicate with the provider via SSL for validating ID tokens
 
-httpclient_pem_location="/var/lib/gems/2.5.0/gems/httpclient-2.8.3/lib/httpclient"
+httpclient_pem_location="/var/lib/ruby/lib/ruby/gems/2.5.0/gems/httpclient-2.8.3/lib/httpclient"
 
 echo “keycloak cert” >> "$httpclient_pem_location/cacert.pem"
 echo =============== >> "$httpclient_pem_location/cacert.pem"
--- a/config/initializers/fips.rb
+++ b/config/initializers/fips.rb
@@ -1,0 +1,24 @@
+require "openssl"
+require "digest"
+
+# Suppress warning messages
+original_verbose, $VERBOSE = $VERBOSE, nil
+
+# override the default Digest with OpenSSL::Digest
+Digest::SHA256 = OpenSSL::Digest::SHA256
+Digest::SHA1 = OpenSSL::Digest::SHA1
+
+# Activate warning messages again
+$VERBOSE = original_verbose
+
+# by default FIPS mode is enabled
+# disable FIPS mode only if OPENSSL_FIPS_ENABLED environment variable is present and has false value
+OpenSSL.fips_mode = !(ENV["OPENSSL_FIPS_ENABLED"].present? && ENV["OPENSSL_FIPS_ENABLED"] == 'false')
+
+# each of the following 3rd party overridden is required since a non FIPS complaint encryption method is used
+# if a non-complaint FIPS method like MD5 is used or a direct use of Digest::encryption-method
+#  (rather than OpenSSL::Digest::encryption-method) is performed
+#  the server will crush on run time
+
+# override ActiveSupport hash_digest_class with FIPS complaint method
+ActiveSupport::Digest.hash_digest_class = OpenSSL::Digest::SHA1.new
--- a/dev/Dockerfile.dev
+++ b/dev/Dockerfile.dev
@@ -1,25 +1,21 @@
-FROM phusion/baseimage:0.11
+    FROM cyberark/phusion-ruby-fips:0.11-latest
 
-RUN DEBIAN_FRONTEND=noninteractive apt-get update -y && \
-    DEBIAN_FRONTEND=noninteractive apt-get install -y \
+RUN DEBIAN_FRONTEND=noninteractive apt-get update -y
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
     build-essential \
-    ruby2.5 ruby2.5-dev \
-    postgresql-client \
-    libpq-dev \
-    unattended-upgrades \
     ldap-utils \
     git \
-    update-notifier-common \
-    vim \
-    curl \
     jq \
     tzdata \
-    wget \
     libfontconfig1 \
-    libfontconfig1-dev \
+    libfontconfig1-dev
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    unattended-upgrades \
+    vim \
+    curl \
+    wget \
+    syslog-ng \
     && rm -rf /var/lib/apt/lists/*
-
-RUN gem install --no-document --version 2.1.4 bundler
 
 RUN mkdir -p /src/conjur-server
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'simplecov'
+
 SimpleCov.command_name "SimpleCov #{rand(1000000)}"
 SimpleCov.merge_timeout 7200
 SimpleCov.start
--- a/.trivyignore
+++ b/.trivyignore
@@ -12,3 +12,19 @@ CVE-2019-10220
 CVE-2019-19813
 CVE-2019-19814
 CVE-2019-19816
+
+# Applications that call the SSL_check_chain() function during or after a TLS 1.3 handshake
+# may crash due to a NULL pointer dereference as a result of incorrect handling of the "signature_algorithms_cert"
+# TLS extension. this issue was fixed in OpenSSL 1.1.1g
+#
+# In order to support fips with openssl we are required to downgrading openssl version to 1.0.2 until openssl will
+# support fips module in newer versions
+# This vulnerability this is not relevant to us as
+# 1. The installed version (1.0.2u) does not support 1.3
+# 2. Trivy detect the usage of openssl 1.0.2 (can be reproduced with
+#    docker run -v /var/run/docker.sock:/var/run/docker.sock
+#   -v $(PWD):/workspace --rm aquasec/trivy -f json -o /workspace/scan_results-conjur-unfixed.json --no-progress
+#   --ignorefile .trivyignore registry.tld/ruby-fips-base-image-phusion:1.0.0)
+#
+# Performed by @yahalomk approved by @shaharglazner
+CVE-2020-1967
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -88,7 +88,7 @@ GEM
     backports (3.17.0)
     base32-crockford (0.1.0)
     base58 (0.2.3)
-    bcrypt-ruby (3.0.1)
+    bcrypt (3.1.13)
     bindata (2.4.7)
     builder (3.2.4)
     byebug (11.1.1)
@@ -258,7 +258,7 @@ GEM
     nio4r (2.5.2)
     nokogiri (1.10.9)
       mini_portile2 (~> 2.4.0)
-    openid_connect (1.1.8)
+    openid_connect (1.2.0)
       activemodel
       attr_required (>= 1.0.0)
       json-jwt (>= 1.5.0)
@@ -396,7 +396,7 @@ GEM
       json (>= 1.8, < 3)
       simplecov-html (~> 0.10.0)
     simplecov-html (0.10.2)
-    slosilo (2.1.1)
+    slosilo (2.2.1)
     spring (2.1.0)
     spring-commands-cucumber (1.0.1)
       spring (>= 0.9.1)
@@ -451,7 +451,7 @@ DEPENDENCIES
   aws-sdk-iam
   base32-crockford
   base58
-  bcrypt-ruby (~> 3.0.0)
+  bcrypt (~> 3.1.2)
   ci_reporter_rspec
   command_class
   conjur-api!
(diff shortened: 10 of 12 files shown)


dev/Dockerfile.dev
```diff
@@ -1,9 +1,8 @@
-FROM phusion/baseimage:0.11
+FROM registry.tld/ruby-fips-base-image-phusion:1.0.0
```
micahlee, line 1: Did you consider or attempting basing this on `registry.tld/ruby-fips-base-image-ubuntu:1.0.0` to get this on a shared based image instead?

config/initializers/fips.rb
```diff
@@ -0,0 +1,46 @@
...
+end
+new_sprockets_config[:digest_class] = OpenSSL::Digest::SHA256
+Sprockets.config = new_sprockets_config.freeze
+
+OpenIDConnect::Discovery::Provider::Config::Resource.module_eval do
+  def cache_key
+    sha256 = Digest::SHA256.hexdigest host
+    "swd:resource:opneid-conf:#{sha256}"
+  end
+end
```
micahlee, line 46: Nitpick: needs a final newline

config/initializers/fips.rb
```diff
@@ -0,0 +1,46 @@
...
+OpenSSL.fips_mode = true
+
+ActiveSupport::Digest.hash_digest_class = OpenSSL::Digest::SHA1.new
+Sprockets::DigestUtils.module_eval do
+  def digest_class
+    OpenSSL::Digest::SHA256
+  end
+end
+
+new_sprockets_config = {}
```
micahlee, line 34: What do you think about making this section a bit more concise with something like:
```ruby
Sprockets.config = Hash[Sprockets.config].tap do |config|
  config[:digest_class] = OpenSSL::Digest::SHA256
end.freeze
```

spec/spec_helper.rb
```diff
@@ -21,6 +24,30 @@
...
+  config.before(:all) do
+    OpenSSL.fips_mode = true
+    ActiveSupport::Digest.hash_digest_class = OpenSSL::Digest::SHA1.new
+    Sprockets::DigestUtils.module_eval do
+      def digest_class
+        OpenSSL::Digest::SHA256
+      end
+    end
+
+    new_sprockets_config = {}
```
micahlee, line 36: Consider using the more concise frozen hash copy here as well.

dev/Dockerfile.dev
nahumcohen: The plan is to consume this image from Dockerhub

micahlee: I meant did you consider not creating two base images:
- `ruby-fips-base-image-ubuntu:1.0.0`
- `ruby-fips-base-image-phusion:1.0.0`

And instead only use one in all places, replacing the use of `phusion` in the OSS repo: `ruby-fips-base-image-ubuntu:1.0.0`.

That would reduce the maintenance overhead of two base images just for a separate dev environment (and it's not clear the dev environment needs to be `phusion` based).

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM registry.tld/ruby-fips-base-image-ubuntu:1.0.0
```
alexkalish, line 1: Is this a private Docker repo URL?  I thought that we were going to publish this so that it's publicly available.

izgeri:
@nahumcohen is there a public github issue you can reference for this work? also, we usually avoid putting links to (private) jira and confluence in public repos in github, since community members can't access those links to get the added context available there

config/initializers/fips.rb
```diff
@@ -0,0 +1,46 @@
+require "openssl"
+require "digest"
+require "digest/md5"
+require "digest/sha1"
+require 'sprockets'
+require 'openid_connect'
+
+# Supress warning messages.
```
alexkalish, line 8: Not sure how @micahlee feels, but I would strongly suggest heavily commenting this file... maybe even have more comment lines than code ones.  Since it's all configuration, this will be difficult to comprehend in the future.  Why is each constant being changed?  What does that affect and what is the result?

Dockerfile
micahlee: That is a really good point! I had missed that this in particular would prevent anyone from build the Conjur image without internal registry access.

config/initializers/fips.rb
```diff
@@ -0,0 +1,46 @@
...
+# amount of log spam and warnings.
+Digest.send(:remove_const, "MD5") if Digest.const_defined?("MD5")
+Digest.const_set("MD5", OpenSSL::Digest::MD5)
+Digest.send(:remove_const, "SHA1") if Digest.const_defined?("SHA1")
+Digest.const_set("SHA1", OpenSSL::Digest::SHA1)
+Digest.send(:remove_const, "SHA256") if Digest.const_defined?("SHA256")
+Digest.const_set("SHA256", OpenSSL::Digest::SHA256)
+
+# override the default Digest with OpenSSL::Digest
+Digest::SHA256 = OpenSSL::Digest::SHA256
```
micahlee, line 20: What is the significance of this line versus the lines above?
```
Digest.send(:remove_const, "SHA256") if Digest.const_defined?("SHA256")
Digest.const_set("SHA256", OpenSSL::Digest::SHA256)
```

dev/Dockerfile.dev
nahumcohen: Currently, OSS is using the Conjur base image and appliance the phusion base image.
Uniting them to one image is required but it is out of scope for this task.

config/initializers/fips.rb
nahumcohen: Chef mentions that
# Remove pre-existing constants if they do exist to reduce the
# amount of log spam and warnings.

https://github.com/chef/chef/blob/4d9857dd8ce9f8f199722c7abad2bdd60a6277d3/chef-config/lib/chef-config/config.rb#L1248
chef-config/lib/chef-config/config.rb:1248

config/initializers/fips.rb
nahumcohen: Thanks for that.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - Change ActiveSupport to use sha1 instead of md5
+- Set openssl version to 1.0.2
+- Enable FIPS
```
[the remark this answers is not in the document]
nahumcohen: Thanks for that.
Did push a fix.
please have a look and tell if it is satisfying.

config/initializers/fips.rb
nahumcohen: Did push a fix.
Please have a look and let know if it is satisfying.

Dockerfile
izgeri: @nahumcohen what is needed to publish the image to our dockerhub? if you need help getting the base image repo reviewed to make it public, and adding a dockerhub to also push this image to, please let me know. I agree with @alexkalish, it is very important that we maintain the ability of community members to build this project locally.

dev/Dockerfile.dev
micahlee: Hi Nahum, 

I'm not referring to the appliance. I understand that it still uses the `phusion`, and needs to for a variety of reasons. 

I'm referring to this `Dockerfile.dev` that we're commenting on for the dev environment in OSS. It should be able to use the `ruby-fips-base-image-ubuntu:1.0.0` base image.

config/initializers/fips.rb
micahlee: Do you have an example of the log spam and warnings this removes in Conjur?

nahumcohen: It was added only because of the details in the link above.
Though, I have tried it without these lines and there are no extra warnings.
Do you think it is better to remove these lines?

dev/Dockerfile.dev
nahumcohen: Micah hi,
Sorry I missed your point twice. I absolutely agree. I've tried to use the ubuntu but faced problems therefore left it. 
It is out of scope for this assignment to change the base image. Should be addressed in a different task

config/initializers/fips.rb
micahlee: It would be worth having others from @cyberark/conjur-core-team weigh in, but if we don't observe the extra warnings these are intended to fix, then I don't think we should include these.

dev/Dockerfile.dev
micahlee: Sounds good! If it was an easy swap, then it would be worth doing. I agree that if it's much more work, it should be separate. Thanks! :)

config/initializers/fips.rb
nahumcohen: @shaharglazner hu,
What do you say regarding the Chef documentation.
I agree with Micah and think it should be removed.

Dockerfile
nahumcohen: This is only WIP PR, next step is to publish base image to dockerhub.
Does the image require any license?

config/initializers/fips.rb
```diff
@@ -0,0 +1,53 @@
...
+    digest_class: OpenSSL::Digest::SHA256
+).freeze
+
+# override OpenIDConnect cache_key with FIPS complaint method
+OpenIDConnect::Discovery::Provider::Config::Resource.module_eval do
+  def cache_key
+    sha256 = Digest::SHA256.hexdigest host
+    "swd:resource:opneid-conf:#{sha256}"
+  end
+end
```
jonahx, line 53: These `module_eval`s make me very nervous.  Monkey patching 3rd party classes in an initializer seems quite dangerous:
1. It's brittle.  We're making assumptions about the classes and how they use these methods, presumably based on their current implementation details.   Even assuming you have verified that their current code uses this method for making _all_ decisions about which digest to use, the class is making no public guarantees that this will continue to be true.  Therefore: this behavior may change at any time, even in a minor version update.
2. This could create some very confusing situations for future developers who don't realize that this class has been monkey-patched.

spec/spec_helper.rb
```diff
@@ -21,6 +24,32 @@
...
+
+    # override OpenIDConnect cache_key with FIPS complaint method
+    OpenIDConnect::Discovery::Provider::Config::Resource.module_eval do
+      def cache_key
+        sha256 = Digest::SHA256.hexdigest host
+        "swd:resource:opneid-conf:#{sha256}"
+      end
+    end
+  end
+
```
jonahx, line 52: Much of this code is repeated from above.  My comments from above would apply here too.  And I'd also wonder if we are missing some abstraction or better design that would allow us to avoid this repetition. 

Also, it's not clear why rspec needs this configuration.  Our application is what needs to be FIPS compliant.  Something has gone wrong somewhere, I would think, if this is requiring changes to our test code too.

Gemfile
```diff
@@ -31,10 +31,11 @@ gem 'sequel-rails'
 
 gem 'activesupport'
 gem 'base32-crockford'
-gem 'bcrypt-ruby', '~> 3.0.0'
+gem 'bcrypt', '~> 3.1.2'
 gem 'gli', require: false
 gem 'listen'
-gem 'slosilo', '~> 2.1'
+#gem 'slosilo', '~> 2.1'
+gem 'slosilo', github: 'cyberark/slosilo', branch: 'sha256'
```
jonahx, line 38: Should publish a new gem version and link to that before merging.

config/initializers/fips.rb
```diff
@@ -0,0 +1,53 @@
...
+
+# override Sprockets digest_class with FIPS complaint method
+Sprockets::DigestUtils.module_eval do
+  def digest_class
+    OpenSSL::Digest::SHA256
+  end
+end
+Sprockets.config = Sprockets.config.merge(
+    digest_class: OpenSSL::Digest::SHA256
+).freeze
```
jonahx, line 45: What are we using sprockets for?  We most likely shouldn't be using it at all, since the only page we're serving afaik is the status page, which shouldn't need it anymore.   Removing it altogether would be prefereable.

nahumcohen: Jonah thanks for the catch.
It was added at the beginning of implementing the feature and was required for a break probably in the status page.

spec/spec_helper.rb
nahumcohen: No need for abstraction.
appears that during development and rspec failure the fips changes were added to spec_helper. But it is taken from fips.rb.
So remove this piece of useless code.

config/initializers/fips.rb
nahumcohen: We left with ActiveSupport and with OpenIDConnect encryption method overridden.
I agree that it has limits and problems but these 3rd parties don't supply FIPS complaint gems so this solution is kind of a compromise.
Is there another alternative that can be implemented?

config/initializers/fips.rb
nahumcohen: Remove the Chef recommendation code after approving with Shahar

config/initializers/fips.rb
jtuttle: It might be worth making a pull request against the [openid_connect](https://github.com/nov/openid_connect) repo to see if the author wants to make it FIPS compliant. Looks like it's under pretty active development so it might not take too long.

If that doesn't work, maybe we can fork the repo?

Dockerfile
jonahx: @nahumcohen Just an FYI: when creating a PR, you can choose to open a "Draft PR" (use arrow dropdown on the GH button).

izgeri: @nahumcohen can we at least put a WIP in the PR title until it's ready for review? Since it's not a draft and doesn't have this heading, there's no indication that this PR is in a pre-mergeable state.

config/initializers/fips.rb
nahumcohen: Open an issue https://github.com/nov/openid_connect/issues/52

nahumcohen: openid_connect PR was merged to master

nahumcohen: openid_connect next version will consist the change

jvanderhoof: @nahumcohen can you please schedule a task to upgrade this version?  I don't want to lose this!

app/domain/util/open_ssl/x509/quick_csr.rb
```diff
@@ -23,7 +23,7 @@ class QuickCsr
         #
         def initialize(
           common_name:,
-          rsa_key: OpenSSL::PKey::RSA.new(1048),
```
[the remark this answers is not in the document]
jvanderhoof: Would it make sense to declare the default key size somewhere so it's easy to reuse?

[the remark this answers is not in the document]
nahumcohen: It is more readable and clear for the developer to keep it as a literal.
If you feel it makes more sense it will be changed.

config/initializers/fips.rb
nahumcohen: The change is in the new openid_connect 1.2.0 that is already released.
We will consume the version with the fix. :)

nahumcohen: And of course the piece of code is removed. Will be pushed ASAP we have a green branch.

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - The Conjur version is now printed on server startup, after running `conjurctl server`
   ([cyberark/conjur#1590](https://github.com/cyberark/conjur/pull/1590))
 
+### Security
+- Upgrade Rails to `v5.2.4.3` to resolve [CVE-2020-8164](https://groups.google.com/forum/#!topic/rubyonrails-security/f6ioe4sdpbY).
```
jvanderhoof, line 13: @andytinkham gave us guidance that we don't need to document dependency upgrades in the security section.  We should be able to remove this line.

andytinkham: To be clear, I said it was TBD and I was leaning towards not including them, so don't take it as gospel truth yet. Naama is raising points in another thread that make this still very much TBD. I'm OK either removing this line right now or just leaving off the "to resolve" bit.

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- Use a new base image ruby-fips-base-image-ubuntu and ruby-fips-base-image-phusion
+  The new image consists of openssl 1.0.2u which is FIPS complaint
```
jvanderhoof, line 15: Would this be more of a change than a security fix?  What about moving this the the `Changes` section, maybe rephrasing it to something like:
```
- Uses OpenSSL 1.0.2u to support FIPS compliance.
```
?

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - The Conjur version is now printed on server startup, after running `conjurctl server`
   ([cyberark/conjur#1590](https://github.com/cyberark/conjur/pull/1590))
 
+### Security
+- Upgrade Rails to `v5.2.4.3` to resolve [CVE-2020-8164](https://groups.google.com/forum/#!topic/rubyonrails-security/f6ioe4sdpbY).
+- Use a new base image ruby-fips-base-image-ubuntu and ruby-fips-base-image-phusion
+  The new image consists of openssl 1.0.2u which is FIPS complaint 
+- Enable FIPS. The enabling is done in fips.rb with the OpenSSL.fips_mode variable
```
jvanderhoof, line 16: Same as above, this looks like a `Change` rather than `Security`.  What do you think about reworking this to something like:
```suggestion
- Conjur can be configured to run in FIPS compliant or Non-FIPS compliant mode depending on requirements.  FIPS Compliant mode is slightly slower then non-FIPS compliant.
```
It feels like customers would benefit from understanding the tradeoffs.

Dockerfile
jvanderhoof: Is there a ticket for making the base image publicly available?  Without it, we're preventing anyone from outside our org from working on Conjur.

Dockerfile
```diff
@@ -10,20 +10,14 @@ EXPOSE 80
 
 RUN apt-get update -y && \
     apt-get -y dist-upgrade && \
-    apt-get install -y build-essential \
+    apt-get install -y libz-dev
+RUN apt-get install -y build-essential \
```
jvanderhoof, line 14: Can we chain the install above rather than executing a second `RUN` command?  It'll remove a layer from the image.

jvanderhoof: Something like:
```
RUN apt-get update -y && \
    apt-get -y dist-upgrade && \
    apt-get install -y libz-dev \
                       build-essential \
                       curl \
                       git \
                       ldap-utils \
                       tzdata \
    && rm -rf /var/lib/apt/lists/*
```

Gemfile
jvanderhoof: Can we get Slosilo released and clean this up before we merge, please?

ci/authn-oidc/keycloak/fetchCertificate
```diff
@@ -3,7 +3,8 @@
 # This script retrieves a certificate from the keycloak OIDC provider.
 # It is needed to communicate with the provider via SSL for validating ID tokens
 
-httpclient_pem_location="/var/lib/gems/2.5.0/gems/httpclient-2.8.3/lib/httpclient"
+httpclient_pem_location="/var/lib/ruby/lib/ruby/gems/2.5.0/gems/httpclient-2.8.3/lib/httpclient"
+#httpclient_pem_location="/var/lib/gems/2.5.0/gems/httpclient-2.8.3/lib/httpclient"
```
jvanderhoof, line 7: can this line be removed?

CHANGELOG.md
nahumcohen: The line is removed

CHANGELOG.md
nahumcohen: Done

CHANGELOG.md
nahumcohen: Done

Dockerfile
nahumcohen: Done

ci/authn-oidc/keycloak/fetchCertificate
nahumcohen: Done

Dockerfile
nahumcohen: On using one layer the OSS build fails.
We are not sure why and would like to postponed it for the future.

andytinkham:
thank you for putting detailed comments in the .trivyignore. That will be helpful when a customer asks us about this issue!

Dockerfile
nahumcohen: Done

PR #1523, opened 2020-05-04
Add a test case for resource list acting_as

izgeri:
### What does this PR do?
During an effort to document the `acting_as` query parameter for the `resources`
route, it was raised that failing to specify the account in the fully qualified
ID of the role results in a 403. We may want to alter this behavior at some
point, but in the meantime it does not appear that we have a test that documents
this functionality. This commit adds a test, so that when/if we change this
behavior in the future, it will be clear that we have broken expected behavior
by breaking this test.

n/a

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/cucumber/api/features/resource_list_by_role.feature
+++ b/cucumber/api/features/resource_list_by_role.feature
@@ -27,3 +27,7 @@ Feature: List resources for another role
     When I successfully GET "/resources?acting_as=cucumber:user:alice"
     Then the resource list should contain "variable" "dev/dev-var"
     And the resource list should not contain "variable" "prod/prod-var"
+
+  Scenario: Attempting to retrieve the resource list for a different role but without giving the account in the ID results in a 403
+    When I GET "/resources?acting_as=user:alice"
+    Then the HTTP response status code is 403


izgeri:
Note - I also filed #1524. If we intend to address that issue and change this behavior, this PR may not need to be merged.

cucumber/api/features/resource_list_by_role.feature
```diff
@@ -27,3 +27,7 @@ Feature: List resources for another role
     When I successfully GET "/resources?acting_as=cucumber:user:alice"
     Then the resource list should contain "variable" "dev/dev-var"
     And the resource list should not contain "variable" "prod/prod-var"
+
+  Scenario: Attempting to retrieve the resource list for a different role but without giving the account in the ID results in a 403
+  When I GET "/resources?acting_as=user:alice"
+  Then the HTTP response status code is 403
```
orenbm, line 33: @izgeri is there also a log message that we want to verify here?

izgeri: do you have an example cuke that you're thinking of? I don't see something like that in the other cukes around resources, so I'm not sure exactly what you mean.

orenbm: ok i ran the test and i see that we don't log anything regarding this error (other than the 403). In any case, we have other cukes that verify log messages, for example [this one](https://github.com/cyberark/conjur/blob/master/cucumber/authenticators_oidc/features/authn_oidc.feature#L93)

PR #1526, opened 2020-05-06
Log when using default authentication container name

orenbm:
Connected to conjurinc/dap-support#69

If the annotation `authentication-container-name` is missing in
the host's annotations then we use a default value `authenticator`.

It can be helpful to the user to see this behaviour in the log. We already
log that an annotation was used to get this value
(e.g `Retrieved value of annotation {authn-k8s/authentication-container-name}`
but now we can also see that the default value is used.

### DIFF
--- a/app/domain/authentication/authn_k8s/application_identity.rb
+++ b/app/domain/authentication/authn_k8s/application_identity.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Authentication
   module AuthnK8s
 
@@ -22,6 +24,9 @@ module Authentication
     # validation of ValidateApplicationIdentity
     class ApplicationIdentity
 
+      AUTHENTICATION_CONTAINER_NAME_ANNOTATION = "authentication-container-name"
+      DEFAULT_AUTHENTICATION_CONTAINER_NAME = "authenticator"
+
       def initialize(host_id:, host_annotations:, service_id:)
         @host_id          = host_id
         @host_annotations = host_annotations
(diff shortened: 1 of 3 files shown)


orenbm:
@shulifink can you approve the following message:
- [ ] `Annotation '{0-authentication-container-annotation-name}' not found. Using default value '{1-default-authentication-container}'`

For example, `Annotation 'authentication-container-name' not found. Using default value 'authenticator'

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -45,11 +45,25 @@ def constraints
...
-          annotation_value("authn-k8s/#{annotation_name}") ||
-          annotation_value("kubernetes/#{annotation_name}") ||
-          "authenticator"
+        @container_name ||= annotation_value("authn-k8s/#{@service_id}/#{authentication_container_name}") ||
+          annotation_value("authn-k8s/#{authentication_container_name}") ||
+          annotation_value("kubernetes/#{authentication_container_name}") ||
+          default_authentication_container_name
+      end
+
+      def authentication_container_name
```
micahlee, line 54: This can also be a [class constant](https://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/constants.html) instead of a method that just returns a literal.

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -45,11 +45,25 @@ def constraints
...
+        "authentication-container-name"
+      end
+
+      def default_authentication_container_name
+        "authenticator".tap do |default_container_name|
```
micahlee, line 59: I don't think the tap is necessary here.

I would consider making a class constant for `default_container_name` as well (e.g. `DEFAULT_AUTHN_CONTAINER_NAME='authenticator'`). Then changing this to:

```ruby
def default_authentication_container_name
  Rails.logger.debug(
    LogMessages::Authentication::ContainerNameAnnotationDefaultValue.new(
      authentication_container_name,
      DEFAULT_AUTHN_CONTAINER_NAME
    )
  )

  DEFAULT_AUTHN_CONTAINER_NAME
end
```

orenbm: you are correct :)

app/domain/authentication/authn_k8s/application_identity.rb
```diff
@@ -22,6 +22,9 @@ module AuthnK8s
     # validation of ValidateApplicationIdentity
     class ApplicationIdentity
 
+      AUTHENTICATION_CONTAINER_NAME_ANNOTATION = "authentication-container-name".freeze
```
micahlee, line 25: This is fine, but FYI you can also denote all string literals in the file as frozen by adding:
```ruby 
# frozen_string_literal: true
```

to the top of the file.

orenbm: oh nice, didn't know that. I'll change it and push again.

jonahx:
@orenbm  Just noticed this: 

https://github.com/cyberark/conjur/blob/248829dee1f72275dc799270f1f4dcef07bd4244/app/domain/logs.rb#L175

I like this method of line breaking line strings, but in the above it will print at “notcall” instead of “not call”.  I’ve been standardizing on always adding the space to the line above.

orenbm:
thanks @jonahx, i fixed some other messages while i'm at it.

shulifink:
> @shulifink can you approve the following message:
> 
> * [ ]  `Annotation '{0-authentication-container-annotation-name}' not found. Using default value '{1-default-authentication-container}'`
> 
> For example, `Annotation 'authentication-container-name' not found. Using default value 'authenticator'

@orenbm  - approved!

Issue #1530, opened 2020-05-06
Separate orchestrator config from policy_version model


jtuttle:
The [`PolicyVersion` class](https://github.com/cyberark/conjur/blob/master/app/models/policy_version.rb) is a Sequel model that represents an entry in the `policy_versions` table. It also includes a [set of fields](https://github.com/cyberark/conjur/blob/master/app/models/policy_version.rb#L33) that describe how policy should be loaded by the [Orchestrator class](https://github.com/cyberark/conjur/blob/master/app/models/loader/orchestrate.rb). Including these fields here is conflating the physical model for the `policy_versions` table with details about how policy should be loaded.

We should split these fields into a separate class (`OrchestratorConfig`?) whose purpose is to describe how policy should be loaded. This will improve separation of duties and prevent us from having to "initialize and configure" instances of `PolicyVersion` like we do [in `lib/root_loader.rb`](https://github.com/cyberark/conjur/blob/master/lib/root_loader.rb#L70).

PR #1534, opened 2020-05-07
cleanup policies_version and orchestrate

h-artzi:
### What does this PR do?
It untangles the perform_automatic_deletion and update_permitted attributes from [policy_version.rb](app/models/policy_version.rb) and the load function within [orchestrate.rb](app/models/loader/orchestrate.rb). In addition, there was some cleanup in order to make [policies_controller.rb](app/controllers/policies_controller.rb) easier to read. There aren't additional tests, this PR is simply a small cleanup. A reviewer should start, by looking at [policy_version.rb](app/models/policy_version.rb) and [orchestrate.rb](app/models/loader/orchestrate.rb). Then proceed to [policies_controller.rb](app/controllers/policies_controller.rb)

#1530 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/app/controllers/policies_controller.rb
+++ b/app/controllers/policies_controller.rb
@@ -16,56 +16,58 @@ class PoliciesController < RestController
   def put
     authorize :update
 
-    load_policy perform_automatic_deletion: true, delete_permitted: true, update_permitted: true
+    policy = save_submitted_policy(delete_permitted: true)
+    replace_policy = Loader::ReplacePolicy.from_policy(policy)
+    created_roles = perform(replace_policy)
+
+    render json: {
+      created_roles: created_roles,
+      version: policy.version
+    }, status: :created
   end
 
   def patch
     authorize :update
 
-    load_policy perform_automatic_deletion: false, delete_permitted: true, update_permitted: true
+    policy = save_submitted_policy(delete_permitted: true)
+    modify_policy = Loader::ModifyPolicy.from_policy(policy)
+    created_roles = perform(modify_policy)
+
+    render json: {
+      created_roles: created_roles,
+      version: policy.version
+    }, status: :created
   end
 
   def post
     authorize :create
 
-    load_policy perform_automatic_deletion: false, delete_permitted: false, update_permitted: false
+    policy = save_submitted_policy(delete_permitted: false)
+    create_policy = Loader::CreatePolicy.from_policy(policy)
+    created_roles = perform(create_policy)
+
+    render json: {
+      created_roles: created_roles,
+      version: policy.version
+    }, status: :created
   end
 
   protected
 
-  def load_policy perform_automatic_deletion:, delete_permitted:, update_permitted:
-    policy_text = request.raw_post
-
-    policy_version = PolicyVersion.new \
-      role: current_user, policy: resource, policy_text: policy_text
-    policy_version.perform_automatic_deletion = perform_automatic_deletion
-    policy_version.delete_permitted = delete_permitted
-    policy_version.update_permitted = update_permitted
-    policy_version.save
-    loader = Loader::Orchestrate.new policy_version
-    loader.load
-
-    created_roles = loader.new_roles.select do |role|
-      %w(user host).member?(role.kind)
-    end.inject({}) do |memo, role|
-      credentials = Credentials[role: role] || Credentials.create(role: role)
-      memo[role.id] = { id: role.id, api_key: credentials.api_key }
-      memo
-    end
-
-    render json: {
-      created_roles: created_roles,
-      version: policy_version.version
-    }, status: :created
+  # Returns newly created roles
+  def perform(policy_action)
+    policy_action.call
+    new_actor_roles = actor_roles(policy_action.new_roles)
+    create_roles(new_actor_roles)
   end
 
   def find_or_create_root_policy
-    Loader::Types.find_or_create_root_policy account
+    Loader::Types.find_or_create_root_policy(account)
   end
 
   private
 
-  def concurrent_load _exception
+  def concurrent_load(_exception)
     response.headers['Retry-After'] = retry_delay
     render json: {
       error: {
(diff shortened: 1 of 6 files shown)


app/controllers/policies_controller.rb
```diff
@@ -34,28 +34,25 @@ def post
...
 
-    policy_version = PolicyVersion.new \
-      role: current_user, policy: resource, policy_text: policy_text
-    policy_version.perform_automatic_deletion = perform_automatic_deletion
-    policy_version.delete_permitted = delete_permitted
-    policy_version.update_permitted = update_permitted
-    policy_version.save
-    loader = Loader::Orchestrate.new policy_version
-    loader.load
+    policy_ver = policy_version(
```
jonahx, line 38: In the method signature, remove poetry mode and add params.  

We have a lot of legacy poetry mode but I think any time we touch something for maintenance we need to fix.

app/controllers/policies_controller.rb
```diff
@@ -80,4 +77,30 @@ def concurrent_load _exception
...
   end
+
+  def policy_version role:, policy:, policy_text:, delete_permitted:
+    policy_version = PolicyVersion.new \
+      role: role, policy: policy, policy_text: policy_text
+    policy_version.delete_permitted = delete_permitted
+    policy_version.save
+  end
+
+  def accepted_roles policy_version:, update_permitted:, perform_automatic_deletion:
```
jonahx, line 88: Fix poetry mode to conform to ruby style guide.

app/controllers/policies_controller.rb
```diff
@@ -80,4 +77,30 @@ def concurrent_load _exception
...
+    loader.load(
+      update_permitted: update_permitted, 
+      perform_automatic_deletion: perform_automatic_deletion
+    )
+    loader.new_roles.select do |role|
+      %w(user host).member?(role.kind)
+    end
+  end
+
+  def create_roles accepted_roles
```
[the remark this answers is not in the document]
jonahx: This CC comment is correct, but it may entail a larger refactoring than we want to do right now.

cucumber/api/features/step_definitions/user_steps.rb
```diff
@@ -1,119 +1,53 @@
...
 end
   
-When(/^I have a password$/) do
-  @current_user.password = "password"
-  @current_user.save
-end
-
-Given(/^I set the password for "([^"]*)" to "([^"]*)"$/) do |login, password|
+Given("I set the password for {string} to {string}") do |login, password|
+  # TODO: investigate smell
```
jonahx, line 49: These guys shouldn't be showing up in your PR, since they were already merged.  Maybe you need a rebase?

app/models/loader/orchestrate.rb
```diff
@@ -88,12 +88,14 @@ def load_first_stage
       load_records
     end
 
+    # TODO: consider renaming this method
```
jonahx, line 91: I'd try to come up with some kind of name that roughly describes each group of methods.  even something approximately correct will be better than "first" "second"...etc

app/models/loader/orchestrate.rb
```diff
@@ -384,4 +380,71 @@ def db
...
+    end
+
+    def self.from_policy(policy_version)
+      ReplacePolicy.new(Loader::Orchestrate.new(policy_version))
+    end
+
+    def call
+      @loader.setup_db_for_new_policy
+
+      # Indicates whether records that exist in the database but not in the policy update
```
jonahx, line 416: Comment out of date

app/models/loader/orchestrate.rb
```diff
@@ -384,4 +380,71 @@ def db
...
+      # should be deleted
+      @loader.delete_removed
+      
+      @loader.delete_shadowed_and_duplicate_rows
+
+      # Indicates whether updates to existing data fields are permitted.
```
jonahx, line 422: comment out of date.  both can be removed.

app/controllers/policies_controller.rb
```diff
@@ -57,7 +57,7 @@ def post
   # Returns newly created roles
   def perform(policy_action)
     policy_action.call
-    new_actor_roles = actor_roles(policy_action.new_roles)
+    new_actor_roles = actor_roles(policy_action.loader)
```
jonahx, line 60: Ah, no, you want to add `new_roles` to the 3 objects you created and _delegate_ the implementation to the underlying loader.  What we're doing above violates demeter's law, and forces too much knowledge out into the client code.  

You can accomplish the delegation with the Forwardable gem, but here since it's just one method I'd prefer to write it manually.

spec/models/loader_orchestrate_spec.rb
```diff
@@ -65,58 +70,51 @@ def verify_data path
...
-        load_policy_update 'host_factory.yml'
+        replace_policy_with 'host_factory_new_layer.yml'
+        replace_policy_with 'host_factory.yml'
         verify_data 'updated/host_factory.txt'
       end
       it "doesn't affect a record in a different account" do
         Role.where(Sequel.function("account", :role_id) => 'acct1').delete
         Role.create(role_id: 'acct1:group:the-policy/group-a')
-        load_policy_update 'simple.yml'
+        replace_policy_with 'simple.yml'
```
jonahx, line 93: I'd use policy files for both accounts, rather than dropping down into model level logic.

spec/models/loader_orchestrate_spec.rb
```diff
@@ -16,22 +16,29 @@ def load_base_policy path
...
-      version.update_permitted = update_permitted
-      version.validate
-      expect(version.errors.to_a).to eq([])
-      expect(version.valid?).to be_truthy
-      version.save
-    end
-    Loader::Orchestrate.new(version).tap do |loader|
-      loader.load
-    end
+  def policy_version path
```
jonahx, line 19: Name suggestion: `save_policy`.

The whole "version" naming was a inherited mistake that we should avoid spreading.

spec/models/loader_orchestrate_spec.rb
h-artzi: At first glance I was not able to find a policy which will be able to complete the following step `Role.where(Sequel.function("account", :role_id) => 'acct1').delete`. If there isn't such a policy, does it make sense to switch the role creation to a policy?

spec/models/loader_orchestrate_spec.rb
```diff
@@ -16,7 +16,7 @@ def load_base_policy path
     RootLoader.load 'rspec', policy_path(path)
   end
 
-  def policy_version path
+  def save_policy path
```
jonahx, line 19: Add parens :)

Might as well update the rest in this file since you're touching it.

Issue #1535, opened 2020-05-07
Unable to use host id-based k8s authentication with container name annotation


micahlee:
## Summary
In the k8s authenticator, if the authenticator container is defined as `authn-k8s/authentication-container-name: authenticator`, but the host still uses host id-based authentication, rather than annotation-based, then the `inject-client-cert` request will fail with:
```
CONJ00045E Host does not have a namespace constraint
```

This is because the check for annotation-based authentication only checks for the `authn-k8s/` prefix, rather than the presence of specific identification annotions:

https://github.com/cyberark/conjur/blob/cc532b9ee2b51b4146fb2c9306fdb9fb6a01e724/app/domain/authentication/authn_k8s/application_identity.rb#L211-L213

## Expected Results

Annotation-based authentication should check for the specific presence of the allowed identity annotations defined in:

https://github.com/cyberark/conjur/blob/cc532b9ee2b51b4146fb2c9306fdb9fb6a01e724/app/domain/authentication/authn_k8s/application_identity.rb#L193-L196

PR #1537, opened 2020-05-11
1535 authn k8s id source bug

micahlee:
### What does this PR do?

This PR updates the k8s authenticator to support specifying the authenticator container
name with an annotation, while also using the host ID to specify the k8s resource constraints.

See connected issue for background on why this change was made.

Connected to #1535

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Policy load API endpoints now default to the `application/x-yaml` content-type
   if no content type is provided in the request
   ([conjurinc/dap-support#74](https://github.com/conjurinc/dap-support/issues/74)).
+- The k8s authenticator correctly authenticates an app using the host ID to specify
+  the k8s resource constraints and an annotation to specify the authenticator
+  container name using the "authn-k8s" prefix
+  ([cyberark/conjur#1535](https://github.com/cyberark/conjur/issues/1535),
+  [conjurinc/dap-support#79](https://github.com/conjurinc/dap-support/issues/79)).
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Policy load API endpoints now default to the `application/x-yaml` content-type
   if no content type is provided in the request
   ([conjurinc/dap-support#74](https://github.com/conjurinc/dap-support/issues/74)).
+- The k8s authenticator correctly authenticates an app using the host ID to
+  specify the k8s resource constraints and an annotation to specify the authenticator
+  container name ([cyberark/conjur#1535](https://github.com/cyberark/conjur/issues/1535),
+  [conjurinc/dap-support#79](https://github.com/conjurinc/dap-support/issues/79)).
```
orenbm, line 18: I would state that `The k8s authenticator correctly authenticates an app using the host ID to
  specify the k8s resource constraints and an annotation to specify the authenticator
  container name using the "authn-k8s" prefix`.

If the customer will specify the authenticator container name using the `kubernetes` prefix like this: `kubernetes/authentication-container-name` then it would have worked before as well.

I didn't expect users to use the `authn-k8s` prefix unless they are defining the application identity in annotations but this fix is in place as we can't assume that.

micahlee: Sure, I'll make that change. 

> I didn't expect users to use the authn-k8s prefix unless they are defining the application identity in annotations but this fix is in place as we can't assume that.

Yes, this comes from a support case where this very scenario occurred :)

micahlee:
@orenbm would you mind approving again with the CHANGELOG update when you have a chance?

Thanks!

Issue #1540, opened 2020-05-12
Remove usage of `tap` when raising error and logging messages


orenbm:
Our codebase has many occurrences where we use `tap` to raise errors or log messages. 

For example:
```
def role
  @role ||= @resource_class[role_id].tap do |role|
    raise Errors::Authentication::Security::RoleNotFound, role_id unless role
  end
end
```

This makes our code harder to understand and we should use a more straight-forward approach.
We should go over the code and remove these occurrences. the example above can change to:
```
def role
  fetched_role or raise Errors::Authentication::Security::RoleNotFound
end

def fetch_role(role_id)
  @fetched_role ||= @resource_class[role_id]
end
```

In logging cases, we cannot use `or` so we can change
```
def role
  @resource_class[role_id].tap do |role|
    Rails.logger.debug("some message regarding #{role}")
  end
end
```

to 
```
def role
  return_obj = @resource_class[role_id]
  Rails.logger.debug("some message regarding #{return_obj}")
  return_obj
end
```

PR #1553, opened 2020-05-17
Do not use `tap` method for logs and errors

orenbm:
Connected to #1540 

We decided not to use `tap` when we want to raise an error
or log a message as it is less straight-forward to the common
Ruby developer

### DIFF
--- a/cucumber/_authenticators_common/features/support/authenticator_helpers.rb
+++ b/cucumber/_authenticators_common/features/support/authenticator_helpers.rb
@@ -11,9 +11,9 @@ module AuthenticatorHelpers
   )
 
   def validated_env_var(var)
-    ENV[var].tap do |env_var_value|
-      raise MissingEnvVariable, var if env_var_value.blank?
-    end
+    env_var_value = ENV[var]
+    raise MissingEnvVariable, var if env_var_value.blank?
+    env_var_value
   end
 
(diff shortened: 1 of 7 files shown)


cucumber/_authenticators_common/features/support/authenticator_helpers.rb
```diff
@@ -11,9 +11,9 @@ module AuthenticatorHelpers
   )
 
   def validated_env_var(var)
-    ENV[var].tap do |env_var_value|
-      raise MissingEnvVariable, var if env_var_value.blank?
-    end
+    env_var_value = ENV[var]
```
sigalsax, line 14: we want also want to add @ here so it is `@env_var_value`?

orenbm: not really. this data is not memoized and is calculated each time. `env_var_value` is merely a placeholder as we want to also log the data.

orenbm:
``` 
def provider_uri
        @provider_uri ||= azure_authenticator_secrets["provider-uri"].tap do |provider_uri_value|
          provider_uri_value << "/" unless provider_uri_value.end_with?('/')
        end
      end
```

this is actually a good use of tap as it manipulates the object before returning it.

Issue #1539, opened 2020-05-12
Enhance logs readability by separating log levels


orenbm:
As a Conjur operator, I want to read the logs, so that I can easily understand the application flow

Currently, when we set the log level to INFO, we do not get any log messages regarding what happened in the application flow. However, if we change the log level to DEBUG we will get too many log messages as we log the DB messages at that level, alongside our own debug level logs. 

More info can be found in the [design](https://github.com/cyberark/conjur/blob/logs-design/design/logging.md)

DoD:
- [ ] Changes were implemented according to the design
- [ ] Add an entry in the CHANGELOG regarding this change
- [ ] Update the docs

orenbm:
@alexkalish @InbalZilberman As this is a customer-facing change, i'd like to get your thoughts on it. thanks :)

jtuttle:
Will stacktrace be included in INFO or just DEBUG? I would lean towards only including it in DEBUG.

micahlee:
I think this would **greatly** enhance the use of the logs in customer interactions, and give customers and field better tools to solve issues they run into on their own.

I only have one comment regarding:
> Change all occurrences of logger.debug to logger.info

I don't think this should just be a universal replacements, but we need clearer guidelines on who the intended audience is for each log level. 

In my opinion, the assumed audience for `DEBUG` should be R&D. If you're looking at that level, the assumption should be the fix is going to require a code change. Field and end-user operators should be able to view (perhaps verbose) information at a higher level to identify problems they are able to fix with configuration changes. 

So an example of what I think should remain a `debug` level log is when we log exception stack traces. There isn't really an action a customer can take with a stack trace, so this is only useful to developers.

I'll include @rafis3 reference to DataDog's articulation of these levels as well, as an example of how this could be defined: https://www.datadoghq.com/blog/managing-rails-application-logs/

jvanderhoof:
Better separation of log verbosity would certainly help.  For reference, here are the above Data Dog recommendations of levels:

- **Debug**: diagnostic information for developers and system administrators, including database calls or inspecting object attributes. This is the most verbose log level.
Info: useful information about normal application operations such as services starting or stopping.
- **Warn**: operations that an application can easily recover from but should be addressed soon. This could include using an out-of-date gem or retrying an operation.
- **Error**: errors that cause an operation to fail (e.g., missing data or files) but not the application. The issue should be resolved soon, but the application can log the exception and continue running.
[...]

PR #1554, opened 2020-05-17
Move log and error messages to centralized files

orenbm:
Connected to #1539 

We maintain our logs and errors in the centralized files `logs.rb` & `errors.rb`. 

We had some log messages written outside these files so we moved them there.

It is best to review this PR via the commits as they are split logically. We will squash-merge this PR when it's ready. 

This PR blocks [this issue](https://github.com/cyberark/conjur/issues/1539) as it cleans up the code a bit so that issue's PR will be easier to review.


app/domain/errors.rb
```diff
@@ -171,10 +171,20 @@ module AuthnOidc
...
+          msg:  "Invalid or expired AWS headers: {0}",
           code: "CONJ00018E"
         )
 
+        VerificationError = ::Util::TrackableLogMessageClass.new(
+          msg:  "Verification of IAM identity failed with exception: {0-exception}",
+          code: "CONJ00063E"
+        )
+
+        VerificationErrorCode = ::Util::TrackableLogMessageClass.new(
```
Tovli, line 183: I suggest you'll use more meaningful names, as our approach of separating the log text to a separate files means that the developer is not aware of what is exactly the text from first glance

orenbm:
@Tovli regarding your comment:

> I would prefer to have explicit collection of info level logs

This PR is a preparation for another one where all the messages will change from DEBUG to INFO so we don't to add a separation of files in this one.

app/controllers/concerns/authorize_resource.rb
```diff
@@ -21,7 +21,13 @@ def authorize_many(resources, privilege)
 
   def auth(user, privilege, resource)
     unless user.allowed_to?(privilege, resource)
-      logger.info "Current user '#{user.role_id}' is not permitted to '#{privilege}' resource '#{resource.resource_id}'"
+      logger.info(
```
nessiLahav, line 24: Why not using logger.error instead of info?

orenbm: According to Datadog's post about [managing Rails application logs](https://www.datadoghq.com/blog/managing-rails-application-logs/), error logs indicate `errors that cause an operation to fail (e.g., missing data or files) but not the application`. In this case the operation didn't fail and we got to the conclusion successfully. We want to notify the client that there was an error in his flow so it's in info

PR #1555, opened 2020-05-18
Add design doc for logging

orenbm:
Connected to #1539

Add a design doc for improving the Conjur log. More info can be found in the connected issue and in the design itself.

It's best to view the doc [here](https://github.com/cyberark/conjur/blob/logs-design/design/logging.md) and add your comments [here](https://github.com/cyberark/conjur/pull/1555/files).

### DIFF
--- a/design/logging.md
+++ b/design/logging.md
@@ -1,0 +1,518 @@
+# Solution Design - Redefine Conjur logs
+
+## Table of Contents
+
+- [Issue description](#issue-description)
+- [Solution](#solution)
+  * [Backwards compatibility](#backwards-compatibility)
+  * [Affected Components](#affected-components)
+- [Security](#security)
+- [Test Plan](#test-plan)
+- [Documentation](#documentation)
+- [Version update](#version-update)
+- [Breaking](#breaking)
+
+## Issue description
+
+As a Conjur operator, I want to read the logs, so that I can easily understand the application flow.
+
+Currently, when we set the log level to INFO (which is the default value),
+we do not get any log messages regarding what happened in the application flow. 
+However, if we change the log level to DEBUG we get too many log messages as we 
+log the DB messages at that level, alongside our own debug level logs. 
+
+For example, let's look at a log sample for the use-case where a user sends an 
+`authn-oidc` request with a username that is not defined in Conjur. If the log level
+is set to INFO (as the default value is), then the log will look like this:
+```
+ INFO 2020/05/18 09:49:20 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Started POST "/authn-oidc/keycloak/cucumber/authenticate" for 127.0.0.1 at 2020-05-18 09:49:20 +0000
+ INFO 2020/05/18 09:49:21 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Processing by AuthenticateController#authenticate_oidc as */*
+ INFO 2020/05/18 09:49:21 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421]   Parameters: {"service_id"=>"keycloak", "account"=>"cucumber"}
+ INFO 2020/05/18 09:49:21 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Completed 401 Unauthorized in 94ms
+```
+
+As can be seen, there is not enough data to understand why the request failed.
+
+If the log level is set to DEBUG, then the log will look like this:
+```
+INFO 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] Started POST "/authn-oidc/keycloak/cucumber/authenticate" for 127.0.0.1 at 2020-05-18 09:45:41 +0000
+ INFO 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] Processing by AuthenticateController#authenticate_oidc as */*
+ INFO 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Parameters: {"service_id"=>"keycloak", "account"=>"cucumber"}
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (0.5ms)  BEGIN
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (0.9ms)  SELECT * FROM "roles" WHERE "role_id" = 'cucumber:user:admin'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (0.7ms)  SELECT * FROM "resources" WHERE "resource_id" = 'cucumber:variable:conjur/authn-oidc/keycloak/provider-uri'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (0.8ms)  SELECT * FROM "resources" WHERE "resource_id" = 'cucumber:variable:conjur/authn-oidc/keycloak/id-token-user-property'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (1.6ms)  SELECT * FROM "secrets" WHERE ("secrets"."resource_id" = 'cucumber:variable:conjur/authn-oidc/keycloak/provider-uri') ORDER BY "version" DESC LIMIT 1
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (0.7ms)  SELECT * FROM "secrets" WHERE ("secrets"."resource_id" = 'cucumber:variable:conjur/authn-oidc/keycloak/id-token-user-property') ORDER BY "version" DESC LIMIT 1
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00023D Concurrency limited cache concurrent requests updated to '1'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00007D Working with Identity Provider https://keycloak:8443/auth/realms/master
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00008D Identity Provider discovery succeeded
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00009D Fetched Identity Provider keys from provider successfully
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00016D Rate limited cache updated successfully
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00021D Concurrency limited cache updated successfully
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00023D Concurrency limited cache concurrent requests updated to '0'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00017D Fetched Identity Provider keys from cache successfully
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00005D Token decoded successfully
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] CONJ00004D Extracted username 'not_in_conjur' from ID token field 'preferred_username'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (0.7ms)  SELECT * FROM "roles" WHERE "role_id" = 'cucumber:user:admin'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (1.4ms)  SELECT * FROM "roles" WHERE "role_id" = 'cucumber:user:admin'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (1.5ms)  SELECT * FROM "roles" WHERE "role_id" = 'cucumber:user:admin'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (0.7ms)  SELECT * FROM "resources" WHERE "resource_id" = 'cucumber:webservice:conjur/authn-oidc/keycloak'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (0.6ms)  SELECT * FROM "roles" WHERE "role_id" = 'cucumber:user:not_in_conjur'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305]   Sequel::Postgres::Database (0.6ms)  SELECT * FROM "roles" WHERE ("role_id" = 'cucumber:user:not_in_conjur') LIMIT 1
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] Authentication Error: #<Errors::Authentication::Security::RoleNotFound: CONJ00007E 'not_in_conjur' not found>
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] /src/conjur-server/app/domain/authentication/validate_webservice_access.rb:59:in `validate_user_is_defined'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] /src/conjur-server/app/domain/authentication/validate_webservice_access.rb:34:in `call'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] (eval):7:in `call'
+DEBUG 2020/05/18 09:45:41 +0000 [pid=298] [origin=127.0.0.1] [request_id=b7390d52-73a1-41ff-add3-9f4e37395e55] [tid=305] /src/conjur-server/app/domain/authentication/validate_security.rb:47:in `validate_user_has_access_to_webservice'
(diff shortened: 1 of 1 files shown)


design/logging.md
```diff
@@ -0,0 +1,495 @@
...
+/var/lib/gems/2.5.0/gems/puma-3.12.4/lib/puma/thread_pool.rb:135:in `block in spawn_thread'
+ INFO 2020/05/17 17:38:07 +0000 [pid=300] [origin=127.0.0.1] [request_id=d3f54689-ed9d-4c66-858a-d01d9c7adbc9] [tid=317] Completed 401 Unauthorized in 447ms
+```
+
+Here we can see that the log is more robust and has all the data that can help a developer
+to investigate the issue, including the stack trace and the DB messages.
+
+### Backwards compatibility
+
+It's important to say that the UX will not change by default. We will do this by changing
```
rafis3, line 446: While this doesn't change for new setups, it will change for users that have a specific value set for `CONJUR_LOG_LEVEL`. I think it's worth mentioning that when not using the default value, there will be an expected behavior change.

design/logging.md
```diff
@@ -0,0 +1,495 @@
...
+the default value for the log level from `info` to `warn`. This way the log will show the
+same data that it shows today.
```
orenbm, line 448: Reading this again now and I see that this is incorrect. The UX will be different if we change the default log level to `warn`. For example, in the use-case described above with `authn-oidc` the log will be empty and will not include these log messages:
```
 INFO 2020/05/18 09:49:20 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Started POST "/authn-oidc/keycloak/cucumber/authenticate" for 127.0.0.1 at 2020-05-18 09:49:20 +0000
 INFO 2020/05/18 09:49:21 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Processing by AuthenticateController#authenticate_oidc as */*
 INFO 2020/05/18 09:49:21 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421]   Parameters: {"service_id"=>"keycloak", "account"=>"cucumber"}
 INFO 2020/05/18 09:49:21 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Completed 401 Unauthorized in 94ms
```
However, before this change it was printed by default without changing the log level. 

We have 2 options here:
1. Still change the default level to `warn` and change the current log behaviour
2. Leave the default log level at `info` which will add more logs to the current log as it will include the logs that we currently log in DEBUG (all the `CONJ` messages).

Option 2 is better in my opinion and will have the following implications:
1. We will not need to make any changes in `evoke`, `conjur-ui` & `ldap-sync`
2. We will need to get a security approval for this change as more data is written to the logs
3. The UX will change as the log will have more log messages in it by default. However, we don't have many logs added as we will add only our logs and not the DB and stacktrace.

@rafis3 wdyt?

design/logging.md
```diff
@@ -0,0 +1,495 @@
...
+### Affected Components
+
+- Conjur
+- DAP
+  - The DAP log contains not only the Conjur log but also the logs of `evoke`, `conjur-ui` & `ldap-sync`. We will update
+    their logs to fit our logs classification defined above.  
+    let's address each component to decide on the action items:
+    - `evoke`: this component doesn't write to the log in INFO level so it won't be affected by the change.
+    - `conjur-ui`: this components writes messages in DEBUG, INFO and WARN and should be addressed.
+    - `ldap-sync`: this components writes messages in DEBUG, INFO and WARN and should be addressed.
```
rafis3, line 459: Syslog-ng logs as well

design/logging.md
rafis3: @orenbm I like the second option better as well. My main concern is that more logs means more space required on disk. For a loaded environment, this will produce more output that the Conjur/DAP instances will need to have enough space to handle. It also means more I/O which can affect performance on very high loads. So in a production environment, I would recommend keeping it at `warn` unless there is a problem. But I'd prefer that by default, when you start working with the product, you'd get an out of the box log level that actually helps you understand the product and solve problems.

design/logging.md
```diff
@@ -0,0 +1,495 @@
...
+As you can see, this log shows too much data on why the request failed.
+This makes the log very hard to debug, and it's clear that we need some solution for this.
+
+## Solution
+
+We will redefine our log levels to the following:
+1. DEBUG - messages that can help Conjur developers investigate an issue. These messages will include
+           stack traces and DB messages.
+2. INFO - messages that can help Conjur users and operators investigate an issue. These messages will include
+          the messages that we print in our logs (messages with the `CONJ` prefix).
```
eladkug, line 228: I think that messages with prefix 'CONJ' with suffix 'E' need to be WARN (CONJ00007E for example). If i take your example:
`WARN 2020/05/17 17:39:59 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] Authentication Error: #<Errors::Authentication::Security::RoleNotFound: CONJ00007E 'not_in_conjur' not found>`.

design/logging.md
```diff
@@ -0,0 +1,542 @@
...
+
+Note: The page above exists only for DAP as we have an issue in Conjur where it's impossible 
+to restart the server. Once that is solved we can create the page for Conjur as well.
+
+We will also add an entry to the CHANGELOG of this project.
+
+## Version update
+
+- Conjur
+- DAP
```
orenbm, line 534: @jvanderhoof @alexkalish do we publish a version for changes like this or are they accumulated to the next release? same question for Conjur and DAP.

design/logging.md
micahlee: I agree generally. However, this particular line has [a long history](https://github.com/cyberark/conjur/issues/1377) a needs approval from our security architects before we can this can show up in the Conjur logs by default (which is currently `WARN` level).

orenbm: the current default log level is `info`. I think that this message should be written in `INFO` (and the code should change accordingly). I address this issue in the doc.

design/logging.md
```diff
@@ -0,0 +1,542 @@
...
+ INFO 2020/05/17 17:38:07 +0000 [pid=300] [origin=127.0.0.1] [request_id=d3f54689-ed9d-4c66-858a-d01d9c7adbc9] [tid=317] Completed 401 Unauthorized in 447ms
+```
+
+Here we can see that the log is more robust and has all the data that can help a developer
+to investigate the issue, including the stack trace and the DB messages.
+
+### Backwards compatibility
+
+As you can see, the UX has changed a bit, as the log will contain more data by default.
+The default log level is INFO, so instead of getting this log for the use-case described above:
```
jvanderhoof, line 461: I'd recommend we change the default log level to `WARN`.  `INFO` feels much too verbose to consider running in production (what I'd expect the default level to be acceptable for).

design/logging.md
```diff
@@ -0,0 +1,542 @@
...
+
+However, I don't think that changing the log level should add any security concern as the
+Conjur log is visible to the same persona that can change the log level, so it really doesn't 
+give the user any data that wasn't available for them before.
+
+## Test Plan
+
+We will not introduce new tests in this effort and the regression tests will be enough.
+
+## Documentation
```
jvanderhoof, line 518: Can we please also add all messages to the [DAP Wiki](https://github.com/cyberark/dap-wiki)?  This source will be used by the field and support (as well as our internal engineers :smile:).

design/logging.md
jvanderhoof: It really depends on the need.  If we want to get a change into the hands of a customer, we'll cut a release.

jvanderhoof:
@orenbm, a big thanks for taking the lead on this.  I really think it's going to impact our team and customers in a really positive way!

design/logging.md
```diff
@@ -0,0 +1,542 @@
...
+- Debug: diagnostic information for developers and system administrators, including 
+  database calls or inspecting object attributes. This is the most verbose log level.
+- Info: useful information about normal application operations such as services starting or stopping.
+- Warn: operations that an application can easily recover from but should be addressed 
+  soon. This could include using an out-of-date gem or retrying an operation.
+- Error: errors that cause an operation to fail (e.g., missing data or files) but 
+  not the application. The issue should be resolved soon, but the application can log the exception and continue running.
+  
+We will go over our log messages and change their log level according to the above.
+DB calls will stay in DEBUG, whereas most of the logs that we write on our own (i.e `CONJ<some-number>D`)
```
Tovli, line 236: Can you elaborate why we are coupling the log code with the log level? Would appreciate a reference (if there's such)

design/logging.md
```diff
@@ -0,0 +1,542 @@
...
+will be logged in INFO, and their codes will change accordingly (i.e `CONJ<some-number>D`).
+
+We will also change the codes for some of our error messages (i.e `CONJ<some-number>E`)
+as they will be written to the log in `INFO` level (today they are written in `DEBUG`). 
+The reason they should be logged in `INFO` and not in `ERROR`
+is that an `ERROR` log indicates `errors that cause an operation to fail (e.g., 
+missing data or files) but not the application`. In some of our error messages,
```
Tovli, line 243: Not critical but note that datadog/rails have also fatal which we don't use so I guess our Errors can also be Fatal

design/logging.md
```diff
@@ -0,0 +1,542 @@
...
+we will not get the useful `INFO` logs that we got earlier (and that _will_ be a breaking change). 
+
+I suggest that we stay with `info` as the default log level and add a note in the 
+documentation stating that for a loaded environment it is best to change the log 
+level to `warn` to reduce the log size.  
+
+### Affected Components
+
+- Conjur
+- DAP
```
Tovli, line 501: you include the UI in this effort?

design/logging.md
orenbm: thanks @jvanderhoof . I was going backwards and forward on this one and it's the biggest question of this design. Let's get some thoughts from @alexkalish, @InbalZilberman and @rafis3 on this and decide.

design/logging.md
jvanderhoof: What do you think about decoupling the log level from the message name? With our current approach, changing a message level requires we change the name message level (`CONJ00012I` -> `CONJ00012W` if we want to change a message from "Info" to "Warn". The log level is present in the error message output (`INFO`, `DEBUG`, etc.).  Can we change the codes to remove the log level (ex. `CONJ<some-number>` instead of `CONJ<some-number>D` for a "Debug" level message)?

jvanderhoof:
With a few minor suggestions, I really like the direction of this proposal. What I'm not clear on is the process for identifying the messages that should be re-classified into each level (debug, info, warn, error). It feels like we'll need to establish a common level across all of Conjur. This is going to be a substantial effort (but will be very valuable for troubleshooting).  

Given your familiarity with `authn-azure`, would it make sense to start in this authenticator?  `authn-K8s` is another great option given the high number of support cases we deal with related to Kubernetes/OCP.  Once we establish a common pattern, we can iteratively push through the rest of the Conjur workflows.

design/logging.md
alexkalish: I agree with @jvanderhoof.  It would be nice to have stable log codes that do not change based on their log level.  But maybe that 🚢 has sailed, since our codes already include them?  Also, I think it's probably more important that error codes do not change, as opposed to informational log messages.

design/logging.md
alexkalish: I think that `WARN` is a fine default.  Using a product for the first time in production, I would prefer to not worry about high log volume.

design/logging.md
```diff
@@ -0,0 +1,542 @@
...
+Currently, when we set the log level to INFO (which is the default value),
+we do not get any log messages regarding what happened in the application flow. 
+However, if we change the log level to DEBUG we get too many log messages as we 
+log the DB messages at that level, alongside our own debug level logs. 
+
+For example, let's look at a log sample for the use-case where a user sends an 
+`authn-oidc` request with a username that is not defined in Conjur. If the log level
+is set to INFO (as the default value is), then the log will look like this:
+```
+ INFO 2020/05/18 09:49:20 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Started POST "/authn-oidc/keycloak/cucumber/authenticate" for 127.0.0.1 at 2020-05-18 09:49:20 +0000
```
alexkalish, line 28: So, I really dislike the default Rails request logging in production.  Spreading a request across multiple lines is fine for development, but is not conducive to log aggregation and reporting.  In the past, I've always configured [lograge](https://github.com/roidrage/lograge) for production environments.  @orenbm, @jvanderhoof: What do you think?

design/logging.md
orenbm: I think that we can still change that to not have the codes as it is better. We can go over our docs are remove the suffix from all the instances.

design/logging.md
orenbm: yes, we don't have any log messages that crash our server.

design/logging.md
orenbm: yes, updated now.

design/logging.md
orenbm: In the new approach - where the default log level is warn - we won't have these messages.

orenbm:
@jvanderhoof @alexkalish @Tovli I changed the default log level to `WARN` and added a section for removing the suffix from the `CONJ` codes. can you please re-review?

design/logging.md
alexkalish: So, if we go to `WARN`, we lose basic request logging?  I missed that detail and I'm not sure if that makes sense.  I would expect basic request logging at the default log level in an application like this.  Maybe we should go back to `INFO`.  Thoughts, @jvanderhoof?

design/logging.md
alexkalish: @orenbm: I think that our codes match the other PAS products, right?  I would like to drop them too, but are we breaking a cross-product requirement?  Could @InbalZilberman perhaps ask about that?

design/logging.md
```diff
@@ -0,0 +1,518 @@
...
+  soon. This could include using an out-of-date gem or retrying an operation.
+- Error: errors that cause an operation to fail (e.g., missing data or files) but 
+  not the application. The issue should be resolved soon, but the application can log the exception and continue running.
+  
+We will go over our log messages and change their log level according to the above.
+DB calls will stay in DEBUG, whereas most of the logs that we write on our own (i.e `CONJ<some-number>D`)
+will be logged in INFO.
+
+We will also remove the suffix from the log codes as the log level is already written 
+in the log message (e.g `INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00008I Identity Provider discovery succeeded`).
```
sgnn7, line 240: This needs to be ensured to retain that info in some way when shipped to syslog/DB

design/logging.md
```diff
@@ -0,0 +1,518 @@
...
+```
+
+Here we can see that the log is more robust and has all the data that can help a developer
+to investigate the issue, including the stack trace and the DB messages.
+
+### Default log level
+
+Currently, the default log level for Conjur OSS and DAP is `INFO`. We will change
+it to `WARN` so the default level will not have all the log messages that are now
+added to the `INFO` level from `DEBUG`.
```
sgnn7, line 457: I'm not sure I'm on board with upping the log level to warn. The reason for my opinion here is that when problems occur, customers don't want to modify their servers to enable debugging but want to only send us the logs which I think will not provide enough troubleshooting information when set to "warn" even with the changes suggested for authentication issues. This makes support issues harder to fix and we would not be able to fix many inconsistent issues since we would be doing things reactively rather than proactively for collecting the info.

From a customer's perspective, they also may want to see when a person logs in, retrieves a secret, etc and other informational info that is not a warning level message and should not be one. I am not a fan of log4j but I do like their [breakdown of log levels](https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm) and some other vantage points from [this](https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels) SO thread. In general though, we also have to consider that we have a security product and that our logging is much more valuable than logging of a random system service.

design/logging.md
orenbm: Yes, you can read about it more [here](https://github.com/cyberark/conjur/blob/logs-design/design/logging.md#backwards-compatibility).

to summarize, if we look at the use-case presented in the doc (unsuccessful request).

The current log is:
```
INFO 2020/05/18 09:49:20 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Started POST "/authn-oidc/keycloak/cucumber/authenticate" for 127.0.0.1 at 2020-05-18 09:49:20 +0000
 INFO 2020/05/18 09:49:21 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Processing by AuthenticateController#authenticate_oidc as */*
 INFO 2020/05/18 09:49:21 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421]   Parameters: {"service_id"=>"keycloak", "account"=>"cucumber"}
 INFO 2020/05/18 09:49:21 +0000 [pid=411] [origin=127.0.0.1] [request_id=0487ea6a-a974-481a-8b7f-d7d62e96aec6] [tid=421] Completed 401 Unauthorized in 94ms
```

If we will have `WARN` as default in the new behaviour, the log will be empty.
If we will have `INFO` as default in the new behaviour, the log will be:
```
INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] Started POST "/authn-oidc/keycloak/cucumber/authenticate" for 127.0.0.1 at 2020-05-17 17:39:58 +0000
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] Processing by AuthenticateController#authenticate_oidc as */*
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413]   Parameters: {"service_id"=>"keycloak", "account"=>"cucumber"}
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00023 Concurrency limited cache concurrent requests updated to '1'
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00007 Working with Identity Provider https://keycloak:8443/auth/realms/master
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00008 Identity Provider discovery succeeded
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00009 Fetched Identity Provider keys from provider successfully
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00016 Rate limited cache updated successfully
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00021 Concurrency limited cache updated successfully
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00023 Concurrency limited cache concurrent requests updated to '0'
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00017 Fetched Identity Provider keys from cache successfully
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00005 Token decoded successfully
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] CONJ00004 Extracted username 'not_in_conjur' from ID token field 'preferred_username'
 INFO 2020/05/17 17:39:59 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] Authentication Error: #<Errors::Authentication::Security::RoleNotFound: CONJ00007 'not_in_conjur' not found>
 INFO 2020/05/17 17:39:59 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] Completed 401 Unauthorized in 143ms
```

So we need to choose between less than today to more than today. I actually agree with you that the UX will be better without those 4 lines because it adds data to the log that we can live without. They are also quite `DEBUG`y. If the user wants more data they can move to `INFO` and get the 4 lines above + our additional info.

I really don't think there's a use case where someone will want just the basic info that we log today without the additional data that we log with the `CONJ` prefix.

design/logging.md
orenbm: Yes, the log level is written also in Syslog.

design/logging.md
orenbm: but that's true also today. Today, we want to get the logs in DEBUG level as without it it doesn't have enough value for us in debugging. So the current default level also doesn't help us.

design/logging.md
orenbm: @sgnn7 can you take a look at this thread? In this one i state the outcome of having `INFO` as the default value and having `WARN` as the default value. 

Please say what you think is better, having the above in mind.

sgnn7: I think that both options in this example (Warn = no logs, Info = too much logging) are not great experience for both the operator and the developer. No messages would mean no tracing of functionality without changing the log level and the other one is just too noisy. The best blend of the two would be taking the output of the INFO and pushing all mostly irrelevant messages to debug level so you would get a much better condensed idea of what happened without flooding the logs:
```
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] Started POST "/authn-oidc/keycloak/cucumber/authenticate" for 127.0.0.1 at 2020-05-17 17:39:58 +0000
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] Processing by AuthenticateController#authenticate_oidc as */*
 INFO 2020/05/17 17:39:58 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413]   Parameters: {"service_id"=>"keycloak", "account"=>"cucumber"}
 INFO 2020/05/17 17:39:59 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] Authentication Error: #<Errors::Authentication::Security::RoleNotFound: CONJ00007 'not_in_conjur' not found>
 INFO 2020/05/17 17:39:59 +0000 [pid=401] [origin=127.0.0.1] [request_id=f39fb7e2-59f9-4f3a-89a9-bb9a50b09860] [tid=413] Completed 401 Unauthorized in 143ms
```

I know some of this seems like we are adding unnecessary info in the logs but they are critical from a support standpoint since they don't require changing the customer's infrastructure to narrow down problems when they happen.

orenbm: Funny enough, this is the current state :)

I actually think that this level doesn't give value as it is somewhere in the middle. I write about it more [here](https://github.com/cyberark/conjur/blob/logs-design/design/logging.md#issue-description).

I would actually like to get some ideas from SEs. I'll send a link to this thread in Slack as i can't tag them.

jodyhuntatx: More info is better unless it creates other complications like filling up disk space or exposing stuff people shouldn't see. One advantage  of docker logs is "--since ..." which lets you easily view recent log messages w/o parsing timestamps. Current default log levels are pretty useless. When you bump the log level to debug you get all the backtrace garbage to wade through. So I'm for whatever gets more useful info into the logs as a responsible default. Currently when debugging K8s 401s I turn set LOG_LEVEL to DEBUG, RAILS_ENV to production (not sure if that's required), deploy the Follower or app, then run "docker logs <container> --since 1m | grep CONJ" and that generally shows what the issue is. In a perfect world I'd just need to do "docker logs <container> --since 1m" w/o changing log levels or grepping.

orenbm: thanks @jodyhuntatx, that's very helpful!

orenbm:
Closing this proposal as we didn't get to an agreement on the desired UX. However, I did create [this PR](https://github.com/cyberark/conjur/pull/1657) that addresses one of the issues raised in this design and that will give us a lot of value.

PR #1561, opened 2020-05-20
Add a solution design template

orenbm:
Having a template for solution designs (and other templates that will follow)
can help us maintain a standard for future designs.

# Notes for the reviewer
- Can you think of a section that is missing?
- Would you remove any section?
- Are the comments clear enough?

Note that the MD file will not include the comments that help the write so they don't need to be deleted, as can be seen [here](https://github.com/cyberark/conjur/blob/solution-design-template/design/templates/solution_design.md).

### DIFF
--- a/design/templates/solution_design.md
+++ b/design/templates/solution_design.md
@@ -1,0 +1,71 @@
+# Solution Design - Template
+[//]: # (Change the title above from "Template" to your design's title) 
+
+## Table of Contents
+[//]: # (You can use this tool to generate a TOC - https://ecotrust-canada.github.io/markdown-toc/) 
+
+## Glossary
+[//]: # (Describe terms that will be used throughout the design)
+[//]: # (You can use this tool to generate a table - https://www.tablesgenerator.com/markdown_tables#)
+
+| **Term** | **Description** |
+|----------|-----------------|
+|          |                 |
+|          |                 |
+
+## Useful links
+[//]: # (Add links that may be useful for the reader)
+
+## Background
+[//]: # (Give relevant background for the designed feature. What is the motivation for this solution?)
+
+## Issue description
+[//]: # (Elaborate on the issue you are writing a solution for)
+
+## Solution
+[//]: # (Elaborate on the solution you are suggesting in this page. Address the functional requirements and the non functional requirements that this solution is addressing. If there are a few options considered for the solution, mention them and explain why the actual solution was chosen over them. Add an execution plan when relevant. It doesn't have to be a full breakdown of the feature, but just a recommendation to how the solution should be approached.)
+
+### Design
+[//]: # (Add any diagrams, charts and explanations about the design aspect of the solution. Elaborate also about the expected user experience for the feature)
+
+### Backwards compatibility
+[//]: # (Address how you are going to handle backwards compatibility, if necessary)
+
+### Performance
+[//]: # (Elaborate on whether your solution will affect the product's performance, and how)
+
+### Affected Components
+[//]: # (Address the components that will be affected by your solution [Conjur, DAP, clients, integrations, etc.])
+
+## Security
+[//]: # (Are there any security issues with your solution? Even if you mentioned them somewhere in the doc it may be convenient for the security architect review to have them centralized here)
+
+## Test Plan
+[//]: # (Fill in the table below to depict the tests that should run to validate your solution)
+[//]: # (You can use this tool to generate a table - https://www.tablesgenerator.com/markdown_tables#)
+
+| **Title** | **Given** | **When** | **Then** | **Comment** |
+|-----------|-----------|----------|----------|-------------|
+|           |           |          |          |             |
+|           |           |          |          |             |
+
+## Logs
+[//]: # (Fill in the table below to depict the log messages that can enhance the supportability of your solution)
+[//]: # (You can use this tool to generate a table - https://www.tablesgenerator.com/markdown_tables#)
+
+| **Scenario** | **Log message** |
+|--------------|-----------------|
+|              |                 |
+|              |                 |
+
+### Audit 
+[//]: # (Does this solution require additional audit messages?)
+
+## Documentation
+[//]: # (Add notes on what should be documented in this solution. Elaborate on where this should be documented. If the change is in open-source projects, we may need to update the docs in github too. If it's in Conjur and/or DAP mention which products are affected by it)
+
+## Open questions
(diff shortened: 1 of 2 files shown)


InbalZilberman:
Hi, I like this template very much
Few things I am missing:
1. Audit section. We need to always think if this new feature has Audit implications.
2. The word options- a good design in my opinion has thought of several options and took from it the best one. Having the word option would encourage imo people to think of more than one way to solve a problem

orenbm:
Thanks @InbalZilberman for your comments. 

> The word options- a good design in my opinion has thought of several options and took from it the best one. Having the word option would encourage imo people to think of more than one way to solve a problem

I'm not sure i follow. In the `Solution` section I say:
`If there are a few options considered for the solution, mention them and explain why the actual solution was chosen over them.`
isn't that what you want the language to be?

design/templates/solution_design.md
```diff
@@ -0,0 +1,74 @@
...
+[//]: # (Add links that may be useful for the reader)
+
+## Background
+[//]: # (Give relevant background for the designed feature. What is the motivation for this solution?)
+
+## Issue description
+[//]: # (Elaborate on the issue you are writing a solution for)
+
+## Solution
+[//]: # (Elaborate on the solution you are suggesting in this page. Address the functional requirements and the non functional requirements that this solution is addressing. If there are a few options considered for the solution, mention them and explain why the actual solution was chosen over them. Add an execution plan when relevant. It doesn't have to be a full breaking of the feature, but just a recommendation to how the solution should be approached.)
```
izgeri, line 26:
```suggestion
[//]: # (Elaborate on the solution you are suggesting in this page. Address the functional requirements and the non functional requirements that this solution is addressing. If there are a few options considered for the solution, mention them and explain why the actual solution was chosen over them. Add an execution plan when relevant. It doesn't have to be a full breakdown of the feature, but just a recommendation to how the solution should be approached.)
```

design/templates/solution_design.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Design
+[//]: # (Add any diagrams, charts and explanations about the design aspect of the solution. Elaborate also about what the expected user experience for the feature)
```
izgeri, line 29:
```suggestion
[//]: # (Add any diagrams, charts and explanations about the design aspect of the solution. Elaborate also about the expected user experience for the feature)
```

design/templates/solution_design.md
```diff
@@ -0,0 +1,74 @@
...
+|              |                 |
+
+### Audit 
+[//]: # (Does this solution require additional audit messages?)
+
+## Documentation
+[//]: # (Add notes on what should be documented in this solution. Elaborate on where this should be documented. If the change is in open-source projects, we may need to update the docs in github too. If it's in Conjur and/or DAP mention which products are affected by it)
+
+## Version update
+[//]: # (Does this solution require a version update? if so, add a list of the projects that should be bumped)
```
izgeri, line 68: this section doesn't make sense to me - in order to use any changes we make to this codebase, we must include the changes in a new version

I think what you are really asking is for other dependencies that may also need to be updated as a result of this change. wouldn't this be covered in affected components?

Maybe that section should be "considerations" and include this bullet (maybe reworded to be a bit more clear), the "affected components" bullet, and the "backwards compatibility" bullet?

design/templates/solution_design.md
```diff
@@ -0,0 +1,74 @@
...
+[//]: # (Elaborate on the issue you are writing a solution for)
+
+## Solution
+[//]: # (Elaborate on the solution you are suggesting in this page. Address the functional requirements and the non functional requirements that this solution is addressing. If there are a few options considered for the solution, mention them and explain why the actual solution was chosen over them. Add an execution plan when relevant. It doesn't have to be a full breaking of the feature, but just a recommendation to how the solution should be approached.)
+
+### Design
+[//]: # (Add any diagrams, charts and explanations about the design aspect of the solution. Elaborate also about what the expected user experience for the feature)
+
+### Backwards compatibility
+[//]: # (Address how you are going to handle backwards compatibility, if necessary)
```
izgeri, line 32: is it necessary to have a whole separate section on this?

design/templates/solution_design.md
orenbm: updated the doc in the original commit, thanks.

design/templates/solution_design.md
orenbm: You are correct. I removed this section as it is covered in "affected components"

design/templates/solution_design.md
orenbm: I think it is important enough to have a sub-section under "Solution". It can be interesting to see how the solution handles the backwards-compatibility.

orenbm:
@izgeri can you please re-review?

InbalZilberman:
@orenbm looks good to me

design/templates/solution_design.md
```diff
@@ -0,0 +1,71 @@
...
+[//]: # (Add any diagrams, charts and explanations about the design aspect of the solution. Elaborate also about the expected user experience for the feature)
+
+### Backwards compatibility
+[//]: # (Address how you are going to handle backwards compatibility, if necessary)
+
+### Performance
+[//]: # (Elaborate on whether your solution will affect the product's performance, and how)
+
+### Affected Components
+[//]: # (Address the componentes that will be affected by your solution [Conjur, DAP, etc.])
```
izgeri, line 38:
```suggestion
[//]: # (Address the components that will be affected by your solution [Conjur, DAP, clients, integrations, etc.])
```

design/templates/solution_design.md
```diff
@@ -0,0 +1,71 @@
...
+### Audit 
+[//]: # (Does this solution require additional audit messages?)
+
+## Documentation
+[//]: # (Add notes on what should be documented in this solution. Elaborate on where this should be documented. If the change is in open-source projects, we may need to update the docs in github too. If it's in Conjur and/or DAP mention which products are affected by it)
+
+## Open questions
+[//]: # (Add any question that is still open. It makes it easier for the reader to have the open questions accumulated here istead of them being acattered along the doc)
+
+## Breaking
```
izgeri, line 70:
```suggestion
## Implementation plan
```

izgeri:
had to rebase against master to merge

izgeri:
had to rebase against master to merge, will merge when green

InbalZilberman:
@orenbm all my concerns were answered

PR #1563, opened 2020-05-21
Remove audit design doc

micahlee:
### What does this PR do?

The OSS Conjur doesn't naturally make the syslog audit log available. This is
only meaningful in the DAP Appliance. The syslog format for audit
is documented for DAP elsewhere, so this is removed from the OSS docs.

The design doc also included development and test instructions for the audit
query engine. These are moved to the README in the audit engine directory.

#### Change log

#### Test coverage

#### Documentation


engines/conjur_audit/README.md
```diff
@@ -0,0 +1,35 @@
...
+To start a development environment that includes and enables Conjur audit:
+
+1. Navigate to the development environment directory from the project root
+    ```sh-session
+    cd dev
+    ```
+
+2. Start the development environment with the audit flag
+    ```sh-session
+    ./start --audit
```
jtuttle, line 17: When I run start w/ `--audit` I get prompted to enter some credentials:

```
+ docker-compose exec client conjur policy load root /src/conjur-server/dev/files/policy.yml
Enter your username to log into Conjur:
```

Tried to login w/ admin using the API key from further up in the output and it didn't work:

```
Enter your username to log into Conjur: admin
Please enter your password (it will not be echoed):
error: Failed to open TCP connection to conjur:3000 (Connection refused - connect(2) for "conjur" port 3000)
```

So it seems these instructions don't currently work as described here. Does it work for you?

micahlee: Ha! No, they don't work for me either. Let me see what I can uncover and if there's a quick resolution for that. Thanks!

micahlee: Yeah, somewhere along the way, defining the `AUDIT_DATABASE_URL` env variable and configuring Conjur were placed in the wrong order.

Defining `AUDIT_DATABASE_URL` after the initial configuration causes the Conjur container to be re-created and lose its configuration, so the admin user doesn't actually exist anymore when the client tries to load the policy.

micahlee:
@jtuttle I added a commit to fix the `--audit` flag on the dev script. This should be ready for another look. Thanks!

engines/conjur_audit/README.md
jtuttle: Aha! Nice find.

PR #1564, opened 2020-05-22
Validate origin before credentials

orenbm:
Connected to #1568

It makes sense to validate the origin of the request before
validating the credentials. Before we validate that the credentials
are correct, we need to validate that the role _can_ authenticate,
same as we do with the security. We first verify that the request can
be authenticated and then validate it is authenticated.

### DIFF
--- a/app/domain/authentication/authenticate.rb
+++ b/app/domain/authentication/authenticate.rb
@@ -22,8 +22,8 @@ module Authentication
     def call
       validate_authenticator_exists
       validate_security
-      validate_credentials
       validate_origin
+      validate_credentials
       audit_success
       new_token
     rescue => e
--- a/app/domain/authentication/authenticate.rb
+++ b/app/domain/authentication/authenticate.rb
@@ -20,8 +20,7 @@ module Authentication
   ) do
 
     extend Forwardable
-    def_delegators :@authenticator_input, :authenticator_name, :account,
-      :username, :webservice
+    def_delegators :@authenticator_input, :authenticator_name, :account, :username, :webservice, :origin, :role
 
     def call
       validate_authenticator_exists
@@ -59,13 +58,19 @@ module Authentication
     end
 
     def validate_origin
-      @validate_origin.(input: @authenticator_input)
+      @validate_origin.(
+        account: account,
+        username: username,
+        origin: origin
+      )
     end
 
     def audit_success
       @log_audit_event.(
         event: ::Authentication::AuditEvent::Authenticate,
-        authenticator_input: @authenticator_input,
+        authenticator_name: authenticator_name,
+        webservice: webservice,
+        role: role,
         success: true,
         message: nil
       )
@@ -74,7 +79,9 @@ module Authentication
     def audit_failure(err)
       @log_audit_event.(
         event: ::Authentication::AuditEvent::Authenticate,
-        authenticator_input: @authenticator_input,
+        authenticator_name: authenticator_name,
+        webservice: webservice,
+        role: role,
         success: false,
         message: err.message
       )
--- a/app/domain/authentication/authenticate.rb
+++ b/app/domain/authentication/authenticate.rb
@@ -19,6 +19,10 @@ module Authentication
     inputs:       %i(authenticator_input authenticators enabled_authenticators)
   ) do
 
+    extend Forwardable
+    def_delegators :@authenticator_input, :authenticator_name, :account,
+      :username, :webservice
+
     def call
       validate_authenticator_exists
       validate_security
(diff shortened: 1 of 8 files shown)


shaharglazner:
LGTM

app/domain/authentication/authenticate.rb
```diff
@@ -22,8 +22,8 @@ module Authentication
     def call
       validate_authenticator_exists
```
Tovli, line 23: Seems that all the methods are taking authenticator_input as a parameter.
Wouldn't it be better to chain them and pass the authenticator input explicitly?
Should make the code more readable

I would also consider pipelining (|>) the methods to reflect the flow

app/domain/authentication/authenticate.rb
```diff
@@ -22,8 +22,8 @@ module Authentication
...
       validate_security
```
Tovli, line 24: Is there a check here that does not relate to security?
As @shaharglazner pointed out. this is not clear

orenbm: what `validate_security` does is validate that the authenticator is enables, it has a defined webservice, the user exists and that it has authenticate permissions on the webservice.

Is it still unclear?

app/domain/authentication/authenticate.rb
orenbm: i added a commit so we don't pass around the implicit `authenticator_input` object when it's not necessary. The only place we now pass it over is in the call to `valid?` which is the generic function of each authenticator. we need to pass the whole object there as we don't know which inputs will be used in each authenticator (and it makes sense there to pass an object named `authenticator_input`)

orenbm:
@micahlee i added 2 commits that i'd like you to review. it touches code that you wrote.

orenbm:
> Why are we using this flow? what was wrong with the previous one? I'm missing the added value of this change

I answered these questions in the referenced issue.

app/domain/authentication/authenticate.rb
orenbm: `validate_authenticator_exists` validates that the authenticator is implemented in Conjur, not that it is enabled in the env.

I agree that `validate_security` is quite vague and we should consider making it clearer.

orenbm: `validate_security` validates several check. we can call each one explicitly instead of stacking them in that class (i.e move all the validations preformed in `validate_security` to the calling class) but i'd like to get some thoughts on this from @micahlee & @jonahx .

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
orenbm: i actually don't have an answer to that. @micahlee @jonahx do you know?

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -8,23 +10,7 @@ module AuthnK8s
...
-
-      def webservice
-        @webservice ||= ::Authentication::Webservice.new(
-          account:            @account,
-          authenticator_name: authenticator_name,
-          service_id:         @service_id
-        )
-      end
-    end
+    AUTHENTICATOR_NAME = 'authn-k8s'
```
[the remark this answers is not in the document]
orenbm: it is declared under the `AuthnK8s` module so it's not really required but i can live with that.

app/domain/authentication/authenticate.rb
orenbm: for visualization, the `authenticate` call method now looks like this:
```
      validate_authenticator_exists
      validate_security
      validate_origin
      validate_credentials
      audit_success
      new_token
```

If we move the validations from `validate_security` to the calling class, the call method will look like this:
```
      validate_authenticator_exists
      validate_webservice_is_whitelisted
      validate_user_has_access_to_webservice
      validate_origin
      validate_credentials
      audit_success
      new_token
```

@Tovli @shaharglazner @micahlee @jonahx wdyt?

micahlee: I think this is clearer than just `validate_security`. I'm also a fan of smaller PRs, so I suggest making this a follow up rather than adding all of the improvements to a single PR. It will make it much easier to review more quickly.

app/domain/authentication/authn_k8s/inject_client_cert.rb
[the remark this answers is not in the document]
micahlee: I don't know the answer. My guess is that it corresponds to the fact that we don't call `ruby` directly to start Conjur, so passing the global flag would be add more complexity than it's worth. I could be wrong though.

PR #1567, opened 2020-05-24
Add solution design for authenticators as plugins

oburstein-hub:
### What does this PR do?
This PR adds a requirements to develop infra for plugin-based authenticators in Conjur

### DIFF
--- a/design/authenticators/authenticators_as_external_plugins.md
+++ b/design/authenticators/authenticators_as_external_plugins.md
@@ -1,4 +1,13 @@
-Problem Statement
+# Solution Design - Authenticator as a Plugin
+
+## Table of Contents
+
+- [Problem Statement](#problemstatement)
+- [Goals](#goals)
+- [Implementation Steps](#implementationsteps)
+
+
+## Problem Statement
 
 Currently adding new authenticator type in Conjur / DAP requires code changes in Conjur.
 
@@ -8,12 +17,13 @@ Adding an ability to implement authenticator as an external plugin will allow ou
 
 and without being familiar with Ruby / Conjur
 
-Goals
+## Goals
 Create programmatic / configurational hooks in Conjur, which will allow to incorporate external authenticator as a plugin to Conjur
 Make authenticator addition to be possible for customers
 Preserve set of authenticators that exist in Conjur
 Provide set of parameters for plugin-based authenticators per user decision
-Implementation Steps
+
+## Implementation Steps
 Identify the functional blocks that are common to all authenticators and split them into 2 kinds
 2.1 Blocks that are not going to be programmed by plugin (example: set / get authenticator in DB)
 2.2 Blocks that should be programmed by plugin (example: validate methods)
--- a/design/authenticators/authenticators_as_external_plugins.md
+++ b/design/authenticators/authenticators_as_external_plugins.md
@@ -24,13 +24,13 @@ Preserve set of authenticators that exist in Conjur
 Provide set of parameters for plugin-based authenticators per user decision
 
 ## Implementation Steps
-Identify the functional blocks that are common to all authenticators and split them into 2 kinds
-2.1 Blocks that are not going to be programmed by plugin (example: set / get authenticator in DB)
-2.2 Blocks that should be programmed by plugin (example: validate methods)
-Verify that base blocks inside Conjur are well separated from authenticator-specific blocks inside Conjur. 
-If they are not - take steps to make them such
-Define hook to load external authenticators
-Define hook to provide external authenticator parameters (probably in yaml)
-Define hook for "validate" step of external authenticator (should contain a pointer to function / executable)
-Verify that configuration of external authenticator is possible in Conjur
-Implement and test sample external authenticator
+- Identify the functional blocks that are common to all authenticators and split them into 2 kinds
+  Blocks that are not going to be programmed by plugin (example: set / get authenticator in DB)
+  Blocks that should be programmed by plugin (example: validate methods)
+- Verify that base blocks inside Conjur are well separated from authenticator-specific blocks inside Conjur. 
+- If they are not - take steps to make them such
+- Define hook to load external authenticators
+- Define hook to provide external authenticator parameters (probably in yaml)
+- Define hook for "validate" step of external authenticator (should contain a pointer to function / executable)
+- Verify that configuration of external authenticator is possible in Conjur
+- Implement and test sample external authenticator
--- a/design/authenticators/AuthenticatorsAsAnExternalPlugin
+++ b/design/authenticators/AuthenticatorsAsAnExternalPlugin
@@ -1,0 +1,26 @@
+Problem Statement
+
+Currently adding new authenticator type in Conjur / DAP requires code changes in Conjur.
+
+Thus is only possible to external users by code contribution to Conjur, and impossible for a concrete DAP version (which uses predefined Conjur version)
+
+Adding an ability to implement authenticator as an external plugin will allow our customers to implement authenticators of their own without changing Conjur codebase 
+
+and without being familiar with Ruby / Conjur
+
+Goals
+Create programmatic / configurational hooks in Conjur, which will allow to incorporate external authenticator as a plugin to Conjur
+Make authenticator addition to be possible for customers
+Preserve set of authenticators that exist in Conjur
+Provide set of parameters for plugin-based authenticators per user decision
+Implementation Steps
+Identify the functional blocks that are common to all authenticators and split them into 2 kinds
+2.1 Blocks that are not going to be programmed by plugin (example: set / get authenticator in DB)
+2.2 Blocks that should be programmed by plugin (example: validate methods)
+Verify that base blocks inside Conjur are well separated from authenticator-specific blocks inside Conjur. 
+If they are not - take steps to make them such
+Define hook to load external authenticators
+Define hook to provide external authenticator parameters (probably in yaml)
+Define hook for "validate" step of external authenticator (should contain a pointer to function / executable)
+Verify that configuration of external authenticator is possible in Conjur
+Implement and test sample external authenticator


orenbm:
thanks @oburstein-hub for taking this!

A few comments before I look into the actual content of the file:
1. Please change the file name to `authenticators_as_external_plugins.md`. Note that i changed the name to be underscored as the rest of the files and changed the extension to Markdown.
2. Please fix the file so it will be easier to read. Add titles, proper bullets, etc. You can see [this file](https://github.com/cyberark/conjur/blob/oburstein-hub-patch-1/design/authenticators/authn_azure/authn_azure_solution_design.md) as an example. Your file now looks [like this](https://github.com/cyberark/conjur/blob/oburstein-hub-patch-1/design/authenticators/AuthenticatorsAsAnExternalPlugin) which makes it hard to review.

The [solution design template](https://github.com/cyberark/conjur/pull/1561) is still in progress but please take a look there to see if you can add any sections to your suggestion that will make the reviewers task easier.

Thanks again for taking this!

orenbm:
adding @cyberark/conjur-core-team to get some thoughts

design/authenticators/authenticators_as_external_plugins.md
```diff
@@ -0,0 +1,36 @@
...
+
+Adding an ability to implement authenticator as an external plugin will allow our customers to implement authenticators of their own without changing Conjur codebase 
+
+and without being familiar with Ruby / Conjur
+
+## Goals
+Create programmatic / configurational hooks in Conjur, which will allow to incorporate external authenticator as a plugin to Conjur
+Make authenticator addition to be possible for customers
+Preserve set of authenticators that exist in Conjur
+Provide set of parameters for plugin-based authenticators per user decision
```
orenbm, line 24: @oburstein-hub can you change these into bullets?

design/authenticators/authenticators_as_external_plugins.md
```diff
@@ -0,0 +1,36 @@
...
+Identify the functional blocks that are common to all authenticators and split them into 2 kinds
+2.1 Blocks that are not going to be programmed by plugin (example: set / get authenticator in DB)
+2.2 Blocks that should be programmed by plugin (example: validate methods)
+Verify that base blocks inside Conjur are well separated from authenticator-specific blocks inside Conjur. 
+If they are not - take steps to make them such
+Define hook to load external authenticators
+Define hook to provide external authenticator parameters (probably in yaml)
+Define hook for "validate" step of external authenticator (should contain a pointer to function / executable)
+Verify that configuration of external authenticator is possible in Conjur
+Implement and test sample external authenticator
```
orenbm, line 36: @oburstein-hub can you change these into bullets?

design/authenticators/authenticators_as_external_plugins.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Problem Statement
+
+Currently adding new authenticator type in Conjur / DAP requires code changes in Conjur.
+
+Thus is only possible to external users by code contribution to Conjur, and impossible for a concrete DAP version (which uses predefined Conjur version)
+
+Adding an ability to implement authenticator as an external plugin will allow our customers to implement authenticators of their own without changing Conjur codebase 
+
+and without being familiar with Ruby / Conjur
```
orenbm, line 18: can the plugins be written in a language other than Ruby?

micahlee: I will give the whole thing some more thought soon, but I do want to second this.

I think the value of plugins for this is greatly increased if they can be implemented in another language of the end-user's choosing. (Note, it's possible that from Conjur's perspective it could still be "all Ruby" through the use of foreign function interfaces or similar).

oburstein-hub: Agree, thats what I planned also

design/authenticators/authenticators_as_external_plugins.md
oburstein-hub: done

orenbm:
@oburstein-hub can you please use the [Solution Design template](https://github.com/cyberark/conjur/blob/master/design/templates/solution_design.md) for this? I think it will make the content easier to follow. thanks!

izgeri:
@oburstein-hub is this a proposal for planned nearterm work, or is it meant to kick off the discussion on enabling conjur plugins (work that I know @jvanderhoof and @jonahx have already done some thinking on)?

if this is meant to start the discussion, would the right place for this file actually be in the new `design/proposals` folder that @rafis3 created recently? we are trying to do a better job of distinguishing between work that is done or being implemented, and work we are considering.

oburstein-hub:
Hi @izgeri . It is a proposal for authentication as external plugins that is handled by Oren
I have placed it here by his request

orenbm:
> Hi @izgeri . It is a proposal for authentication as external plugins that is handled by Oren
> I have placed it here by his request

Oh i didn't understand this is the case. I that case I will try to get to it in the near future and make this design more robust.

design/authenticators/authenticators_as_external_plugins.md
```diff
@@ -0,0 +1,36 @@
...
+- Identify the functional blocks that are common to all authenticators and split them into 2 kinds
+  Blocks that are not going to be programmed by plugin (example: set / get authenticator in DB)
+  Blocks that should be programmed by plugin (example: validate methods)
+- Verify that base blocks inside Conjur are well separated from authenticator-specific blocks inside Conjur. 
+- If they are not - take steps to make them such
+- Define hook to load external authenticators
+- Define hook to provide external authenticator parameters (probably in yaml)
+- Define hook for "validate" step of external authenticator (should contain a pointer to function / executable)
+- Verify that configuration of external authenticator is possible in Conjur
+- Implement and test sample external authenticator
```
jonahx, line 36: This approach is unlikely to be flexible enough to handle all possible future authenticators.  Authentication schemes differ enough that we don't want to go for a "function template" solution:  That is, we can't find the steps that "all" authenticators take, make them into into abstract methods, and have each authenticator "fill them in".

Something like the original design that Kevin proposed seems more robust to me:

In his proposal, an authenticator was just any dockerized webservice, which could be configured to be "attached" to Conjur.  That webservice could have whatever API it wanted, including multiple routes if needed, and could be written in any language.  The webservice could then "mint" new conjur tokens at will, by requesting them from the core conjur service via a unix websocket.  

Thus the basic architecture is:

1. A generic webservice with whatever API it needed which is responsible for determining when something is authenticated.
2. A mechanism (provided by Conjur core) for this webservice to create a conjur token once it decides something is authenticated.

This makes no assumption about the generic "structure" of an authenticator, and this is good, because it is unlikely any such abstraction will stand up to the test of time.

Tovli: Thanks @jonahx - so essentially we will have an authentication service decoupled from the core project
Going that direction, I wonder what would you think on generating the tokens by this service and have conjur just validate them.
@oburstein-hub @orenbm - WDYT?

oburstein-hub: Generating tokens sounds a good idea, but it should be detailed in the design

jonahx: @Tovli In Kevin's design, conjur-core was still responsible for generating the tokens (so that only conjur-core had access to the signing key), but the webservice plugins could ask conjur to generate tokens for them, via a local unix socket.  The local unix socket isn't the only way to do this, ofc, but I like the idea that minting tokens remains the centralized responsibility of conjur-core.

izgeri:
@oburstein-hub can the title of this be prefaced by "WIP" please? This PR has been open for a while, and the document is very lightweight still. I'd expect before we merge this the doc will be quite detailed and follow more closely @orenbm's template for design documents.

Alternatively, if this is not something we'll have time to work on soon, perhaps we should close the PR for now.

oburstein-hub:
Are not going to work on it soon

design/authenticators/authenticators_as_external_plugins.md
orenbm: @jonahx before we drop the suggested approach, i want to clarify that the authentication flow can be separated into 2 parts:
  - General validations
  - Specific validations

This can be seen in the `call` function of [`Authentication:: Authenticate`](https://github.com/cyberark/conjur/blob/4a02078f7921d195f1e02cf7070041d4c1b05ebc/app/domain/authentication/authenticate.rb):
```
def call
  validate_authenticator_exists
  validate_webservice_is_whitelisted
  validate_user_has_access_to_webservice
  validate_origin
  validate_credentials
  audit_success
  new_token
rescue => e
  audit_failure(e)
  raise e
end
```

The authentication performs all the check until `validate_credentials`, which is implemented as follows:
```
def validate_credentials
  raise Errors::Authentication::InvalidCredentials unless authenticator.valid?(@authenticator_input)
end
```

The `valid?` function is implemented inside the authenticator class and it performs the specific authentication (e.g verify the ID token in OIDC).

So, the way i see it, a new authenticator needs only to implement the specific validations of the new authenticator. I created such a template just to pass the idea (it is not implemented as a plugin). Please see the classes i added [here](https://github.com/cyberark/conjur/tree/authenticator-template/app/domain/authentication/template).

We can change the way that we build the `authenticators` array that has all the authenticator classes so that these classes can be loaded in runtime instead of compile time.

Won't this approach be generic enough for us?

orenbm: i am thinking of something similar to what we have in secretless but i'm not sure we have this in Ruby

jonahx: As I said over chat, it all depends on what the goal is.  From my understanding, being able to write these in any language would be a huge boon -- to the sales engineers, to customers, and for the community efforts.  So if we take that as a given, then our current approach can't support that.

I don't think there's anything wrong with our current approach.  It's simple and I helped design it.  But I think the way things have played out, it's become clear that "plugins you can write in any language" is probably where we ultimately want to go.

PR #1570, opened 2020-05-26
Disable FIPS mode using OPENSSL_FIPS_ENABLED env var

sashaCher:
### What does this PR do?
OpenSSL FIPS mode is enabled by default
It will be disabled if and only if OPENSSL_FIPS_ENABLED env var is present and has false value


sgnn7:
Drive-by comment: spelling in the commit for `mode` is wrong. I fixed the title of the PR but it needs the commit fixed too.

uCatu:
Can't approve until CI is green :)

uCatu:
> Can't approve until CI is green :)
@sashaCher

Issue #1573, opened 2020-05-26
Audit events are simple and testable


jonahx:
The current objects and class hierarchies used for creating audit events are very difficult to understand and use.  They are also not testable.

This work aims to fix both those issues, and thereby ease adding new audit events.

PR #1574, opened 2020-05-27
1573 simplify audit

jonahx:
### What does this PR do?

The previous audit event classes were based on the unnecessary
Util::Struct class, and were full of confusing indirection: mutiple
levels of inheritance, as well as mixins (the CanFail module).

This made them difficult to understand, maintain, change, and test.

This commit contains the bulk of the work the removes the Util::Struct
dependency and all uses of inheritance and mixins.  Each class can now
be read and full understood on its own, with its full interface visible
in the class file itself.  Repeated code is handled through delegation,
in particular the Authn class, which the authentication events delegate
to.  Similarly, the CanFail module has been replaced by the
AttemptedAction class, an object we now delegate to rather than mixing
in.

We also simplify the interface itself for logging events, so that
callers only need to pass the logger a single parameter, the event
itself.

Finally, since we support two different loggers -- the Rails logger
which uses the standard ruby logger interface, as well as a syslog
logger -- we introduce adapter classes for each logger type, so that the
translation logic will be isolated.
- _Appliance build with this change:_

https://jenkins.conjur.net/job/conjurinc--appliance/job/1573-simplify-audit/

#1573 

#### Change log

#### Test coverage

Tests already exists for this functionality.  Adding additional tests will happen in a new PR.

#### Documentation


app/models/audit/event/authn2.rb
```diff
@@ -0,0 +1,76 @@
...
+      def message_id
+        "authn"
+      end
+
+      def structured_data
+        {
+          SDID::SUBJECT => {role: @role&.id},
+          SDID::AUTH => auth_stuctured_data,
+        }.merge(
+          possibly_failing_event.action_sd(@operation)
```
micahlee, line 53: Is there a reason that `@operation` can't be passed to the constructor for `PossiblyFailingEvent`?

I'm not aware of any reason that would change in a given instance of an Audit event. So this could become just:
```
 possibly_failing_event.action_sd
```

PR #1577, opened 2020-05-29
[WIP] Solution Proposal for Variable Provisioning

micahlee:
### What does this PR do?

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/design/proposals/variable_provision.md
+++ b/design/proposals/variable_provision.md
@@ -1,0 +1,299 @@
+# Solution Design - Variable Provisioning
+
+## Problem Definition
+
+The lifecycle of a secret value in Conjur suffers from a couple of shortcomings
+that make it less effective than it could be. These include:
+
+1. **Browsing-up**: Secret values are often generated on a less-trusted, or
+   less-secure platform than Conjur itself, and then loaded into the more-secure
+   platform.
+
+2. **Domain Knowledge**: Secret values for access often require a degree of
+   domain knowledge to create and configure (e.g. x509 certificates, OAuth tokens)
+   which can lead to accidental misconfigurations or malformed values.
+
+3. **Provenance**: By only storing values produced outside of Conjur, the
+   audit information around secrets lack key provenance information regarding
+   their creation and attributes.
+
+Variable Provisioning is feature proposal to address these shortcomings by allowing
+Conjur to produce secret values internally, based on variable annotations defined
+in Conjur policy.
+
+## Table of Contents
+
+- [Glossary](#glossary)
+- [Useful links](#useful-links)
+- [Background](#background)
+- [Issue description](#issue-description)
+  * ['Browsing-up` for secrets management](#-browsing-up--for-secrets-management)
+  * [Secrets Domain Knowledge Required](#secrets-domain-knowledge-required)
+  * [Human Intervention Required](#human-intervention-required)
+  * [Credential Provenance](#credential-provenance)
+- [Solution](#solution)
+  * [Design](#design)
+  * [Backwards compatibility](#backwards-compatibility)
+  * [Performance](#performance)
+  * [Affected Components](#affected-components)
+- [Security](#security)
+- [Test Plan](#test-plan)
+- [Logs](#logs)
+  * [Audit](#audit)
+- [Documentation](#documentation)
+- [Version update](#version-update)
+- [Open questions](#open-questions)
+- [Breaking](#breaking)
+
+## Background
+
+The purpose of Variables in Conjur is to put values in them for applications to
+use for service access. These values often consist of:
+
+- Passwords
+- API Keys
+- x509 Private Keys
+- x509 Certificates
+
+The typical workflow for a Variable is:
+
+- The Variable is created when it is defined in a Conjur policy document and loaded
+  into Conjur:
+
+  ```YAML
+  # my-variable.yaml
+
+  - !variable my-api-key
+  ```
+
+  ```sh-session
+  $ conjur policy load root my-variable.yaml
+  ...
+  ```
+
+- The owner of the value uses the Conjur API or CLI to store the value in the Variable:
+  
+  ```sh-session
+  $ conjur variable values add my-api-key abcdef12345
+  ...
+  ```
+
+- Once stored, access to the variable may be granted to applications, and those
+  applications can use value using the Conjur API, Summon, Secretless, etc.
+
+## Issue description
+
+There are a couple of issues this solution intends to address:
+
+### 'Browsing-up` for secrets management
+
+"Browse-up" is defined as [[ref]](https://www.ncsc.gov.uk/whitepaper/security-architecture-anti-patterns#section_3):
+
+> When administration of a system is performed from a device which is less trusted
+> than the system being administered.
+
+Unless performed only on dedicated, controlled workspaces, the `variable values add`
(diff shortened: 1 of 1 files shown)


design/proposals/variable_provision.md
```diff
@@ -0,0 +1,295 @@
...
+- [Test Plan](#test-plan)
+- [Logs](#logs)
+  * [Audit](#audit)
+- [Documentation](#documentation)
+- [Version update](#version-update)
+- [Open questions](#open-questions)
+- [Breaking](#breaking)
+
+
+## Glossary
```
jvanderhoof, line 28: Can we lead with the problem to grab the reader? Maybe something like:
```md
## Problem Definition
Conjur currently suffers a couple of shortcomings related to the creation of variable values:
1. `Browsing-up`: secret values are generated on a less secure platform to load into a more secure platform (Conjur)
1. Domain Knowledge: generating certificates require a deep understanding which increases the likelihood of incorrect configuration
1. Provenance: we currently lack visibility into the creation of a variable. 
...
```

micahlee: Are you suggesting moving the `Issue description` to the top and renaming it? Or adding a `Problem Definition` that summarizes the issue(s)?

jvanderhoof: Kind of both 😛 . I like the idea of starting a proposal like this by defining the problem so we can first establish this is a real problem that substantially impacts customers.  Let's move the `Issue Description` section up. 

As a side note, I don't feel like I have a good vocabulary for what to call the problem definition ☹️

design/proposals/variable_provision.md
```diff
@@ -0,0 +1,295 @@
...
+
+The purpose of Variables in Conjur is to put values in them for applications to
+use for service access. These values often consist of:
+
+- Passwords
+- API Keys
+- x509 Private Keys
+- x509 Certificates
+
+The typical workflow for a Variable is:
```
jvanderhoof, line 54: Just a note (mostly to myself), it would be good to visualize what a mess this flow actually looks like.  Particularly with multiple actors (ex. Security engineer to create policy, database admin to set value, application to use the value).

design/proposals/variable_provision.md
```diff
@@ -0,0 +1,295 @@
...
+"Browse-up" is defined as [[ref]](https://www.ncsc.gov.uk/whitepaper/security-architecture-anti-patterns#section_3):
+
+> When administration of a system is performed from a device which is less trusted
+> than the system being administered.
+
+Unless performed only on dedicated, controlled workspaces, the `variable values add`
+step in our existing workflow necessitates that all secrets stored in Conjur (a
+"trusted system") must first travel through a user's workstation (a "less trusted system").
+
+### Secrets Domain Knowledge Required
```
jvanderhoof, line 95: Not sure if it's part of this section, but generating certificates for the K8s Authenticator seems a lot more complex due by asking customers to generate these certificates.

design/proposals/variable_provision.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Human Intervention Required
+
+As application deployments scale up, the need to automate the creation and rotation
+of credentials will continue to grow as human provisioning of credentails becomes
+a bottleneck. Our API enables end-users to bolt this automation onto Conjur, but
+we have an opportunity to enable more just-in-time creation of these credentails
+on behalf the user to make this kind of automation more widely available.
+
+### Credential Provenance
```
jvanderhoof, line 118: This really struck me.  I've never thought about the fact that these credentials likely live an entire life before they come into the DAP/Conjur world. During that time, no one really knows where they are with complete certainty.

design/proposals/variable_provision.md
```diff
@@ -0,0 +1,295 @@
...
+The solution I propose to address these issues is **variable provisioning**, in which
+Conjur is configured (through the policy where a variable is defined) to provision
+the secret value for a variable when the policy for that variable is loaded.
+
+For example, defining a variable to store an RSA private key could be represented
+in policy as:
+
+```yaml
+- !variable
+  id: private-key
```
jvanderhoof, line 135: How could we capture the public key in a variable as well?

design/proposals/variable_provision.md
```diff
@@ -0,0 +1,295 @@
...
+    provision/x509-certificate/basic-constraints/ca: true
+    provision/x509-certificate/basic-constraints/critical: true
+    provision/x509-certificate/key-usage/critical: true
+    provision/x509-certificate/key-usage/key-cert-sign: true
+    provision/x509-certificate/key-usage/crl-sign: true
+    provision/x509-certificate/private-key/variable: ca/private-key
+    provision/x509-certificate/issuer/private-key/variable: ca/private-key # Self-signed CA
+    provision/x509-certificate/issuer/certificate/self: true
+    provision/x509-certificate/ttl: P1Y
+```
```
jvanderhoof, line 169: Thinking about IP address whitelisting, I was wondering about using a `Group` as a mechanism for DRYing up some of these annotation.  What do you think about something like:
```yml
- !group 
  id: certificate-base 
  annotations:
    provision/provisioner: x509-certificate
    provision/x509-certificate/subject/c: US
    provision/x509-certificate/basic-constraints/ca: true
    provision/x509-certificate/basic-constraints/critical: true
    provision/x509-certificate/key-usage/critical: true
    provision/x509-certificate/key-usage/key-cert-sign: true
    provision/x509-certificate/key-usage/crl-sign: true
    provision/x509-certificate/private-key/variable: ca/private-key
    provision/x509-certificate/issuer/private-key/variable: ca/private-key # Self-signed CA
    provision/x509-certificate/issuer/certificate/self: true
    provision/x509-certificate/ttl: P1Y

- !variable
  id: certificate
  annotations:
    provision/x509-certificate/subject/cn: conjur-master-ca

- !grant
  member: !variable certificate
  role: !group certificate-base 
```

design/proposals/variable_provision.md
micahlee: Yes, this would completely eliminate the need for certificate creation (either manually, or the rake task). The parameters for the certificates would just become a part of the template for the authenticator policy.

I actually used this from the prototype in one of the openshift operator demos to completely eliminate that step from the openshift follower installation flow.

design/proposals/variable_provision.md
micahlee: At the moment in prototype, the `rsa` provisioner is the equivalent of:
```
openssl genrsa ...
```

The resulting PEM actually contains both the private key and public key, so you could have an rsa public key provisioner that just extracts the public key portion if you needed it, similar to how we reference the private key in the `x509-certificate` provisioner.

I'm not aware of many (any?) use cases where you want the public key apart from a X.509 certificate that identifies it though...

design/proposals/variable_provision.md
micahlee: I see where you're going with this, and the issue you're trying to resolve. I'm wondering how you deal with precedence/aggregation when you're a member of multiple groups that define these annotations?

Annotation defaults is more of a `derivativeOf` relationship, rather than the `memberOf` relationship that groups convey.

As a side note, my solution for this was the Policy Factory... to use templates for the defaults, and then create policy instances/resources from the template.

jvanderhoof: Rethinking this, I like your original solution.  It's much more explicit. There are alternatives to streamlining the generation of a large number of resources.

Issue #1354, opened 2020-02-24
Simplify `application_identity.rb` for authn-azure & authn-k8s


orenbm:
Our current code for the Application Identity of both authn-k8s & authn-azure may be hard to understand. We should try to simplify it a bit.

Things that we can do:
- Create a class for ApplicationIdentity that will hold all the constraints as attributes, instead of the class that we have today that holds a Hash for the constraints.
- Create a CommandClass called `ExtractApplicationIdentityFromRole` which will do all the calculations and will create the class mentioned above. This will move most of the logic from ApplicationIdentity to this CommandClass.
- Create a CommandClass called ValidateRoleAnnotations that will run before we extract the application identity from the role. This way we extract some of the logic to a separate class that will verify that the annotations are permitted.

In authn-azure we can also create a class for TokenIdentity instead of using a Hash. 

Once this is done we should write UTs for the classes.

DoD:
- [ ] Application Identity classes have been refactored
- [ ] A review was done by @jonahx 
- [ ] UTs have been added to all relevant classes

PR #1580, opened 2020-06-01
Refactor application identity

orenbm:
Connected to #1354 

### What does this PR do?
This PR aligns the code of Application Identity validations between authenticators. It does not generalize the validations at this phase, and it only refactors the code of the validations in authn-k8s and authn-azure to better demonstrate the similarities between the two. After this effort is done we may generalize the validations but this step is a crucial one towards that. It also has enough value on its own as after this change it will be easier for future developers to add application identity validations for a new authenticator.

This PR also changes the term "Application Identity" to "Resource Restrictions" as it better tells the story. Although the term Application Identity is documented in our official docs, it is not clear to readers of the code and "ResourceRestrictions" are more self explanatory, as that is what application identity is all about. It defines the resources from which an authentication request is restricted to (like we have with IP restrictions).

#### Change log

#### Test coverage
This PR does not fix UTs and they will fail at first. I want to first get feedback on the refactor itself, and once it is approved, to update UTs. This is done to avoid back-and-forth with fixing UTs. Integration tests are passing.

#### Documentation


app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -2,30 +2,43 @@ module Authentication
   module AuthnAzure
 
     Log = LogMessages::Authentication::AuthnAzure
+    Err = Errors::Authentication::AuthnAzure
 
     class ApplicationIdentity
 
-      def initialize(role_annotations:, service_id:, logger:)
+      attr_reader :resources, :resource_types
```
sigalsax, line 9: IMO helpful to add some background and explanatory comments here

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -2,30 +2,43 @@ module Authentication
...
 
-      # check the `service-id` specific constraint first to be more granular
-      def constraint_value constraint_name
-        annotation_value("authn-azure/#{@service_id}/#{constraint_name}") ||
-          annotation_value("authn-azure/#{constraint_name}")
+      def init_resources
+        @resource_types = []
+        @resources = @azure_resource_types.each_with_object([]) do |resource_type, resources|
+          resource_name = resource_name(resource_type)
+          if resource_name
```
sigalsax, line 26: very readable and clear what you are trying to achieve here. A+ for @orenbm !

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -9,119 +9,158 @@ module AuthnAzure
...
+            )
+          )
         end
-        @logger.debug(Log::ValidatedApplicationIdentity.new)
+        @logger.debug(Log::ExtractedApplicationIdentityFromToken.new)
       end
 
-      def annotation_type_constraints constraints
-        constraints.map { |constraint| annotation_type_constraint(constraint) }
+      def parse_xms_mirid
```
sigalsax, line 143: `parse_xms_mirid` doesn't seem like a fitting name because the function it is calling to doesn't parse it

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -9,119 +9,158 @@ module AuthnAzure
...
+      # xms_mirid is a term in Azure to define a claim that describes the resource that holds the encoding of the instance's
+      # among other details the subscription_id, resource group, and provider identity needed for authorization.
+      # xms_mirid is one of the fields in the JWT token. This function will extract the relevant information from
+      # xms_mirid claim and populate a representative hash with the appropriate fields.
+      def extract_resources_from_token
+        parse_xms_mirid
+        @resources_from_token = [
+          AzureResource.new(
+            resource_type: "subscription-id",
+            resource_name: xms_mirid.subscriptions
```
sigalsax, line 113: what if xms_mirid.subscriptions doesn't exist?

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -9,119 +9,158 @@ module AuthnAzure
...
-      # converts the constraint to be in annotation style (e.g resource-group
-      # instead of resource_group) to enhance supportability
-      def annotation_type_constraint constraint
-        constraint.to_s.tr('_', '-')
+      def validate_application_identity_matches_request
+        extract_resources_from_token
+
+        application_identity.resources.each do |resource_from_role|
+          @resources_from_token.each do |resource_from_token|
+            if resource_from_token.resource_type == resource_from_role.resource_type
```
sigalsax, line 156: why do we have an if unless here? 
`if resource_from_token.resource_type == resource_from_role.resource_type` do logging?

If so, I don't think that is very clear

app/domain/authentication/authn_azure/application_identity.rb
orenbm: would this help?
```
      # resources: list of AzureResource objects included in the Application Identity
      # resource_types: list of the Azure resource types included in the Application Identity
      attr_reader :resources, :resource_types
```

If not, what data do you think is missing?

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: then we won't get to this line. We validate the required fields in the XmsMirid object creation.

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: we want to compare between the values of the resource object of the same resource type. For example, if the subscription id is `sub_1` and the resource_group is `rg_1`, i would like to compare `sub_1` with the subscription id in the token, and not with the resource_group in the token.

So first, we find the corresponding resource type and then compare between its value. 

I'll change it from:
```
if resource_from_token.resource_type == resource_from_role.resource_type
              unless resource_from_token.resource_name == resource_from_role.resource_name
                raise Err::InvalidApplicationIdentity, resource_from_role.resource_type
              end
            end
```

to:
```
if resource_from_token.resource_type == resource_from_role.resource_type && 
              resource_from_token.resource_name != resource_from_role.resource_name
                raise Err::InvalidApplicationIdentity, resource_from_role.resource_type
              end
            end
```.

Is this better?

File: app/domain/authentication/authn_azure/azure_resource.rb [hunk 6]
```diff
@@ -0,0 +1,13 @@
+module Authentication
+  module AuthnAzure
+
+    class AzureResource
+      attr_reader :resource_type, :resource_name
+
+      def initialize(resource_type:, resource_name:)
+        @resource_type = resource_type
```
sigalsax, line 8: type & name are similar in their meaning. IMO it will be easier to diffentiate with 

```
resource-name ("subscription-id") // this can also be resource-type btw. which ever one you see fit
resource-value (" xms_mirid.subscription")
```

orenbm: yeah originally it was resource name and value as you did. I changed it to align with k8s who already had the use of the resource type. I think i can have it the other way around and align k8s to have name and value.

orenbm: According to [k8s docs](https://kubernetes.io/docs/reference/kubectl/overview/#resource-types), it seems that "resource type" is more appropriate.

app/domain/authentication/authn_azure/application_identity.rb
sigalsax: I feel like resources and resource_type are too similiar and took me a while to understand even with your comment which leads me to believe naming will help solve the problem.

resources is just a list of AzureResources that comes from the application identity of the host.
so what about `ApplicationIdentitiesObject` or `AzureApplicationIdentitiesList` or `HostApplicationIdentitiesList`... Whatever you choose it should be clear that this object holds a list of Host application identities 

As mentioned in a comment below, I think it would be more representative if `resource_types` would be `resource_name` or even `application_identity_name`

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: much!

app/domain/authentication/authn_azure/azure_resource.rb
orenbm: or maybe Resource Kind

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -9,119 +9,152 @@ module AuthnAzure
     # MissingRequiredFieldsInXmsMirid, MissingProviderFieldsInXmsMirid, MissingConstraint,
```
sigalsax, line 9: "MissingRequiredFieldsInXmsMirid" can be removed as it's not used in file

app/domain/authentication/authn_azure/validate_application_identity.rb
sigalsax: you right you right

app/domain/authentication/authn_azure/application_identity.rb
orenbm: I would rather not to add Application Identity to the name as i want to rename `ApplicationIdentity` to something else. I thought of doing this in another PR to make this PR smaller but maybe it makes sense to have them in the same PR. 

what do you think about `ResourceRestrictions`? This name is more self-explanatory as this is what the application identity is. It collects restrictions of resources from which the host can authenticate.

app/domain/authentication/authn_azure/azure_resource.rb
sigalsax: I don't mind if it is `resource type` / `kind` as long as the second is `resource-value` b/c when I think `value`, I think key -> value which is what you are trying to get build

orenbm: makes sense, thanks

app/domain/authentication/authn_azure/application_identity.rb
sigalsax: I am interested to hear your thought process around ResourceRestrictions...  I remember us talking in the scope of Azure about how this collection (subscription id, resource group, etc) is the native identity of the resource so wouldn't at least `resource_identity` be fitting?

I understood that subscription id, resource group, etc are affiliations. They are define the azure identity of a VM, not what it restricts it from authenticating but again, let's talk about this more

orenbm: well, the word "identity" is not in place as these objects are not part of any identity. Is it part of the Conjur identity? Is it part of the Azure identity? The answer to both questions is "no". It is not the id of either.

Then what are these characteristics? If you explain them to me in words, you will say something like "they define objects in the native environment from which the host can authenticate with Conjur", right? For example, if the host has an annotation `authn-azure/resource-group: some-rg` then it means that this host can authenticate with the Azure authenticator only with an Azure AD token that was enrolled for an Azure resource in the `some-rg` resource group. So it is basically a restriction on the resource.

Another thing that led me to this is that when I tried to explain what an Application Identity is, I gave the analogy of IP restrictions. Where if a host has the field `restricted_to: 1.1.1.1` then it means that it can authenticate with Conjur only from the IP 1.1.1.1. The language is pretty much the same. And that led me to understand that application identities are basically restrictions on the resource that the host is authenticating from, and not an identity at all.

sigalsax: Is `some-rg` not part of it's identity? The resource is _**part**_ of a grouping called `some-rg` which is why "affiliation" is more descriptive . Or another example, for the system assigned ID, the OID of the resource, it makes up its identity no? 

It seems to me that "affiliation" would be more fitting. What about `ResourceAffiliation`?

orenbm: Why is it better than Restriction? I don't like words that send me to google translate :)

Restriction is _exactly_ what the thing is.

orenbm: we'll wait for @jonahx to give his opinion. I think that he has the value of someone who didn't implement the Application Identity and has a fresh view on it.

app/domain/authentication/authn_azure/azure_resource.rb
jonahx: The class is already named `AzureResource`, so repeating that in the member names is called stutter.

`type` and `value` read smoother.

app/domain/authentication/authn_azure/application_identity.rb
jonahx: I agree `Restriction` feels like an accurate, good choice here.

jonahx: > I understood that subscription id, resource group, etc are affiliations. They are define the azure identity of a VM, not what it restricts it from authenticating but again, let's talk about this more

To answer @sigalsax 's question directly, I think in another context the affiliations perspective might be appropriate, as might an "attribute" perspective.  But in the context of an authenticator, we're saying, "if you, the application, don't check of all these boxes, you're not allowed to authenticate".  Hence "restriction" feel accurate.  I think "requirements" could work for the same reason.

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -2,30 +2,45 @@ module Authentication
   module AuthnAzure
 
     Log = LogMessages::Authentication::AuthnAzure
+    Err = Errors::Authentication::AuthnAzure
```
jonahx, line 5: Doesn't look like either of these are being used.

app/domain/authentication/authn_azure/application_identity.rb
```diff
@@ -2,30 +2,45 @@ module Authentication
...
+            )
+            @resource_types.push(resource_type)
+          end
+        end
+      end
+
+      # check the `service-id` specific resource first to be more granular
+      def resource_value resource_type
+        annotation_value("authn-azure/#{@service_id}/#{resource_type}") ||
+          annotation_value("authn-azure/#{resource_type}")
```
jonahx, line 43: The method `annotation_value` below hits the database, making this no longer a value object.

jonahx: Also, I believe the way this is written and the how sequel works, this is making a separate database call to try every single possible valid annotation.  Which is both inefficient but also feels like a roundabout way to do it.  Instead, get all the annotations in a single batch inside the command class `ValidateApplicationIdentity` (where database calls are allowed), then pass them into a modified version of this class to get back a collection of `AzureResource` objects.

You'll then be inferring the resource type from the annotation itself, rather than trying all combinations of possible types and checking if an annotation exists for each of them.

Now, as for the collection of `AzureResource`s you return, you could make an custom collection object with a `resource_types` method similar to what you have now, but I don't think that's the simplest approach.  You could also just return an array of `AzureResource`s, and getting the `resource_types` for when you use them in:
```ruby
identifiers_constraints = application_identity.resource_types & identifiers
```
is just a simple `map(&:resource_type)` away.  

But I'd also strongly consider another approach.... will post separately.

app/domain/authentication/authn_azure/validate_application_identity.rb
```diff
@@ -6,122 +6,155 @@ module AuthnAzure
...
+              resource_value: xms_mirid.providers.last
+            )
+          )
+        else
+          @resources_from_token.push(
+            AzureResource.new(
+              resource_type: "system-assigned-identity",
+              resource_value: @oid_token_field
+            )
+          )
```
jonahx, line 137: I see another missing concept here, in both the above and in what you are pulling out of the annotations, and that is the idea of `ResourceSet`.  Afaict from the code, a valid ResourceSet consists of:
1. subscription-id
2. resource-group
3. one of either `user-assigned-identity` or `system-identity`.

You can then create / validate such objects from either a token or from annotations.

You could them replace code like this:

```ruby
       application_identity.resources.each do |resource_from_role|
          @resources_from_token.each do |resource_from_token|
            if resource_from_token.resource_type == resource_from_role.resource_type &&
              resource_from_token.resource_value != resource_from_role.resource_value
                raise Err::InvalidApplicationIdentity, resource_from_role.resource_type
            end
          end
        end
```
with something like:
```ruby
if token_resource_set != annotation_resource_set
  raise Err::InvalidApplicationIdentity
end
```
and define a `mismatches(other_resource_set)` method on it, if it's important to include where they don't match.

I think changes like these would help clarify the high level logic, which I think is still simpler than it seems in this version of the code.

app/domain/authentication/authn_azure/application_identity.rb
orenbm: Thanks, I will change it to `ResourceRestriction` and we can continue from there.

app/domain/authentication/authn_azure/application_identity.rb
orenbm: i am not sure that it calls the database each time because we initialize the `ApplicationIdentity` object like this:
```
def application_identity
        @application_identity ||= @application_identity_class.new(
          role_annotations: role.annotations,
          service_id:       @service_id,
          azure_resource_types: AZURE_RESOURCE_TYPES,
          logger:           @logger
        )
      end
```

then, inside `ApplicationIdentity` each call to `@role_annotations` will just call the given value to `role_annotations`. I may be wrong of course.

Anyway, i changed the calling method to:
```
      def application_identity
        @application_identity ||= @application_identity_class.new(
          role_annotations: role_annotations,
          service_id:       @service_id,
          azure_resource_types: AZURE_RESOURCE_TYPES,
          logger:           @logger
        )
      end

      def role_annotations
        @role_annotations ||= role.annotations
      end
```

so it fixes the issue and the method `annotation_value` does not hit the DB.

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: I am not sure i understand your suggestion here but i just want to note that it's not really a `!=` check as they don't need to match. The logic says that the `token_resource_set` should be a super-set of the `annotations_resource_set`.

jonahx: So you may not want to use `!=` as the operator, but my point should stand.  Instead of a custom `==` operator you can create a `subset?(other)` method.  The only question is if the idea of `ResourceSet` (or another more appropriate name if you have one) is really at play here.  It looks like it is.   But let me know what else was unclear.

orenbm:
@sigalsax @jonahx please re-review. I made quite a lot of changes (per your requests) so take your time with it.

thanks :)

app/domain/authentication/authn_azure/validate_application_identity.rb
orenbm: yes, it seems correct. just wanted to verify.

orenbm: do you mean something like this:
```
module Authentication
  module AuthnAzure

    class ResourceSet

      attr_reader :resources

      def self.from_role(role_annotations:, service_id:, azure_resource_types:, logger:)
        resources = azure_resource_types.each_with_object([]) do |resource_type, resources|
          resource_value = resource_value(role_annotations, service_id, resource_type)
          if resource_value
            resources.push(
              AzureResource.new(
                type: resource_type,
                value: resource_value
              )
            )
          end
        end

        new(resources: resources, logger: logger)
      end
      
      def self.from_token(xms_mirid_token_field:, oid_token_field:, logger:)
        xms_mirid = XmsMirid.new(xms_mirid_token_field)
        
        resources ||= [
          AzureResource.new(
            type: "subscription-id",
            value: xms_mirid.subscriptions
          ),
          AzureResource.new(
            type: "resource-group",
            value: xms_mirid.resource_groups
          )
        ]

        # determine which identity is provided in the token. If the token is
        # issued to a user-assigned identity then we take the identity name.
        # If the token is issued to a system-assigned identity then we take the
        # Object ID of the token.
        if xms_mirid.providers.include? "Microsoft.ManagedIdentity"
          resources.push(
            AzureResource.new(
              type: "user-assigned-identity",
              value: xms_mirid.providers.last
            )
          )
        else
          resources.push(
            AzureResource.new(
              type: "system-assigned-identity",
              value: oid_token_field
            )
          )
        end
        logger.debug(Log::ExtractedResourceRestrictionsFromToken.new)

        new(resources: resources, logger: logger)
      end
      
      def valid?
        validate_permitted_scope
        validate_required_constraints_exist
        validate_constraint_combinations
      end
      
      def subset?(other_resource_set)
        @resources.each do |resource|
          other_resource_set.each do |other_resource|
            if other_resource.type == resource.type &&
              other_resource.value != resource.value
              raise Err::InvalidResourceRestrictions, resource.type
            end
          end
        end
      end
      
      private

      def initialize(resources:, logger:)
        @resources = resources
        @logger           = logger
      end

      # check the `service-id` specific resource first to be more granular
      def resource_value role_annotations, service_id, resource_type
        annotation_value(role_annotations, "authn-azure/#{service_id}/#{resource_type}") ||
          annotation_value(role_annotations, "authn-azure/#{resource_type}")
      end

      def annotation_value role_annotations, name
        annotation = role_annotations.find { |a| a.values[:name] == name }

        # return the value of the annotation if it exists, nil otherwise
        if annotation
          @logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
          annotation[:value]
        end
      end

      # validating that the annotations listed for the Conjur resource align with the permitted Azure constraints
      def validate_permitted_scope
        validate_prefixed_permitted_annotations("authn-azure/")
        validate_prefixed_permitted_annotations("authn-azure/#{@service_id}/")
      end

      # check if annotations with the given prefix is part of the permitted list
      def validate_prefixed_permitted_annotations prefix
        @logger.debug(LogMessages::Authentication::ValidatingAnnotationsWithPrefix.new(prefix))

        prefixed_annotations(prefix).each do |annotation|
          annotation_name = annotation[:name]
          next if prefixed_permitted_constraints(prefix).include?(annotation_name)
          raise Errors::Authentication::ConstraintNotSupported.new(annotation_name.gsub(prefix, ""), AZURE_RESOURCE_TYPES)
        end
      end

      def prefixed_annotations prefix
        role_annotations.select do |a|

            # verify we take only annotations from the same level
            annotation_name.split('/').length == prefix.split('/').length + 1
        end
      end

      # add prefix to all permitted constraints
      def prefixed_permitted_constraints prefix
        AZURE_RESOURCE_TYPES.map { |k| "#{prefix}#{k}" }
      end

      def validate_required_constraints_exist
        validate_resource_constraint_exists "subscription-id"
        validate_resource_constraint_exists "resource-group"
      end

      def validate_resource_constraint_exists resource_type
        resource = resource_restrictions.resources.find { |a| a.type == resource_type }
        raise Err::RoleMissingConstraint, resource_type unless resource
      end

      # validates that the resource restrictions do include logical resource constraint
      # combinations (e.g user_assigned_identity & system_assigned_identity)
        identifiers = %w(user-assigned-identity system-assigned-identity)

        identifiers_constraints = resource_restrictions.resources.map(&:type) & identifiers
        unless identifiers_constraints.length <= 1
          raise Errors::Authentication::IllegalConstraintCombinations, identifiers_constraints
        end
      end

    end
  end
end
```

if so, it doesn't do much more than move the logic from `validate_resource_restrictions` to this class.

orenbm: ok I looked into it and I don't think it will be an easy change. In Azure it works in a way (although it mostly moves logic from `validate_resource_restrictions.rb` to this new class) but in k8s it will be a harder refactor. It is not conceptually wrong to do it in k8s but the refactor will need to go deeper for that, because of the way the k8s authenticator validates the resource restrictions.

So at this point I'd rather leave it the way it is now.

jonahx: I would suggest that instead of trying to find the abstraction first, that you make the code for each authenticator as clear and simple as possible.

If, after that, there is still commonality, you can factor it out.  

If not, then it was probably a premature abstraction.  

In general, it's dangerous to let your abstractions dictate how you code, especially when they don't match your mental model.  Instead, they should arise naturally out of the code.  I still maintain that in the case of these authenticators the individual code for each authenticator is not yet simple enough, and so I would first work to make that happen.

orenbm: please see this commit with the changes: https://github.com/cyberark/conjur/pull/1580/commits/2beaffa81f83a5cfb9fce2da05ab07a18132e723

do you think it is easier to understand now?

app/domain/authentication/authn_azure/resource_restrictions.rb
```diff
@@ -1,29 +1,70 @@
...
+        end
       end
 
       private
 
-      # check the `service-id` specific constraint first to be more granular
-      def constraint_value constraint_name
-        annotation_value("authn-azure/#{@service_id}/#{constraint_name}") ||
-          annotation_value("authn-azure/#{constraint_name}")
+      def init_resources
```
sigalsax, line 50: double like for clarity

app/domain/authentication/authn_azure/validate_resource_restrictions.rb
```diff
@@ -3,124 +3,93 @@
...
+      def resource_restrictions
+        @resource_restrictions ||= @resource_restrictions_class.new(
+          role_annotations: role_annotations,
+          service_id:       @service_id,
+          logger:           @logger
+        )
+      end
+
+      def validate_resource_restrictions_configuration
+        resource_restrictions.validate_configuration
```
sigalsax, line 37: What is your motivation in placing the logic in ResourceRestrictions and not in ValidateResourceRestrictions? 
The functions in ResourecRestrictions are performing validation on the xms mirid fields so stylistically wondering why they aren't in their calling class

app/domain/authentication/authn_k8s/k8s_resource.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    class K8sResource
+      attr_reader :type, :value
+
+      def initialize(type:, value:)
+        @type = type
+        @value = value
+      end
+
+      def ==(other_resource)
```
sigalsax, line 12: why an operator and not `matches` here?

sigalsax: or `equal` / `isEqual`

app/domain/authentication/authn_k8s/resource_restrictions.rb
```diff
@@ -3,140 +3,89 @@
...
+    #   - type: "namespace"
+    #   - value: "some-namespace"
+    #
+    # then this Conjur host can authenticate with Conjur only from a pod that is
+    # part of the namespace "some-namespace"
+    class ResourceRestrictions
+
+      attr_reader :resources
+
+      K8S_RESOURCE_TYPES = %w(namespace service-account pod deployment stateful-set deployment-config)
```
sigalsax, line 21: I like how both authenticators follow the same pattern in their structure

app/domain/authentication/authn_k8s/resource_restrictions.rb
```diff
@@ -3,140 +3,89 @@
...
+      #              "namespace/blah/some-value")
+      def validate_constraints_are_permitted
+        resource_restrictions_in_annotations? ? validate_permitted_annotations : validate_host_id
       end
 
-      # Return the last part of the host id (which is the actual hostname).
-      # The host id is build as "account_name:kind:identifier" (e.g "org:host:some_hostname").
-      def hostname
-        @hostname ||= @host_id.split(':')[2]
+      # We expect the resource restrictions to be defined by the host's annotations
```
sigalsax, line 83: > We expect the resource restrictions to be defined by the host's annotations
"and not in the host", correct?

app/domain/authentication/authn_k8s/resource_restrictions.rb
```diff
@@ -3,140 +3,89 @@
...
+    # This class represents the restrictions that are set on a Conjur host regarding
+    # the K8s resources that it can authenticate with Conjur from.
+    # It consists a list of K8sResource objects which represent the resource
+    # restriction that need to be met in an authentication request.
+    #
+    # For example, if `resources` includes the K8sResource:
+    #   - type: "namespace"
+    #   - value: "some-namespace"
+    #
+    # then this Conjur host can authenticate with Conjur only from a pod that is
```
sigalsax, line 15: I would mention here for clarity purposes that in authn-k8s there are two ways that resource restrictions can be provided for authentication 1. via host annotations (newer) 2. via host id (older)

app/domain/authentication/authn_k8s/resource_restrictions.rb
```diff
@@ -181,52 +127,59 @@ def prefixed_permitted_annotations prefix
         permitted_annotations.map { |k| "#{prefix}#{k}" }
       end
 
+      def permitted_annotations
+        @permitted_annotations ||= K8S_RESOURCE_TYPES | [@authentication_container_name_annotation]
```
sigalsax, line 131: what is `|`?

app/domain/authentication/authn_k8s/validate_resource_restrictions.rb
```diff
@@ -103,20 +97,29 @@ def pod_namespace
         @spiffe_id.namespace
       end
 
-      def k8s_resource_object resource_type, resource_name, namespace
-        @k8s_resource_object = k8s_object_lookup.find_object_by_name(
-          resource_type,
-          resource_name,
-          namespace
+      # @return The Conjur resource for the webservice.
```
sigalsax, line 100: is the `.` in comment purposeful?

spec/app/domain/authentication/authn_k8s/validate_pod_request_spec.rb
```diff
@@ -96,16 +97,153 @@
...
 
         expect { subject }.to(
           raise_error(
-            validate_application_identity_error
+            /FAKE_RESOURCE_RESTRICTIONS_ERROR/
           )
         )
       end
+
+      #context "with different annotations for container name" do
```
sigalsax, line 111: I assume this is the test you have left?

app/domain/authentication/authn_k8s/validate_pod_request.rb
```diff
@@ -5,14 +5,18 @@
 module Authentication
   module AuthnK8s
 
+    AUTHENTICATION_CONTAINER_NAME_ANNOTATION = "authentication-container-name"
```
sigalsax, line 8: inject_client_cert.rb has the following comment

```

# This code is implemented similarly also in application_identity.rb
# We have it here too as we need the container name for the injection
# and it simplifies the code to have this specific 2 methods duplicated
# rather than passing around the ApplicationIdentity object
```
I think we should also update Application Identity here

app/domain/authentication/authn_k8s/k8s_resource.rb
orenbm: implementing this operator let's us compare between hashs of AzureResource easily

app/domain/authentication/authn_k8s/resource_restrictions.rb
orenbm: read the next line :)

```
# We expect the resource restrictions to be defined by the host's annotations
      # if any of the constraint annotations is present.
```

app/domain/authentication/authn_azure/validate_resource_restrictions.rb
orenbm: update after chat: moved `validate_configuration` to be part of the initializer and `match_request` to `validate_resource_restrictions`

app/domain/authentication/authn_k8s/resource_restrictions.rb
orenbm: unifies both arrays

spec/app/domain/authentication/authn_k8s/validate_pod_request_spec.rb
orenbm: yes, implemented a new class - `extract_container_name` and moves these UTS THERE

app/domain/authentication/authn_k8s/validate_pod_request.rb
orenbm: done, thanks

README.md
```diff
@@ -10,7 +10,7 @@
 [commons]: https://discuss.cyberarkcommons.org/c/conjur/5 "Find answers on CyberArk Commons"
 [twitter]: https://twitter.com/intent/user?screen_name=ConjurInc "Follow Conjur on Twitter"
 
-Conjur provides secrets management and application identity for modern infrastructure:
+Conjur provides secrets management and resource restrictions for modern infrastructure:
```
sigalsax, line 13: > Conjur provides secrets management and resource restrictions for modern infrastructure:

Sounds a bit weird no?

orenbm: named back to application identity.

orenbm: it is another application identity

app/domain/authentication/authn_k8s/validate_pod_request.rb
```diff
@@ -74,8 +75,8 @@ def validate_resource_restrictions
       def validate_container
         unless container
           raise Errors::Authentication::AuthnK8s::ContainerNotFound,
-            container_name,
-            k8s_host.conjur_host_id
+                container_name,
```
sigalsax, line 78: seems like a large tab no?

orenbm: code climate asked for them to be aligned with `Errrors`

PR #1585, opened 2020-06-03
Fixing apt-get installation order

sashaCher:
By default `curl` and `git` try to install a higher versions of packages those are installed by `libz-dev` then `libz-dev` installs by itself.
Installing `libz-dev` in advance solves the problem.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,13 +10,12 @@ EXPOSE 80
 
 RUN apt-get update -y && \
     apt-get -y dist-upgrade && \
-    apt-get install -y build-essential \
+    apt-get install -y libz-dev
+RUN apt-get install -y build-essential \
                        curl \
                        git \
                        ldap-utils \
                        tzdata \
-                       # needed to build some gem native extensions:
-                       libz-dev \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,13 +10,12 @@ EXPOSE 80
 
 RUN apt-get update -y && \
     apt-get -y dist-upgrade && \
-    apt-get install -y build-essential \
+    apt-get install -y libz-dev
```
aloncarmel111, line 13:
```suggestion
    apt-get install -y libz-dev \
                        build-essential                                    
```

sashaCher: @aloncarmel111 Is it really worth additional 30 minutes of changing-pushing-waiting for build etc?

aloncarmel111: actually build-essential is already installed in base-image so we can removed this

sashaCher: Is it a necessary part of base image?
Base should be as small as possible.
My suggestion is to remove it from the base.

aloncarmel111: ok let's create followup task to removed build-essential from base-image

Issue #1589, opened 2020-06-04
Version is printed only on server startup


orenbm:
We recently started to print the Conjur version on server startup:
```
root@094a586bdea9:/opt/conjur-server# conjurctl server
CONJ00037I Conjur v1.7.1 starting up...
=> Booting Puma
=> Rails 5.2.4.3 application starting in production
=> Run `rails server -h` for more startup options
```

However, this message is written for every `conjurctl` command. For example:
```
root@094a586bdea9:/opt/conjur-server# conjurctl role retrieve-key cucumber:user:some-user
CONJ00037I Conjur v1.7.1 starting up...
error: role does not exist: cucumber:user:some-user
error: key retrieval failed
```

This can be an issue for users using the `conjurctl role` command in their CI as instead of getting an API key: 
```
CONJUR_ADMIN_AUTHN_API_KEY=$(docker-compose exec -T conjur conjurctl role retrieve-key myAccout:user:admin)
# $CONJUR_ADMIN_AUTHN_API_KEY is "<api_key>"
```

they will get the message + the api key:
```
CONJUR_ADMIN_AUTHN_API_KEY=$(docker-compose exec -T conjur conjurctl role retrieve-key myAccout:user:admin)
# $CONJUR_ADMIN_AUTHN_API_KEY is "CONJ00037I Conjur v1.7.1 starting up... 
#                                                                      <api_key>"
```

We should fix this so the message is printed only for server startup.

PR #1590, opened 2020-06-04
Print Conjur version only on server startup

orenbm:
Connected to #1589

We recently started to print the Conjur version on server startup:
```
root@094a586bdea9:/opt/conjur-server# conjurctl server
CONJ00037I Conjur v1.7.1 starting up...
=> Booting Puma
=> Rails 5.2.4.3 application starting in production
=> Run `rails server -h` for more startup options
```

However, this message is written for every `conjurctl` command. For example:
```
root@094a586bdea9:/opt/conjur-server# conjurctl role retrieve-key cucumber:user:some-user
CONJ00037I Conjur v1.7.1 starting up...
error: role does not exist: cucumber:user:some-user
error: key retrieval failed
```

This PR changes the behaviour so that the message is printed only on server startup.

After the change, the output of `conjurctl server` is:
```
root@094a586bdea9:/opt/conjur-server# conjurctl server
Conjur v1.7.1 starting up...
=> Booting Puma
=> Rails 5.2.4.3 application starting in production
=> Run `rails server -h` for more startup options
```

and the output of `conjurctl role` is:
```
root@094a586bdea9:/opt/conjur-server# conjurctl role retrieve-key cucumber:user:some-user
error: role does not exist: cucumber:user:some-user
error: key retrieval failed
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Fixed
+- The Conjur version is now printed on server startup, after running `conjurctl server`
+  ([cyberark/conjur#1590](https://github.com/cyberark/conjur/pull/1590))
 
 ## [1.7.1] - 2020-06-03
 
--- a/bin/conjur-cli.rb
+++ b/bin/conjur-cli.rb
@@ -72,6 +72,9 @@ command :server do |c|
     end
 
     Process.fork do
+      conjur_version = File.read(File.expand_path("../VERSION", File.dirname(__FILE__))).strip
+      puts "Conjur v#{conjur_version} starting up..."
+
       exec "rails server -p #{options[:port]} -b #{options[:'bind-address']}"
     end
     Process.fork do
--- a/cucumber/api/features/retrieve_api_key.feature
+++ b/cucumber/api/features/retrieve_api_key.feature
@@ -1,0 +1,14 @@
+Feature: Retrieving an API key with conjurctl
+
+  Background:
+    # We need to be in production environment to test this to demonstrate a real use-case
+    Given I set environment variable "RAILS_ENV" to "production"
+    And I set environment variable "CONJUR_LOG_LEVEL" to "info"
+
+  Scenario: Retrieve an API key
+    When I retrieve an API key for user "cucumber:user:admin" using conjurctl
+    Then the API key is correct
+
+  Scenario: Retrieve an API key of a non-existing user fails
+    When I retrieve an API key for user "cucumber:user:non-existing-user" using conjurctl
+    Then the stderr includes the error "role does not exist"
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -222,11 +222,6 @@ unless defined? LogMessages::Authentication::OriginValidated
         code: "CONJ00023D"
       )
 
-      ConjurVersionStartup = ::Util::TrackableLogMessageClass.new(
-        msg:  "Conjur v{0-conjur-version} starting up...",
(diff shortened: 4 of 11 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Fixed
+- The Conjur version is now printed on server startup, after running `conjurctl server`
+  ((cyberark/conjur#1590)[https://github.com/cyberark/conjur/pull/1590])
```
jtuttle, line 10: The parens are backwards here. The issue should have square brackets and the link should have parens.

bin/conjur-cli.rb
```diff
@@ -72,6 +72,9 @@ def test_select
     end
 
     Process.fork do
+      conjur_version = (File.read(File.expand_path("../VERSION", File.dirname(__FILE__)))).strip
```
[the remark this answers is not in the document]
jtuttle: Can we use some helper variables to make this line less complex? Might solve this codeclimate warning too.

cucumber/api/features/retrieve_api_key.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Retrieving an API key with conjurctl
+
+  # We need to be in production environment to test this
```
eladkug, line 3: Can you explain why we need to be in production?

cucumber/api/features/retrieve_api_key.feature
```diff
@@ -0,0 +1,8 @@
...
+  Scenario: Retrieve an API key
+    Given I set environment variable "RAILS_ENV" to "production"
+    And I set environment variable "CONJUR_LOG_LEVEL" to "info"
```
eladkug, line 6: Why setting the conjur_log_level is relevant here?

cucumber/api/features/retrieve_api_key.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Retrieving an API key with conjurctl
```
eladkug, line 1: I would add another test to verify we have error when the API is wrong(role does not exist).

cucumber/api/features/retrieve_api_key.feature
orenbm: because that's the environment our clients will be. When we are in dev env the error doesn't occur

cucumber/api/features/retrieve_api_key.feature
orenbm: because if we are in debug mode the output will contain some junk. we change the env so we are in the expected environment of the client using this tool.

eladkug: o.k. Since its not suppose to influence the results.
In general i want to work with relevant steps only(in this case its not critical) because if from some reason we will fail on this step we cant say "retrieve an API key" is not working.

cucumber/api/features/retrieve_api_key.feature
orenbm: done

cucumber/api/features/retrieve_api_key.feature
```diff
@@ -0,0 +1,14 @@
...
+    Given I set environment variable "RAILS_ENV" to "production"
+    And I set environment variable "CONJUR_LOG_LEVEL" to "info"
+
+  Scenario: Retrieve an API key
+    When I retrieve an API key for user "cucumber:user:admin" using conjurctl
+    Then the API key is correct
+
+  Scenario: Retrieve an API key of a non-existing user fails
+    When I retrieve an API key for user "cucumber:user:non-existing-user" using conjurctl
+    Then the stderr includes the error "role does not exist"
```
eladkug, line 14: new line is missing

PR #1591, opened 2020-06-04
Raise proper error in case user not defined with `authn`

orenbm:
Connected to #1593 

In case a user sends an authn request with the `authn` authenticator,
and the user is not defined in Conjur, the log message will show:
`Authentication Error: #<NoMethodError: undefined method `valid_origin?' for nil:NilClass>`.

This is not informative and doesn't tell the user the real story. This commit will change
the message to:
`Authentication Error: #<Errors::Authentication::Security::RoleNotFound: CONJ00007E ‘cucumber:user:non-existing-user’ not found>`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - The Conjur version is now printed on server startup, after running `conjurctl server`
   ([cyberark/conjur#1590](https://github.com/cyberark/conjur/pull/1590))
+- Raise proper error of an authn request with a non-existing user to the `authn`
+  authenticator ([cyberark/conjur#1591](https://github.com/cyberark/conjur/pull/1591))
 
 ## [1.7.1] - 2020-06-03
 
--- a/app/domain/authentication/validate_origin.rb
+++ b/app/domain/authentication/validate_origin.rb
@@ -4,7 +4,7 @@ module Authentication
 
   Err ||= Errors::Authentication
   # Possible Errors Raised:
-  # InvalidOrigin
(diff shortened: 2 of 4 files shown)


app/domain/authentication/validate_origin.rb
```diff
@@ -22,7 +22,15 @@ def call
     private
 
     def role
-      @role_cls.by_login(@username, account: @account)
+      return @role if @role
+
+      @role = @role_cls.by_login(@username, account: @account)
+      raise Errors::Authentication::Security::RoleNotFound, role_id unless @role
```
izgeri, line 28: should this just be
```suggestion
      raise Err::Security::RoleNotFound, role_id unless @role
```
since `Err ||= Errors::Authentication` above?

also, does `Security::RoleNotFound` need to be added to the comment at the top of the file?

orenbm: thanks, nice catch

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Fixed
+- Raise proper error of an authn request with a non-existing user to the `authn`
+  authenticator ((cyberark/conjur#1591)[https://github.com/cyberark/conjur/pull/1591])
```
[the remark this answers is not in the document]
orenbm: i created the bug and add it to the PR description. Do we want it also in the Changelog? I think that the PR link is enough.

orenbm:
> Before we merge this, I asked a question in Slack that needs to be resolved from a security standpoint.

Thanks, I'll look there. Please approve the PR once it is resolved.

Issue #1550, opened 2020-05-15
Requestor IP address is included in all Audit events


jvanderhoof:
As a security administrator, I want all audit events to include the requester's IP address so that I can identify the origin of the request that created the audit event.

GIVEN a running Conjur instance
WHEN a request that generates an audit event (or events) is made
THEN those events include the IP address of the requestor <sup>1</sup>

**Notes**
- We might want to consider using something like [Nginx Real IP](https://nginx.org/en/docs/http/ngx_http_realip_module.html).

--- 

<sup>1</sup> We'll never be 100% successful at getting the requestor's IP address, but we should strive to achieve this as best we can.

PR #1595, opened 2020-06-05
Add Client IP to Audit Events

micahlee:
### What does this PR do?

This PR updates all audit events to include the caller's client IP address.

The container image `registry.tld/conjur-appliance:5.9.0-20200609134725-fe79d95` may be used to evaluate this change manually, for example in [dap-intro](https://github.com/conjurdemos/dap-intro/tree/master/demos/cluster).

- _Example Audit Event_

Example audit record after this change, with the addition of `[client@43868 ip="172.24.0.5"]`:
```
<37>1 2020-05-28T20:43:35.602+00:00 ce1fd1ccf9ac conjur 728c4c1f-b027-4bf2-861b-47f3b756d559 policy [auth@43868 user="demo:user:admin"][subject@43868 role="demo:group:staging/myapp/secrets-users" member="demo:group:developers"][action@43868 operation="add"][client@43868 ip="172.24.0.5"][policy@43868 id="demo:policy:root" version="3"][meta sequenceId="150"] demo:user:admin added membership of demo:group:developers in demo:group:staging/myapp/secrets-users
```

- _Appliance Build_

https://jenkins.conjur.net/job/conjurinc--appliance/job/1550-audit-client-ip-refactored/

Connected to #1550

#### Change log

#### Test coverage

#### Documentation

R&D Docs Issue: https://github.com/cyberark/dap-wiki/issues/45

### DIFF
--- a/app/domain/authentication/authenticate.rb
+++ b/app/domain/authentication/authenticate.rb
@@ -22,7 +22,7 @@ module Authentication
     extend Forwardable
     def_delegators(
       :@authenticator_input, :authenticator_name, :account, :username,
-      :webservice, :origin, :role
+      :webservice, :client_ip, :role
     )
 
     def call
@@ -73,7 +73,7 @@ module Authentication
       @validate_origin.(
         account: account,
         username: username,
-        origin: origin
+        client_ip: client_ip
       )
     end
 
@@ -83,7 +83,7 @@ module Authentication
           authenticator_name: authenticator_name,
           service:            webservice,
           role:               role,
-          client_ip:          origin,
+          client_ip:          client_ip,
           success:            true,
           error_message:      nil
         )
@@ -96,7 +96,7 @@ module Authentication
           authenticator_name: authenticator_name,
           service:            webservice,
           role:               role,
-          client_ip:          origin,
+          client_ip:          client_ip,
           success:            false,
           error_message:      err.message
         )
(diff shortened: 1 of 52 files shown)


app/domain/authentication/authenticate.rb
```diff
@@ -86,6 +87,7 @@ def audit_failure(err)
           authenticator_name: authenticator_name,
           service: webservice,
           role: role,
+          client_ip: origin,
```
jonahx, line 90: It just occurs to me... is there a reason to have the two terms `client_ip` and `origin`?

Given `origin` is more vague, can we use `client_ip` everywhere?

shaharglazner:
@micahlee 
I saw that most of the audit client IP is based on request.ip, what will happen when the client go through a LB before reaching Conjur? Which IP will be audited?

micahlee:
@shaharglazner 
> I saw that most of the audit client IP is based on request.ip, what will happen when the client go through a LB before reaching Conjur? Which IP will be audited?

It depends on how `HTTP_X_FORWARDED_FOR` is handled in the LB, but you can read through a pretty good description of how it works in [this stack overflow thread](https://stackoverflow.com/questions/10997005/whats-the-difference-between-request-remote-ip-and-request-ip-in-rails) (that references the Rails and Rack source).

From my understanding, this should really probably be `request.remote_ip` instead, but that should be changed throughout, and not just in audit. We used `request.ip` here because that's currently the same source of IP address we use for `restricted_to/origin` checking for authentication. This way our treatment of the client IP address is consistent throughout.

Issue #1548, opened 2020-05-15
Changing a Password is audited


jvanderhoof:
As a security administrator, I want to know when users attempt to change their passwords so that I can detect any suspicious behavior.

GIVEN a running Conjur instance
WHEN I change my password with an invalid password <sup>1</sup>
THEN I see an unsuccessful password change audit event <sup>2</sup>

AND WHEN I change my password with a valid password <sup>3</sup>
THEN I see a successful password change audit event <sup>4</sup>

**Notes**
Conjur does not currently support changing another user's password. The only way to "fail" when setting a password is to attempt to change a password to something that does not meet the complexity requirements.

---

<sup>1</sup> Change the `admin` user's password to a password that does not meet the complexity requirements:
```sh
$ curl -k -X PUT --data password --user admin:ADmin123!!!! \
http://localhost:3000/authn/cucumber/password
```

<sup>2</sup> Unsuccessful "Change Password" audit event should follow the [Audit Event convention](https://github.com/cyberark/dap-wiki/blob/api-docs/reference-architectures/audit.md). Before we implement, let's get agreement on the format. A possible format could be:
- subject@43868 
    - user: `cucumber:user:admin`
- auth@43868 
    - user: `cucumber:user:admin`
- action@43868 
    - result: `failure`
    - operation: `update` 
- MSGID: password
- MESSAGE: "cucumber:user:admin failed to update password for cucumber:user:admin due to complexity requirements"
- LEVEL: "notice"

<sup>3</sup> Change the `admin` user's password to a password that meets the complexity requirements:
```sh
$ curl -k -X PUT --data ADmin123!!-new --user admin:ADmin123!!!!
http://localhost:3000/authn/cucumber/password
```

<sup>4</sup> Successful "Change Password" audit event should follow the [Audit Event convention](https://github.com/cyberark/dap-wiki/blob/api-docs/reference-architectures/audit.md). Before we implement, let's get agreement on the format. A possible format could be:
- subject@43868
    - user: `cucumber:user:admin`
- auth@43868
    - user: `cucumber:user:admin`
- action@43868
    - result: `success`
    - operation: `update`
- MSGID: password
- MESSAGE: "cucumber:user:admin updated password for cucumber:user:admin"
- LEVEL: "notice"

micahlee:
Possible locations to hook in the audit event:

* In the controller:
[...]

PR #1600, opened 2020-06-08
Add Audit Event for Password Change

micahlee:
**NOTE:** This PR is blocked until the [parent branch](https://github.com/cyberark/conjur/pull/1574) is merged.

---

### What does this PR do?

This PR adds an audit log message for password changes events using the API endpoint `PUT /authn/:account/password` 

The container image `registry.tld/conjur-appliance:5.9.0-20200610203923-d5e10a1` may be used to evaluate this change manually, for example in [dap-intro](https://github.com/conjurdemos/dap-intro/tree/master/demos/cluster).

- _Example audit message_

**Successful Password Change:**
```
<86>1 2020-06-10T21:04:53.745+00:00 60aa5330d24b conjur fb98c0b8-136e-4f6f-ab35-1f545acadba0 password [auth@43868 user="demo:user:admin"][subject@43868 user="demo:user:admin"][action@43868 result="success" operation="change"][meta sequenceId="3"] demo:user:admin successfully changed their password
```

**Failed Password Change:**
```
<84>1 2020-06-10T21:03:32.205+00:00 60aa5330d24b conjur 74acf74b-7743-41d7-be1d-1dc060356d26 password [auth@43868 user="demo:user:admin"][subject@43868 user="demo:user:admin"][action@43868 result="failure" operation="change"][meta sequenceId="5"] demo:user:admin failed to change their password: password CONJ00046E The password you have chosen does not meet the complexity requirements. Choose a password that includes: 12-128 characters, 2 uppercase letters, 2 lowercase letters, 1 digit, 1 special character
```

- _Appliance Build_

https://jenkins.conjur.net/job/conjurinc--appliance/job/1548-audit-password-change/

Connected to #1548

#### Change log

#### Test coverage

#### Documentation

R&D Docs (`dap-wiki`) PR: [cyberark/dap-wiki#29](https://github.com/cyberark/dap-wiki/pull/29)

### DIFF
--- a/app/controllers/credentials_controller.rb
+++ b/app/controllers/credentials_controller.rb
@@ -36,8 +36,11 @@ class CredentialsController < ApplicationController
   def update_password
     password = request.body.read
 
-    @role.credentials.password = password
-    @role.credentials.save(raise_on_save_failure: true)
+    Commands::Credentials::ChangePassword.new.call(
+      role: @role,
+      password: password
+    )
+        
     head 204
   end
   
--- a/app/domain/commands/credentials/change_password.rb
+++ b/app/domain/commands/credentials/change_password.rb
@@ -19,6 +19,10 @@ module Commands
 
       def call
         change_password
+        audit_success
+      rescue => e
+        audit_failure(e)
+        raise e
       end
 
       private
@@ -30,6 +34,22 @@ module Commands
 
       def credentials
         @role.credentials
+      end
+
+      def audit_success
+        @audit_log.log(
+          ::Audit::Event::Password.new(user: @role, success: true)
+        )
+      end
+
+      def audit_failure(err)
+        @audit_log.log(
+          ::Audit::Event::Password.new(
+            user: @role,
+            success: false,
+            error_message: err.message
+          )
+        )
       end
     end
   end
--- a/app/domain/commands/credentials/change_password.rb
+++ b/app/domain/commands/credentials/change_password.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'command_class'
+
+module Commands
+  module Credentials
+
+    ChangePassword ||= CommandClass.new(
+      dependencies: {
+        audit_log: ::Audit.logger
+      },
+
+      # `role` is not a pure input, but comes with an implicit database
+      # dependency that it will use to save the new password.
+      # This should be addressed to make the database dependency explicit:
+      # https://github.com/cyberark/conjur/issues/1611
+      inputs: %i(role password)
+    ) do
+
+      def call
+        change_password
+      end
+
+      private
+
+      def change_password
+        credentials.password = @password
+        credentials.save(raise_on_save_failure: true)
+      end
+
+      def credentials
+        @role.credentials
+      end
+    end
+  end
+end
--- a/app/models/audit/event/password.rb
+++ b/app/models/audit/event/password.rb
@@ -1,0 +1,71 @@
+module Audit
+  module Event
+    class Password
+
+      def initialize(user:, success:, error_message: nil)
+        @user = user
+        @success = success
+        @error_message = error_message
+      end
+
+      # Note: We want this class to be responsible for providing `progname`.
+      # At the same time, `progname` is currently always "conjur" and this is
+      # unlikely to change.  Moving `progname` into the constructor now
+      # feels like premature optimization, so we ignore reek here.
+      # :reek:UtilityFunction
+      def progname
+        Event.progname
+      end
+
+      def severity
+        attempted_action.severity
+      end
+
+      def to_s
+        message
+      end
+
+      # It's clearer to simply call the #id attribute multiple times, rather
+      # than factor it out, even though it reeks of :reek:DuplicateMethodCall
+      def message
+        attempted_action.message(
+          success_msg: "#{@user.id} successfully changed their password",
+          failure_msg: "#{@user.id} failed to change their password",
+          error_msg: @error_message
+        )
+      end
+
+      def message_id
+        'password'
+      end
+
+      # It's clearer to simply call the #id attribute multiple times, rather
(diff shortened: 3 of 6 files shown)


app/controllers/credentials_controller.rb
```diff
@@ -36,8 +36,11 @@ class CredentialsController < ApplicationController
   def update_password
     password = request.body.read
 
-    @role.credentials.password = password
-    @role.credentials.save(raise_on_save_failure: true)
+    Credentials::ChangePassword.new.(
```
jonahx, line 39: I've moved from `.(` to `.call` for invoking procs for clarity.  It's too easy to misread the shorter syntax.  IIRC ruby style guide may advocate this is as well.

app/models/audit/event/password.rb
```diff
@@ -0,0 +1,61 @@
+module Audit
+  module Event
+    class Password
+
+      def initialize(
+        operation: 'change',
+        user: ,
+        success:,
+        error_message: nil
+      )
```
jonahx, line 10: I know it doesn't matter from the caller's perspective since we're using keyword args, but I'd put all the optional params at the end just for consistency / readability.

Also nit: I think lines 5-10 can fit on a single line under 80 chars.

jonahx: Oh also: If there are other operations here, I think I might prefer making the callers _always_ be explicit about, rather than having a default value for the most common case.  That way, readers can glean everything they know from the calling site, rather than having to jump to definition to see the operation.

Wdyt?

micahlee: So I put this in the draft to let it simmer for a minute in my mind. I think I'm actually inclined to remove `operation` from the inputs, because the only operation we know of right now for passwords is `change`.

The mental exercise here was understanding how it could be parameterized, since that was one of the immediate hurdles when trying to split the single `authn` event into `authn`, `login`, and `inject_client_cert` in the earlier PR.

I think it would be pretty obvious in the current structure how to make this an option of the caller, if that became a requirement in the future. Although, I think the current pattern with `authn` would lead to new operations of `password` just becoming classes themselves.

jonahx: I agree.  YAGNI wins especially when the future change, should we need it, will be trivial to make.

app/models/audit/event/password.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def message_id
+        'password'
+      end
+
+      def structured_data
+        {
+          SDID::AUTH => { user: @user.id },
+          SDID::SUBJECT => { user: @user.id },
+        }.tap do |sd|
```
jonahx, line 35: merge here will avoid the need for local var `sd`

app/domain/commands/credentials/change_password.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'command_class'
+
+module Commands
+  module Credentials
+
+    ChangePassword ||= CommandClass.new(
+      dependencies: {},
```
jonahx: Two problems here:
1. A CommandClass with no dependencies doesn't make any sense, conceptually.
2. We actually, _do_ have dependencies here, but they sneak in implicitly via our _input_, namely the `role`.   Spefically, `credentials.save(raise_on_save_failure: true)` is a database interaction.

Unfortunately, I think fixing this would require completely reworking the credentials controller, which will be too big of a sidetrack right now.

I recommend instead that we keep your code mostly as is, but just make it an ordinary object rather than a `CommandClass`, and add a comment to a placeholder issue for refactoring the controller.  This will allow your test code to remain unchanged, or close to it as well.

spec/app/domain/commands/credentials/change_password_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  it 'bubbles up exceptions' do 
+    # Assume the database update fails. This could be caused by an
+    # invalid password, database issues, etc.
+    allow(credentials).to receive(:password=)
+    allow(credentials).to receive(:save).and_raise(err_message)
+
+    # Expect the command to raise the original exception
+    expect { subject.call(role: role,  password: password) }.to raise_error(err_message)
+  end
+end
```
jonahx, line 57: Given the existing code, and our decision to delay changing it, these tests look perfect.

app/domain/commands/credentials/change_password.rb
micahlee: So the way this is set up, with factoring out the command class first, and then adding audit to it, it only appears to have no dependencies in this step.

The audit log is added in the next commit as a dependency to this command class.

Does this change your assessment at all?

micahlee: Could you share, hypothetically, what your ideal command class for this would look like (even in just pseudocode)? If you could have exactly what you wanted as inputs from the CredentialsController?

That would help me better grasp what the desired end state would be from your perspective.

app/domain/commands/credentials/change_password.rb
```diff
@@ -6,12 +6,18 @@ module Commands
   module Credentials
 
     ChangePassword ||= CommandClass.new(
-      dependencies: {},
+      dependencies: {
+        audit_log: ::Audit.logger
+      },
```
jonahx, line 11: Ok so this somewhat changes my comment on previous commit, but it remains the case that one of our dependencies is still sneaking in through the input.  

My recommendation might change now to:

1. Keep the command class.
2. Add a comment stating the problem above dependencies and inputs and link to issue.

app/domain/commands/credentials/change_password.rb
```diff
@@ -24,6 +30,25 @@ def change_password
       def credentials
         @role.credentials
       end
+
+      def audit_success
+        @audit_log.log(
+          ::Audit::Event::Password.new(
+            user: @role,
+            success: true
+          )
```
jonahx, line 39: Nit: Seems like this would fit on one line

Issue #1549, opened 2020-05-15
Rotating an API key is audited


jvanderhoof:

jvanderhoof:
As a security administrator, I want to know when an API key is rotated so that I can detect any suspicious behavior.

GIVEN a running Conjur instance
WHEN I rotate my own API key
THEN I see a corresponding audit event
AND WHEN I rotate another user or host's API key
THEN I see a corresponding audit event

**Notes**
Please document the audit event formats before starting the code implementation.

PR #1609, opened 2020-06-11
Add Audit Event for API Key Rotation

micahlee:
### What does this PR do?

This PR adds an audit log message for API key rotations events using the API endpoint `PUT /:authenticator/:account/api_key`.

The container image `registry.tld/conjur-appliance:5.9.0-20200612203143-b8c6a2f` may be used to evaluate this change manually, for example in [dap-intro](https://github.com/conjurdemos/dap-intro/tree/master/demos/cluster).

- _Example audit message_

**Successful API Key Rotation for Own Role:**
```
<86>1 2020-06-12T13:28:52.245+00:00 6cc8d001f840 conjur 1d9b2ba1-7a81-45fd-95e8-632e351d9a5a api-key [auth@43868 user="demo:user:admin"][subject@43868 role="demo:user:admin"][action@43868 result="success" operation="rotate"][meta sequenceId="3"] demo:user:admin successfully rotated their API key
```

**Successful API Key Rotation for Another Role:**
```
<86>1 2020-06-12T13:35:52.062+00:00 6cc8d001f840 conjur fb291c77-da64-4b86-8374-c06d9da537e6 api-key [auth@43868 user="demo:user:admin"][subject@43868 role="demo:user:marcel.calisto"][action@43868 result="success" operation="rotate"][meta sequenceId="2"] demo:user:admin successfully rotated the api key for demo:user:marcel.calisto
```

- _Appliance Build_

https://jenkins.conjur.net/job/conjurinc--appliance/job/1549-audit-api-key-rotation/

Connected to #1549

#### Change log

#### Test coverage

#### Documentation

R&D Docs (`dap-wiki`) PR: [cyberark/dap-wiki#29](https://github.com/cyberark/dap-wiki/pull/29)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,16 +21,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   absence of the `Content-Type` header
   ([cyberark/conjur#1622](https://github.com/cyberark/conjur/issues/1622))
 
+### Added
+- Password changes (`PUT /authn/:account/password`) now produce audit events with
+  message ID `password` ([cyberark/conjur#1548](https://github.com/cyberark/conjur/issues/1548))
+- API key rotations (`PUT /:authenticator/:account/api_key`) now produce audit events with
+  message ID `api-key` ([cyberark/conjur#1549](https://github.com/cyberark/conjur/issues/1549))
+
 ## [1.7.3] - 2020-06-11
 
 ### Fixed
 - Host Factory Host creation no longer makes unecessary database queries, causing
   performance issues with large numbers of created hosts
   ([cyberark/conjur#1605](https://github.com/cyberark/conjur/issues/1605))
-
-### Added
-- Password changes (`PUT /authn/:account/password`) now produce audit events with
-  message ID `password` ([cyberark/conjur#1548](https://github.com/cyberark/conjur/issues/1548))
 
 ## [1.7.2] - 2020-06-08
 
--- a/app/domain/commands/credentials/rotate_api_key.rb
+++ b/app/domain/commands/credentials/rotate_api_key.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require 'command_class'
+
+module Commands
+  module Credentials
+
+    # In might seem unnecessary to move this behavior into a command class.
+    # However, we factor this out now to prepare for A) adding audit decorations
+    # to this operation, and B) those audit events including the requestor's
+    # client IP address, which will be an input to this CommandClass, rather than
+    # to the Credential model's method.
+    RotateApiKey ||= CommandClass.new(
+      dependencies: {},
+      inputs: %i(role)
+    ) do
+
+      def call
+        rotate_api_key
+      end
+
+      private
+
+      def rotate_api_key
+        credentials.rotate_api_key
+        credentials.save
+      end
+
+      def credentials
+        @role.credentials
+      end
+    end
+  end
+end
--- a/app/domain/commands/credentials/rotate_api_key.rb
+++ b/app/domain/commands/credentials/rotate_api_key.rb
@@ -11,12 +11,18 @@ module Commands
     # client IP address, which will be an input to this CommandClass, rather than
     # to the Credential model's method.
     RotateApiKey ||= CommandClass.new(
-      dependencies: {},
-      inputs: %i(role)
+      dependencies: {
+        audit_log: ::Audit.logger
+      },
+      inputs: %i(role_to_rotate authenticated_role)
     ) do
 
       def call
         rotate_api_key
+        audit_success
+      rescue => e
+        audit_failure(e)
+        raise e
       end
 
       private
@@ -27,7 +33,28 @@ module Commands
       end
 
       def credentials
-        @role.credentials
+        @role_to_rotate.credentials
+      end
+
+      def audit_success
+        @audit_log.log(
+          ::Audit::Event::ApiKey.new(
+            authenticated_role_id: @authenticated_role.id,
+            rotated_role_id: @role_to_rotate.id,
+            success: true
+          )
+        )
+      end
+
+      def audit_failure(err)
+        @audit_log.log(
+          ::Audit::Event::ApiKey.new(
+            authenticated_role_id: @authenticated_role.id,
+            rotated_role_id: @role_to_rotate.id,
+            success: false,
+            error_message: err.message
+          )
+        )
       end
     end
   end
--- a/app/models/audit/event/api_key.rb
+++ b/app/models/audit/event/api_key.rb
@@ -1,0 +1,95 @@
+module Audit
+  module Event
+
+    # As a value object, it is okay that this class has :reek:TooManyInstanceVariables
+    class ApiKey
+      def initialize(authenticated_role_id:, rotated_role_id:, success:, error_message: nil)
+        @authenticated_role_id = authenticated_role_id
+        @rotated_role_id = rotated_role_id
+        @success = success
+        @error_message = error_message
+
+        # Implements `==` for audit events
+        @comparable_evt = ComparableEvent.new(self)
+      end
+
+      # :reek:UtilityFunction
+      def progname
+        Event.progname
+      end
+
+      def severity
+        attempted_action.severity
+      end
+
+      def to_s
+        message
+      end
+
+      def message
+        attempted_action.message(
+          success_msg: success_message,
+          failure_msg: failure_message,
+          error_msg: @error_message
+        )
+      end
+
+      def message_id
+        'api-key'
+      end
+
+      def structured_data
+        {
+          SDID::AUTH => { user: @authenticated_role_id },
+          SDID::SUBJECT => { role: @rotated_role_id }
+        }.merge(
+          attempted_action.action_sd
+        )
+      end
+
+      def facility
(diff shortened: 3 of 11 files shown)


CHANGELOG.md
```diff
@@ -6,17 +6,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- Password changes (`PUT /authn/:account/password`) now produce audit events with
+  message ID `password` ([cyberark/conjur#1548](https://github.com/cyberark/conjur/issues/1548))
```
jonahx, line 11: What about POSTing or any other HTTPVERBing?  Did any of those already have events, or do we only support PUT?

CHANGELOG.md
```diff
@@ -6,17 +6,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- API key rotations (`PUT /:authenticator/:account/api_key`) now produce audit events with
+  message ID `api-key` ([cyberark/conjur#1549](https://github.com/cyberark/conjur/issues/1549))
```
jonahx, line 13: Possibly incorrect nit, but... Is this a rotation?   In my head I think of rotations only as actions performed by a rotator.

app/domain/commands/credentials/rotate_api_key.rb
```diff
@@ -0,0 +1,61 @@
...
+require 'command_class'
+
+module Commands
+  module Credentials
+
+    # In might seem unnecessary to move this behavior into a command class.
+    # However, we factor this out now to prepare for A) adding audit decorations
+    # to this operation, and B) those audit events including the requestor's
+    # client IP address, which will be an input to this CommandClass, rather than
+    # to the Credential model's method.
```
jonahx, line 12: 👍

app/domain/commands/credentials/rotate_api_key.rb
```diff
@@ -0,0 +1,61 @@
...
+    RotateApiKey ||= CommandClass.new(
+      dependencies: {
+        audit_log: ::Audit.logger
+      },
+      inputs: %i(role_to_rotate authenticated_role)
```
jonahx, line 17: Good names 👍

app/models/audit/event/api_key.rb
```diff
@@ -0,0 +1,91 @@
+module Audit
+  module Event
+    class ApiKey
+
+      def initialize(auth_role:, subject_role:, success:, error_message: nil)
```
jonahx, line 5: The names from the other class were more immediately clear to me: `role_to_rotate authenticated_role`

app/models/audit/event/api_key.rb
```diff
@@ -0,0 +1,91 @@
...
+          "#{@auth_role.id} failed to rotate their API key"
+        else
+          "#{@auth_role.id} failed to rotate the api key for #{@subject_role.id}"
+        end
+      end
+      
+      # True if the role requesting the rotation is the same role
+      # the rotation is for. In other words, when you rotate your own
+      # API key
+      def own_key?
```
jonahx, line 79: It's verbose, but I think I'd prefer something like `user_rotating_their_own_key?` so that it's 100%  self-explanatory when reading at the call site.

app/models/audit/event/api_key.rb
```diff
@@ -0,0 +1,91 @@
+module Audit
+  module Event
+    class ApiKey
+
+      def initialize(auth_role:, subject_role:, success:, error_message: nil)
+        @auth_role = auth_role
+        @subject_role = subject_role
```
jonahx, line 7: This should take the `role_id` instead of the role, since we only ever use the `role_id`.  As a side benefit, we'll make reek happy.
happyreek

CHANGELOG.md
micahlee: It's only `PUT`: https://github.com/cyberark/conjur/blob/db6126023a664f4b498c18e2686a3ca7ef00c650/config/routes.rb#L42

CHANGELOG.md
micahlee: Yeah, this isn't performed by rotators, probably primarily because those are only implemented for `Secret`'s and this is a different concept in our data model (`Credentials`). 

"Rotation" is how we refer to this though: https://docs.conjur.org/Latest/en/Content/Developer/Conjur_API_Rotate_Personal_API_Key.htm

app/models/audit/event/api_key.rb
micahlee: So you think we should use those here as well? 

I was trying to adapt the names to represent the domain of audit, and how they'll manifest in the structured data, but I don't think I have any issue with these names instead.

jonahx: Ah yeah, I still do prefer using the other names too.

I agree with your point about adapting names for their domain, but I'd see the boundary starting only once you call `structured_data`.  That is, the structured_data method converts the human friendly names we created the event with into the jargon-y audit terms.

micahlee: That makes sense, thanks!

app/models/audit/event/api_key.rb
```diff
@@ -0,0 +1,97 @@
...
+      def success_message
+        if user_rotating_their_own_key?
+          "#{@auth_role_id} successfully rotated their API key"
+        else
+          "#{@auth_role_id} successfully rotated the api key for #{@subject_role_id}"
+        end
+      end
+
+      # It's clearer to simply call the #id attribute multiple times, rather
+      # than factor it out, even though it reeks of :reek:DuplicateMethodCall
```
jonahx, line 73: You can now remove this note :)

spec/app/domain/commands/credentials/rotate_api_key_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    Commands::Credentials::RotateApiKey.new(
+      audit_log: audit_log
+    )
+  end
+
+  context 'when rotating own API key' do
+    it 'updates the key' do
+      # Expect it to rotate the api key on the credentials model, and to save it
+      expect(credentials).to receive(:rotate_api_key)
+      expect(credentials).to receive(:save)
```
jonahx, line 43: Per the very first part of our convo....

Since normally it's incorrect to use an input _mock_ (those would typically be doubles, and only the deps would be mocks with receive assertions), let's add a comment here that we're only breaking the rule because the correct fix (refactoring the entire credentials controller) was out of scope.

spec/app/domain/commands/credentials/rotate_api_key_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+      # Expect it to log a successful audit message
+      expect(audit_log).to receive(:log).with(audit_success)
```
jonahx, line 46: 💥

spec/app/domain/commands/credentials/rotate_api_key_spec.rb
micahlee: Ah, that's right. Will do.

spec/app/domain/commands/credentials/rotate_api_key_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+      # Call the command
+      subject.call(role_to_rotate: role, authenticated_role: role)
+    end
+
+    it 'bubbles up exceptions' do 
+      # Assume the database update fails. This could be caused by an
+      # invalid password, database issues, etc.
+      allow(credentials).to receive(:rotate_api_key)
+      allow(credentials).to receive(:save).and_raise(err_message)
```
jonahx, line 56: Refer to suggested comment above here too.

spec/app/domain/commands/credentials/rotate_api_key_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      expect(audit_log).to receive(:log).with(audit_failure)
+
+      # Expect the command to raise the original exception
+      expect do
+        subject.call(role_to_rotate: role, authenticated_role: role)
+      end.to raise_error(err_message)
+    end
+  end
+
+  context 'when rotating another\'s API key' do
```
jonahx, line 68: Super nit but... 
```ruby
context "when rotating another's API key" do
```

spec/app/domain/commands/credentials/rotate_api_key_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      # Expect it to log a failed audit message
+      expect(audit_log).to receive(:log).with(audit_failure)
+
+      # Expect the command to raise the original exception
+      expect do
+        subject.call(role_to_rotate: other_role, authenticated_role: role)
+      end.to raise_error(err_message)
+    end
+  end
+end
```
jonahx, line 98: These read very clean now.

spec/models/audit/event/api_key_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      let(:subject_role) { other_role }
+
+      it 'renders to string correctly' do
+        expect(subject.to_s).to eq(
+          'rspec:user:my_user failed to rotate the api key for rspec:user:other_user: failed rotation'
+        )
+      end
+    end
+  end
+end
```
jonahx, line 74: Also nice.  Value objects are pleasure to test....

PR #1620, opened 2020-06-15
Fix the FIPS change log entry

nahumcohen:
### What does this PR do?
Fix the FIPS change log entry


orenbm:
@nahumcohen @uCatu can you please add a link to the issue for each entry like we do in the rest of the changelog?

nahumcohen:
> @nahumcohen @uCatu can you please add a link to the issue for each entry like we do in the rest of the changelog?

Done on top

PR #1623, opened 2020-06-15
Remove request parameters

micahlee:

#1625 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/cucumber/api/features/authenticate.feature
+++ b/cucumber/api/features/authenticate.feature
@@ -20,6 +20,11 @@ Feature: Exchange a role's API key for a signed authentication token
       cucumber:user:alice successfully authenticated with authenticator authn
     """
 
+  Scenario: Authenticating with no Content-Type header succeeds without writing API key to the logs
+    And I save my place in the log file
+    And I can authenticate Alice with no Content-Type header
+    Then Alice's API key does not appear in the log
+
   Scenario: Attempting to use an invalid API key to authenticate result in 401 error
     When I POST "/authn/cucumber/alice/authenticate" with plain text body "wrong-api-key"
     Then the HTTP response status code is 401
--- a/cucumber/api/features/step_definitions/logs_steps.rb
+++ b/cucumber/api/features/step_definitions/logs_steps.rb
@@ -12,6 +12,12 @@ And(/^The following appears in the (audit )?log after my savepoint:$/) do |_is_a
   expect(num_matches_since_savepoint(message)).to be >= 1
 end
 
+And(/Alice's API key does not appear in the log/) do
+  expect(
+    num_matches_since_savepoint(api_key_for_role_id("cucumber:user:alice"))
+  ).to be(0)
+end
+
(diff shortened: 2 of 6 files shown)


cucumber/api/features/authenticate.feature
```diff
@@ -20,6 +20,11 @@ Feature: Exchange a role's API key for a signed authentication token
       cucumber:user:alice successfully authenticated with authenticator authn
     """
 
+  Scenario: Authenticating with no Content-Type header succeeds without leakage.
```
jtuttle, line 23: Minor but we don't usually punctuate single sentence descripts (i.e. no period). Also could we say what exactly isn't being leaked and where it's not getting leaked to? So something like "succeeds without writing API key to the logs".

cucumber/api/features/authenticate.feature
```diff
@@ -20,6 +20,11 @@ Feature: Exchange a role's API key for a signed authentication token
...
+    And I save my place in the log file
+    And I can authenticate Alice with no Content-Type header
+    Then The Alice's API key does not appear in the log
```
jtuttle, line 26: There seems to be an extra `The` in this line...is this working? I would have thought that would cause it to not match the step def.

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -289,6 +289,10 @@ def random_hex nbytes = 12
 
   protected
 
+  def api_key_for(role)
```
jtuttle, line 292: Can we call this `api_key_for_role` so that it's clear what the param is from the place where it's being called? Maybe even `api_key_for_role_name` since it's supposed to be a string and not a `Role` object.

cucumber/api/features/authenticate.feature
jonahx: Done.  Unfortunately Gherkin has no way to break up long lines, but we clarity is more important than line overflow.

cucumber/api/features/authenticate.feature
jonahx: 😆 will fix.

It works because my regex isn't anchoring to start `^` and and end `$`:
```
And(/Alice's API key does not appear in the log/)
```

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -289,6 +289,10 @@ def random_hex nbytes = 12
 
   protected
 
+  def api_key_for_role_id(role)
```
jtuttle, line 292: Parameter should now be `role_id`

cucumber/api/features/step_definitions/logs_steps.rb
```diff
@@ -12,6 +12,12 @@
   expect(num_matches_since_savepoint(message)).to be >= 1
 end
 
+And(/Alice's API key does not appear in the log/) do
+  expect(
+    num_matches_since_savepoint(api_key_for_role_id("cucumber:user:alice"))
+  ).to be 0
```
jtuttle, line 18: We are leaning towards preferring parens around method calls, right? So this should be `be(0)`.

Issue #1621, opened 2020-06-15
Policy with special characters fails to load


jtuttle:
## Summary
Policy with special characters fails to load.

## Steps to Reproduce (needs to be tested)
1. Start Conjur + CLI with dev scripts
2. load the following policy
```
- &active-variables
  - !variable
    id: AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name/password
    annotations:
      cyberark-vault: 'true'
      cyberark-vault/accounts: AutomationVault/safeToSync/AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name
  - !variable
    id: AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name/username
    annotations:
      cyberark-vault: 'true'
      cyberark-vault/accounts: AutomationVault/safeToSync/AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name
  - !variable
    id: AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name/address
    annotations:
      cyberark-vault: 'true'
      cyberark-vault/accounts: AutomationVault/safeToSync/AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name
```

## Expected Results
200 response

## Actual Results (including error logs, if applicable)
422 response

PR #1629, opened 2020-06-16
Fix Policy Load Bugs

micahlee:
### What does this PR do?

This PR fixes two issues in policy loading:

* Not setting the default content type for policy (`yaml`) before other Rack middleware that uses it.
* A logic bug in the default content type middleware for a missing `Content-Type` header.

Connected to #1622
Connected to #1621

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/config/initializers/default_content_type.rb
+++ b/config/initializers/default_content_type.rb
@@ -1,5 +1,0 @@
-require 'rack/default_content_type'
-
-Rails.application.configure do
-  config.middleware.use ::Rack::DefaultContentType
-end
--- a/config/initializers/rack_middleware.rb
+++ b/config/initializers/rack_middleware.rb
@@ -1,0 +1,30 @@
+require 'rack/default_content_type'
+
+# This is where we introduce custom middleware that interacts with Rack
+# and Rails to change how requests are handled.
+Rails.application.configure do
+
+  # This configures which paths do and do not require token authentication.
+  # Token authentication is optional for authn routes, and it's not applied at
+  # all to authentication, host factories, or static assets (e.g. images, CSS)
+  config.middleware.use(Conjur::Rack::Authenticator,
+  optional: [
+    /^\/authn-[^\/]+\//,
+    /^\/authn\//,
+    /^\/public_keys\//
+  ],
+  except: [
+    /^\/authn-[^\/]+\/.*\/authenticate$/,
+    /^\/authn\/.*\/authenticate$/,
+    /^\/host_factories\/hosts$/,
+    /^\/assets\/.*/,
+    /^\/authenticators$/,
+    /^\/$/
+  ])
+
+  # We want to ensure requests have an expected content type
+  # before other middleware runs to make sure any body parsing
+  # attempts are handled correctly. So we add this middleware
+  # to the start of the Rack middleware chain.
+  config.middleware.insert_before(0, ::Rack::DefaultContentType)
+end
--- a/cucumber/api/features/step_definitions/policy_load_steps.rb
+++ b/cucumber/api/features/step_definitions/policy_load_steps.rb
@@ -1,0 +1,18 @@
+When(/^I load a large policy with POST$/) do
+  # Generate a large policy with 1000 variables, sampled from
+  # an example PAS synchronizer policy
+  policy_body = "- &my-variables\n" + [*1..1000].map do |i|
+    <<-POLICY
+  - !variable
+    id: epv_safe_#{i}/password
+    annotations:
+      cyberark-vault: 'true'
+    POLICY
+  end.join("\n")
+
+  path = '/policies/cucumber/policy/dev/db'
+
+  try_request true do
+    post_json path, policy_body
+  end
+end
--- a/cucumber/api/features/step_definitions/policy_load_steps.rb
+++ b/cucumber/api/features/step_definitions/policy_load_steps.rb
@@ -1,3 +1,7 @@
+# Let's you reference the global variables such as `$?` using less 
+# cryptic names like `$CHILD_STATUS`
+require 'English'
+
 When(/^I load a large policy with POST$/) do
   # Generate a large policy with 1000 variables, sampled from
   # an example PAS synchronizer policy
@@ -15,4 +19,40 @@ When(/^I load a large policy with POST$/) do
   try_request true do
     post_json path, policy_body
   end
+end
+
+When('I use curl to load a policy with special characters and no content type') do
+  # Policy with special characters, sampled from a PAS synchronizer test policy
+  policy_body = <<~POLICY
+    - !variable
+      id: AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name/password
+  POLICY
+
+  auth_token = current_user_credentials.dig(:headers, :authorization)
+
+  url_path = '/policies/cucumber/policy/dev/db'
+
+  # `--no-buffer` allows to close the pipe in popen without getting a curl error
+  # `--output ...` prevents curl from trying to write to stdout, causing an io error
+  command = <<~COMMAND
+    curl --silent \
+      --no-buffer \
+      --fail \
+      --output /dev/null \
+      --header 'Content-Type:' \
+      --header 'Authorization: #{auth_token}' \
+      --data-binary '@-' \
+      #{full_conjur_url(url_path)}
+  COMMAND
+
+  IO.popen(command, "r+") do |pipe|
+    pipe.puts policy_body
+    pipe.close_write
+  end
+
+  @command_result = $CHILD_STATUS.success?
+end
+
+Then('the command is successful') do
+  expect(@command_result).to be(true)
 end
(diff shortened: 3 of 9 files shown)


config/initializers/default_content_type.rb
```diff
@@ -1,5 +1,9 @@
 require 'rack/default_content_type'
 
 Rails.application.configure do
-  config.middleware.use ::Rack::DefaultContentType
+  # We want to ensure requests have an expected content type
+  # before other middleware runs to make sure any body parsing
+  # attempts are handled correctly. So we add this middleware
+  # to the start of the Rack middleware chain.
+  config.middleware.insert_before(0, ::Rack::DefaultContentType)
```
jonahx, line 8: Here's one I consider important:

I'd like to see _everything_ that messes with the middleware chain in a single initializer.  This is a case where the benefit of "having a central view of all the magic/danger" far outweighs the cleanliness of separate files.

One place, clearly marked, that says:

"This is where all the nonstandard Rails stuff you might not expect is hapenning.  Read this to have a clear view of all the magic we've introduced for Conjur"

cucumber/api/features/step_definitions/policy_load_steps.rb
```diff
@@ -0,0 +1,69 @@
+require 'English'
```
jonahx, line 1: Can we add a comment above this require:

```
# Let's you reference the global variables such as `$_` using less 
# cryptic names like `$LAST_LINE_READ`
```

cucumber/api/features/step_definitions/policy_load_steps.rb
```diff
@@ -0,0 +1,69 @@
...
+
+When(/^I load a large policy with POST$/) do
+  # Generate a large policy with 1000 variables
+  policy_body = "- &my-variables\n" + [*1..1000].map do |ndx|
+    %Q(
```
jonahx, line 6: Nit: Just use the standard `i` instead of `ndx`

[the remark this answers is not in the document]
jonahx: CC might like a heredoc better in this case.

cucumber/api/features/step_definitions/policy_load_steps.rb
```diff
@@ -0,0 +1,69 @@
...
+      - !variable
+        id: obj_Safe_00c_0_#{ndx}/password
+        annotations:
+          cyberark-vault: 'true'
+          cyberark-vault/accounts: AutomationVault/Safe_00c/obj_Safe_00c_0_10
+    )
```
jonahx, line 12: Where does this name come from: `Safe_00c/obj_Safe_00c_0_10`?

cucumber/api/features/step_definitions/policy_load_steps.rb
```diff
@@ -0,0 +1,69 @@
...
+          cyberark-vault: 'true'
+          cyberark-vault/accounts: AutomationVault/safeToSync/AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name
+      - !variable
+        id: AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name/address
+        annotations:
+          cyberark-vault: 'true'
+          cyberark-vault/accounts: AutomationVault/safeToSync/AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name
+  )
+
+  authorization = current_user_credentials.dig(:headers, :authorization)
```
jonahx, line 42: I would name this variable `auth_token` or `api_key` (whichever it is)

cucumber/api/features/step_definitions/policy_load_steps.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  url_path = '/policies/cucumber/policy/dev/db'
+  url = URI.parse(Conjur.configuration.appliance_url + url_path)
```
jonahx, line 45: This feels like something generally useful: I'd add it into rest_helpers (or anywhere you think is more appropriate) so that here you can just write something like:
```ruby
full_url(url_path)
```

cucumber/api/features/step_definitions/policy_load_steps.rb
```diff
@@ -0,0 +1,69 @@
...
+      --header 'Content-Type:' \
+      --header 'Authorization: #{authorization}' \
+      --data-binary '@-' \
+      #{url}
+  )
+
+  IO.popen(command, "r+") do |pipe|
+    pipe.puts policy_body
+    pipe.close_write
+  end
```
jonahx, line 62: Clever solution.

cucumber/api/features/step_definitions/policy_load_steps.rb
micahlee: This comes from the synchronizer test policy. It's the name of a vault and safe in Enterprise Password Vault (EPV). It doesn't have a significant meaning here, other than is representative of the failing policy document.

config/initializers/default_content_type.rb
micahlee: To be clear, you want to change this in this PR, correct? The file itself already exists, this PR is currently only changing where the middleware is inserted.

cucumber/api/features/step_definitions/policy_load_steps.rb
```diff
@@ -0,0 +1,69 @@
...
+        id: AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name/password
+        annotations:
+          cyberark-vault: 'true'
+          cyberark-vault/accounts: AutomationVault/safeToSync/AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name
+      - !variable
+        id: AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name/username
+        annotations:
+          cyberark-vault: 'true'
+          cyberark-vault/accounts: AutomationVault/safeToSync/AccountWithSpecialCharacters~!@#$%^&(){}[]-+=,Name
+      - !variable
```
jtuttle, line 39: These other two variables aren't really doing anything differently, right? They all use the same set of special characters. I think we only need the one use case here unless I'm missing something.

lib/rack/default_content_type.rb
```diff
@@ -60,8 +60,9 @@ def call(env)
 
     # :reek:UtilityFunction
     # :reek:NilCheck
+    # :reek:DuplicateMethodCall
     def content_type_missing?(env)
-      env['CONTENT_TYPE']&.empty?
+      env['CONTENT_TYPE'].nil? || env['CONTENT_TYPE'].empty?
```
jtuttle, line 65: I believe you could just make this `env['CONTENT_TYPE'].blank?` or you could use `!env['CONTENT_TYPE'].present?`. `nil` should respond to both of those without a `NoMethodError`.

micahlee: @jonahx has commented on that before, that we shouldn't use the Rails helpers in Rack middleware. Otherwise `blank?` would be the correct solution.

config/initializers/default_content_type.rb
micahlee: @jonahx @jtuttle , this is done. Double check that it is what you had in mind. Thanks!

cucumber/api/features/step_definitions/policy_load_steps.rb
micahlee: Yes, you're right. I was able to simplify both sets of policies to a more simplified version that still caused the failure.

config/initializers/rack_middleware.rb
```diff
@@ -0,0 +1,30 @@
...
+    /^\/public_keys\//
+  ],
+  except: [
+    /^\/authn-[^\/]+\/.*\/authenticate$/,
+    /^\/authn\/.*\/authenticate$/,
+    /^\/host_factories\/hosts$/,
+    /^\/assets\/.*/,
+    /^\/authenticators$/,
+    /^\/$/
+  ]
```
jonahx, line 23: Can we get some parens around this function call?

jonahx: I am concerned that we don't know exactly what this is doing.  It appears to be existing code that we're just moving so doesn't have to be a blocker here, but something so central, with serious security implications, should be well-understand by everyone on the team, and we should have an explanatory comment here or at least a link to the docs

lib/rack/default_content_type.rb
jonahx: These reek UtilityFunction callouts look valid to me.  

There are two options, I'll start with the one I'd recommend.

1.  Remove the methods and inline their logic in `call` -- they are both one-liners anyway.  Assign the result of what the calculate to a local var corresponding to the current method names. 

2. Create an object which is initialized with the `env`, and then the references to `@env` will be references to an instance var. 

This is a good equivalence to keep in mind generally.  And generally, the decision between 1. and 2. comes down to... "If complex enough (more than a few methods) go with 2.  Otherwise 2. is overkill so go with 1."

PR #1632, opened 2020-06-17
Update to allow releasing from a branch

ismarc:
### What does this PR do?
These changes make debian packages built from a branch available to
any project consuming 'stable' components, removing the need to
have matching branch names in all downstream projects.  It also
allows the `release` script to be run from a branch and have it
work appropriately with that branch rather than just `master`

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/push-image.sh
+++ b/push-image.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 # Push the 'conjur' image to various Docker registries
-# Push stable images on master branch
+# Push tagged images on master branch
 # Release images can be created by passing the desired tag to this script
 # Ex: ./push-image 4.9.5.1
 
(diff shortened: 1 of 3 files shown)


push-image.sh
```diff
@@ -25,23 +25,20 @@ function main() {
...
 
-  # if on master (as determined by examining Jenkins-set envar BRANCH_NAME)
-  # assume tests have passed and push to latest and stable tags, too
-  if [ "$BRANCH_NAME" = "master" ]; then
+  git fetch --tags || : # Jenkins brokenness workaround
+  local git_description=`git describe`
+
+  # if on a tag matching the VERSION, assume tests have passed and push to latest and stable tags
+  # and push releases to DockerHub
+  if [[ $git_description = v$VERSION ]]; then
```
izgeri, line 33: If we're going to be changing this logic (which looks like it's still only publishing images on tags), can we just update this project to run this script only on tag pipelines?

ismarc: Absolutely, I initially was going to gut out all of the `$git_description` and similar tag checks and only run it for tag builds, but decided against it due to the time pressures of this release and the potential risk with a larger change.  I'm more than happy to to make those changes if folks think it's not too high of a risk.

izgeri: ah, I see. for now what you've implemented (if I'm reading it correctly) will publish new images on a tag, even if that tag is not on master. is that correct?

fwiw I'm ok with moving forward without the tag triggers if that's too risky for this release, though from what I've seen on other projects the change to implement this is small.

ismarc: > will publish new images on a tag, even if that tag is not on master
It's slightly more restrictive than that.  It will publish new images if it's on a tag (independent of branch) and that tag matches what's listed in the `VERSION` file.  This should be safe for arbitrary tags to exist without publishing it, and only publishing the specific tag + `VERSION` file matches.

Issue #1633, opened 2020-06-18
Enhance logs around client certificate installation for k8s


orenbm:
I helped with a case lately where it took us time to understand the real-time flow of the client certificate installation. After we understood the issue we analyzed the flow and tried to see which logs would have helped us understand the issue quicker. 

We decided that we will log:
- Conjur server
  - The place where the cert is installed: namespace, pod name, container name, path
  - success message after installation
- authenticator client
  - success message for reading the certificate
  - message about deleting this cert from file

DoD:
- [ ] All messages above are logged

PR #1634, opened 2020-06-18
Enhance logs around client cert installment

orenbm:
Connected to #1633

The client certificate installment process can be tricky
and it can help customers to know more details about it.
We not log the container name and the path inside it to
which the certificate will be copied.

We also log a success message to indicate that to the user.

### DIFF
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -129,8 +129,14 @@ unless defined? LogMessages::Authentication::OriginValidated
         )
 
         CopySSLToPod = ::Util::TrackableLogMessageClass.new(
-          msg:  "Copying SSL certificate to {0-pod-namespace}/{1-pod-name}",
+          msg:  "Copying SSL certificate to {0-container-name}:{1-cert-file-path} " \
+            "in {2-pod-namespace}/{3-pod-name}",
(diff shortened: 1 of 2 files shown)


orenbm:
@shulifink can you review the logs please?

app/domain/logs.rb
```diff
@@ -129,10 +129,16 @@ module AuthnK8s
         )
 
         CopySSLToPod = ::Util::TrackableLogMessageClass.new(
-          msg:  "Copying SSL certificate to {0-pod-namespace}/{1-pod-name}",
+          msg:  "Copying SSL certificate to {0-container-name}:{1-cert-file-path} " \
```
shulifink, line 132: approved

app/domain/logs.rb
```diff
@@ -129,10 +129,16 @@ module AuthnK8s
...
+            "in {2-pod-namespace}/{3-pod-name}",
           code: "CONJ00015D"
         )
 
+        CopySSLToPodSuccess = ::Util::TrackableLogMessageClass.new(
+          msg:  "Copied SSL certificate successfully",
```
shulifink, line 138: approved

PR #1635, opened 2020-06-18
Proxy Support Design

jvanderhoof:
This PR includes a design for passing the client IP address through trusted Load Balancers.

### Review Period
Ideally a decision in the next week. There is a team ready to begin work once the design has been approved.

### Mandatory Reviewers
|Name|Focus|
|-|-|
|@rafis3|Overall approach and usability|
|@shaharglazner / @andytinkham|Security concerns|
 
### Appreciated Reviewers
|Name|Focus|
|-|-|
|@InbalZilberman|Overall approach and usability|
|@micahlee|Overall approach and usability|
|@alexkalish|Overall approach and usability|
|@Tovli|Overall approach and usability|
|@nessiLahav|Overall approach and usability|


InbalZilberman:
Thanks for sharing!
I feel this approach place too much burden on the end user who is most of the time is not aware of any load balancer and actually should not be aware of it.
I fail to understand the exact technical problem we come to solve here. Does it happen in the master or in the follower or both?

Tovli:
@InbalZilberman - Would it be acceptable if the CIDR restrictions property was used instead of the suggested policy?
So instead of the current approach where the customer is entering the client ip/range into the CIDR. In the new approach (s)he will need to enter all the IPs in the request chain. Thus, restrict authn just to these specific IPs (LBs and client)

So the UX stays the same for hosts creation:
`- !host
  id: www-01
  owner: !group ops-admin
  annotations:
    description: Hypertext web server
  restricted_to: [172.17.0.3, 10.0.0.0/24]`

design/diagrams/src/proxy-support-overview.puml
```diff
@@ -0,0 +1,12 @@
+@startuml
+
+participant "client\n(10.0.0.1)" as client
+participant "Layer 7 Load Balancer 1\n(10.1.0.1)" as lb_1
+participant "Layer 7 Load Balancer 2\n(10.2.0.1)" as lb_2
+participant "Conjur" as conjur
+
+client --> lb_1 : X-Forwarded-For: null
+lb_1 --> lb_2 : X-Forwarded-For: "10.0.0.1, 10.1.0.1"
```
Tovli, line 9: AFAIK LB is not adding its own IP to the header as you already have it as part of the tcp protocol

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
...
+
+The following visualization offers a typical experience as a requests moves through a Layer 7 load balancers.
+
+![](/design/diagrams/out/proxy-support-overview/proxy-support-overview.png)
+
+As we can see, each load balancer adds its IP address to the `X-Forwarded-For` header.
+
+With our current implementation:
+```ruby
+# Rack::Request
```
Tovli, line 28: Can you express how this example will change given your proposal?

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
...
+
+we can apply the following policy:
+```yaml
+- !policy
+  id: conjur/settings/proxies
+  body:
+    - !host 
+      id: dap-master.mycompany.com
+      annotations:
+        proxy/ip-addresses: ['1.2.3.4']
```
sashaCher, line 57: Will it support CIDR notation?
I suppose that it's a possible configuration to have a pool of load balancers.

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+With the above proxied request, Conjur will:
+1. Check if the `X-Forwarded-For` header has any values. In this example it will have the following string: `10.0.0.1, 1.2.3.4`. 
+2. Perform a Regex match using the host ID in the `conjur/settings/proxies` namespace against the request hostname. In this example, we'd match on `dap-master.mycompany.com`.
+3. Grab the array of IP addresses stored in the `proxy/ip-addresses` annotation (`['1.2.3.4']`).
+4. Starting at the **end** of the list of `X-Forwarded-For` entries, return the first IP address that is not present in the `proxy/ip-addresses` annotation array.
+   1. Skip `1.2.3.4`
+   2. Return `10.0.0.1`
+
```
nessiLahav, line 67: So if i understand it correctly, all the hosts which defined in conjur/settings/proxies policy are permitted proxies addresses? If yes:

1. Can we configure IP instead of hostname ?
2. In step 2, we are fetching the hostname(or the IP) from the TCP packet ?
3. In step 2, if there is no match, we will return a new error of unpermitted proxy?
4. In step 1, If  `X-Forwarded-For` value is empty the behavior will stay as today ?

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
...
+```yaml
+- !policy
+  id: conjur/settings/proxies
+  body:
+    - !host
+      id: '*' # matches any request hostname
+      annotations:
+        proxy/ip-addresses: ['1.2.3.4', '10.0.0.1']
+```
+
```
nessiLahav, line 80: I think we should prevent this configuration, because if we are not validating the IP/Hostname of the TCP packet and only relying on the ` X-Forwarded-For` header, it will be easy to bypass the CIDR restriction by simply sending whatever you want in this header

design/diagrams/src/proxy-support-overview.puml
```diff
@@ -0,0 +1,12 @@
...
+lb_2 --> conjur : X-Forwarded-For: "10.0.0.1, 10.1.0.1, 10.2.0.1"
+
```
nessiLahav, line 11: @shaharglazner Do we prevent MITM ? lets assume we have a malicious proxy in the chain which delete himself from the` X-Forwarded-For `header, this proxy will pass CIDR restrictions and get all the secrets before returning them to the original client

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
...
+## Current Request Flow
+
+The following visualization offers a typical experience as a requests moves through a Layer 7 load balancers.
+
+![](/design/diagrams/out/proxy-support-overview/proxy-support-overview.png)
+
+As we can see, each load balancer adds its IP address to the `X-Forwarded-For` header.
+
+With our current implementation:
```
nessiLahav, line 26: Where is code located ? (i cant find it)

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
...
+- !policy
+  id: conjur/settings/proxies
+  body:
+    - !host 
+      id: dap-master.mycompany.com
+      annotations:
+        proxy/ip-addresses: ['1.2.3.4']
+```
+
+With the above proxied request, Conjur will:
```
nessiLahav, line 60: Where this logic will be implemented ? under validate origin flow ?

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
...
+1. Check if the `X-Forwarded-For` header has any values. In this example it will have the following string: `10.0.0.1, 1.2.3.4`. 
+2. Perform a Regex match using the host ID in the `conjur/settings/proxies` namespace against the request hostname. In this example, we'd match on `dap-master.mycompany.com`.
+3. Grab the array of IP addresses stored in the `proxy/ip-addresses` annotation (`['1.2.3.4']`).
```
nessiLahav, line 63: WDYT to use `restricted_to` instead of annotation ?

design/proposals/proxy-support/proxy-support.md
jvanderhoof: https://github.com/cyberark/conjur/blob/0a724b7f3c0af0ae681a0261ad87d4aaa6b1fab0/app/controllers/authenticate_controller.rb#L55

design/proposals/proxy-support/proxy-support.md
jvanderhoof: https://github.com/cyberark/conjur/blob/0a724b7f3c0af0ae681a0261ad87d4aaa6b1fab0/app/controllers/authenticate_controller.rb#L55

jvanderhoof: `Rack::Request` is where the method `ip` is located.

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
+# Maintaining Requester IP Addresses <!-- omit in toc --> 
+
+When Conjur (or DAP) are run behind a load balancer, users are unable to use CIDR restrictions on Hosts or Layers. Furthermore, Load Balancer IP address appears in Audit logs instead of the true client IP address. This is because Conjur references the last entry in the `X-Forwarded-For` header.
```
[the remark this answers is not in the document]
jvanderhoof: CIDR restrictions are supported for `users` and `groups`.  This functionality does not currently work when a user is accessing an instance running behind a load balancer 😄 .

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
...
+Given the following setup
+![](/design/diagrams/out/proxy-master-node/proxy-master-node.png)
+
+we can apply the following policy:
+```yaml
+- !policy
+  id: conjur/settings/proxies
+  body:
+    - !host 
+      id: dap-master.mycompany.com
```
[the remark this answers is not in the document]
jvanderhoof: I had thought about using `webservices` to accomplish the same thing.  After digging into auto-failover (which uses Hosts, I changed it up.

Thinking forward, by giving each DAP node a host (Master, Standbys, and Followers), we make it easy to deploy nodes using the seed service (which requires a valid host). 

What about adding proxies as a Group or a Layer?

design/proposals/proxy-support/proxy-support.md
jvanderhoof: Yes. I didn't include it because I'm not sure how common it would be for all load balancers to be in the same network space.  I've always set them up in an application of environment subnet. In that case, a single IP would be fine.

design/proposals/proxy-support/proxy-support.md
jvanderhoof: I was thinking we rewrite the current `request.ip` method to strip "trusted proxies", then return the first "non-trusted" IP (which should be the requesting client).

design/proposals/proxy-support/proxy-support.md
jvanderhoof: How about `restricted_proxies`?  `restricted_to` is host-specific and denotes a limited pool of IPs the request can originate from.  In this case, we're looking to remove trusted proxies to allow the true requesting client identity to come through.

design/proposals/proxy-support/proxy-support.md
```diff
@@ -0,0 +1,157 @@
...
+      annotations:
+        proxy/ip-addresses: ['1.2.3.4']
+```
+
+With the above proxied request, Conjur will:
+1. Check if the `X-Forwarded-For` header has any values. In this example it will have the following string: `10.0.0.1, 1.2.3.4`. 
+2. Perform a Regex match using the host ID in the `conjur/settings/proxies` namespace against the request hostname. In this example, we'd match on `dap-master.mycompany.com`.
```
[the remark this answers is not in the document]
jvanderhoof: How about using the request URL?  Shouldn't that have the full request address (which was proxied through the load balancer)?

design/proposals/proxy-support/proxy-support.md
jvanderhoof: > Can we configure IP instead of hostname ?
Theoretically, yes 😄 

> In step 2, we are fetching the hostname(or the IP) from the TCP packet ?
I was thinking the originating request's URL, but ideally, the most secure approach is best.

> In step 2, if there is no match, we will return a new error of unpermitted proxy?
Yes.

> In step 1, If X-Forwarded-For value is empty the behavior will stay as today ?
Yes.

design/proposals/proxy-support/proxy-support.md
jvanderhoof: I was told by @rafis3 this approach is common in other Cyberark products. Proxy Protocol is another alternative, but both approaches require the proxy is trusted.

design/proposals/proxy-support/proxy-support.md
nessiLahav: @jvanderhoof  WDYT on the following alternative:
Instead of a configure the permitted proxies ips for each follower name,
We will define it generally by the proxy range with a list of ips permitted in the XFF header (restricted_proxies), and it will look like:

```
{
- !policy
  id: conjur/settings/proxies
  body:
    - !proxy
      restricted_to: [155.0.6.6\32]
      restricted_proxies: [1.0.0.6\32, 2.0.0.6\32]
}
```

In the above example only 1 proxy can access directly to the follower (155.0.6.6) and we will validate the XFF base on the restricted_proxies list

I also add an diagram that explains the flow:
image

Issue #972, opened 2019-04-18
Rails autoload is not re-loading error classes


orenbm:
We have an issue where due to Rails auto-loading mechanism, the errors in `authentication/errors.rb` are loaded more than once and thus aren't rescued properly in the code. we need to implement a proper solution for this issue.

This is related to the `warning: already initialized constant...` warning we get every time we run the server. 

more info can be found here:

- https://medium.com/invoca-engineering-blog/ruby-constants-with-rails-and-how-to-avoid-the-already-initialized-constant-x-warning-753d28f17691
- https://guides.rubyonrails.org/autoloading_and_reloading_constants.html

@micahlee created this simplified example to demonstrate what’s going on:

https://repl.it/repls/DrearySkyblueTransversal

orenbm:
Closed by #1645

PR #1645, opened 2020-06-22
Remove Err and Log definitions

orenbm:
Note: it is best to review this PR through the commits

#### Remove Err and Log definitions from top of classes

Each time we used an error class, or a log message class,
we would define its module at the top of the file. This was done
so when you call the class it would be concise without the whole path.

This approach led to issues where because we used memoization to init these
constants, they would sometimes mismatch what is written. For example,
if we had `Err ||= Errors::Authentication::AuthnAzure` in the top of the class
then sometimes the actual value of `Err` was `Errors::Authentication` because it was
already initialized with that value and the memoization meant that we don't initialize
it again.

A solution for that was to remove the memoization whenever this happened but then we had
another issue, where we got "already initialized constant" warnings. This happens because
now we initialize the same constant over and over again (e.g Err = Errors::Authentication::AuthnAzure).

Adding to the above the advantage of the explicit readability of calling error classes with their
full path led to the conclusion that we should remove the definition of `Err` and `Log` from the
top of classes, and call the full path each time.

#### Remove possible errors raised comments

We currently add a comment to each command class where we
specify the possible errors that are raised by this class.

This comment is hard-coded and it is the responsibility of the
developers to maintain it. The problem with this comment is that
it is not generated from anywhere and is not verified by code.

This leads us to a lot of cases where we move errors and forget to
update the comment so the comments are not accurate. Also, because
it is not verified then some classes have it and some classes do not.

It is better to not have any data at all than to have data that is incorrect,
thus we should remove these comments until they are forced by code of the
CommandClass interface.

Connected to #972 

#### Change log

#### Test coverage
- [x] The changes in this PR do not require additional tests

#### Documentation

### DIFF
--- a/app/domain/authentication/authenticate.rb
+++ b/app/domain/authentication/authenticate.rb
@@ -4,7 +4,6 @@ require 'command_class'
 
 module Authentication
 
-  Err ||= Errors::Authentication
   # Possible Errors Raised:
   # AuthenticatorNotFound, InvalidCredentials
 
@@ -45,11 +44,11 @@ module Authentication
     end
 
     def validate_authenticator_exists
-      raise Err::AuthenticatorNotFound, authenticator_name unless authenticator
+      raise Errors::Authentication::AuthenticatorNotFound, authenticator_name unless authenticator
     end
 
     def validate_credentials
-      raise Err::InvalidCredentials unless authenticator.valid?(@authenticator_input)
+      raise Errors::Authentication::InvalidCredentials unless authenticator.valid?(@authenticator_input)
     end
 
     def validate_webservice_is_whitelisted
--- a/app/domain/authentication/authenticate.rb
+++ b/app/domain/authentication/authenticate.rb
@@ -4,9 +4,6 @@ require 'command_class'
 
 module Authentication
 
-  # Possible Errors Raised:
-  # AuthenticatorNotFound, InvalidCredentials
-
   Authenticate ||= CommandClass.new(
     dependencies: {
       token_factory:                       TokenFactory.new,
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -4,7 +4,7 @@ module LogMessages
 
   module Conjur
 
-    PrimarySchema = Util::TrackableErrorClass.new(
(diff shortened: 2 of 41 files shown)


app/domain/authentication/authenticate.rb
```diff
@@ -45,11 +41,11 @@ def authenticator
     end
 
     def validate_authenticator_exists
-      raise Err::AuthenticatorNotFound, authenticator_name unless authenticator
+      raise Errors::Authentication::AuthenticatorNotFound, authenticator_name unless authenticator
```
jonahx, line 44: I like this change, but I don't like these long lines.  How about:

```ruby
      raise(
        Errors::Authentication::AuthenticatorNotFound,
        authenticator_name
      ) unless authenticator
```
whenever the long gets too long?

orenbm: sure, i'll do it.

app/domain/authentication/authenticate.rb
```diff
@@ -4,7 +4,6 @@
 
 module Authentication
 
-  Err ||= Errors::Authentication
```
sigalsax, line 7: Found the following as part of the design for authn-azure and azuthn-status

```
    Err = Errors::Authentication::AuthnAzure
```

```
    Err = Errors::Authentication::Authn<Type>
```

Should we update those too?

sigalsax: along with

```
 # Possible Errors Raised:
```

app/domain/logs.rb
```diff
@@ -4,7 +4,7 @@ module LogMessages
 
   module Conjur
 
-    PrimarySchema = Util::TrackableErrorClass.new(
+    PrimarySchema = ::Util::TrackableLogMessageClass.new(
```
sigalsax, line 7: why was :: added? From this commit, it doesn't seem that there was a change in directory

orenbm: i rooted the module so there is no confusion with other Util modules

app/domain/authentication/authenticate.rb
orenbm: Nah, i don't think we need to update old designs when we perform refactors

PR #1652, opened 2020-06-26
Trying to make hot reloading work in development

jonahx:
So far:

1. Reloading _sort of_ works again, at the expense of unititialized
   constant errors.
2. I found a hack to suppress those errors which should be ok.
3. But even with that there are still issues making this non-viable as
   of now.

To reproduce the issues:

1. Run:

       bundle exec cucumber -p api ./cucumber/api/features/login.feature:9

   It will pass.

2. Add "raise 'Blah'" above line 60 in "authenticate_controller.rb".
   Don't restart "conjurctl server".  Run the cuke command from step 1.
   again.  It will fail.  All as expected.

3. Remove the "raise 'Blah'".  Run it again without restarting.  Now the
   cuke will still fail, not as expected.  The error from the server
   logs is:

        DEBUG 2020/06/26 20:40:13 +0000 [pid=1147] [origin=127.0.0.1] [request_id=6664a988-22a3-4688-845e-02b7c5732d8a] [tid=1155] A copy of Audit::Log::SyslogAdapter has been removed from the module tree but is still active!
        /var/lib/ruby/lib/ruby/gems/2.5.0/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:495:in `load_missing_constant'
        /var/lib/ruby/lib/ruby/gems/2.5.0/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:195:in `const_missing'
        /src/conjur-server/app/models/audit/log/syslog_adapter.rb:30:in `log'
        /src/conjur-server/app/domain/authentication/login.rb:86:in `audit_failure'
        /src/conjur-server/app/domain/authentication/login.rb:33:in `rescue in call'
        /src/conjur-server/app/domain/authentication/login.rb:25:in `call'
        (eval):7:in `call'

### What does this PR do?

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/app/models/schemata.rb
+++ b/app/models/schemata.rb
@@ -19,17 +19,30 @@ class Schemata
   def initialize
     @search_path = Sequel::Model.db.search_path
 
-    # Verify that all schemata in the search path actually exist, except the automatic entry "$user"
+    # Verify that all schemata in the search path actually exist, except the
+    # automatic entry "$user"
     @search_path.each do |schema|
       next if schema == :"$user"
-      raise "Schema #{schema.inspect} from the search path is not listed current_schemas(false)" unless Sequel::Model.db.current_schemata.member?(schema)
+      unless Sequel::Model.db.current_schemata.member?(schema)
+        raise "Schema #{schema.inspect} from the search path is not listed " \
+          "current_schemas(false)"
+      end
     end
 
     # Verify that there is a primary schema in the DB
-    primary_schema = db.select(Sequel::function(:current_schema)).single_value
-    raise "No primary schema is available from search path #{search_path.inspect}" if primary_schema.nil?
+    primary_schema = db.select(
+      Sequel::function(:current_schema)
+    ).single_value
 
-    Rails.logger.info(LogMessages::Conjur::PrimarySchema.new(primary_schema.inspect))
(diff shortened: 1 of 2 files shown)


config/environments/development.rb
```diff
@@ -8,7 +8,27 @@
...
+    CONJUR_IGNORE_RE = Regexp.union(
+      /previous definition of [^ ]* was here/,
+      /already initialized constant/
+    )
+
+    def warn(msg)
+      return if msg =~ CONJUR_IGNORE_RE
+      super(msg)
+    end
+  end
```
orenbm, line 31: @jonahx what is the benefit of this change? don't we want to see these errors in DEV so we can verify they are not showing in PROD? 

I am worried that some warnings will slip to PROD without us noticing.

app/models/schemata.rb
```diff
@@ -19,17 +19,30 @@ class Schemata
   def initialize
```
orenbm, line 19: would love if this file change was in another commit stating that you just fixed indentations. took me a while to verify this is all that is going here.

jonahx: Yes ofc this is just a rough draft

PR #1653, opened 2020-06-29
Print Conjur server FIPS mode to Conjur log

aloncarmel111:
# What does this PR do?
We would like to add to the logs whether FIPS mode is enabled or disabled to have better supportability
connected to #1654 

In case FIPS mode is enabled - CONJ00038I Server set OpenSSL FIPS mode to true
In case FIPS mode is disabled - CONJ00038I Server set OpenSSL FIPS mode to false

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - All audit events now contain the IP address of the client that initiated the
   API request (e.g. `[client@43868 ip="172.24.0.5"]`)
   ([cyberark/conjur#1550](https://github.com/cyberark/conjur/issues/1550))
+- Print Conjur server FIPS mode status ([cyberark/conjur#1654](https://github.com/cyberark/conjur/issues/1654))
 
 ## [1.7.4] - 2020-06-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - All audit events now contain the IP address of the client that initiated the
   API request (e.g. `[client@43868 ip="172.24.0.5"]`)
   ([cyberark/conjur#1550](https://github.com/cyberark/conjur/issues/1550))
-- Print Conjur server FIPS mode status ([cyberark/conjur#1654](https://github.com/cyberark/conjur/issues/1654))
+- Print Conjur server FIPS mode status. [cyberark/conjur#1654](https://github.com/cyberark/conjur/issues/1654)
 
 ## [1.7.4] - 2020-06-17
 
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -10,7 +10,7 @@ module LogMessages
     )
 
     FipsModeStatus = ::Util::TrackableLogMessageClass.new(
-      msg:  "Server set OpenSSL FIPS mode to {0}",
+      msg:  "OpenSSL FIPS mode set to {0}",
       code: "CONJ00038I"
     )
 
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -10,7 +10,7 @@ module LogMessages
     )
 
     FipsModeStatus = ::Util::TrackableLogMessageClass.new(
-      msg:  "Conjur server set OpenSSL to run in {0} FIPS mode",
+      msg:  "Server set OpenSSL FIPS mode to {0}",
       code: "CONJ00038I"
     )
 
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -9,6 +9,11 @@ module LogMessages
       code: "CONJ00034I"
     )
 
+    FipsModeStatus = ::Util::TrackableLogMessageClass.new(
+      msg:  "Conjur server set OpenSSL to run in {0} FIPS mode",
+      code: "CONJ00038I"
+    )
+
   end
 
   module Authentication
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -23,5 +23,5 @@ on_worker_boot do
 end
 
 before_fork do
-  Rails.logger.info(LogMessages::Conjur::FipsModeStatus.new(OpenSSL.fips_mode ? "" : "non"))
+  Rails.logger.info(LogMessages::Conjur::FipsModeStatus.new(OpenSSL.fips_mode))
 end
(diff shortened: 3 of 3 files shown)


config/puma.rb
```diff
@@ -22,3 +22,6 @@
   Sequel::Model.db.disconnect
 end
 
+before_fork do
```
uCatu, line 25: Amm this is intresting, I don't know if I would print this log as part of `puma` since the server might fail to start and I'll have this log line hanging... 

Where do we print other configuration properties?

In addition, why we use `before_fork` It will run each time workder duplicate itself, I think it's also a deprecated method

WDYT to combine it with `on_worker_boot` above?
```
Code to run when a worker boots to setup
the process before booting the app.
```
https://github.com/nathansamson/puma/blob/2562379635c01906a3124a7600f593570cac92e0/lib/puma/dsl.rb#L175


Also suggest:
```
Rails.logger.info(LogMessages::Conjur::FipsModeStatus.new(OpenSSL.fips_mode))
```
If `OpenSSL.fips_mode` will always return true/false

app/domain/logs.rb
```diff
@@ -9,6 +9,11 @@ module Conjur
       code: "CONJ00034I"
     )
 
+    FipsModeStatus = ::Util::TrackableLogMessageClass.new(
+      msg:  "Conjur server set OpenSSL to run in {0} FIPS mode",
```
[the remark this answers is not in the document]
aloncarmel111: Done, waiting for @shulifink  review

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - All audit events now contain the IP address of the client that initiated the
   API request (e.g. `[client@43868 ip="172.24.0.5"]`)
   ([cyberark/conjur#1550](https://github.com/cyberark/conjur/issues/1550))
+- Print Conjur server FIPS mode status ([cyberark/conjur#1654](https://github.com/cyberark/conjur/issues/1654))
```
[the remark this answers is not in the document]
aloncarmel111: Done, @izgeri  can we update the contributing guide https://github.com/cyberark/community/blob/master/Conjur/CONTRIBUTING.md

config/puma.rb
aloncarmel111: `on_worker_boot` will run in a worker when it boots 
see: https://www.rubydoc.info/gems/puma/Puma%2FDSL:on_worker_boot

`before_fork `run on master before  forks workers (once on boot) 
see: https://www.rubydoc.info/gems/puma/Puma/DSL#before_fork-instance_method

seems that we don't printing any configuration properties to the logs

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - All audit events now contain the IP address of the client that initiated the
   API request (e.g. `[client@43868 ip="172.24.0.5"]`)
   ([cyberark/conjur#1550](https://github.com/cyberark/conjur/issues/1550))
+- Print Conjur server FIPS mode status. [cyberark/conjur#1654]
```
uCatu, line 25: From @izgeri thread:
```
- Change log entries should be treated as sentences and include punctuation (especially a period at the end)
- There’s no need to put the links in parentheses
I’ll be adding this to our changelog guidelines```
I think your previous format was good just without parentheses

Comment (aloncarmel111) [reply CHANGELOG.md]: Done

File: CHANGELOG.md (hunk 3 above)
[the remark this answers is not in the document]
Comment (aloncarmel111) [reply CHANGELOG.md]: Checking the new guidelines,  there’s no need to put the links in parentheses

File: app/domain/logs.rb (hunk 2 above)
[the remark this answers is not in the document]
Comment (shulifink) [reply app/domain/logs.rb]: I think the queslifier for the server should be in there ie DAP Server or Conjur Server, or even just DAP or Conjur 
Question: Is this for DAP and OSS? Or only DAP

If for DAP only,  "DAP set OpenSSL .....
If for OSS as well, can you put a variable which know to pick up DAP or Conjur when it's in the relevant environment?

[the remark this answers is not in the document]
Comment (uCatu) [reply app/domain/logs.rb]: HI @shulifink,
This message is for both :)
I was against coupling server name to this log message, I don't see any extra benefit from doing so, the log file is attached to his server - DAP/Conjur and can't be switched. 

WDYT?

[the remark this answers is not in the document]
Comment (shulifink) [reply app/domain/logs.rb]: So then simply:
OpenSSL FIPS mode set to true
OpenSSL FIPS mode set to false

File: app/domain/logs.rb [hunk 5]
```diff
@@ -9,6 +9,11 @@ module Conjur
       code: "CONJ00034I"
     )
 
+    FipsModeStatus = ::Util::TrackableLogMessageClass.new(
+      msg:  "Server set OpenSSL FIPS mode to {0}",
```
micahlee, line 13: Looking at the other log messages it seems that either
```
msg:  "OpenSSL FIPS mode set to {0}",
```
or
```
msg:  "Set OpenSSL FIPS mode to {0}",
```
would be more appropriate and consistent with other messages.

Is there a reason we need to explicitly call out the `Server`?

uCatu: @micahlee 
https://github.com/cyberark/conjur/pull/1653#discussion_r450121946
Same suggestion as Shuli :)

config/puma.rb
uCatu: @micahlee @sashaCher 
Can you share your opinion on using `puma.rb` and `before_fork` to display the logs message, is it the right place?
As brought up we also use `webrick` in some cases

app/domain/logs.rb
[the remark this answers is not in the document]
aloncarmel111: Done

app/domain/logs.rb
aloncarmel111: Done

config/puma.rb
micahlee: > Can you share your opinion on using puma.rb and before_fork to display the logs message, is it the right place?

Hmm.. I don't know about the Rails and Puma/Webrick relation to know the answer to this. @jonahx @jtuttle or @jvanderhoof, do any of you know or have any thoughts?

jtuttle: Why do you need to print this when puma forks a process? Can you remove the `before_fork` and just have it print once during initialization or does that not work for some reason?

aloncarmel111: It's works fine and print this log once during initialization without the `before_fork` method, added new commit to solved this issue.
@jtuttle @micahlee @uCatu

Issue #1655, opened 2020-06-29
Print proper message when user not exist in authn request


orenbm:
We currently do not verify that the user exists in authentication (and login) requests
with the default authenticator. This means that in case a user sends an authn request
with default authenticator and a Conjur user that does not exist we will write an 
`InvalidCredentials` message to the log. This doesn't tell the story properly and a 
`RoleNotFound` message will be better.

Note: this message will not be returned in the response and will only be written
to the log.

PR #1657, opened 2020-06-29
Log authentication and login errors in info

orenbm:
Connected to #1377, #1655

### What does this PR do?
Once an authentication or login request fails, we write the reason
to the log. We currently log it in DEBUG mode which is hard to filter
out. As this info is necessary, and customers would like to get it easily,
we would like to print it in INFO level, which is the default one.

#### Change log

#### Test coverage
We can't test this feature in Conjur as the log level in the tests are at DEBUG. However, [this PR](https://github.com/conjurinc/appliance/pull/1277) in the appliance will test it.

#### Documentation
- [ ] Created [this doc issue](https://github.com/cyberark/conjur-docs/issues/977). It can be done once this PR is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   FIPS Compliant mode is slightly slower then non-FIPS compliant
   ([cyberark/conjur#1527](https://github.com/cyberark/conjur/issues/1527))
 - Bump conjur-rack from 4.0.0 to 4.2.0 that consumes FIPS compliant slosilo(https://github.com/cyberark/conjur/issues/1527))
+- Print login and authentication error to the log in INFO level.
+  [cyberark/conjur#1377](https://github.com/cyberark/conjur/issues/1377)
 
 ### Added
 - Password changes (`PUT /authn/:account/password`) now produce audit events with
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Bump conjur-rack from 4.0.0 to 4.2.0 that consumes FIPS compliant slosilo(https://github.com/cyberark/conjur/issues/1527))
 - Print login and authentication error to the log in INFO level.
   [cyberark/conjur#1377](https://github.com/cyberark/conjur/issues/1377)
+- Print proper message when user does not exist in authn or login request with
+  default authenticator.
+  [cyberark/conjur#1655](https://github.com/cyberark/conjur/issues/1655)
 
 ### Added
 - Password changes (`PUT /authn/:account/password`) now produce audit events with
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -128,10 +128,7 @@ class AuthenticateController < ApplicationController
   private
 
   def handle_login_error(err)
-    logger.info("Login Error: #{err.inspect}")
-    err.backtrace.each do |line|
-      logger.debug(line)
-    end
+    log_error(err, "Login")
 
     case err
     when Errors::Authentication::Security::AuthenticatorNotWhitelisted,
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   FIPS Compliant mode is slightly slower then non-FIPS compliant
   ([cyberark/conjur#1527](https://github.com/cyberark/conjur/issues/1527))
 - Bump conjur-rack from 4.0.0 to 4.2.0 that consumes FIPS compliant slosilo(https://github.com/cyberark/conjur/issues/1527))
+- Print login and authentication error to the log in INFO level 
+  ([cyberark/conjur#1377](https://github.com/cyberark/conjur/issues/1377))
```
micahlee, line 17: The guidance for changelog entries [was recently updated](https://github.com/cyberark/community/pull/71). So these should now end with punctuation and not put the link in parens. For example:

```suggestion
- Print login and authentication error to the log in INFO level. 
  [cyberark/conjur#1377](https://github.com/cyberark/conjur/issues/1377)
```

app/domain/authentication/security/validate_role_can_access_webservice.rb
```diff
@@ -18,13 +18,14 @@ module Security
     ) do
 
       def call
-        # No checks required for default conjur authn
+        validate_account_exists
+        validate_role_is_defined
```
micahlee, line 22: It looks like these caused some rspec failures that will require test updates.

CHANGELOG.md
orenbm: thanks, didn't know that.

app/controllers/authenticate_controller.rb
```diff
@@ -128,23 +128,24 @@ def k8s_inject_client_cert
   private
 
   def handle_login_error(err)
-    logger.debug("Login Error: #{err.inspect}")
+    logger.info("Login Error: #{err.inspect}")
     err.backtrace.each do |line|
       logger.debug(line)
```
Tovli, line 133: Why do we print info level only for the first issue in the trace?

app/controllers/authenticate_controller.rb
```diff
@@ -128,23 +128,24 @@ def k8s_inject_client_cert
   private
 
   def handle_login_error(err)
-    logger.debug("Login Error: #{err.inspect}")
+    logger.info("Login Error: #{err.inspect}")
```
Tovli, line 131: I think it is confusing to have a `login error` in INFO level. maybe it will be better to write `login failed` as this is the information you want to give.
Unless it is really an error and not a valid part of the business logic

app/controllers/authenticate_controller.rb
```diff
@@ -128,23 +128,24 @@ def k8s_inject_client_cert
...
     else
       raise err
     end
   end
 
   def handle_authentication_error(err)
-    logger.debug("Authentication Error: #{err.inspect}")
+    logger.info("Authentication Error: #{err.inspect}")
     err.backtrace.each do |line|
       logger.debug(line)
```
Tovli, line 150: I see this is repeating itself, shouldn't we refactor this into a method?

app/domain/authentication/security/validate_role_can_access_webservice.rb
```diff
@@ -18,13 +18,14 @@ module Security
...
+
+        # No further checks required for default conjur authn
         return if default_conjur_authn?
 
-        validate_account_exists
         validate_webservice_exists
-        validate_user_is_defined
-        validate_user_has_access
+        validate_role_has_access
```
Tovli, line 28: I wonder if moving from user to role does not create confusion for the future devs

app/controllers/authenticate_controller.rb
orenbm: it is not an error but customers (and SEs) already know to look for these messages so i don't want to change that.

app/controllers/authenticate_controller.rb
orenbm: we print just the one line that customers will look for and can help them. This line prints the stack trace which is more for developers and thus is printed at DEBUG level.

app/controllers/authenticate_controller.rb
orenbm: Done

app/domain/authentication/security/validate_role_can_access_webservice.rb
orenbm: Done, thanks

orenbm:
> This will probably also require an issue/PR for the appliance tests, correct?

Yes @micahlee, I already created it and linked to it from the PR description.

orenbm:
@micahlee can you re-review please?

Issue #1640, opened 2020-06-19
Audit Event Inputs are Refactored to Values rather than Models


micahlee:
See comments on this PR:
https://github.com/cyberark/conjur/pull/1637

There are several inputs to audit events where we pass a data model (e.g. `Role`) when all we need is the `role_id` (`Role#id`) instead. These should be refactored to use the simpler, more obvious form.

Related to: #1608

PR #1658, opened 2020-06-29
Pass role_id instead of role to authn events

jonahx:
All of these events only ever use the "role.id" method of the "role"
    object that is passed to them, so passing the entire "role" object is
    unnecessary and misleading.
    
    We also introduce an Audit::Event::Authn::RoleId to encapsulate the
    conditional logic around the meaning of "role id" in the context of
    authentication events.

### What does this PR do?

To simplify the interface of the audit events and consolidate the many unnecessary nil-checks being done in every audit message into a single place.

#1640 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/app/domain/authentication/authn_oidc/authenticator.rb
+++ b/app/domain/authentication/authn_oidc/authenticator.rb
@@ -143,11 +143,11 @@ module Authentication
         @audit_log.log(
           ::Audit::Event::Authn::Authenticate.new(
             authenticator_name: authenticator_name,
-            service:            webservice,
-            role:               role,
-            client_ip:          client_ip,
-            success:            true,
-            error_message:      nil
+            service: webservice,
+            role_id: audit_role_id,
+            client_ip: client_ip,
+            success: true,
+            error_message: nil
           )
         )
       end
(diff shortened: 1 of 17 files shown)


app/domain/authentication/authn_oidc/authenticator.rb
```diff
@@ -156,15 +156,23 @@ def audit_failure(err)
         @audit_log.log(
           ::Audit::Event::Authn::Authenticate.new(
             authenticator_name: authenticator_name,
-            service:            webservice,
-            role:               role,
-            client_ip:          client_ip,
-            success:            false,
-            error_message:      err.message
+            service: webservice,
+            role_id: audit_role_id,
```
jtuttle, line 160: You make a helper method for creating the `RoleId` object in this class but in-line it in `authenticate.rb`. We should probably pick one pattern and try to stick to it.

app/models/audit/event/authn/role_id.rb
```diff
@@ -0,0 +1,23 @@
...
+      # user or host authenticating doesn't exist.  In this case, the "role_id"
+      # becomes "what the role id would be if the user _did_ exist". In all
+      # other cases, it's the actual "role_id" for the existing user.
+      #
+      # The control parameter smell here is intentional: This is the object
+      # that is hiding the "nil"-ugliness.
+      # :reek:ControlParameter
+      class RoleId
+        # Note: "username" may refer to a user or host.
+        def self.new(role:, account:, username:)
```
jtuttle, line 15: I don't particularly like this method being called `new` because naively I would expect it to act like a constructor and return an object of type `Audit::Event::Authn::RoleId`. I'm not sure I have a better idea, though.

jonahx: It's a little funny, true.  There are two alternatives here:

1. Make it a utility function.  I am not a big fan of utility functions in ruby so lean away from this.
2. Have it return a proper `RoleId` instance, which has a `to_s` method which returns the the actual role id -- ie, what we're returning in the above.  Conceptually, this is cleaner and more in the OO / ruby spirit.  Wdyt?

jtuttle: Yeah I think 2 makes sense. It's a little more clear what's going on in that case.

Issue #1663, opened 2020-07-05
CIDR support proxy - fetch permitted proxies impl


nessiLahav:
Implement how to define and fetch the configuration of the permitted proxies according to the [design](https://github.com/conjurinc/appliance/tree/master/design/cidr_proxy/designs)  

DoD:

- [x] We are able to define and fetch permitted proxies list

PR #1678, opened 2020-07-13
Cidr proxy fetch permitted proxies

nessiLahav:
### What does this PR do?
Implement `FetchTrustedProxies` command class

Connected to #1663 

#### Change log
Will be updated in the end of the feature development 

#### Test coverage
This PR includes UTs, integration tests will be developed in a different PR

#### Documentation
Will be updated in the end of the feature development

### DIFF
--- a/app/domain/conjur/fetch_trusted_proxies.rb
+++ b/app/domain/conjur/fetch_trusted_proxies.rb
@@ -4,19 +4,25 @@ require 'command_class'
 
 module Conjur
 
-  FetchPermittedProxies ||= CommandClass.new(
+  FetchTrustedProxies ||= CommandClass.new(
     dependencies: {
-        role_cls:                          ::Role,
-        validate_account_exists:           ::Authentication::Security::ValidateAccountExists.new
+        role_cls:                 ::Role,
+        validate_account_exists:  ::Authentication::Security::ValidateAccountExists.new,
+        logger:                   Rails.logger
     },
     inputs: %i(account)
   ) do
 
     def call
       validate_account_exists
-      fetch_permitted_proxies_list
+      fetch_trusted_proxies_list
       delete_duplications_in_list
-      permitted_proxies_list
+      trusted_proxies_list
+    rescue => e
+      raise Errors::Conjur::TrustedProxiesMissing.new(
+          host_id,
+          e.inspect
+      )
     end
 
     private
@@ -27,18 +33,25 @@ module Conjur
       )
     end
 
-    def permitted_proxies_list
+    def trusted_proxies_list
       @proxies_list
     end
 
     def delete_duplications_in_list
+      @logger.debug(LogMessages::Conjur::DeletingTrustedProxiesDuplications.new)
       if @proxies_list
         @proxies_list = @proxies_list.uniq { |cidr| [cidr.to_s]}
       end
+      @logger.debug(LogMessages::Conjur::TrustedProxiesAmount.new(@proxies_list.length))
     end
 
-    def fetch_permitted_proxies_list
+    def fetch_trusted_proxies_list
+      @logger.debug(LogMessages::Conjur::FetchingTrustedProxies.new(host_id))
       @proxies_list = role.restricted_to
+      unless @proxies_list
+        @proxies_list = Array.new
+      end
+      @logger.debug(LogMessages::Conjur::FetchingTrustedProxies.new(@proxies_list.length))
     end
 
     def role
--- a/app/domain/conjur/fetch_trusted_proxies.rb
+++ b/app/domain/conjur/fetch_trusted_proxies.rb
@@ -6,12 +6,15 @@ module Conjur
 
   FetchTrustedProxies ||= CommandClass.new(
     dependencies: {
-        role_cls:                 ::Role,
-        validate_account_exists:  ::Authentication::Security::ValidateAccountExists.new,
-        logger:                   Rails.logger
+      role_cls:                 ::Role,
+      validate_account_exists:  ::Authentication::Security::ValidateAccountExists.new,
+      logger:                   Rails.logger
     },
-    inputs: %i(account)
+    inputs:       %i(account)
   ) do
+
+    SETTINGS_POLICY_ID = "conjur/settings"
+    TRUSTED_PROXIES_HOST_NAME = "trusted_proxies"
 
     def call
       validate_account_exists
@@ -19,9 +22,9 @@ module Conjur
       delete_duplications_in_list
       trusted_proxies_list
     rescue => e
-      raise Errors::Conjur::TrustedProxiesMissing.new(
-          host_id,
-          e.inspect
+      raise Errors::Conjur::TrustedProxiesFetchFailed.new(
+        host_id,
+        e.inspect
       )
     end
 
@@ -29,29 +32,26 @@ module Conjur
 
     def validate_account_exists
       @validate_account_exists.(
-          account: @account
+        account: @account
       )
-    end
-
-    def trusted_proxies_list
-      @proxies_list
-    end
-
-    def delete_duplications_in_list
-      @logger.debug(LogMessages::Conjur::DeletingTrustedProxiesDuplications.new)
-      if @proxies_list
-        @proxies_list = @proxies_list.uniq { |cidr| [cidr.to_s]}
-      end
-      @logger.debug(LogMessages::Conjur::TrustedProxiesAmount.new(@proxies_list.length))
     end
 
     def fetch_trusted_proxies_list
       @logger.debug(LogMessages::Conjur::FetchingTrustedProxies.new(host_id))
-      @proxies_list = role.restricted_to
-      unless @proxies_list
-        @proxies_list = Array.new
+      @trusted_proxies_list = role.restricted_to || []
+      @logger.debug(LogMessages::Conjur::FetchingTrustedProxies.new(@trusted_proxies_list.length))
+    end
+
+    def delete_duplications_in_list
+      @logger.debug(LogMessages::Conjur::DeletingTrustedProxiesDuplications.new)
+      if @trusted_proxies_list.any?
+        @trusted_proxies_list = @trusted_proxies_list.uniq {|cidr| [cidr.to_s]}
       end
-      @logger.debug(LogMessages::Conjur::FetchingTrustedProxies.new(@proxies_list.length))
+      @logger.debug(LogMessages::Conjur::TrustedProxiesAmount.new(@trusted_proxies_list.length))
+    end
+
+    def trusted_proxies_list
+      @trusted_proxies_list
     end
 
     def role
@@ -67,15 +67,8 @@ module Conjur
     end
 
     def host_id
-      return "host/" + policy_id + "/" + host_name
-    end
-
-    def policy_id
-      return "settings"
-    end
-
-    def host_name
-      return "trusted_proxies"
+      @host_id ||= "host/#{SETTINGS_POLICY_ID}/#{TRUSTED_PROXIES_HOST_NAME}"
     end
   end
 end
+
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -20,9 +20,9 @@ module Errors
       code: "CONJ00046E"
     )
 
-    TrustedProxiesMissing = ::Util::TrackableErrorClass.new(
-        msg:  "Failed to fetch trusted proxies list. Reason: '{1}'",
-        code: "CONJ00065E"
+    TrustedProxiesFetchFailed = ::Util::TrackableErrorClass.new(
+      msg:  "Failed to fetch trusted proxy list. Reason: '{0}'",
+      code: "CONJ00065E"
     )
 
   end
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -20,6 +20,11 @@ module Errors
       code: "CONJ00046E"
     )
 
+    TrustedProxiesMissing = ::Util::TrackableErrorClass.new(
+        msg:  "Failed to fetch trusted proxies list. Reason: '{1}'",
+        code: "CONJ00065E"
+    )
+
   end
 
   module Authentication
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -15,23 +15,23 @@ module LogMessages
     )
 
     FetchingTrustedProxies = ::Util::TrackableLogMessageClass.new(
-        msg:  "Fetching trusted proxies list from {0-host-id}",
-        code: "CONJ00039D"
+      msg:  "Fetching trusted proxy list from {0-host-id}",
+      code: "CONJ00039D"
     )
 
     FetchedTrustedProxies = ::Util::TrackableLogMessageClass.new(
-        msg:  "{0-list-length} trusted proxies ip addresses has fetched",
-        code: "CONJ00040D"
+      msg:  "{0-list-length} trusted proxies were fetched",
+      code: "CONJ00040D"
     )
 
     DeletingTrustedProxiesDuplications = ::Util::TrackableLogMessageClass.new(
-        msg:  "Deleting duplications from trusted proxies list",
-        code: "CONJ00041D"
+      msg:  "Deleting duplications from trusted proxy list",
+      code: "CONJ00041D"
     )
 
     TrustedProxiesAmount = ::Util::TrackableLogMessageClass.new(
-        msg:  "Trusted proxies list contains {0-list-length} ip addresses",
-        code: "CONJ00042D"
+      msg:  "Trusted proxy list contains {0-list-length} IP addresses",
+      code: "CONJ00042D"
     )
 
   end
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -14,6 +14,26 @@ module LogMessages
       code: "CONJ00038I"
     )
 
+    FetchingTrustedProxies = ::Util::TrackableLogMessageClass.new(
+        msg:  "Fetching trusted proxies list from {0-host-id}",
+        code: "CONJ00039D"
+    )
+
+    FetchedTrustedProxies = ::Util::TrackableLogMessageClass.new(
+        msg:  "{0-list-length} trusted proxies ip addresses has fetched",
+        code: "CONJ00040D"
+    )
+
+    DeletingTrustedProxiesDuplications = ::Util::TrackableLogMessageClass.new(
+        msg:  "Deleting duplications from trusted proxies list",
+        code: "CONJ00041D"
+    )
+
+    TrustedProxiesAmount = ::Util::TrackableLogMessageClass.new(
+        msg:  "Trusted proxies list contains {0-list-length} ip addresses",
+        code: "CONJ00042D"
+    )
+
   end
 
   module Authentication
--- a/spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
+++ b/spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
@@ -9,17 +9,104 @@ RSpec.describe 'Conjur::FetchTrustedProxies' do
 
   let(:valid_host) { "host/settings/trusted_proxies" }
 
+  let(:ip_1) { Util::CIDR.new("1.0.0.0/8") }
+  let(:ip_2) { Util::CIDR.new("1.2.3.0/24") }
+  let(:ip_3) { Util::CIDR.new("1.2.3.4") }
+
   let(:mocked_role_class_empty_list) { double("RoleClass") }
   let(:mocked_role_empty_list) { double("RoleReturnsEmptyList") }
-  let(:empty_trusted_proxies_list) { Array.new }
+  let(:empty_trusted_proxies_list) { [] }
+
+  let(:mocked_role_class_return_nil) { double("RoleClass") }
+  let(:mocked_role_return_nil) { double("RoleReturnsNil") }
+
+  let(:mocked_role_class_throw_error_in_restricted_to) { double("RoleClass") }
+  let(:mocked_role_throw_error_in_restricted_to) { double("RoleRestrictedToError") }
+
+  let(:mocked_role_class_throw_error_in_by_login) { double("RoleClassByLoginError") }
+
+  let(:mocked_role_class_by_login_returns_nil_and_roleid_valid) { double("RoleClassByLoginNilAndRoleIdValid") }
+  let(:mocked_role_class_by_login_returns_nil_and_roleid_nil) { double("RoleClassByLoginNilAndRoleIdNil") }
+  let(:mocked_role_class_by_login_returns_nil_and_roleid_throws_error) { double("RoleClassByLoginNilAndRoleIdError") }
+
+  let(:mocked_role_class_list_no_duplications) { double("RoleClass") }
+  let(:mocked_role_list_no_duplications) { double("RoleReturnsListWithoutDuplications") }
+  let(:trusted_proxies_list_no_duplications) { [ip_1, ip_2 , ip_3] }
+
+  let(:mocked_role_class_list_duplications_1) { double("RoleClass") }
+  let(:mocked_role_list_duplications_1) { double("RoleReturnsListWithDuplications") }
+  let(:trusted_proxies_list_duplications_1) { [ip_1, ip_1 , ip_1] }
+  let(:expected_list_1) { [trusted_proxies_list_duplications_1[0]] }
+
+  let(:mocked_role_class_list_duplications_2) { double("RoleClass") }
+  let(:mocked_role_list_duplications_2) { double("RoleReturnsListWithDuplications") }
+  let(:trusted_proxies_list_duplications_2) { [ip_1, ip_2 , ip_3, ip_1, ip_2 , ip_3] }
+  let(:expected_list_2) { [trusted_proxies_list_duplications_2[0],
+                           trusted_proxies_list_duplications_2[1],
+                           trusted_proxies_list_duplications_2[2]] }
 
 
   before(:each) do
     allow(mocked_role_class_empty_list).to receive(:by_login)
-                                           .with(valid_host, {:account=> valid_account})
-                                           .and_return(mocked_role_empty_list)
+                                               .with(valid_host, {:account=> valid_account})
+                                               .and_return(mocked_role_empty_list)
     allow(mocked_role_empty_list).to receive(:restricted_to)
-                                           .and_return(empty_trusted_proxies_list)
+                                         .and_return(empty_trusted_proxies_list)
+
+    allow(mocked_role_class_throw_error_in_restricted_to).to receive(:by_login)
+                                                                 .with(valid_host, {:account=> valid_account})
+                                                                 .and_return(mocked_role_throw_error_in_restricted_to)
+    allow(mocked_role_throw_error_in_restricted_to).to receive(:restricted_to)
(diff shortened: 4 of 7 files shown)


app/domain/errors.rb
```diff
@@ -20,6 +20,11 @@ module Conjur
       code: "CONJ00046E"
     )
 
+    TrustedProxiesMissing = ::Util::TrackableErrorClass.new(
+        msg:  "Failed to fetch trusted proxies list. Reason: '{1}'",
+        code: "CONJ00065E"
+    )
```
nessiLahav, line 26: @shulifink please review :)

app/domain/logs.rb
```diff
@@ -14,6 +14,26 @@ module Conjur
       code: "CONJ00038I"
     )
```
nessiLahav, line 16: @shulifink please the following new logs :)

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
...
+
+require 'command_class'
+
+module Conjur
+
+  FetchTrustedProxies ||= CommandClass.new(
+    dependencies: {
+        role_cls:                 ::Role,
+        validate_account_exists:  ::Authentication::Security::ValidateAccountExists.new,
+        logger:                   Rails.logger
```
orenbm, line 11: fix indentation - should be 2 spaces per indent

spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+                                                  .and_return(trusted_proxies_list_duplications_2)
+
+    allow(mocked_account_validator).to receive(:call)
+                                           .with({:account=>valid_account})
+                                           .and_return(true)
+    allow(mocked_account_validator).to receive(:call)
+                                           .with({:account=>invalid_account})
+                                           .and_raise('ACCOUNT_NOT_EXIST_ERROR')
+  end
+  #  ____  _   _  ____    ____  ____  ___  ____  ___
```
nessiLahav, line 118: @eladkug Please review the following UTs, also see commit message, thanks :)

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
...
+      raise Errors::Authentication::Security::RoleNotFound, role_id unless @role
+      @role
+    end
+
+    def role_id
+      @role_id ||= @role_cls.roleid_from_username(@account, host_id)
+    end
+
+    def host_id
+      return "host/" + policy_id + "/" + host_name
```
orenbm, line 70: please change to `return "host/#{policy_id}/#{host_name}"`

orenbm: and use memoization

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    def policy_id
+      return "settings"
```
orenbm, line 74: `return` is redundant

orenbm: also - this should be a const, not a method

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    def host_name
+      return "trusted_proxies"
```
orenbm, line 78: this should also be a const

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
...
+          host_id,
+          e.inspect
+      )
+    end
+
+    private
+
+    def validate_account_exists
+      @validate_account_exists.(
+          account: @account
```
orenbm, line 32: fix indentation

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
```
Tovli, line 1: The file name is fetch_trusted_proxies but it also do delete duplication... 
What is the responsibility of this file?

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    def validate_account_exists
+      @validate_account_exists.(
+          account: @account
+      )
+    end
+
+    def trusted_proxies_list
+      @proxies_list
+    end
```
orenbm, line 38: please change one of these. it should either be `proxies_list` or `trusted_proxies_list`

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
...
+    private
+
+    def validate_account_exists
+      @validate_account_exists.(
+          account: @account
+      )
+    end
+
+    def trusted_proxies_list
+      @proxies_list
```
oburstein-hub, line 37: please, elaborate why is this function needed

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub, line 70: should define constant to "host"

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub, line 74: should define constant to setting - maybe even instead of function (if possible)

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
...
+    inputs: %i(account)
+  ) do
+
+    def call
+      validate_account_exists
+      fetch_trusted_proxies_list
+      delete_duplications_in_list
+      trusted_proxies_list
+    rescue => e
+      raise Errors::Conjur::TrustedProxiesMissing.new(
```
orenbm, line 22: this is something that we don't do in other places. can you elaborate on the cases where this error would occur? I am trying to see if we can avoid it and place the `raise` inside one of the methods.

Also - please at lease add a `begin` after `validate_account_exists` so we don't get a `TrustedProxiesMissing` error when the error is that the account doesn't exist.

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub, line 78: should define constant to "trusted_proxies" - maybe even instead of function (if possible)

app/domain/conjur/fetch_trusted_proxies.rb
Tovli: I did not get why you created a method that return a string that is constructed from two consts
Did you meant it to be configurable in the future?

app/domain/errors.rb
oburstein-hub: code strings should be constants

app/domain/errors.rb
```diff
@@ -20,6 +20,11 @@ module Conjur
       code: "CONJ00046E"
     )
 
+    TrustedProxiesMissing = ::Util::TrackableErrorClass.new(
```
orenbm, line 23: looking at the message, it seems that the name is not good enough. how about `TrustedProxiesFetchFailed`?

app/domain/conjur/fetch_trusted_proxies.rb
Tovli, line 22: It seems this error will be raised also for DB error (or any other error).
Is it really trusted proxies missing?
Also - why is it an error to not have trusted proxies? it is a valid configuration AFAIK

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: @Tovli good point, i am fetching the list from the db which may contains duplication because we do not prevent it in load policy.
So the answer is 2:
1. Fetch the list from DB
2. Prepare the list for validation process - delete duplication

spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+
+  context "A valid trusted proxies configuration" do
+    context "trusted proxies contains an empty list of IPs addresses" do
+      context "when role.restricted_to returns empty array" do
+        subject do
+          Conjur::FetchTrustedProxies.new(
+              role_cls:                 mocked_role_class_empty_list,
+              validate_account_exists:  mocked_account_validator,
+              ).call(
+              account: valid_account
```
orenbm, line 131: fix indentation

spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+          )
+        end
+
+        it "returns an empty list" do
+          expect( subject ).to eq(empty_trusted_proxies_list)
+        end
+      end
+    end
+
+    context "trusted proxies contains an list of IPs addresses without duplications" do
```
orenbm, line 156: a list

orenbm: of IP addresses

spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+          )
+        end
+
+        it "raises an error" do
+          expect { subject }.to raise_error(::Errors::Conjur::TrustedProxiesMissing)
+        end
+      end
+    end
+  end
+end
```
orenbm, line 297: add missing newline

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    def delete_duplications_in_list
+      @logger.debug(LogMessages::Conjur::DeletingTrustedProxiesDuplications.new)
+      if @proxies_list
+        @proxies_list = @proxies_list.uniq { |cidr| [cidr.to_s]}
+      end
+      @logger.debug(LogMessages::Conjur::TrustedProxiesAmount.new(@proxies_list.length))
+    end
+
+    def fetch_trusted_proxies_list
```
oburstein-hub, line 48: This is a fetch from policy configuration, right ?
Are you assuming that policy configuration of trusted-proxies will contain duplications ?
If yes - just notice that it is not a best practice for configuring entities. The best practice is to disallow duplication when user tries to enter it

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub: I think all these methods must be converted to string constants unless these are overrides / implementations of base class / interface

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub: I think that verify_account_missing should throw another exception - it has nothing to do with valid proxies

nessiLahav: This is an internal action against the DB to fetch the configured trusted proxies so no matter what is the error i want that the main error in log will be on incorrect proxy configuration with an additional reason .
So validate account is also include here

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub: Cannot we prevent a duplication on load policy ? This is a better practice of object modelling, and there is a code which handles such things in conjur-parser

nessiLahav: @oburstein-hub everything is possible :) time is the problem

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: Fix it, review again :)

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub: How long will it take and what is your deadline to submit it ?

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: Lets talk about this with @shaharglazner and @InbalZilberman as far as i understand this security feature is always enabled so if you are unable to fetch the trusted proxies from the DB we should fail the request

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: Its aligned to our code, i dont remember the motivation

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: Please take it with @InbalZilberman to prioritise it and understand the risks of such change in load policy

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: Please take it with @Tovli and @InbalZilberman, i dont think it is in my scope

spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
```diff
@@ -34,17 +121,177 @@
...
+              account: valid_account
+          )
+        end
+
+        it "raises an error" do
+          expect { subject }.to raise_error(::Errors::Conjur::TrustedProxiesMissing)
+        end
+      end
+
+      context "then role_cls.roleid_from_username returns nil" do
```
eladkug, line 281: then role_cls.roleid_from_username throws an error

spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
```diff
@@ -34,17 +121,177 @@
...
+              account: valid_account
+          )
+        end
+
+        it "returns list with no duplication" do
+          expect( subject ).to eq(expected_list_1)
+        end
+      end
+
+      context "with 3 same IP addresses" do
```
eladkug, line 187: There is 3 duplication ([ip_1, ip_2 , ip_3, ip_1, ip_2 , ip_3] ) and all over the test you wrote mocked_role_class_list_duplications_2, expected_list_2 , need to make it clear.

app/domain/conjur/fetch_trusted_proxies.rb
orenbm: fwiw i'm ok with calling the class `FetchTrustedProxies` while it also removes the duplications before actually returning the proxies.

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'command_class'
+
+module Conjur
+
+  POLICY_ID = "settings"
```
orenbm, line 7: i would call it `SETTINGS_POLICY_ID` and `TRUSTED_PROXIES_HOST_NAME`.

Also, i think it should go inside the class and not defined in the scope of the `Conjur` module.

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,75 @@
...
+  HOST_NAME = "trusted_proxies"
+
+  FetchTrustedProxies ||= CommandClass.new(
+      dependencies: {
+          role_cls: ::Role,
+          validate_account_exists: ::Authentication::Security::ValidateAccountExists.new,
+          logger: Rails.logger
```
orenbm, line 14: fix indentation

app/domain/conjur/fetch_trusted_proxies.rb
orenbm: but why? if the reason of the error is that the account doesn't exist then that's the action item that the user needs to do. why bother them with a `TrustedProxiesMissing` error?

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def call
+      validate_account_exists
+      fetch_trusted_proxies_list
+      delete_duplications_in_list
+      trusted_proxies_list
+    rescue => e
+      raise Errors::Conjur::TrustedProxiesFetchFailed.new(
+          host_id,
+          e.inspect
```
orenbm, line 27: fix indentation

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,75 @@
...
+    def validate_account_exists
+      @validate_account_exists.(
+          account: @account
+      )
+    end
+
+    def fetch_trusted_proxies_list
+      @logger.debug(LogMessages::Conjur::FetchingTrustedProxies.new(host_id))
+      @trusted_proxies_list = role.restricted_to
+      @trusted_proxies_list ||= []
```
orenbm, line 42: can you add a comment to explain this line? Do you initialize to an empty array in case the trusted_proxies role doesn't have a `restricted_to` field?

app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -0,0 +1,75 @@
...
+      @logger.debug(LogMessages::Conjur::FetchingTrustedProxies.new(@trusted_proxies_list.length))
+    end
+
+    def delete_duplications_in_list
+      @logger.debug(LogMessages::Conjur::DeletingTrustedProxiesDuplications.new)
+      if @trusted_proxies_list
+        @trusted_proxies_list = @trusted_proxies_list.uniq {|cidr| [cidr.to_s]}
+      end
```
orenbm, line 50: is there a case where we won't go inside this `if` clause? `@trusted_proxies_list` is set to `[]` if `role.restricted_to` is nil so `@trusted_proxies_list` will never be nil. 

did you mean `if @trusted_proxies_list.any?` ? this will return false on an empty list.

app/domain/conjur/fetch_trusted_proxies.rb
Tovli: @nessiLahav - regarding what @oburstein-hub wrote - Please open an issue for any work we won't do due to time so we can do it at a later phase. You don't need me or Inbal for this

Regarding the naming - do we usually use verbs to decribe classes? It feels odd to me as I would expect to have an object that deals with request origin and in it a method to fetch the client ip that part of its work is to validate the proxies
WDYT?

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub: I think we should distinguish here between the case of DB failure, and the case of unconfigured trusted proxies
DB failure case should be handled as an exception.
Unconfigured trusted proxies handling should be defined by Inbal and Shahar

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub: It looks like it does not do anything. Maybe you should consider to remove it

orenbm: the motivation is to let the `call` method tell a story. this way the `call` method is:
```
validate_account_exists
fetch_trusted_proxies_list
delete_duplications_in_list
trusted_proxies_list
```

it can be also:
```
validate_account_exists
fetch_trusted_proxies_list
delete_duplications_in_list
@trusted_proxies_list
```

and the method will be removed but i think there's no harm in having it

app/domain/conjur/fetch_trusted_proxies.rb
orenbm: we use verbs for command classes.

app/domain/errors.rb
shulifink: fine

spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
shulifink: ACCOUNT_DOES_NOT_EXIST_ERROR. or NO_ACCOUNT_ERROR

app/domain/logs.rb
```diff
@@ -14,6 +14,26 @@ module Conjur
...
 
+    FetchingTrustedProxies = ::Util::TrackableLogMessageClass.new(
+        msg:  "Fetching trusted proxies list from {0-host-id}",
```
shulifink: fine

app/domain/logs.rb
```diff
@@ -14,6 +14,26 @@ module Conjur
...
+        code: "CONJ00039D"
+    )
+
+    FetchedTrustedProxies = ::Util::TrackableLogMessageClass.new(
+        msg:  "{0-list-length} trusted proxies ip addresses has fetched",
```
shulifink: Let's discuss...

app/domain/logs.rb
```diff
@@ -14,6 +14,26 @@ module Conjur
...
+        code: "CONJ00040D"
+    )
+
+    DeletingTrustedProxiesDuplications = ::Util::TrackableLogMessageClass.new(
+        msg:  "Deleting duplications from trusted proxies list",
```
shulifink: fine

app/domain/logs.rb
```diff
@@ -14,6 +14,26 @@ module Conjur
...
+        code: "CONJ00041D"
+    )
+
+    TrustedProxiesAmount = ::Util::TrackableLogMessageClass.new(
+        msg:  "Trusted proxies list contains {0-list-length} ip addresses",
```
shulifink: Trusted proxies list contains {0-list-length} IP addresses

(IP is always uppercase)

spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+            account: valid_account
+        )
+      end
+
+      it "returns expected list" do
+        expect( subject ).to eq(trusted_proxies_list_no_duplications)
+      end
+    end
+
+    context "trusted proxies contains an list of IPs addresses with duplications" do
```
shulifink: trusted proxies contains a list of IP addresses without duplications

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: Because from the POV of this command class the account is not related to the user, we get this account as an input and maybe in the future it will be different from some reason.
And from UX wise i do not think its bothering the customer to have more context to the `account doesn't exist` error

shulifink:
@nessiLahav , Did I review everything I was supposed to?

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: I do not think we should distinguish between the errors in this flows, any errors that prevent from me to return the trusted proxy list should be wrapped under `TrustedProxiesFetchFailed` error

orenbm: this is a new behaviour in our code. In other places we do not put an error inside an error in case it's related. The `account` comes from the authentication request so if I send a request with a wrong account i want the authentication error in the log to be concise and point me to the error. what do i care that you found that the account is missing while you are fetching the proxies and not, for example, while you are verifying that the webservice exists? 

For me as a user, all i want to know is that the account is incorrect and that i need to fix it.

orenbm: the stack trace will state the place where the error was raised but the one-liner should only say that the account doesn't exist IMO

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: correct

app/domain/logs.rb
nessiLahav: Fixed :)

app/domain/errors.rb
```diff
@@ -20,6 +20,11 @@ module Conjur
       code: "CONJ00046E"
     )
 
+    TrustedProxiesFetchFailed = ::Util::TrackableErrorClass.new(
+      msg:  "Failed to fetch trusted proxies list. Reason: '{1}'",
```
orenbm, line 24: why is this 1 and not 0?

spec/app/domain/conjur/fetch_trusted_proxies_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+          )
+        end
+
+        it "raises an error" do
+          expect {subject}.to raise_error(::Errors::Conjur::TrustedProxiesFetchFailed)
+        end
+      end
+    end
+  end
+end
```
orenbm, line 297: still missing

app/domain/errors.rb
```diff
@@ -20,6 +20,11 @@ module Conjur
       code: "CONJ00046E"
     )
 
+    TrustedProxiesMissing = ::Util::TrackableErrorClass.new(
+        msg:  "Failed to fetch trusted proxies list. Reason: '{1}'",
```
InbalZilberman: @nessiLahav how about
Failed to fetch trusted proxies list provided by  host conjur/setting/trusted-proxies

app/domain/logs.rb
```diff
@@ -14,6 +14,26 @@ module Conjur
...
 
+    FetchingTrustedProxies = ::Util::TrackableLogMessageClass.new(
+        msg:  "Fetching trusted proxies list from {0-host-id}",
+        code: "CONJ00039D"
+    )
+
+    FetchedTrustedProxies = ::Util::TrackableLogMessageClass.new(
+        msg:  "{0-list-length} trusted proxies ip addresses has fetched",
+        code: "CONJ00040D"
+    )
```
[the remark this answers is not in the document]
InbalZilberman: @nessiLahav can we log these addresses?

app/domain/logs.rb
```diff
@@ -14,6 +14,26 @@ module Conjur
...
+
+    DeletingTrustedProxiesDuplications = ::Util::TrackableLogMessageClass.new(
+        msg:  "Deleting duplications from trusted proxies list",
+        code: "CONJ00041D"
+    )
```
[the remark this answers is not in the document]
InbalZilberman: @nessiLahav can we log these non duplicate addresses?

app/domain/conjur/fetch_trusted_proxies.rb
oburstein-hub: I agree that this is a new behaviour in our code , and that we decide about distinguishing between 2 cases. I think that we find that account is missing during fetching the proxies while verification

app/domain/errors.rb
nessiLahav: @InbalZilberman if the reason is that the policy or the host is not defined it will be written in the error under: `
Reason: '#<Errors::Authentication::Security::RoleNotFound: CONJ00007E 'cucumber:host:settings/trusted_proxies' not found>'>
`
@InbalZilberman do you still think we should change the message ?

app/domain/logs.rb
nessiLahav: @shulifink this is the amount of IP addresses that fetched, what do you want to discuss about?

orenbm:
@nessiLahav when you merge this PR please squash the commits into one and provide a [good commit message](https://chris.beams.io/posts/git-commit/).

PR #1683, opened 2020-07-13
Add file to enable GH action for enforcing rebases

jonahx:
See:
https://github.com/jonahx/enforce-rebase

### What does this PR do?
- Adds a GH action workflow file that will allow us to enable enforced rabasing

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,16 @@
+name: Enforce Rebasing
+
+on: [push]
+
+jobs:
+  rebase_job:
+    runs-on: ubuntu-latest
+    name: Enforce rebasing
+
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


sgnn7:
Drive-by comment: I think this is already a flag that can be flipped on per-repo by an admin - we have something like that set in secretless-broker iirc (@izgeri might know what setting it it)

izgeri:
noted in slack:

@jonahx's action has a special section in its [README](https://github.com/jonahx/enforce-rebase#how-is-this-different-from-the-require-linear-history-status-check) about how the action differs from the "require linear history" status check

I'm not sure I understand how it's different from the "Require branches to be up to date before merging” status check, though - that’s the one that's on in secretless
https://docs.github.com/en/github/administering-a-repository/types-of-required-status-checks

PR #1688, opened 2020-07-15
Add safer data-handling in ConjurCA

### LINKED ISSUE #1315: Repos::ConjurCA doesn't raise a proper error when required secrets are empty
## Issue description

The class `Repos::ConjurCA` retrieves its required secrets by creating a Resource object and calling its `last_secret` method.

This is not good for supportability as in case the variable has no value we get a "NoMethodError: undefined method 'value' for nil:NilClass" error which doesn't tell the story.

In `FetchRequiredSecrets` we first verify that the variable exists and that it has a value before retrieving it. We should use this class in `Repos::ConjurCA` so we have this capability out of the box.

## Steps to reproduce the issue
1. Create an authn-k8s authenticator without loading a value into `conjur/authn-k8s/<service-id>/ca/cert`
2. Run the authn-client to authenticate with k8s
3. The Nil error will show in the logs

## What's the expected result?
- a RequiredResourceMissing or RequiredSecretMissing error is raised

## What's the actual result?
- a "NoMethodError: undefined method 'value' for nil:NilClass" error is raised

liavyona:
### What does this PR do?
In order to avoid NoMethodError that can be raised when the ConjurCA relevant resources is missing/have no value, added custom exceptions which inform the proper error.

resolves #1315 

### Checklists
- Reproduce the bug by deploy current version
- Deploy a new env with the these changes
- Test authn k8s works fine
- Test the new exceptions raised when ConjurCa has malformed data.

#### Change log
- This PR does not include user-facing changes and doesn't require a CHANGELOG update

### DIFF
--- a/app/domain/repos/conjur_ca.rb
+++ b/app/domain/repos/conjur_ca.rb
@@ -14,12 +14,18 @@ module Repos
     #
     def self.ca(resource_id)
       ca_info = ::Conjur::CaInfo.new(resource_id)
-      stored_cert = Resource[ca_info.cert_id].last_secret.value
-      stored_key = Resource[ca_info.key_id].last_secret.value
+      ca_cert_id = ca_info.cert_id
+      ca_key_id = ca_info.key_id
+      ca_secrets = fetch_required_secrets.(resource_ids: [ca_cert_id, ca_key_id])
+      stored_cert = ca_secrets[ca_cert_id]
+      stored_key = ca_secrets[ca_key_id]
       ca_cert = OpenSSL::X509::Certificate.new(stored_cert)
       ca_key = OpenSSL::PKey::RSA.new(stored_key)
       ::Util::OpenSsl::CA.new(ca_cert, ca_key)
     end
 
+    def self.fetch_required_secrets
+      @fetch_required_secrets ||= ::Conjur::FetchRequiredSecrets.new
+    end
   end
 end


app/domain/errors.rb
```diff
@@ -306,6 +306,16 @@ module AuthnK8s
         msg:  "Invalid Kubernetes host id: {0}. Must end with <namespace>/<resource_type>/<resource_id>",
         code: "CONJ00048E"
       )
+
+      CaResourceMissing = ::Util::TrackableErrorClass.new(
+        msg:  "Resource {0-resource-id} doesn't have saved resource for cert_id or key_id ",
```
uCatu, line 311: Do we have validation for log messages from Docs or is it not needed here?

app/domain/errors.rb
```diff
@@ -306,6 +306,16 @@ module AuthnK8s
...
+        code: "CONJ00063E"
```
uCatu, line 312: Those error codes need to be documeneted in a KB/Troubleshooting guide?

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,28 @@ def self.create(resource_id)
     #
     def self.ca(resource_id)
       ca_info = ::Conjur::CaInfo.new(resource_id)
-      stored_cert = Resource[ca_info.cert_id].last_secret.value
-      stored_key = Resource[ca_info.key_id].last_secret.value
+      stored_cert, stored_key = fetch_ca_secrets(ca_info, resource_id)
```
uCatu, line 17: What is the reasoning for using oneliner?

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,28 @@ def self.create(resource_id)
...
       ca_cert = OpenSSL::X509::Certificate.new(stored_cert)
       ca_key = OpenSSL::PKey::RSA.new(stored_key)
       ::Util::OpenSsl::CA.new(ca_cert, ca_key)
     end
 
+    class << self
+      private
+
+      def self.fetch_secret
+        @fetch_secret ||= ::Conjur::FetchRequiredSecrets.new
```
orenbm, line 27: let's give this variable the same name as the class it represents - `@fetch_required_secrets`

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,28 @@ def self.create(resource_id)
...
+      end
+
+      def self.fetch_ca_secrets(ca_info, resource_id)
+        begin
+          fetch_secret.(resource_ids: [ca_info.cert_id, ca_info.key_id])
+        rescue Errors::Conjur::RequiredResourceMissing
+          raise Errors::Authentication::AuthnK8s::CaResourceMissing, resource_id
```
orenbm, line 34: what is `resource_id` in this context? what will be printed to the log?

orenbm: Also - which log message will be the most helpful for the user?

orenbm: which message will give them the best pin-pointed action item?

app/domain/repos/conjur_ca.rb
orenbm, line 17: i think this is quite risky, as it assumes the order of the retrieval. what happens if someone will change the behaviour of `FetchRequiredSecrets` so that it retrieves the secrets in parallel? In this case the first secret in the array may not be `stored_cert`. How can you overcome this?

app/domain/errors.rb
uCatu: From where are we getting the `resource-id`?

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,28 @@ def self.create(resource_id)
...
       ca_info = ::Conjur::CaInfo.new(resource_id)
-      stored_cert = Resource[ca_info.cert_id].last_secret.value
-      stored_key = Resource[ca_info.key_id].last_secret.value
+      stored_cert, stored_key = fetch_ca_secrets(ca_info, resource_id)
       ca_cert = OpenSSL::X509::Certificate.new(stored_cert)
       ca_key = OpenSSL::PKey::RSA.new(stored_key)
       ::Util::OpenSsl::CA.new(ca_cert, ca_key)
     end
 
+    class << self
```
uCatu, line 23: Are we sure we want to use this architacutre of class + private? Is it TDD - can we write tests?

app/domain/repos/conjur_ca.rb
liavyona: Changed

app/domain/repos/conjur_ca.rb
liavyona: Refactor exceptions messages

app/domain/repos/conjur_ca.rb
liavyona: Changed

app/domain/errors.rb
liavyona: From the creation of the exception

app/domain/errors.rb
liavyona: @orenbm ?

app/domain/repos/conjur_ca.rb
liavyona: Changed

app/domain/repos/conjur_ca.rb
liavyona: Changed

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,26 @@ def self.create(resource_id)
...
       ca_key = OpenSSL::PKey::RSA.new(stored_key)
       ::Util::OpenSsl::CA.new(ca_cert, ca_key)
     end
 
+    def self.fetch_required_secrets
+      @fetch_required_secrets ||= ::Conjur::FetchRequiredSecrets.new
+    end
+
+    def self.fetch_ca_resource(ca_resource_id)
+      begin
```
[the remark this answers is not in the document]
liavyona:
```    
def self.fetch_ca_resource(ca_resource_id)
      begin
        fetch_required_secrets.(resource_ids: [ca_resource_id])[ca_resource_id]
      rescue Errors::Conjur::RequiredResourceMissing
        raise Errors::Authentication::AuthnK8s::CaResourceMissing, ca_resource_id
      rescue Errors::Conjur::RequiredSecretMissing
        raise Errors::Authentication::AuthnK8s::CaResourceValueMissing, ca_resource_id
      end
 end
```
@orenbm  When I delete the begin it has errors with the indentation of rescue (and to fix this they need to be in the same indent as ```def``` which basically won't work). What's the right way?

app/domain/errors.rb
orenbm: not at the moment. there is an effort led by @netacoral but i am not sure that it has started.

app/domain/repos/conjur_ca.rb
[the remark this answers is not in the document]
orenbm: this should do it:
```
def self.fetch_ca_resource(ca_resource_id)
  fetch_required_secrets.(resource_ids: [ca_resource_id])[ca_resource_id]
rescue Errors::Conjur::RequiredResourceMissing
  raise Errors::Authentication::AuthnK8s::CaResourceMissing, ca_resource_id
rescue Errors::Conjur::RequiredSecretMissing
  raise Errors::Authentication::AuthnK8s::CaResourceValueMissing, ca_resource_id
 end
```

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,26 @@ def self.create(resource_id)
     #
     def self.ca(resource_id)
       ca_info = ::Conjur::CaInfo.new(resource_id)
-      stored_cert = Resource[ca_info.cert_id].last_secret.value
-      stored_key = Resource[ca_info.key_id].last_secret.value
+      stored_cert = fetch_ca_resource(ca_info.cert_id)
+      stored_key = fetch_ca_resource(ca_info.key_id)
```
orenbm, line 18: this will make 2 calls to the DB instead of 1 so it's not the ideal solution. I would retrieve them together and then take each var from the retrieved list. you can see how we did it in `FetchAuthenticatorSecrets`.

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,26 @@ def self.create(resource_id)
...
+      @fetch_required_secrets ||= ::Conjur::FetchRequiredSecrets.new
+    end
+
+    def self.fetch_ca_resource(ca_resource_id)
+      begin
+        fetch_required_secrets.(resource_ids: [ca_resource_id])[ca_resource_id]
+      rescue Errors::Conjur::RequiredResourceMissing
+        raise Errors::Authentication::AuthnK8s::CaResourceMissing, ca_resource_id
+      rescue Errors::Conjur::RequiredSecretMissing
+        raise Errors::Authentication::AuthnK8s::CaResourceValueMissing, ca_resource_id
```
orenbm, line 34: do you see value in raising the error `CaResourceValueMissing` instead of the error `RequiredResourceMissing`? what value does the user get from this message that they don't get from the original?

orenbm:
@liavyona you will also need to rebase on master

liavyona:
> Beautiful!
> 
> Care to fix your git history?

Done

liavyona:
> @liavyona you will also need to rebase on master

Done

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,18 @@ def self.create(resource_id)
     #
     def self.ca(resource_id)
       ca_info = ::Conjur::CaInfo.new(resource_id)
-      stored_cert = Resource[ca_info.cert_id].last_secret.value
-      stored_key = Resource[ca_info.key_id].last_secret.value
+      ca_cert_id = ca_info.cert_id
+      ca_key_id = ca_info.key_id
+      ca_secrets = fetch_required_secrets.(resource_ids: [ca_cert_id, ca_key_id])
```
uCatu, line 19:
```suggestion
      ca_secrets = fetch_required_secrets.(resource_ids: [ca_info.cert_id, ca_info.key_id])
```

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,18 @@ def self.create(resource_id)
...
+      stored_cert = ca_secrets[ca_cert_id]
```
uCatu, line 20:
```suggestion
      stored_cert = ca_secrets[ca_info.cert_id]
```

app/domain/repos/conjur_ca.rb
```diff
@@ -14,12 +14,18 @@ def self.create(resource_id)
...
+      stored_key = ca_secrets[ca_key_id]
```
uCatu, line 21:
```suggestion
      stored_key = ca_secrets[ca_info.key_id]
```

Issue #1689, opened 2020-07-15
`TRUSTED_PROXIES` environment variable is not honored in v1.6.0+


micahlee:
### Description of the Problem
Prior to v1.6.0, the `TRUSTED_PROXIES` environment variable for Conjur was used to inform Rack and Rails which IP addresses are "trusted" when processing the `X-Forwarded-For` HTTP header and the TCP remote IP. The first IP address that is not in the trusted proxies list becomes the `request.ip` used by Rack and Rails.

This was accomplished in versions before v1.6.0 by patching the `Rack::Request#trusted_proxy?` method in the `conjur-rack` gem: https://github.com/cyberark/conjur-rack/blob/master/lib/conjur/rack.rb

However, in version v1.6.0, specifically in upgrading our version of Rails from 4 to 5 (which also upgraded Rack), this method in Rack no longer exists at this location. It was moved to another `Rack::Request::Helpers` module. So this patch no longer works as expected, and only the default Rails IP safelist is used: https://github.com/rack/rack/blob/master/lib/rack/request.rb#L19

### Expected Behavior

To align with the behavior pre-Rails 5, if `TRUSTED_PROXIES` is set in the environment, then only this list of IP address or IP address ranges (in CIDR format) are considered trusted by Rack and Rails for `request.ip`. If the `TRUSTED_PROXIES` is not set, it falls back to the default Rails behavior.

PR #1690, opened 2020-07-15
Restore `TRUSTED_PROXIES` behavior for Rails 5

micahlee:
### What does this PR do?

This PR re-implements the `TRUSTED_PROXIES` Rack IP filter for Rails 5. This replaces the implementation previously provided by the `conjur-rack` gem for Rails 4.

This PR also adds rspec tests that verify the expected `request.ip` with the given scenarios for `REMOTE_ADDR`, `TRUSTED_PROXIES`, and `X-Forwarded-For`.

Connected to #1689 

#### Change log

#### Test coverage

#### Documentation

#### TODO:
- [x] Cleanup fix code
- [x] Add tests for the `request.ip` behavior
- [x] Create issue/PR to remove monkey patch in `conjur-rack` for `trusted_proxy?` 
    Issue: https://github.com/cyberark/conjur-rack/issues/21
- [x] Update CHANGELOG

### DIFF
--- a/lib/conjur/trusted_proxy_filter.rb
+++ b/lib/conjur/trusted_proxy_filter.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Conjur
+  # TrustedProxyFilter wraps the default Rack IP filtering. This allows
+  # the trusted proxies to be configured using the environment variable,
+  # `TRUSTED_PROXIES`.
+  #
+  # The value of `TRUSTED_PROXIES` must be a comma-separated list of IP
+  # addresses or IP address ranges using CIDR notation.
+  #
+  # Example: TRUSTED_PROXIES=4.4.4.4,192.168.100.0/24
+  class TrustedProxyFilter
+    def initialize(wrapped_filter, env: ENV, options: { disable_cache: true })
+      @wrapped_filter = wrapped_filter
+      @env = env
+      @options = options
+      @cached_trusted_proxies = nil
+    end
+
+    def call(ip)
+      return @wrapped_filter.call(ip) unless trusted_proxies
+
+      trusted_proxies.any? { |cidr| cidr.include?(ip) }
+    end
+
+    def trusted_proxies
+      @cached_trusted_proxies || load_trusted_proxies_from_env
+    end
+
+    # Reek flags @env['TRUSTED_PROXIES'] as :reek:DuplicateMethodCall. Refactoring
+    # this would not enhance the readability or performance.
+    def load_trusted_proxies_from_env
+      return nil unless @env['TRUSTED_PROXIES']
+
+      proxy_ips = Set.new(@env['TRUSTED_PROXIES'].split(',') + ['127.0.0.1'])
+        .collect { |cidr| IPAddr.new(cidr.strip) }
+
+      @cached_trusted_proxies = proxy_ips unless @options[:disable_cache]
+      proxy_ips
+    end
+  end
+end
--- a/spec/request/request_ip_spec.rb
+++ b/spec/request/request_ip_spec.rb
@@ -1,0 +1,164 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+# These are integration tests that describe the expected client IP address
+# behavior for Conjur, considering the entire HTTP server pipeline and how it
+# handles the TCP remote IP address, the `TRUSTED_PROXIES` environment variable,
+# and the `X-Forwarded-For` HTTP header.
+#
+# These would generally be better suited as cucumber behavior tests rather than
+# RSpec tests. However, in this case we use RSpec for these integration tests for
+# two reasons:
+#
+# 1. Rspec tests are generally faster than Cucumber tests, so if we can test
+#    this appropriately in RSpec alone, we can keep the CI execution time lower.
+#
+# 2. Our cucumber tests currently can't test this as thoroughly as we can in RSpec,
+#    because of the way we are able to manipulate the environment and the TCP
+#    remote IP with RSpec and the Rails "request" test type. To do this with
+#    Cucumber would require greater control of the Docker network and containers
+#    for the Conjur cucumber tests than we have now.
+RSpec.describe "request IP address determination", type: :request do
+  def request_env(remote_addr)
+    {
+      # We can't modify the access token middleware to add an exception to our
+      # test route, so we need to create an access token for our request to use.
+      'HTTP_AUTHORIZATION' => access_token_for('admin'),
+      'REMOTE_ADDR' => remote_addr
+    }
+  end
+
+  def request_ip(remote_addr:, x_forwarded_for: nil, trusted_proxies: nil)
+    ENV['TRUSTED_PROXIES'] = trusted_proxies if trusted_proxies
+  
+    headers = {}
+    headers['X-Forwarded-For'] = x_forwarded_for if x_forwarded_for
+  
+    get '/whoami', env: request_env(remote_addr), headers: headers
+  
+    JSON.parse(response.body)['client_ip']
+  end
+
+  # --------------------------------------------------------------------
+  # Test Scenarios
+  # --------------------------------------------------------------------
+
+  # Without any other configuration in play, we expect to get the remote
+  # TCP connection IP address as the request IP address.
+  it 'returns the remote_addr with no additional config' do
+    expect(request_ip(remote_addr: '44.0.0.1')).to eq('44.0.0.1')
+  end
+
+  it 'ignores the XFF header when the remote addr is untrusted' do
+    expect(
+      request_ip(
+        remote_addr: '44.0.0.1',
+        x_forwarded_for: '3.3.3.3'
+      )
+    ).to eq('44.0.0.1')
+  end
+
+  # By default "non-routable" IP addresses are trusted (according to this
+  # regular expression: https://github.com/rack/rack/blob/master/lib/rack/request.rb#L19)
+  #
+  # `127.0.0.1` is important as the address of the nginx proxy when used in DAP
+  it 'trusts the loopback address by default to provide XFF' do
+    expect(
+      request_ip(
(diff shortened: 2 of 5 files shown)


spec/request/request_ip_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+      # Loopback is always trusted, even if not explicity provided in
+      # TRUSTED_PROXIES, so we accept the forwarded IP address.
+      context 'when X-Forwarded-For is set' do
+        let(:x_forwarded_for) { '3.3.3.3' }
+        let(:expected_request_ip) { '3.3.3.3' }
+        include_examples "expected request IP address"
+      end
+    end
+
+    # The following spec exercise how variations on the X-Forwarded-For header
```
Tovli, line 113: Hi,
I don't see a test that simulate our issue of having duplicate headers.
Currently we get from LB `X-Forwarded-For` but nginx is adding `X-Forwarded-For`
Need to either [fix this issue in nginx](https://github.com/conjurinc/appliance/pull/1300) or test for it so we can better simulate the production use

spec/request/request_ip_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+  end
+
+  # The value to user for the TRUSTED_PROXIES environment variable
+  let(:trusted_proxies) { nil }
+
+  # The value to use for the X-Forwarded-For HTTP header
+  let(:x_forwarded_for) { nil }
+
+  # --------------------------------------------------------------------
+  # Test Scenarios
```
Tovli, line 69: I'm missing a test with CIDR range in the trusted proxies

spec/request/request_ip_spec.rb
micahlee: @Tovli, I am not trying to test for that yet, nor address that in this PR.

This PR is to resolve a regression found in Conjur from Rails 5 while work on the Client IP Sourcing doc here: https://github.com/cyberark/dap-wiki/pull/85.

I haven't gotten to the DAP portion in that document yet, and I expect this to come up then.

As a side note, I do agree that the issue should probably be fixed in DAP's nginx reverse proxy though.

spec/request/request_ip_spec.rb
micahlee: Thanks, I've added tests to include trusted IP range examples using CIDR notation.

lib/conjur/trusted_proxy_filter.rb
```diff
@@ -0,0 +1,44 @@
...
+      proxy_ips = Set.new(env_trusted_proxies.split(',') + ['127.0.0.1'])
+        .collect { |cidr| IPAddr.new(cidr.strip) }
+
+      @cached_trusted_proxies = proxy_ips unless @options[:disable_cache]
+      proxy_ips
+    end
+
+    def env_trusted_proxies
+      @env['TRUSTED_PROXIES']
+    end
```
jonahx, line 42: I think I'd just inline this.

micahlee: Yeah... I wish there was a better way to configure code climate for this. Inlining was my default, but code climate didn't like that "`@env['TRUSTED_PROXIES']` is called twice."

I'll inline again and add the exception comment.

micahlee:
@jonahx 

> I'd still like to move unit testing the object directly for the matrix tests, per our previous convo. Were you planning to do that in a different PR?

I'm torn on this. On one hand, I agree that the layers of unit tests would be the more proper way to do this. On the other hand, I do still have concerns about the fidelity of the tests, considering how mutable the ActiveDispatch::Request is as it passes through the Rack/Rails pipeline. I really do want the subject of these tests to be that request pipeline, rather than only the `TrustedProxiesFilter`. Or to assume too much about what the final state of `ActiveDispatch::Request` will be after the pipeline executes.

I _am_ willing to explore that refactor, but I do think that's better suited as a follow up PR. The ROI at this moment of that refactor seems low compared to the ROI of finishing the client IP sourcing document (which depends on this regression fix) on these points:

A) These tests do clearly communicate what the expected outcomes are for the inputs we care about.
B) These tests run fast now, so a refactor wouldn't greatly improve test cycle times.
C) We do get a high degree of assurance from these tests regarding the `request.ip` outcome, given the full Rack request pipeline is exercised. 

I created the issue here: https://github.com/cyberark/conjur/issues/1696. If there is an epic for the overall spec refactor you discussed during our team meeting Wednesday. I think this would be very appropriate in that epic.

> Though I really don't like the rspec convention whereby a function call with 3 parameters becomes 3 let statements followed by an include_examples to "call the function"

I agree with you, and I'd like to hear your suggestions for how to do it better.

* I considered inlining everything with something like:
    ```ruby
    request_env = { 'HTTP_AUTHORIZATION' => token_auth_header, 'REMOTE_ADDR' => '4.4.4.4' }
    headers = { 'X-Forwarded-For' => '3.3.3.3,5.5.5.2,5.5.5.3'
    ENV['TRUSTED_PROXIES'] = '4.4.4.4,5.5.5.0/24'
    get '/request_ip', env: request_env, headers: headers
    expect(body_hash['ip']).to eq('3.3.3.3')
    ```

  But this felt overly cumbersome, not very DRY, and made it hard to read the values that mattered.

* I considered writing my own helper method in vanilla Ruby to produce the result. Something akin to:
    ```ruby
    expect(request_ip(
      remote_addr: '4.4.4.4', 
      x_forwarded_for: '3.3.3.3,5.5.5.2,5.5.5.3',
      trusted_proxies: '4.4.4.4,5.5.5.0/24'
      )
    ).to eq('3.3.3.3')
    ```
  But this didn't feel remarkably different in outcome (or LOC) from the `let`s and `include_examples`, or in the spirit of `rspec`.

* Some of the `let`s are unnecessary since they are inherited from the containing context, so they could be removed. But I thought it was easier to read with each test's context clearly stated.

What do you think? Do any of those stand out as a clear improvement over the current code? Do you have any other suggestions I haven't considered yet?

alexkalish:
@micahlee: Are you planning on covering the use `TRUSTED_PROXIES` (including `evoke`, etc) in depth in your IP related `dap-wiki` PR?  If not, can you file a separate issue to do so?  I want ensure that we document this feature well in `dap-wiki` so that it can get into the docs for 11.7.  Thanks!

micahlee:
@alexkalish 
> Are you planning on covering the use TRUSTED_PROXIES (including evoke, etc) in depth in your IP related dap-wiki PR?

I'm not sure if this PR is where you want this discussion, rather than on the dap-wiki PR, but [the existing draft](https://github.com/cyberark/dap-wiki/pull/85) includes this as a reference discussion. [Jason's PR](https://github.com/cyberark/dap-wiki/pull/87) includes this in a tutorial/guide format (although I now see he's removed Conjur OSS from his doc). 

It would be helpful to know more specifically what gaps there are that need to be filled for allowing this to be officially documented. 

You can see my current "to do" list on[ the PR description here](https://github.com/cyberark/dap-wiki/pull/85). If Jason does not intend to cover OSS configuration of `TRUSTED_PROXIES` in his tutorial. If that is the case, I can create an issue for that.

alexkalish:
@micahlee: Actually, I forgot that @jvanderhoof was already working on this PR: https://github.com/cyberark/dap-wiki/pull/87.  Maybe that could serve to document this feature.  Could you take a look?

jonahx:
@micahlee 
- Your plan re: the refactor or lack thereof is ok with me.  It's good food for thought for future tests either way and eventually it would be nice to get standards out of it, but now is not the time for that.
- I prefer both of your alternatives to the `let` versions.  And I think "inherited" lets are especially problematic, and would always take repetition over that.
- I like the version with your bespoke helper most.  I think it not being "rspec"-ish is a non-issue.  In this case it seems to be a badge of honor.

micahlee:
@jonahx , this is still ready for your re-review when you're able. I added one more fixup commit today to re-used a shared method to create the test access token that I added in https://github.com/cyberark/conjur/pull/1698

I've left the fixups as separate commits to make the updates easier to review. Once you've had a chance to look at them, I'll squash them back down.

Thanks!

spec/request/request_ip_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+        remote_addr: '44.0.0.1',
+        x_forwarded_for: '3.3.3.3'
+      )
+    ).to eq('44.0.0.1')
+  end
+
+  # By default "non-routable" IP addresses are trusted (according to this
+  # regular expression: https://github.com/rack/rack/blob/master/lib/rack/request.rb#L19)
+  #
+  # `127.0.0.1` is important as the address of the nginx proxy when used in DAP
```
jonahx, line 49: Love these contextual comments.

spec/request/request_ip_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+  # --------------------------------------------------------------------
+
+  # Without any other configuration in play, we expect to get the remote
+  # TCP connection IP address as the request IP address.
+  it 'returns the remote_addr with no additional config' do
+    expect(request_ip(remote_addr: '44.0.0.1')).to eq('44.0.0.1')
+  end
+
+  # When X-Forwarded-For is set from an untrusted remote IP, we ignore it
+  it 'ignores the XFF header when the remote addr is untrusted' do
```
jonahx, line 37: The comment here seems to just repeat the test description, so can be removed.

spec/request/request_ip_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+        remote_addr: '127.0.0.1',
+        x_forwarded_for: '3.3.3.3'
+      )
+    ).to eq('3.3.3.3')
+  end
+
+  # If TRUSTED proxies are set, and are remote IP address is not trusted
+  # we expect to get the remote IP
+  it "doesn't trust the remote_addr if not included in TRUSTED_PROXIES" do
+    expect(
```
jonahx, line 62: This comment too feels like repetition of the test case.

PR #1693, opened 2020-07-15
Process X-Forwarded-For Request Header Command Class

eranha:
### What does this PR do?
Adding a new [CommandClass](https://github.com/jonahx/command_class) to the Authentication::Security namespace according to the [design](https://github.com/conjurinc/appliance/blob/master/design/cidr_proxy/designs/ldd-fetch-request-proxies-list.md).

As part of the CIDR support for hosts [epic](https://github.com/conjurinc/appliance/issues/1273), this PR includes a [CommandClass](https://github.com/jonahx/command_class) that takes part in the proxy validation flow.
`ProcessRequestProxies` is a [CommandClass](https://github.com/jonahx/command_class), with the responsibility of performing input validation in the XFF value, and processing the IP address list for further validation.

Read the [design](https://github.com/conjurinc/appliance/blob/master/design/cidr_proxy/designs/ldd-fetch-request-proxies-list.md).
Bear in mind that this [CommandClass](https://github.com/jonahx/command_class), **is pending** an input validation dependency that should be developed.

Connected to #[Issue](https://github.com/conjurinc/AAM/issues/255).

#### Test coverage
- [x ] This PR includes new unit tests to go with the code changes.

#### Documentation
- [x ] This PR does not require updating any documentation.


app/domain/authentication/security/process_request_proxies.rb
```diff
@@ -0,0 +1,58 @@
...
+        process_xff_header
+      end
+
+      private
+
+      Struct.new('RemoteAddress', :client_ip, :proxy_list)
+
+      def validate_request
+        raise ArgumentError, '@request' if @request.nil?
+        raise ArgumentError, '@request.remote_addr' if nil_or_empty @request.remote_addr
```
orenbm, line 23: don't we want to raise our own error?

app/domain/authentication/security/process_request_proxies.rb
```diff
@@ -0,0 +1,58 @@
...
+    ) do
+
+      def call
+        validate_request
+        process_xff_header
+      end
+
+      private
+
+      Struct.new('RemoteAddress', :client_ip, :proxy_list)
```
orenbm, line 19: this struct should be a [Value Object](https://revs.runtime-revolution.com/value-objects-in-ruby-on-rails-9df64bc8db34)

app/domain/authentication/security/process_request_proxies.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      def validate_request
+        raise ArgumentError, '@request' if @request.nil?
+        raise ArgumentError, '@request.remote_addr' if nil_or_empty @request.remote_addr
+      end
+
+      def process_xff_header
+        remote_addr = @request.remote_addr
```
orenbm, line 27: this can be done also by `def_delegators`

app/domain/authentication/security/process_request_proxies.rb
```diff
@@ -0,0 +1,58 @@
...
+        xff_header = @request.headers['X-Forwarded-For']
```
orenbm, line 28: do we want to set `'X-Forwarded-For'` as a class const?

app/domain/authentication/security/process_request_proxies.rb
```diff
@@ -0,0 +1,58 @@
...
+        xff_header_ip_list = (nil_or_empty xff_header) ? [] : xff_header.to_s.split(/, /)
+        empty_ip_list = xff_header_ip_list.empty?
```
orenbm, line 30: this is redundant IMO. you can later call `xff_header_ip_list.empty?` in the `if` clause and it will be clearer.

app/domain/authentication/security/process_request_proxies.rb
```diff
@@ -0,0 +1,58 @@
...
+        ip_list_length = xff_header_ip_list.length
+        single_ip_in_list = ip_list_length == 1
+        multi_ips_in_list = ip_list_length > 1
```
orenbm, line 33: this is redundant as well. you can just query `ip_list_length` in the `if`s below

app/domain/authentication/security/process_request_proxies.rb
orenbm: how would you call the class that holds this value object?

app/domain/authentication/security/process_request_proxies.rb
eranha: This is not likely to happen. I consider it a `RuntimeExcpeiton` from which the user cannot recover.

app/domain/authentication/security/process_request_proxies.rb
eranha: > this struct should be a [Value Object](https://revs.runtime-revolution.com/value-objects-in-ruby-on-rails-9df64bc8db34)
Do you mean to create a separate class in a separate file for that? 


> how would you call the class that holds this value object?
It contains info about the request so maybe RequestIpInfo?

orenbm: yes, the value object should be in a separate class.
so if it is called `RequestIpInfo` then this command class should be `ExtractRequestIpInfo` as it is more informative than `ProcessRequestHeader.

app/domain/authentication/security/process_request_proxies.rb
eranha: Actually it was at the beginning

app/domain/authentication/security/process_request_proxies.rb
orenbm: i am not a fan of this check if the scenario doesn't really exist.

orenbm: and if it _does_ happen i would rather have a meaningful log message than an `ArgumentError`

app/domain/authentication/security/extract_reques_remote_addr_info.rb
```diff
@@ -0,0 +1,51 @@
+require 'command_class'
+
+module Authentication
+  module Security
+    ExtractRequestRemoteAddrInfo = CommandClass.new(
+      dependencies: {
+        # InputValidation
```
orenbm, line 7: what's this? when will this happen?

orenbm: also - which input are we validating?

app/domain/authentication/security/extract_request_remote_addr_info.rb
```diff
@@ -0,0 +1,51 @@
+require 'command_class'
+
+module Authentication
+  module Security
+    ExtractRequestRemoteAddrInfo = CommandClass.new(
```
orenbm, line 5: where will this class be used? 

I am not sure that this needs to be a command class, but rather you can take the only line in this call method `extract_request_remote_addr_info` and just implement it in the calling class

app/domain/authentication/security/extract_request_remote_addr_info.rb
```diff
@@ -0,0 +1,51 @@
...
+      dependencies: {
+        # InputValidation
```
orenbm, line 7: what's the dependency here? when will it happen?

app/domain/authentication/security/extract_request_remote_addr_info.rb
```diff
@@ -0,0 +1,51 @@
...
+      extend Forwardable
+      def_delegators :@request, :client_ip, :headers
+
+      def call
+        extract_request_remote_addr_info
+      end
+
+      private
+
+      X_FORWARDED_FOR = 'X-Forwarded-For'
```
orenbm, line 21: i would change to `X_FORWARDED_FOR_HEADER_NAME` to clarify what it is.

app/domain/authentication/security/request_remote_addr_info.rb
```diff
@@ -0,0 +1,11 @@
...
+  module Security
+    class RequestRemoteAddrInfo
+      attr_reader :client_ip, :xff_ip_list
+
+      def initialize(client_ip, xff_ip_list)
+        @client_ip, @xff_ip_list = client_ip, xff_ip_list
+      end
+    end
+  end
+end
```
orenbm, line 11: missing new line

app/domain/authentication/security/request_remote_addr_info.rb
```diff
@@ -0,0 +1,11 @@
+module Authentication
+  module Security
+    class RequestRemoteAddrInfo
+      attr_reader :client_ip, :xff_ip_list
+
+      def initialize(client_ip, xff_ip_list)
+        @client_ip, @xff_ip_list = client_ip, xff_ip_list
```
orenbm, line 7: can you split into 2 lines? it is more readable

spec/app/domain/authentication/security/extract_reques_remote_addr_info_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+    '2.2.2.2'
+  end
+
+  let(:proxy_2) { 'proxy_2' }
+
+  def proxy_2
+    '3.3.3.3'
+  end
+
+
```
orenbm, line 24: remove these 2 lines

spec/app/domain/authentication/security/extract_reques_remote_addr_info_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+  let(:request_with_empty_xff) { 'request_with_empty_xff' }
+
+  def request_with_empty_xff
+    double('request_with_empty_xff').tap do |request|
+      allow(request).to receive(:headers).and_return(request)
+      allow(request).to receive(:[]).and_return(nil)
+      allow(request).to receive(:client_ip).and_return(remote_addr)
+
```
orenbm, line 33: redundant empty line

spec/app/domain/authentication/security/extract_reques_remote_addr_info_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  let(:request_with_single_entry_xff) { 'request_with_single_entry_xff' }
+
+  def request_with_single_entry_xff
+    double('request_with_single_entry_xff').tap do |request|
+      allow(request).to receive(:headers).and_return(request)
+      allow(request).to receive(:[]).and_return(proxy_1)
+      allow(request).to receive(:client_ip).and_return(remote_addr)
+
```
orenbm, line 44: here as well

spec/app/domain/authentication/security/extract_reques_remote_addr_info_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+    double('request_with_multi_entries_xff').tap do |request|
+      allow(request).to receive(:headers).and_return(request)
+      allow(request).to receive(:[]).and_return(proxy_1 + ', ' + proxy_2, + ', ' + proxy_2)
+      allow(request).to receive(:client_ip).and_return(remote_addr)
+    end
+  end
+
+  context "A request with an empty X-Forwarded-For header" do
+    subject do
+      Authentication::Security::ExtractRequestRemoteAddrInfo.new(
```
orenbm, line 60: redundant parentheses after `new`

orenbm: please fix in the rest of the file

app/domain/authentication/security/extract_reques_remote_addr_info.rb
eranha: It was planned that Moti will fo that, now I don't know. It should validate the IPs.

app/domain/authentication/security/extract_request_remote_addr_info.rb
eranha: Another `CommandClass` `ValidateProxy` will use it.

app/domain/authentication/security/extract_request_remote_addr_info.rb
eranha: Duplicate comment.

app/domain/authentication/security/extract_request_remote_addr_info.rb
orenbm: i'll approve this PR to not block it but once we create the PR that consumes this class i would like us to challenge its existence. We may be better off implementing this as a method in `ValidateProxy`

app/domain/authentication/security/extract_request_remote_addr_info.rb
```diff
@@ -0,0 +1,51 @@
...
+  module Security
+    ExtractRequestRemoteAddrInfo = CommandClass.new(
+      dependencies: {
+        # InputValidation
+      },
+      inputs: [:request]
+    ) do
+
+      extend Forwardable
+      def_delegators :@request, :client_ip, :headers
```
nessiLahav, line 13: you should use here `remote_addr` instead of `client_ip`

app/domain/authentication/security/extract_request_remote_addr_info.rb
nessiLahav: I think the name of the command class is confusing because the `remote_addr` is the IP in the TCP packet,
but here we are performing logic both on the XFF and the remote_addr to return a client ip and a proxy list. which should be validated later

app/domain/authentication/security/extract_request_remote_addr_info.rb
```diff
@@ -0,0 +1,51 @@
...
+            []
+          )
+        elsif xff_ip_list_length == 1
+          RequestRemoteAddrInfo.new(
+            xff_header_ip_list.first,
+            [client_ip]
+          )
+        elsif xff_ip_list_length > 1
+          RequestRemoteAddrInfo.new(
+            xff_header_ip_list.shift,
```
nessiLahav, line 40: You should delete all occurrences of client IP in the list

app/domain/authentication/security/extract_request_remote_addr_info.rb
```diff
@@ -0,0 +1,51 @@
...
+            xff_header_ip_list.uniq
```
nessiLahav, line 41: You should add `remote_addr` to the proxy list , please go over again on the design

nessiLahav:
@eranha please add some logs

spec/app/domain/authentication/security/extract_reques_remote_addr_info_spec.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Authentication::Security::ExtractRequestRemoteAddrInfo do
```
nessiLahav, line 5: Add tests of :
1. duplication proxies ip address in the XFF
2. duplication of the first IP (the client) in XFF

app/domain/authentication/security/extract_request_remote_addr_info.rb
eranha: it was remote_addr when I added the delegate I switched by mistake

spec/app/domain/authentication/security/extract_reques_remote_addr_info_spec.rb
eranha: There is a test for duplication of IPs.

app/domain/authentication/security/extract_request_origin_ip_addr_info.rb
```diff
@@ -0,0 +1,54 @@
...
+        extract_request_origin_addr_info
+      end
+
+      private
+
+      X_FORWARDED_FOR_HEADER_NAME = 'X-Forwarded-For'
+
+      def extract_request_origin_addr_info
+        xff_header = headers[X_FORWARDED_FOR_HEADER_NAME]
+        xff_header_ip_list = (nil_or_empty? xff_header) ? [] : xff_header.to_s.split(/, /).uniq
```
nessiLahav, line 25: @eranha probably the whole feature will be deleted, but still i think there is a possible bug here,
If you first perform `.uniq` and then extract the first ip from the list, there is a possible bug where the first ip will be deleted before the extraction

Issue #1681, opened 2020-07-13
CIDR support proxy - validate trusted proxies impl


nessiLahav:
Implement `ValidateTrustedProxies` command class according to the design

DoD:

- [x]  we have full UT coverage

Tovli:
We found out that we already implemented similar feature using evoke proxy to set the trusted proxies
This will be closed

PR #1694, opened 2020-07-16
Cidr proxy validate trusted proxies

nessiLahav:
### What does this PR do?
Implements  `ValidateTrustedProxies` CommandClass  according to the [design](https://github.com/conjurinc/appliance/blob/master/design/cidr_proxy/designs/ldd-validate-permitted-proxies-list.md
).

Connected to #1681 

#### Test coverage
- [x ] This PR includes unit tests.
Note: integration tests will be developed in another PR.

#### Documentation
Document will be updated in the end of the feature.

### DIFF
--- a/app/domain/conjur/fetch_trusted_proxies.rb
+++ b/app/domain/conjur/fetch_trusted_proxies.rb
@@ -23,7 +23,6 @@ module Conjur
       trusted_proxies_list
     rescue => e
       raise Errors::Conjur::TrustedProxiesFetchFailed.new(
-        host_id,
         e.inspect
       )
     end
@@ -39,7 +38,7 @@ module Conjur
     def fetch_trusted_proxies_list
       @logger.debug(LogMessages::Conjur::FetchingTrustedProxies.new(host_id))
       @trusted_proxies_list = role.restricted_to || []
-      @logger.debug(LogMessages::Conjur::FetchingTrustedProxies.new(@trusted_proxies_list.length))
+      @logger.debug(LogMessages::Conjur::FetchedTrustedProxies.new(@trusted_proxies_list.length))
     end
 
     def delete_duplications_in_list
--- a/app/domain/conjur/validate_trusted_proxies.rb
+++ b/app/domain/conjur/validate_trusted_proxies.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'command_class'
+
+module Conjur
+
+  ValidateTrustedProxies ||= CommandClass.new(
+    dependencies: {
+      fetch_trusted_proxies:  ::Conjur::FetchTrustedProxies,
+      logger:                 Rails.logger
+    },
+    inputs:       %i(account proxies_list)
+  ) do
+
+    def call
+      fetch_trusted_proxies_list
+      validate_proxies
+    end
+
+    private
+
+    def fetch_trusted_proxies_list
+      @trusted_proxies_list ||= @fetch_trusted_proxies.(
+        account: @account
+      )
+    end
+
+    def validate_proxies
+      #TODO: validation logic
+    end
+  end
+end
+
--- a/app/domain/conjur/validate_trusted_proxies.rb
+++ b/app/domain/conjur/validate_trusted_proxies.rb
@@ -13,6 +13,7 @@ module Conjur
   ) do
 
     def call
+      return if @proxy_list.blank?
       fetch_trusted_proxies_list
       validate_proxies
     end
@@ -26,18 +27,18 @@ module Conjur
     end
 
     def validate_proxies
-      return if @trusted_proxies_list.blank? || @proxy_list.blank?
+      return if @trusted_proxies_list.blank?
 
       @logger.debug(LogMessages::Conjur::ValidatingProxyList.new(@proxy_list.map {|ip| ip.to_s}))
       @proxy_list.each do |ip_addr|
-        valid_proxy?(ip_addr)
+        valid_proxy(ip_addr)
       end
       @logger.debug(LogMessages::Conjur::ProxyListValidated.new)
     end
 
-    def valid_proxy?(ip_addr)
-      @trusted_proxies_list.each do |cidr|
-        return if cidr.include?(ip_addr)
+    def valid_proxy(ip_addr)
+      @trusted_proxies_list.each do |trusted_proxy|
+        return if trusted_proxy.include?(ip_addr)
       end
       raise Errors::Conjur::InvalidProxy, ip_addr.to_s
     end
--- a/app/domain/conjur/validate_trusted_proxies.rb
+++ b/app/domain/conjur/validate_trusted_proxies.rb
@@ -6,10 +6,10 @@ module Conjur
 
   ValidateTrustedProxies ||= CommandClass.new(
     dependencies: {
-      fetch_trusted_proxies:  ::Conjur::FetchTrustedProxies,
+      fetch_trusted_proxies:  ::Conjur::FetchTrustedProxies.new,
       logger:                 Rails.logger
     },
-    inputs:       %i(account proxies_list)
+    inputs:       %i(account proxy_list)
   ) do
 
     def call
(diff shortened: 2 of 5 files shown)


app/domain/conjur/fetch_trusted_proxies.rb
```diff
@@ -23,7 +23,6 @@ def call
       trusted_proxies_list
     rescue => e
       raise Errors::Conjur::TrustedProxiesFetchFailed.new(
```
orenbm, line 25: don't we want to log the host id? can't it help the customer?

app/domain/conjur/validate_trusted_proxies.rb
```diff
@@ -0,0 +1,46 @@
...
+    private
+
+    def fetch_trusted_proxies_list
+      @trusted_proxies_list ||= @fetch_trusted_proxies.(
+        account: @account
+      )
+    end
+
+    def validate_proxies
+      return if @trusted_proxies_list.blank? || @proxy_list.blank?
```
orenbm, line 29: i see that we return if `@proxy_list` is blank. in that case we might perform a redundant call to the DB. how about we move the `@proxy_list` emptiness check to the beginning of the class?

orenbm: Also - i remember @jonahx saying that we should avoid `Rails` methods and use only pure `Ruby` methods. `blank?` is a `Rails` method so it is better to use `@proxy_list.nil? || @proxy_list.empty?`.

@jonahx i can't find a reference as to why we should prefer ruby over rails. Can you elaborate?

app/domain/conjur/validate_trusted_proxies.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      @logger.debug(LogMessages::Conjur::ValidatingProxyList.new(@proxy_list.map {|ip| ip.to_s}))
+      @proxy_list.each do |ip_addr|
+        valid_proxy?(ip_addr)
+      end
+      @logger.debug(LogMessages::Conjur::ProxyListValidated.new)
+    end
+
+    def valid_proxy?(ip_addr)
```
orenbm, line 38: it seems from the method name that you return a boolean but you actually return nothing or raise on failure. `validate_proxy` is more fitting.

app/domain/conjur/validate_trusted_proxies.rb
```diff
@@ -0,0 +1,46 @@
...
+      @trusted_proxies_list.each do |cidr|
```
orenbm, line 39: quite funny that the array is `trusted_proxies_list` and each object inside it is `cidr`. can you align those? choose between (`trusted_proxies_list` and `trusted_proxy`) or (`cidr_list` and `cidr`).

orenbm: i would use the former

app/domain/conjur/validate_trusted_proxies.rb
```diff
@@ -0,0 +1,46 @@
...
+        return if cidr.include?(ip_addr)
+      end
+      raise Errors::Conjur::InvalidProxy, ip_addr.to_s
+    end
```
orenbm, line 43: each object in `@proxy_list` is an array?

app/domain/conjur/validate_trusted_proxies.rb
```diff
@@ -0,0 +1,46 @@
...
+    def fetch_trusted_proxies_list
+      @trusted_proxies_list ||= @fetch_trusted_proxies.(
+        account: @account
+      )
+    end
+
+    def validate_proxies
+      return if @trusted_proxies_list.blank? || @proxy_list.blank?
+
+      @logger.debug(LogMessages::Conjur::ValidatingProxyList.new(@proxy_list.map {|ip| ip.to_s}))
```
orenbm, line 31: how will this log look like? can you show me an example?

nessiLahav: `CONJ00043D Validating proxy list [1.1.1.1, 2.2.2.2]`

app/domain/conjur/validate_trusted_proxies.rb
nessiLahav: each object is an `Util::CIDR` class, see examples in the UT file

app/domain/conjur/fetch_trusted_proxies.rb
nessiLahav: i think no,
In most of the errors we already have this info, for example:
```
CONJ00039D Fetching trusted proxy list from host/conjur/settings/trusted_proxies
Authentication Error: #<Errors::Conjur::TrustedProxiesFetchFailed: CONJ00065E Failed to fetch trusted proxy list. Reason: '#<Errors::Authentication::Security::RoleNotFound: CONJ00007E 'cucumber:host:conjur/settings/trusted_proxies' not found>
```
So it will be strange to write it twice.

Also you can see it in the log flow, for example:
```
CONJ00039D Fetching trusted proxy list from host/conjur/settings/trusted_proxies
CONJ00040D 340 trusted proxies were fetched
CONJ00041D Deleting duplications from trusted proxy list
CONJ00042D Trusted proxy list contains 7 IP addresses
CONJ00043D Validating proxy list [1.1.1.1, 2.2.2.2]
Authentication Error: #<Errors::Conjur::InvalidProxy: CONJ00066E Invalid proxy '1.1.1.1'>

```

app/domain/conjur/validate_trusted_proxies.rb
nessiLahav: There are a lot of places in our code which we are using `blank?`

app/domain/errors.rb
```diff
@@ -25,6 +25,11 @@ module Conjur
       code: "CONJ00065E"
     )
 
+    InvalidProxy = ::Util::TrackableErrorClass.new(
+      msg:  "Invalid proxy '{0-proxy-ip-address}'",
```
nessiLahav, line 29: @shulifink can you review please?

app/domain/logs.rb
```diff
@@ -34,6 +34,15 @@ module Conjur
       code: "CONJ00042D"
     )
 
+    ValidatingProxyList = ::Util::TrackableLogMessageClass.new(
+      msg:  "Validating proxy list {0-proxy-list}",
```
nessiLahav, line 38: @shulifink can you review please?

app/domain/logs.rb
```diff
@@ -34,6 +34,15 @@ module Conjur
...
+      code: "CONJ00043D"
+    )
+
+    ProxyListValidated = ::Util::TrackableLogMessageClass.new(
+      msg:  "Proxy list validated",
```
nessiLahav, line 43: @shulifink can you review please?

spec/app/domain/conjur/validate_trusted_proxies_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+
+    allow(mocked_fetch_trusted_proxy_returns_3_ranges).to receive(:call)
+                                                              .with({:account => valid_account})
+                                                              .and_return(trusted_proxies_list_3_ranges)
+  end
+  #  ____  _   _  ____    ____  ____  ___  ____  ___
+  # (_  _)( )_( )( ___)  (_  _)( ___)/ __)(_  _)/ __)
+  #   )(   ) _ (  )__)     )(   )__) \__ \  )(  \__ \
+  #  (__) (_) (_)(____)   (__) (____)(___/ (__) (___/
+
```
nessiLahav, line 89: @eladkug can you review the following tests please ?

app/domain/errors.rb
shulifink: Fine

shulifink: Is this the same as the trusted proxy list? If so, do we want "Invalid trusted proxy '{0-proxy-ip-address}'"

app/domain/logs.rb
shulifink: Is this the same as the trusted proxy list?
If so, do we want "Validating trusted proxy list {0-proxy-list}"?

app/domain/logs.rb
shulifink: Fine.
Same comments as above

spec/app/domain/conjur/validate_trusted_proxies_spec.rb
shulifink: cool!

Issue #1697, opened 2020-07-17
Add `/whoami` API endpoint


micahlee:
When setting up network based authentication (with `restricted_to`) or debugging IP address issues with authentication and audit, it is helpful to observe how Conjur views your API request.

To make this more observable, we should add a new API endpoint `/whoami` to return the following properties of the request:

* `client_ip`: The rails `request.ip` attribute. This is used for network restrictions and audit. This can contain an unexpected value if load balancers or reverse proxies are not properly configured.

* `user_agent`: The rails `request.user_agent` attribute.

* `account`: The account attribute of the access token (if provided on the request, `null` otherwise).

* `login`: The login (username) attribute of the access token (if provided on the request, `null` otherwise).

* `token_issued_at`: The issued timestamp (`iat`) when the access token was created (if provided on the request, `null` otherwise).

DoD:
- [x] API endpoint exists
- [x] API endpoint is tested in CI
- [ ] API endpoint is documented

PR #1698, opened 2020-07-17
Add `/whoami` API endpoint

micahlee:
### What does this PR do?

This PR adds a new API endpoint:
```log
GET /whoami
```

This endpoint optionally accepts an `Authorization` header with a Conjur access token, and returns the following JSON structure:

```json
{
    "client_ip": "192.168.192.1",
    "user_agent": "PostmanRuntime/7.26.1",
    "account": "cucumber",
    "login": "admin",
    "token_issued_at": "2020-07-17T21:11:56.000+00:00"
}
```

When no access token, or an invalid access token, is provided the response is:
```
401 Unauthorized

Unauthorized: Invalid token
```

Connected to #1697 

#### Change log

#### Test coverage

#### Documentation

Issue for API docs is: https://github.com/cyberark/dap-wiki/issues/90

### DIFF
--- a/app/controllers/status_controller.rb
+++ b/app/controllers/status_controller.rb
@@ -1,8 +1,26 @@
 # frozen_string_literal: true
 
+require 'date'
+
 class StatusController < ApplicationController
+  include TokenUser
+
   def index
     render 'index', layout: false
   end
 
+  # /whoami returns basic information about the request client and access token
+  # that Conjur receives.
+  #
+  # This is useful for troubleshooting authentication with access tokens and
+  # configuring proxies or load balancers.
+  def whoami
+    render json: { 
+      client_ip: request.ip,
+      user_agent: request.user_agent,
+      account: token_user.account,
+      username: token_user.login,
+      token_issued_at: Time.at(token_user.token.claims["iat"])
+    }
+  end
 end
--- a/spec/request/whoami_spec.rb
+++ b/spec/request/whoami_spec.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe StatusController, type: :request do
+  describe "GET #whoami" do
+    it "responds with unauthorized when no access token is provided" do
+      get '/whomai'
+      expect(response.status).to equal(401)
+    end
+
+    it "returns the client information" do
+      request_env = {
+        'HTTP_AUTHORIZATION' => access_token_for('alice'),
+        'REMOTE_ADDR' => '4.4.4.4'
+      }
+      
+      get('/whoami', env: request_env)
+
+      response_json = JSON.parse(response.body)
+      expect(response_json).to include({
+        'client_ip' => '4.4.4.4',
+        'account' => 'rspec',
+        'username' => 'alice'
+      })
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   API request (e.g. `[client@43868 ip="172.24.0.5"]`)
   ([cyberark/conjur#1550](https://github.com/cyberark/conjur/issues/1550))
 - Print Conjur server FIPS mode status. [cyberark/conjur#1654](https://github.com/cyberark/conjur/issues/1654)
+- Add `/whoami` API endpoint for improved supportability and debugging for access
+  tokens and client IP address determination. [cyberark/conjur#1697](https://github.com/cyberark/conjur/issues/1697)
 
 ## [1.7.4] - 2020-06-17
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,6 +13,7 @@ end
 Rails.application.routes.draw do
   scope format: false do
     get '/' => 'status#index'
+    get '/whoami' => 'status#whoami'
     get '/authenticators' => 'authenticate#index'
 
     constraints id: /[^\/\?]+/ do
--- a/cucumber/api/features/step_definitions/policy_load_steps.rb
+++ b/cucumber/api/features/step_definitions/policy_load_steps.rb
@@ -1,4 +1,4 @@
-# Let's you reference the global variables such as `$?` using less 
+# This allows to you reference some global variables such as `$?` using less 
 # cryptic names like `$CHILD_STATUS`
 require 'English'
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+# This allows to you reference some global variables such as `$?` using less 
+# cryptic names like `$CHILD_STATUS`
+require 'English'
+
 require 'simplecov'
 
 SimpleCov.command_name "SimpleCov #{rand(1000000)}"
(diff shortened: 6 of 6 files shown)


andytinkham:
The first thing that comes to mind - if an access token is leaked somehow, does this amplify the impacts of that leak by allowing a connection between an access token and an account, or by allowing an attacker to figure out how much longer an access token might be usable for?

InbalZilberman:
As the endpoint is going to be on Conjur/DAP
I find it a bit misleading that we don't return the Conjur identity. Meaning host or user id. WDYT? Resemble the CLI experience

uCatu:
What do we think about this endpoint security wise, Can it be abused?
Are those fields: `account` & `login`, `ip` are also visible if user decode that JWT with external tool?

config/initializers/rack_middleware.rb
```diff
@@ -11,7 +11,8 @@
   optional: [
     /^\/authn-[^\/]+\//,
     /^\/authn\//,
-    /^\/public_keys\//
+    /^\/public_keys\//,
+    /^\/whoami$/
```
uCatu, line 15: If we make this token required for `whoami` can it work for you? related to comment above

app/controllers/status_controller.rb
```diff
@@ -1,8 +1,40 @@
...
+  end
+
+  def token_attributes
+    if token_user?
+      {
+        account: token_user.account,
+        login: token_user.login,
+        token_issued_at: Time.at(token_user.token.claims["iat"])
+      }
+    else
```
uCatu, line 32: Can you manage with `401` in case user doesn't have `token_user` - we can add error message if we need to distingish between `401` use cases

Randering and suppliing 200 OK (cache etc) seems less tight

app/controllers/status_controller.rb
```diff
@@ -1,8 +1,40 @@
...
+      user_agent: request.user_agent,
+    }.merge(token_attributes)
+  end
+
+  def token_attributes
+    if token_user?
+      {
+        account: token_user.account,
+        login: token_user.login,
+        token_issued_at: Time.at(token_user.token.claims["iat"])
```
uCatu, line 30:
```suggestion
        token_issued_at: Time.at token_user.token.claims["iat"] || 'unknown issuing time'
```
https://github.com/cyberark/slosilo/blob/4121d4967fc0bc329d73d31e15ee8e9c31962832/lib/slosilo/key.rb#L148

Can this use-case be valid? failing to parse `iat` leasing us to publising stacktrace?

micahlee:
@andytinkham @uCatu 

>  if an access token is leaked somehow, does this amplify the impacts of that leak by allowing a connection between an access token and an account

> Are those fields: account & login, ip are also visible if user decode that JWT with external tool?

These field values are only base64 encoded in the access token itself, so the endpoint doesn't reveal any token information the holder of a token doesn't have.

app/controllers/status_controller.rb
micahlee: No, the token is validated before the endpoint is reached. If the `iat` is missing, the token is invalid and the Rack middleware will return `unauthorized`. So we can be sure `iat` will be present when returning this response.

app/controllers/status_controller.rb
```diff
@@ -1,8 +1,40 @@
...
     render 'index', layout: false
   end
 
+  # /whoami returns basic information about the request client from Conjur's
+  # perspective, including a role_id from a provided access token (if provided)
+  # and the origin IP address of the request.
+  #
+  # This is useful for debugging authentication with access tokens and configuring
+  # proxies or load balancers.
+  def whoami
```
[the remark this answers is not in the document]
micahlee: This is not a `debug` mode endpoint, but is generally useful for troubleshooting authentication as an end-user/operator.

config/initializers/rack_middleware.rb
micahlee: Sure, we can make this require authentication. I'm not sure I agree with the cost/risk of echoing the client IP address, but for my use of this, I can use authenticated requests.

CC: @alexkalish @andytinkham if either of you have comments/thoughts.

micahlee:
> I find it a bit misleading that we don't return the Conjur identity. Meaning host or user id. WDYT? Resemble the CLI experience

Sure, `login` is the convention used in the JWT token itself. This is the same as the `username` in the CLI `conjur authn whoami` output, and I can use that instead.

spec/request/whoami_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+        'REMOTE_ADDR' => '4.4.4.4'
+      }
+      
+      get('/whoami', env: request_env)
+
+      response_json = JSON.parse(response.body)
+      expect(response_json).to include({
+        'client_ip' => '4.4.4.4',
+        'account' => 'rspec',
+        'username' => 'alice'
```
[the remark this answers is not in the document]
micahlee: The keys here are strings, rather than symbols (because that's how the json is parsed). This requires the old style syntax.

config/initializers/rack_middleware.rb
andytinkham: I'm less concerned about echoing the client IP back (my understanding is that it would basically be the equivalent of https://whatsmyip.com/your-ip-address, right?). I think having it require authentication addresses my concern about increasing the impact of an API token leak.

app/controllers/status_controller.rb
```diff
@@ -1,8 +1,26 @@
...
+  #
+  # This is useful for troubleshooting authentication with access tokens and
+  # configuring proxies or load balancers.
+  def whoami
+    render json: { 
+      client_ip: request.ip,
+      user_agent: request.user_agent,
+      account: token_user.account,
+      username: token_user.login,
+      token_issued_at: Time.at(token_user.token.claims["iat"])
```
uCatu, line 23: Now authentication is required for this endpoint and there is no way for tempering and generating a JWT that contains illegal formating of `iat` that might cause `Time.at` to raise exception right?

micahlee: That's right. The validation itself uses `Time.at`. If that raises an exception, that validation is going to fail, never reaching our use of the claim.

See: https://github.com/cyberark/slosilo/blob/c2bb674c75c1a6ff2af66dfb5da68cb433dcc5f4/lib/slosilo/key.rb#L148

micahlee:
Thanks @uCatu @jtuttle!

I just cleaned up this PR to assimilate the fixup commits. Let me know if you see anything else, or please re-approve if you're satisfied. Thanks!

config/initializers/rack_middleware.rb
micahlee: @andytinkham 

> I'm less concerned about echoing the client IP back (my understanding is that it would basically be the equivalent of https://whatsmyip.com/your-ip-address, right?)

Similar, yes. However, assuming this in an internal network, if the XFF is not configured correctly, it could potentially reveal the address of a load balancer, for example.

> I think having it require authentication addresses my concern about increasing the impact of an API token leak.

Ironically, this doesn't do anything to address the token info concern. If you have a leaked access token that is still valid, you can use it to call this endpoint. But this doesn't reveal any information not contained in the token in plain-text directly.

micahlee:
Note, I initially used an incorrect method from slosilo to generate a JWT access token in the tests. I've added an issue to consider if we can deprecate/remove this method here: https://github.com/cyberark/slosilo/issues/20

PR #1701, opened 2020-07-21
Command class readme

jonahx:
### What does this PR do?

Adds README for `app/domain` and `CommandClass`

### DIFF
--- a/app/domain/README.md
+++ b/app/domain/README.md
@@ -1,26 +1,156 @@
-## What is this folder?
+# Why `app/domain`?
 
-- decouple from rails
-- easily test
-- common set of standards
-- Make a coordinated business use case into an explicit, first-class object.
-  - What does it coordinate?
-    - Eg, Database query, audit, a log, and database update
-- Link to Tim Riley talk
+## Introduction
 
-## CommandClass
+This document explains the coding standards used in `app/domain`, and the
+motivation behind them.  By design, these standards are differerent from those
+of a typical Rails app. 
 
-- Jason convo:
-  - Just a proc
-  - Think of it like a configurable method
-  - Could be instantiated in initializers
-- Creates object under the hood.
-  - Makes the parameters available as methods
+You can think of `app/domain` as a haven where pure Conjur business logic (the
+Conjur "domain") lives as plain ruby code, unpolluted by framework-specific
+concepts like controllers or routes.  The code in this folder should have no
+dependencies on Rails itself (if you find any, they are likely mistakes we
+should fix).  
 
-- To undertand the motivation for the `CommandClass` gem, it's helpful to compare it
-to how we'd do the same thing in vanilla ruby: 
+Currently, `app/domain` contains mostly authentication logic.  But its scope is
+not limited to authenticators.
 
-Stage 1
+At the highest level, this approach aims for:
+
+- Explicit, maintainable code that is easy to understand
+- Fast unit tests
+- Intuitive names that correspond with our documentation and the way we
+  naturally talk
+
+We'll dig into the technical specifics below.
+
+## The "Functional Object-Oriented" Style
+
+I'll offer a TLDR version here.  For more in depth discussion of the same
+principles, the video [Functional Architecture for the Practical
+Rubyist](https://www.youtube.com/watch?v=7qnsRejCyEQ&feature=youtu.be) is a
+good place to start.
+
+The basic idea is to reap many benefits of functional programming while 
+remaining true to ruby's objected-oriented paradigm.
+
+In brief, this means building features out of three basic kinds of objects, and
+maintaing a strict sepration of responsibilities:
+
+- **Command objects**
+  - These go by many names: service objects, interactors, use cases
+  - Coordinates high-level logic involving multiple collaborators
+    - For example: `CreateUser` might need to coordinate user input validation,
+      a database transaction, a log entry, and a confirmation email
+  - Makes all collaborators explicit as dependencies
+  - Easy to unit test with doubles or mocks
+- **Value objects**
+  - These are essentially data types.
+  - Methods return data the object was initialized with, or transformations of
+    it
+  - For example: `User`, `Address`, `Url`, etc.
+  - Immutable
+  - No side effects
+  - Class names are nouns
+  - Method names are nouns describing what they return
+- **Objects with side effects**
+  - Repositories for performing database transactions
+  - Objects that communicate with network services (`Aws`, `Github`, etc)
+  - Objects that interact with the file system
+  - Class names are usually nouns
+  - Method names are verbs (to emphasize their side-effects)
+    - For example: `github.fetch_issues`
+    - By contrast, a value object would use a name like `github.issues`.  The
+      "fetch" indicates the network side effect and possibility of error.
+
+Other principles include:
+
+- Avoid inheritance
+  - Instead, make shared code into its own object, which is passed into other
+    objects that need it.  Composition over inheritance.
+- All objects are immutable
+- Careful, consistent naming practices that align with documentation
+
+### Specific Technologies
+
+Writing in the functional OO style does not require any specific gems, though
+many gems make it easier to write value objects and command objects.
+
+In general, there is a tradeoff when using gems:
+
+- Positive: 
+  - They cut down on boilerplate and make your code more concise
+- Negative: 
+  - Every new developer has another extra thing to learn
+  - The code is often not as obvious as plain ruby
+
+You'll encounter two gems when working in Conjur:
+
+- `command_class` for writing command objects
+  - This is still used extensively.  Read below for more information.
+- `dry-struct` for value objects
+  - While this is a good gem, I've stopped using inside Conjur for the reasons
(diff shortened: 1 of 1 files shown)


app/domain/README.md
```diff
@@ -0,0 +1,296 @@
+# Why `app/domain`?
+
+## Introduction
+
+This document explains the coding standards used in `app/domain`, and the
```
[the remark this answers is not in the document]
jonahx: I definitely could have put this under `lib`.  That may have been the better choice.  I don't have a strong feeling one way or another and would have no objection to moving it.

app/domain/README.md
```diff
@@ -0,0 +1,296 @@
...
+  - Repositories for performing database transactions
+  - Objects that communicate with network services (`Aws`, `Github`, etc)
+  - Objects that interact with the file system
+  - Class names are usually nouns
+  - Method names are verbs (to emphasize their side-effects)
+    - For example: `github.fetch_issues`
+    - By contrast, a value object would use a name like `github.issues`.  The
+      "fetch" indicates the network side effect and possibility of error.
+
+Other principles include:
```
micahlee, line 66: This would benefit from examples, both for good and bad cases.

The authentication audit events would be a good example for avoiding inheritance, also the equality operator.

The use of database lookups (though subtle with sequel) are common "bad" examples for immutability.

Not sure about naming examples off-hand, but this might benefit the most where a few examples communicate the "spirit" of the principal more clearly than the description itself.

app/domain/README.md
```diff
@@ -0,0 +1,296 @@
...
+pattern, even though it doesn't matter much practically.
+
+The inputs, on the other hand, are known only at runtime. They're often, but
+not necessarily, given as the params of an http request.  
+
+### Pitfalls To Avoid
+
+- Hardcoding dependencies that should be injected.
+  - For example: Directly referencing `Rails.logger` or a model class.
+  - All code with side effects should be injected as a depenency.
```
micahlee, line 128:
```suggestion
  - All code with side effects should be injected as a dependency.
```

app/domain/README.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+We've eliminated the argument-passing clutter and we're still thread safe.
+_But_ the price is boilerplate and complexity: An additional nested class to
+write and keep track of, and the duplication of the dependencies as attributes
+in both `DoSomeUseCase` and the nested `Run` class.
+
+So we now have two options for vanilla ruby alternatives to the `CommandClass`:
+
+1. One which is coneptually simple but cluttered by passing runtime arguments
```
micahlee, line 272:
```suggestion
1. One which is conceptually simple but cluttered by passing runtime arguments
```

app/domain/README.md
```diff
@@ -0,0 +1,296 @@
...
+   between all the methods.
+1. One which avoids that clutter but at the expense of other complexity and
+   boilerplate.
+
+The `CommandClass` is simply a third option, which takes the second approach --
```
micahlee, line 277: Would it be difficult to include an example of code that you would _not_ convert to a command class?

jonahx:
@micahlee Only just now saw your comments.  Really appreciate them.  I'll try to think of examples per your suggestions.

app/domain/README.md
```diff
@@ -0,0 +1,296 @@
...
+remaining true to ruby's objected-oriented paradigm.
+
+In brief, this means building features out of three basic kinds of objects, and
+maintaing a strict sepration of responsibilities:
+
+- **Command objects**
+  - These go by many names: service objects, interactors, use cases
+  - Coordinates high-level logic involving multiple collaborators
+    - For example: `CreateUser` might need to coordinate user input validation,
+      a database transaction, a log entry, and a confirmation email
```
orenbm, line 44: i would add here that `CreateUser` will consume some of its logic from other Command objects with smaller resplonsibility

app/domain/README.md
```diff
@@ -0,0 +1,296 @@
...
+
+- Positive: 
+  - They cut down on boilerplate and make your code more concise
+- Negative: 
+  - Every new developer has another extra thing to learn
+  - The code is often not as obvious as plain ruby
+
+You'll encounter two gems when working in Conjur:
+
+- `command_class` for writing command objects
```
orenbm, line 89: do you want to move this repo to a cyberark one?

app/domain/README.md
```diff
@@ -0,0 +1,296 @@
...
+  - This is still used extensively.  Read below for more information.
+- `dry-struct` for value objects
+  - While this is a good gem, I've stopped using inside Conjur for the reasons
+    listed above
+  - I now write value objects using plain ruby
```
orenbm, line 94: do we have an effort to remove this altogether? how hard do you think it will be?

app/domain/README.md
```diff
@@ -0,0 +1,296 @@
...
+to capture.  Nothing beats a set of fresh, critical eyes to validate
+readability.
+
+A few heuristics to keep in mind:
+
+- Avoid putting too many responsibilities into a command class.  Its single
+  verb phrase name should naturally describe it.
+  - Split up a `CommandClass` when needed
+- Avoid putting too little in one.  Not everything should be a `CommandClass`.
+  Keep in mind why we write them.
```
orenbm, line 146: i'd like to get an example here as this is something that i never really understood. From my point of view, if you have a responsibility that needs to be done in several places then it makes sense to make it a command class. i have an example in another PR which i will link to once it is merged.

orenbm: here is the example. we have this class: https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/extract_container_name.rb

it is used in both [here](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/inject_client_cert.rb#L177) and [here](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/validate_pod_request.rb#L88)

you will see that the command class has only a dependency on `Rails.logger` which may point that it should not be a command class. However, it does have some logic and constants that i think should not be implemented in both places. 

i'd like to get your thoughts on this, and even more important - extract a rule of thumb for this README.

jonahx: This is a good question, and it's come a few places in your PRs as well as others.  I'll try a short-ish answer here.  Tell me what you think, or if it raises other questions....

> From my point of view, if you have a responsibility that needs to be done in several places then it makes sense to make it a command class.

This is mostly try if:

1. The logic has side-effects.
2. The logic has minimum complexity

The reason I add the 2. caveat is if it's really simple, like 1 line, it might be simpler to repeat it.  Again, introducing an abstraction is not cost-free.  Just like it's possible to have too many tiny methods in a class, it's possible to have too many tiny command classes... you may have removed duplication, but you force the reader to know about more things, and to be able to remember and use them properly.  This is a deep subject, and I think the best way to get a handle on it personally and as a team is to have discussions about multiple specific examples.

> here is the example. we have this class: https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/extract_container_name.rb

I remember seeing this one before. 

So I think my take on this one is that we're missing two things:

1. A value object like `K8sHost`, which is either created with `service_id` and has a `container_name` method, or is created without a `service_id` and has a `container_name(service_id)` method, depending on your needs.
2. A repository object which would return a `K8sHost`, given a `host_id` (or given a `host_id`/`service_id` combo)

We don't use `Repo` objects formally in Conjur, though I have may have created 1 or 2, and we treat the Rails model objects like repos in many places.  But the basic idea is that the repository knows all the database storage details, and it just give you back a nice, clean application object that works the way you want it to.

So in this particular case,  all this junk:

```ruby
 def container_name
        annotation_value("authn-k8s/#{@service_id}/#{AUTHENTICATION_CONTAINER_NAME_ANNOTATION}") ||
          annotation_value("authn-k8s/#{AUTHENTICATION_CONTAINER_NAME_ANNOTATION}") ||
          annotation_value("kubernetes/#{AUTHENTICATION_CONTAINER_NAME_ANNOTATION}") ||
          default_authentication_container_name
      end
```
is just "details about how Conjur happens to store k8s info in its database using annotations".  Your application code does't care about that crap.  That's just database implementation detail.  So we want our Repo, and _only_ our repo, to know about this stuff.  We'd just query it and get back `K8sHost`, say.  

So this stuff fits more naturally at the at lower level than in a `CommandClass` object, which (per this README) are intended for the higher level coordination of multiple tasks into a single use-case.

app/domain/README.md
jonahx: Well, all my work on it was done on my own time.  If there's a pressing need to I suppose I would, but it's really no different from our perspective than using any 3rd party gem and we depend on hundreds of them.

app/domain/README.md
jonahx: Probably not that hard.  I don't think there are that many instances but I'd have to check.

jonahx:
@orenbm and @micahlee I've addressed all the comments except for the naming example (I have some ideas) and integrating the long reply I made to Oren's last comment.  Let me know what you think.

PR #1705, opened 2020-07-22
Use generic authentication code in authn-oidc

orenbm:
Prior to this commit, the OIDC authenticator implemented all the
general authentication process (webservice & user validations,
token generation, audit, etc.) whereas the rest of the authenticators
used a general authentication class that performed these actions for them.

This helps us in maintaining the code so that if a change is made somewhere
(e.g how we audit success) then we need to update it in one place instead of
in all the authenticator classes.

The reason why the OIDC authenticator implemented everything by itself is because
it gets the username from the provided ID token, and not from the URL like the rest
of the authenticators do.

So in order to use the general authentication class we first need to decode the ID token
and extract the username from it, and then use the general class.

Note: the build will break because of UTs. I didn't bother fixing them before getting your thoughts.


app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -0,0 +1,102 @@
...
+        @conjur_username ||= @decoded_token[id_token_username_field]
+      end
+
+      def id_token_username_field
+        oidc_authenticator_secrets["id-token-user-property"]
+      end
+
+      def admin?(username)
+        username == "admin"
+      end
```
jonahx, line 95: Remove and inline this method.

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -0,0 +1,102 @@
...
+        if admin?(conjur_username)
+          raise Errors::Authentication::AuthnOidc::AdminAuthenticationDenied
+        end
+
+        @logger.debug(LogMessages::Authentication::AuthnOidc::ExtractedUsernameFromIDToked.new(conjur_username, id_token_username_field))
+      end
+
+      def conjur_username
+        @conjur_username ||= @decoded_token[id_token_username_field]
+      end
```
jonahx, line 87: No need to memoize here.

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -0,0 +1,102 @@
...
+                     :account, :username, :webservice, :credentials, :client_ip,
+                     :role
+
+      def call
+        validate_account_exists
+        validate_credentials_include_id_token
+        verify_and_decode_token
+        validate_conjur_username
+        add_username_to_input
+        @authenticator_input
```
jonahx, line 25: Replace the last two lines above with:
```ruby
@authenticator_input.update(username: conjur_username)
```
and then you can delete `add_username_to_input` method.

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      def validate_credentials_include_id_token
+        id_token_field_name = "id_token"
+
+        # check that id token field exists and has some value
+        unless decoded_credentials.include?(id_token_field_name) &&
+            !decoded_credentials[id_token_field_name].empty?
+          raise Errors::Authentication::RequestBody::MissingRequestParam, id_token_field_name
+        end
```
jonahx, line 43: Equivalent shorter logic:
```ruby
        unless decoded_credentials.fetch(id_token_field_name, "") != ""
```

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
jonahx: On 2nd thought, I might like the original better here.

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -0,0 +1,102 @@
...
+          raise Errors::Authentication::AuthnOidc::IdTokenFieldNotFoundOrEmpty,
+                id_token_username_field
+        end
+
+        if admin?(conjur_username)
+          raise Errors::Authentication::AuthnOidc::AdminAuthenticationDenied
+        end
+
+        @logger.debug(LogMessages::Authentication::AuthnOidc::ExtractedUsernameFromIDToked.new(conjur_username, id_token_username_field))
+      end
```
jonahx, line 83: Break line after `new(`

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
jonahx, line 25: Change the final two lines to this single line:

```ruby
input_with_username
```
and then change `add_username_to_input` to:

```ruby
      def input_with_username
        @authenticator_input.update(username: conjur_username)
      end
```

Issue #1213, opened 2019-10-23
No indication for the user when providing a wrong Conjur data key


rafis3:
## Issue description
Conjur uses the data key in order to encrypt sensitive information in the database. In case a wrong data key is provided, Conjur needs to indicate this with an elaborate and specific error message.

## Steps to reproduce the issue
1. Create a new Conjur instance. For example, follow the steps in [this](https://www.conjur.org/get-started/quick-start/oss-environment/) guide.
1. Stop Conjur.
1. Restart Conjur with a different data key. If using the example above, when running the command: `export CONJUR_DATA_KEY="$(< data_key)"`, provide a modified value to the environment variable and then start Conjur.

## What's the expected result?
The expected result is that when Conjur starts, it would be able to verify that the data key is valid and if not, provide an informative error to the user.

## What's the actual result?
The Conjur server remains up, without any indication to a problem. Only when the data key would be used, a general OpenSSL error would be given in the logs.

## Additional details / screenshot
Example of such failure:
```
Sequel::Postgres::Database (6.7ms)  SELECT * FROM "secrets" WHERE ("secrets"."resource_id" = 'myorg:variable:conjur/authn-k8s/prod/ca/cert') ORDER BY "version" DESC LIMIT 1
Authentication Error: OpenSSL::Cipher::CipherError
```

jtuttle:
Going to approach this by improving the error for now. Coming up with a way to verify the data key is a more difficult ask that will require some design. @rafis3 If you have any good ideas about this do let us know. My initial thought was to have a dummy file in the Conjur dir that gets created using the data key when conjur starts. If it already exists, Conjur tries to decode it on startup using the data key and errors if it doesn't work.

rafis3:
@jtuttle I was thinking that we can keep it in the database, instead of persisting a file, to keep Conjur stateless.
And the value can be the sha512 of the data key, encrypted with the data key. So when Conjur starts, it will decrypt the value, using its data key and then hash the data key value and compare it to the decrypted value. This would allow Conjur to verify the value while also using an unpredictable, instance specific, non constant value.

What do you think?

jtuttle:
@rafis3 Interesting, so it would be a special Conjur "variable" where value is set as you described? That seems like it could work.

PR #1709, opened 2020-07-23
Validate Conjur Data Key on Server Startup

eranha:
Conjur uses the data key in order to encrypt sensitive information in the database. In case a wrong data key is provided, there is no indication for the user. This task validate the key and indicates if the key is invalid with an elaborated and specific error message.

### What does this PR do?
A rake task has been added to the `conjurctl server` command to validate the CONJUR_DATA_KEY and exits the process in case the key is invalid.
The validation is done by creating secret once when the db is created and retrieving that secret each time the server boots.

1. Create a new Conjur instance. For example, follow the steps in this guide.
2. Stop Conjur.
3. Restart Conjur with a different data key. If using the example above, when running the command: export CONJUR_DATA_KEY="$(< data_key)", provide a modified value to the environment variable and then start Conjur.

N\A
Connected to #[GitHub issue](https://github.com/cyberark/conjur/issues/1213)

#### Change log

#### Test coverage

#### Documentation


lib/tasks/data_key_validation.rake
```diff
@@ -0,0 +1,52 @@
...
+
+    unless Resource[resource_id]
+      # create the master key validator resource
+      # and store a secret using the master key
+      role_id = "!:!:data-key-validator"
+
+      begin
+        Role.create(role_id: role_id)
+        Resource.create(resource_id: resource_id, owner_id: role_id)
+        Secret.create(resource_id: resource_id, value: "Don't Panic.")
```
rafis3, line 33: Are these resources visible to the end user? If so, I think we should find another way to test the validity of the data key to avoid introducing unexpected resources for the user.

lib/tasks/data_key_validation.rake
```diff
@@ -0,0 +1,52 @@
...
+      rescue => e
+        $stderr.puts "Error creating a resource for rake: data_key_validation, Error: #{e.inspect}"
+      end
+    else
+      begin
+        # Attempt to retrieve the secret encrypt with the original data key.
+        # A failure to decrypt the value will indicate the key is invalid.
+        Resource[resource_id].secret.value
+      rescue => e
+        if e.inspect == 'OpenSSL::Cipher::CipherError'
```
rafis3, line 43: Is there a possibility that this error type could be raised not because of the data key being invalid? I want to make sure there will not be a case where we write a misleading error message.

lib/tasks/data_key_validation.rake
```diff
@@ -0,0 +1,52 @@
...
+          raise Errors::System::InvalidConjurDataKey
+        else
+          $stderr.puts "rake: data_key_validation, error retrieving Resource[#{resource_id}].secret.value, #{e.inspect}"
```
rafis3, line 46: What is the value in printing the secret value? How does it help the user troubleshoot the problem?

lib/tasks/data_key_validation.rake
```diff
@@ -0,0 +1,52 @@
...
+          exit 1
+        end
+      end
+    end
+ end
```
rafis3, line 51:
```suggestion
  end
```

lib/tasks/data_key_validation.rake
eranha: The key is used only in secrets and Host Tokens.

lib/tasks/data_key_validation.rake
eranha: Only if the key is Invalid.

lib/tasks/data_key_validation.rake
eranha: It does not print the secret value.

eranha:
> Can you please try to add a test for this? I see we already have one for rake tasks (`cucumber/policy/features/rake.feature`).

bin/conjur-cli.rb
```diff
@@ -66,6 +66,10 @@ def test_select
       system "rake account:create[#{account}]" or exit $?.exitstatus
     end
 
+    puts "Validating Conjur data key..."
+    system "rake data_key_validation:validate" or exit $?.exitstatus
+
+
```
orenbm, line 72: double empty-line

lib/tasks/data_key_validation.rake
```diff
@@ -0,0 +1,52 @@
+
```
orenbm, line 1: i see that we already have a `rake` task named `data-key.rake`. how about we add a task `:validate` into that file instead of creating a new file?

lib/tasks/data_key_validation.rake
orenbm: i'd still like to get an answer to @rafis3 's question - are these resources visible to the end user? are they visible only to the admin user?

lib/tasks/data_key_validation.rake
```diff
@@ -0,0 +1,52 @@
...
+namespace :data_key_validation do
+  desc "Validates the value of ENV['CONJUR_DATA_KEY']"
+  # This task validate the master data key which is stored in `ENV['CONJUR_DATA_KEY']`
+  # Once the db is created, a secret will be created using the master key.
+  # Each time the server is restarted this task will attempt to retrieve the
+  # secret, if the provided key does not match the original master the secret retrieval
+  # will fail.
```
orenbm, line 8: i would write that we have 2 cases - a first run of the server and a restart of the server. Then i would describe what this task does in each case

lib/tasks/data_key_validation.rake
orenbm: also - what's `Don't Panic.`? can we have something more official like `data-key-validator-secret`?

lib/tasks/data_key_validation.rake
```diff
@@ -0,0 +1,52 @@
...
+
+    if data_key.nil? || data_key.empty?
+      raise Errors::System::EmptyConjurDataKey
+      exit 1
+    end
+
+    # try get the master key validator resource
+    resource_id = "!:!:data-key-validator-resource"
+
+    unless Resource[resource_id]
```
orenbm, line 25: if you `unless` and then `else` it is better to turn these around and have an `if` instead so it is more readable. so you can do:
```
if Resource[resource_id]
  # we are in a server restart - verify you can decrypt the secret
else
  # we are in the first run - create the secret

File: lib/tasks/data_key_validation.rake (hunk 3 above)
Comment (orenbm) [reply lib/tasks/data_key_validation.rake:46]: i would change it to have words instead of code. `error retrieving data-key validator secret`

Comment (orenbm) [reply lib/tasks/data_key_validation.rake:46]: also - why is this not in `errors.rb`?

File: lib/tasks/data_key_validation.rake (hunk 1 above)
Comment (eranha) [reply lib/tasks/data_key_validation.rake:33]: > also - what's `Don't Panic.`? can we have something more official like `data-key-validator-secret`?

It's a quote from Hitchhiker's guide to the galaxy. No one is suppose to read this value. Consider it an easter egg.

File: lib/tasks/data_key_validation.rake (hunk 3 above)
Comment (eranha) [reply lib/tasks/data_key_validation.rake:46]: > also - why is this not in `errors.rb`?

I don't expect this error to rise, it's a fallback.

File: lib/tasks/data_key_validation.rake (hunk 1 above)
Comment (eranha) [reply lib/tasks/data_key_validation.rake:33]: > i'd still like to get an answer to @rafis3 's question - are these resources visible to the end user? are they visible to the admin user?

It will have to wait for the next boost.

File: app/domain/errors.rb [hunk 9]
```diff
@@ -353,4 +353,18 @@ module Util
       code: "CONJ00044E"
     )
   end
+
+  module System
+    EmptyConjurDataKey = ::Util::TrackableErrorClass.new(
+      msg:  "CONJUR_DATA_KEY is empty. " \
+            "Try loading the master key as an environment variable.",
```
shulifink, line 360: @eranha  (cc: @rafis3 ) Just want to make sure that this "master key" is the key from the encryption of the server keys....

app/domain/errors.rb
```diff
@@ -353,4 +353,18 @@ module Util
...
+      code: "CONJ00065E"
+    )
+
+    InvalidConjurDataKey = ::Util::TrackableErrorClass.new(
+      msg:  "Invalid CONJUR_DATA_KEY. " \
+            "The provided key does not match the key that the db was created with.",
```
shulifink, line 366: The key provided does not match the key that the database was created with.

bin/conjur-cli.rb
```diff
@@ -66,6 +66,9 @@ def test_select
       system "rake account:create[#{account}]" or exit $?.exitstatus
     end
 
+    puts "Validating Conjur data key..."
```
shulifink, line 69: Conjur? @rafis3  please confirm that this is okay? Should it be product (DAP/Conjur) specific, depending on the context?

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
     key = Base64.strict_encode64(Slosilo::Symmetric.new.random_key)
     print key
   end
+
+  desc "Validates the value of ENV['CONJUR_DATA_KEY']"
+  # This task validate the master data key which is stored in `ENV['CONJUR_DATA_KEY']`
```
shulifink, line 13: validate**s** the (master?) data key **that** is stored..."

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
...
+  # The task runs in two modes:
+  # First mode:
+  # ==========
+  #   Once the db is created, a role + resource + secret will be created using the master key.
```
shulifink, line 17: db = database

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
...
+      # Server restart - verify you can decrypt the secret
+      begin
+        # Attempt to retrieve the secret encrypt with the original data key.
+        # A failure to decrypt the value will indicate the key is invalid.
+        Resource[resource_id].secret.value
+      rescue => e
+        if e.inspect == 'OpenSSL::Cipher::CipherError'
+          raise Errors::System::InvalidConjurDataKey
+        else
+          $stderr.puts "Error retrieving data-key validator secret in Resource[#{resource_id}], Error: #{e.inspect}"
```
shulifink, line 51: fine

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
...
+      # create the master key validator resource
+      # and store a secret using the master key
+      role_id = "!:!:data-key-validator"
+
+      begin
+        Role.create(role_id: role_id)
+        Resource.create(resource_id: resource_id, owner_id: role_id)
+        Secret.create(resource_id: resource_id, value: "Don't Panic.")
+      rescue => e
+        $stderr.puts "Error creating a resource for rake: data_key_validation, Error: #{e.inspect}"
```
shulifink, line 66: What is rake? I saw it was capitalized earlier. Should it be Rake here too?

app/domain/errors.rb
rafis3: @shulifink good catch! No, the master key is used in DAP to encrypt the server keys. This key, which is the data key, is one of the sever keys that the master key encrypts. So we should rewrite this message and explicitly call it data key.

bin/conjur-cli.rb
rafis3: Conjur is a component rather than a product. DAP contains the Conjur component. So to me, specifying that it's a Conjur data key is completely fine. But we should be consistent in the other locations where we reference it.

lib/tasks/data-key.rake
rafis3: Again, the master key is something else. Here we refer to the data key.

lib/tasks/data-key.rake
eranha: Rake is a software task management and build automation tool.

lib/tasks/data_key_validation.rake
orenbm: i prefer that we have `data-key-validator-secret` instead of a joke that not everyone will understand

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
     key = Base64.strict_encode64(Slosilo::Symmetric.new.random_key)
     print key
   end
+
+  desc "Validates the value of ENV['CONJUR_DATA_KEY']"
+  # This task validate the master data key that is stored in `ENV['CONJUR_DATA_KEY']`
```
orenbm, line 13: validates

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
...
+  # The task runs in two modes:
+  # First mode:
+  # ==========
+  #   Once the database is created, a role + resource + secret will be created using the master key.
+  #
+  # Second mode:
+  # ===========
+  #   Each time the server restarts this task will attempt to retrieve the
+  #   secret, if the provided key does not match the original master the secret retrieval
```
orenbm, line 22: the secret. (period)

orenbm: the original master key

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
...
+  #   will fail and the task will exit with error.
+  #
+  # Side Note
+  # =========
+  #  What's the 'environment' task in Rake?
+  #  the :validate => :environment means "do environment before validate".
+  #  You can get access to your models, and in fact, your whole environment
+  #  by making tasks dependent on the environment task.
```
orenbm, line 30: LOVE this comment

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
...
+  task :validate => [:environment] do
+    data_key = ENV['CONJUR_DATA_KEY']
+
+    if data_key.nil? || data_key.empty?
+      raise Errors::System::EmptyConjurDataKey
+      exit 1
```
orenbm, line 36: will we get to this line after we raised an error?

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
...
+    end
+
+    # try get the master key validator resource
+    resource_id = "!:!:data-key-validator-resource"
+    if Resource[resource_id]
+      # Server restart - verify you can decrypt the secret
```
orenbm, line 42: verify we

lib/tasks/data-key.rake
eranha: COnjur data key to be consistent

lib/tasks/data-key.rake
eranha: When I tested it the process exited with error.

jonahx: This `exit` is unreachable.  I'm guessing the process exited with an error because of the `raise`, not because of line 36, which you can remove.

lib/tasks/data-key.rake
```diff
@@ -8,4 +8,63 @@ namespace :"data-key" do
...
+    # try get the Conjur data key validator resource
+    resource_id = "!:!:data-key-validator-resource"
+    if Resource[resource_id]
+      # Server restart - verify we can decrypt the secret
+      begin
+        # Attempt to retrieve the secret encrypt with the original data key.
+        # A failure to decrypt the value will indicate the key is invalid.
+        Resource[resource_id].secret.value
+      rescue => e
+        if e.inspect == 'OpenSSL::Cipher::CipherError'
```
jonahx, line 48: Instead of using `inspect` to do an implicit string conversion and matching on strings, just rescue the specific error `OpenSSL::Cipher::CipherError`, and then add a 2nd rescue for everything else.  Then you can delete the `if... else` currently inside the `rescue`.

orenbm:
@eranha i see that a few files were generated and added to this PR. do we want to push them? should we add them to `.gitignore`?

app/domain/errors.rb
```diff
@@ -353,4 +353,18 @@ module Util
       code: "CONJ00044E"
     )
   end
+
+  module System
+    EmptyConjurDataKey = ::Util::TrackableErrorClass.new(
+      msg:  "CONJUR_DATA_KEY is empty. " \
+            "Try loading the data key as an environment variable.",
+      code: "CONJ00065E"
```
orenbm, line 361: is this the correct code? not saying it isn't, just want to verify it doesn't exist already

eranha:
Some questions were raised regarding:
`resource_id = "!:!:data-key-validator-resource"` and if this resource is exposed to end user, requires more research hence moved to next boost.

Tovli:
This issue is on hold as it needs more investigation

Issue #1050, opened 2019-06-06
Conjur should return a usable as-is access token


orenbm:
At this point, Conjur returns an access token that users/hosts need to base64 encode it before they can use it in API calls. We should improve the ux of this.

there is a slack thread here: https://conjurhq.slack.com/archives/C0A2NF134/p1557841105256800

orenbm:
@liavyona can this issue be closed? if so, please link to the PR in a comment a close.

izgeri:
is the solution design in #1740 going to be merged as well?

orenbm:
@liavyona this can be closed, correct?

Issue #151, opened 2017-07-26
Authenticate method optionally returns a base64 encoded token


jvanderhoof:
Connected to #1050

orenbm:
@liavyona can this issue be closed? if so, please link to the PR in a comment a close.

PR #1710, opened 2020-07-26
Add encoded optional param to authenticate request

liavyona:
### What does this PR do?
Added optional parameter to 'authenticate' request which instruct the server to encode with base64 its response

Connected to #151 
Connected to #1050 

#### Test coverage
- A UT should be added

### DIFF
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -4,6 +4,8 @@ class AuthenticateController < ApplicationController
   include BasicAuthenticator
   include AuthorizeResource
 
+  @@logger = Rails.logger
+
   def index
     authenticators = {
       # Installed authenticator plugins
@@ -70,7 +72,15 @@ class AuthenticateController < ApplicationController
       authenticators: installed_authenticators,
       enabled_authenticators: Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV)
     )
-    render json: authn_token
+    content_type = :json
+    if encoded_response?
+      # The response will be a base64 encoded version of the JWT.
+      @@logger.debug("Responding with base64 encoded JWT")
+      content_type = :plain
+      authn_token = ::Base64.encode64(authn_token.to_json).gsub("\n", "").strip
+      response.set_header("Content-Encoding", "base64")
+    end
+    render content_type => authn_token
   rescue => e
     handle_authentication_error(e)
   end
@@ -224,5 +234,9 @@ class AuthenticateController < ApplicationController
 
   def enabled_authenticators
     Authentication::InstalledAuthenticators.enabled_authenticators(ENV)
+  end
+
+  def encoded_response?
+    request.accept_encoding.include?("base64")
   end
 end
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -4,8 +4,6 @@ class AuthenticateController < ApplicationController
   include BasicAuthenticator
   include AuthorizeResource
 
-  @@logger = Rails.logger
-
   def index
     authenticators = {
       # Installed authenticator plugins
@@ -74,10 +72,9 @@ class AuthenticateController < ApplicationController
     )
     content_type = :json
     if encoded_response?
-      # The response will be a base64 encoded version of the JWT.
-      @@logger.debug("Responding with base64 encoded JWT")
+      logger.debug(LogMessages::Authentication::EncodedJWTResponse.new)
       content_type = :plain
-      authn_token = ::Base64.encode64(authn_token.to_json).gsub("\n", "").strip
+      authn_token = ::Base64.strict_encode64(authn_token.to_json)
       response.set_header("Content-Encoding", "base64")
     end
     render content_type => authn_token
@@ -237,6 +234,8 @@ class AuthenticateController < ApplicationController
   end
 
   def encoded_response?
-    request.accept_encoding.include?("base64")
+    return false unless request.accept_encoding
+    encodings = request.accept_encoding.split(",")
+    encodings.any? { |encoding| encoding.squish.casecmp?("base64") }
   end
 end
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -54,6 +54,11 @@ module LogMessages
       code: "CONJ00030D"
     )
 
+    EncodedJWTResponse = ::Util::TrackableLogMessageClass.new(
+      msg:  "Responding with a base64 encoded access token",
+      code: "CONJ00039D"
+    )
+
     module OAuth
 
       IdentityProviderUri = ::Util::TrackableLogMessageClass.new(
--- a/cucumber/api/features/authenticate.feature
+++ b/cucumber/api/features/authenticate.feature
@@ -103,10 +103,6 @@ Feature: Exchange a role's API key for a signed authentication token
       cucumber:user:alice failed to authenticate with authenticator authn
     """
 
-  Scenario: successful encoded requests in limited time
-    When I can authenticate Alice 1000 times in 10 threads with Accept-Encoding header "base64"
-    Then The avg authentication request responds in less than 0.75 seconds
-
   Scenario: User cannot login as a same-named user in a different account
 
     User logins are scoped per account. Conjur cannot be tricked into authenticating a user
--- a/cucumber/api/features/authenticate.feature
+++ b/cucumber/api/features/authenticate.feature
@@ -21,6 +21,20 @@ Feature: Exchange a role's API key for a signed authentication token
       cucumber:user:alice successfully authenticated with authenticator authn
     """
 
+  Scenario: Authenticate response should be encoded if asked
+    When I POST "/authn/cucumber/alice/authenticate?encoded=true" with plain text body ":cucumber:user:alice_api_key"
+    Then the HTTP response content type is "text/plain"
+    And the HTTP response is base64 encoded
+    And user "alice" has been authorized by Conjur
+    And there is an audit record matching:
+    """
+      <86>1 * * conjur * authn
+      [auth@43868 authenticator="authn"]
+      [subject@43868 role="cucumber:user:alice"]
+      [action@43868 operation="authenticate" result="success"]
+      cucumber:user:alice successfully authenticated with authenticator authn
+    """
+
   Scenario: Authenticating with no Content-Type header succeeds without writing API key to the logs
     And I save my place in the log file
     And I can authenticate Alice with no Content-Type header
--- a/cucumber/api/features/authenticate.feature
+++ b/cucumber/api/features/authenticate.feature
@@ -5,7 +5,6 @@ Feature: Exchange a role's API key for a signed authentication token
   The token is a signed JSON structure that contains the role id. The 
   token can be sent as the `Authorization` header to other Conjur REST
   functions as proof of authentication.
-
   Background:
     Given I create a new user "alice"
 
@@ -21,11 +20,51 @@ Feature: Exchange a role's API key for a signed authentication token
       cucumber:user:alice successfully authenticated with authenticator authn
     """
 
-  Scenario: Authenticate response should be encoded if asked
-    When I POST "/authn/cucumber/alice/authenticate?encoded=true" with plain text body ":cucumber:user:alice_api_key"
+  Scenario: Authenticate response should be encoded if Accept-Encoding equals base64
+    When I successfully authenticate Alice with Accept-Encoding header "base64"
     Then the HTTP response content type is "text/plain"
     And the HTTP response is base64 encoded
     And user "alice" has been authorized by Conjur
+    And there is an audit record matching:
+    """
+      <86>1 * * conjur * authn
+      [auth@43868 authenticator="authn"]
+      [subject@43868 role="cucumber:user:alice"]
+      [action@43868 operation="authenticate" result="success"]
+      cucumber:user:alice successfully authenticated with authenticator authn
+    """
+
+  Scenario: Authenticate response should be encoded if Accept-Encoding includes base64
+    When I successfully authenticate Alice with Accept-Encoding header "base64,gzip,defalte,br"
+    Then the HTTP response content type is "text/plain"
+    And the HTTP response is base64 encoded
+    And user "alice" has been authorized by Conjur
+    And there is an audit record matching:
+    """
+      <86>1 * * conjur * authn
+      [auth@43868 authenticator="authn"]
+      [subject@43868 role="cucumber:user:alice"]
+      [action@43868 operation="authenticate" result="success"]
+      cucumber:user:alice successfully authenticated with authenticator authn
+    """
+
+  Scenario: Authenticate response should be encoded if Accept-Encoding includes base64 with mixed case and spaces
+    When I successfully authenticate Alice with Accept-Encoding header "gzip      ,  bASe64 ,defalte,br"
+    Then the HTTP response content type is "text/plain"
+    And the HTTP response is base64 encoded
+    And user "alice" has been authorized by Conjur
+    And there is an audit record matching:
+    """
+      <86>1 * * conjur * authn
+      [auth@43868 authenticator="authn"]
+      [subject@43868 role="cucumber:user:alice"]
+      [action@43868 operation="authenticate" result="success"]
+      cucumber:user:alice successfully authenticated with authenticator authn
+    """
+
+  Scenario: Authenticate response should be json if Accept-Encoding doesn't include base64
+    When I successfully authenticate Alice with Accept-Encoding header "gzip,defalte,br"
+    Then the HTTP response content type is "application/json"
     And there is an audit record matching:
     """
       <86>1 * * conjur * authn
@@ -51,6 +90,22 @@ Feature: Exchange a role's API key for a signed authentication token
       [action@43868 operation="authenticate" result="failure"]
       cucumber:user:alice failed to authenticate with authenticator authn
     """
+
+  Scenario: Attempting to use an invalid API key to authenticate with Accept-Encoding base64 result in 401 error
+    When I authenticate Alice with Accept-Encoding header "base64" with plain text body "wrong-api-key"
+    Then the HTTP response status code is 401
+    And there is an audit record matching:
+    """
+      <84>1 * * conjur * authn
+      [auth@43868 authenticator="authn"]
+      [subject@43868 role="cucumber:user:alice"]
+      [action@43868 operation="authenticate" result="failure"]
+      cucumber:user:alice failed to authenticate with authenticator authn
+    """
+
+  Scenario: successful encoded requests in limited time
+    When I can authenticate Alice 1000 times in 10 threads with Accept-Encoding header "base64"
+    Then The avg authentication request responds in less than 0.75 seconds
 
   Scenario: User cannot login as a same-named user in a different account
 
--- a/cucumber/api/features/step_definitions/request_steps.rb
+++ b/cucumber/api/features/step_definitions/request_steps.rb
@@ -162,6 +162,12 @@ Then(/^the HTTP response content type is "([^"]*)"$/) do |content_type|
   expect(@content_type).to match(content_type)
 end
 
+Then(/^the HTTP response is base64 encoded$/) do
+  # Override encoded response with decode one to use other helpers
+  @result = @response_body = Base64.strict_decode64(@result)
+  expect(JSON.parse(@result).is_a?(Hash)).to be true
+end
+
 Then(/^the result is true$/) do
   expect(@result).to be true
 end
--- a/cucumber/api/features/step_definitions/request_steps.rb
+++ b/cucumber/api/features/step_definitions/request_steps.rb
@@ -103,6 +103,23 @@ When(/I can authenticate Alice with no Content-Type header/) do
   end
 end
 
+When(/^I( (?:successfully|can))? authenticate Alice (?:(\d+) times? in (\d+) threads? )?with Accept-Encoding header "([^"]*)"(?: with plain text body "([^"]*)")?$/) do |can, requests_num, threads_num, header, body|
+  body ||= ":cucumber:user:alice_api_key"
+  requests_num ||= 1
+  threads_num ||= 1
+  authenticate_with_performance(
+    requests_num,
+    threads_num,
+    authentication_func: :authn_request,
+    authentication_func_params: {
+      url: "/authn/cucumber/alice/authenticate",
+      api_key: body,
+      encoding: header,
+      can: can
+    }
+  )
+end
+
 When(/^I( (?:can|successfully))? POST "([^"]*)" with body:$/) do |can, path, body|
   try_request can do
     post_json path, body
@@ -151,11 +168,11 @@ Then(/^the result is the API key for user "([^"]*)"$/) do |login|
 end
 
 Then(/^it's confirmed$/) do
-  expect(@status).to be_blank
+  expect(@http_status).to be_blank
 end
 
 Then(/^the HTTP response status code is (\d+)$/) do |code|
-  expect(@status).to eq(code.to_i)
+  expect(@http_status).to eq(code.to_i)
 end
 
 Then(/^the HTTP response content type is "([^"]*)"$/) do |content_type|
@@ -163,9 +180,11 @@ Then(/^the HTTP response content type is "([^"]*)"$/) do |content_type|
 end
 
 Then(/^the HTTP response is base64 encoded$/) do
+  expect(@result.headers[:content_encoding]).to eq("base64")
+
   # Override encoded response with decode one to use other helpers
-  @result = @response_body = Base64.strict_decode64(@result)
-  expect(JSON.parse(@result).is_a?(Hash)).to be true
+  @response_body = Base64.strict_decode64(@result)
+  expect(JSON.parse(@response_body).is_a?(Hash)).to be true
 end
 
 Then(/^the result is true$/) do
--- a/cucumber/api/features/support/rest_helpers.rb
+++ b/cucumber/api/features/support/rest_helpers.rb
@@ -99,6 +99,10 @@ module RestHelpers
   def set_result result
     @response_api_key = nil
     @status = result.code
+
+    # Authenticators helpers using http_status
+    @http_status = result.code
+
     @content_type = result.headers[:content_type]
     if /^application\/json/.match?(@content_type)
       @result = JSON.parse(result)
--- a/cucumber/api/features/support/rest_helpers.rb
+++ b/cucumber/api/features/support/rest_helpers.rb
@@ -54,7 +54,7 @@ module RestHelpers
   # NOTE: The way this works is tricky.  We are relying on a behavior of
   # RestClient, which will automatically add a Basic Auth header to your
   # request, if you've set the user and password fields on its Request object:
-  # 
+  #
   # See: https://github.com/rest-client/rest-client/blob/f450a0f086f1cd1049abbef2a2c66166a1a9ba71/spec/unit/request_spec.rb#L442
   #
   # However, it will only do this IF the request does not already have its
@@ -98,9 +98,6 @@ module RestHelpers
 
   def set_result result
     @response_api_key = nil
-    @status = result.code
-
-    # Authenticators helpers using http_status
     @http_status = result.code
 
     @content_type = result.headers[:content_type]
@@ -145,6 +142,13 @@ module RestHelpers
     UNIXSocket.open socket_file do |sock|
       sock.puts command
       sock.read
+    end
+  end
+
+  def authn_request(url:, api_key:, encoding:, can:)
+    headers["Accept-Encoding"] = encoding
+    try_request can do
+      post_json(url, api_key)
     end
   end
 
@@ -281,7 +285,7 @@ module RestHelpers
   rescue RestClient::Exception
     puts $ERROR_INFO
     @exception = $ERROR_INFO
-    @status = $ERROR_INFO.http_code
+    @http_status = $ERROR_INFO.http_code
     raise if can
     set_result @exception.response  
   end
--- a/cucumber.yml
+++ b/cucumber.yml
@@ -7,6 +7,9 @@ api: >
   --format pretty
   -r cucumber/api/features/support/logs_helpers.rb
   -r cucumber/api/features/step_definitions/logs_steps.rb
+  -r cucumber/_authenticators_common/features/support/conjur_token.rb
+  -r cucumber/_authenticators_common/features/support/authenticator_helpers.rb
+  -r cucumber/_authenticators_common/features/step_definitions/authn_common_steps.rb
   -r cucumber/api
   cucumber/api
 
--- a/cucumber/authenticators_azure/features/authn_azure_performance.feature
+++ b/cucumber/authenticators_azure/features/authn_azure_performance.feature
@@ -34,6 +34,11 @@ Feature: Azure Authenticator - Performance tests
     When I authenticate 1000 times in 10 threads via Azure with token as host "test-app"
     Then The avg authentication request responds in less than 0.75 seconds
 
+  Scenario: successful requests with Accept-Encoding base64
+    And I fetch a non-assigned-identity Azure access token from inside machine
+    When I authenticate 1000 times in 10 threads via Azure with token as host "test-app" with Accept-Encoding header "base64"
+    Then The avg authentication request responds in less than 0.75 seconds
+
   Scenario: Unsuccessful requests with an invalid token
     And I fetch a non-assigned-identity Azure access token from inside machine
     When I authenticate 1000 times in 10 threads via Azure with invalid token as host "test-app"
--- a/cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb
+++ b/cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb
@@ -19,7 +19,7 @@ Given(/I fetch a (non|system|user)-assigned-identity Azure access token from ins
   end
 end
 
-Given(/I authenticate (?:(\d+) times? in (\d+) threads? )?via Azure with (no |empty |invalid )?token as (user|host) "([^"]*)"/) do |num_requests, num_threads, token_state, role_type, username|
+Given(/I authenticate (?:(\d+) times? in (\d+) threads? )?via Azure with (no |empty |invalid )?token as (user|host) "([^"]*)"(?: with Accept-Encoding header "([^"]*)")?/) do |num_requests, num_threads, token_state, role_type, username, header|
   username = role_type == "user" ? username : "host/#{username}"
 
   token = case token_state
@@ -41,10 +41,11 @@ Given(/I authenticate (?:(\d+) times? in (\d+) threads? )?via Azure with (no |em
     num_threads,
     authentication_func: :authenticate_azure_token,
     authentication_func_params: {
-      service_id:  AuthnAzureHelper::SERVICE_ID,
-      account:     AuthnAzureHelper::ACCOUNT,
-      username:    username,
-      azure_token: token
+      service_id:             AuthnAzureHelper::SERVICE_ID,
+      account:                AuthnAzureHelper::ACCOUNT,
+      username:               username,
+      azure_token:            token,
+      accept_encoding_header: header
     }
   )
 end
--- a/cucumber/authenticators_azure/features/support/authn_azure_helper.rb
+++ b/cucumber/authenticators_azure/features/support/authn_azure_helper.rb
@@ -35,12 +35,14 @@ module AuthnAzureHelper
     raise "Failed to fetch azure token with reason: #{err}"
   end
 
-  def authenticate_azure_token(service_id:, account:, username:, azure_token:)
+  def authenticate_azure_token(service_id:, account:, username:, azure_token:, accept_encoding_header:)
     username = username.gsub("/", "%2F")
     path = "#{conjur_hostname}/authn-azure/#{service_id}/#{account}/#{username}/authenticate"
 
     payload = {}
     payload["jwt"] = azure_token
+
+    headers["Accept-Encoding"] = accept_encoding_header
 
     post(path, payload)
   end


app/controllers/authenticate_controller.rb
```diff
@@ -70,7 +70,7 @@ def authenticate
       authenticators: installed_authenticators,
       enabled_authenticators: Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV)
     )
-    render json: authn_token
+    render json: (params[:encoded] ?  authn_token : Base64.strict_encode64(authn_token.to_json))
```
orenbm, line 73: another option is to pass this param into the `Authenticate` command class and perform this encoding (if needed) there. This way we still have the logic inside the command class.

wdyt?

orenbm: it will also be easier to UT

liavyona: Agreed. Thanks.

app/controllers/authenticate_controller.rb
```diff
@@ -68,7 +68,8 @@ def authenticate
     authn_token = Authentication::Authenticate.new.(
       authenticator_input: authenticator_input,
       authenticators: installed_authenticators,
-      enabled_authenticators: Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV)
+      enabled_authenticators: Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV),
+      encoded_response: params[:encoded]
```
orenbm, line 72: this can move to the `authenticator_input` object like we do with the rest of the params.

btw, i don't see that you defined this param in `routes.rb`. did you verify that this works? what do you think about adding an integration test?

app/domain/authentication/authenticate.rb
```diff
@@ -28,7 +28,8 @@ def call
       validate_origin
       validate_credentials
       audit_success
-      new_token
+      token = new_token
+      @encoded_response ? Base64.strict_encode64(token.to_json) : token
```
orenbm, line 32: how about moving this logic inside `new_token`?

app/controllers/authenticate_controller.rb
liavyona: 1. Ok.
2. It works. its optional POST parameter - ..../authenticate?encoded=. I didn't see any example in routes.rb to define it.

orenbm: you are correct, i looked again and you don't need to add anything to `routes.rb`

app/controllers/authenticate_controller.rb
```diff
@@ -83,7 +83,8 @@ def authenticator_input
       username:           params[:id],
       credentials:        request.body.read,
       client_ip:          request.ip,
-      request:            request
+      request:            request,
+      encoded_token:      ! params[:encoded].nil?
```
orenbm, line 87: how will the request look like? can you share the URL?

app/domain/authentication/authenticator_input.rb
```diff
@@ -24,17 +25,17 @@ def update(hash)
 
     def webservice
       @webservice ||= ::Authentication::Webservice.new(
-        account:            @account,
-        authenticator_name: @authenticator_name,
-        service_id:         @service_id
+        account:            account,
+        authenticator_name: authenticator_name,
+        service_id:         service_id
```
orenbm, line 30: why did this change?

app/domain/authentication/authenticator_input.rb
```diff
@@ -24,17 +25,17 @@ def update(hash)
...
       )
     end
 
     # :reek:NilCheck
     def role
-      return nil if @username.nil?
+      return nil if username.nil?
 
-      ::Role.by_login(@username, account: @account)
+      ::Role.by_login(username, account: account)
```
orenbm, line 38: why did this change?

app/domain/authentication/webservice.rb
```diff
@@ -13,7 +13,6 @@
 
 module Authentication
   class Webservice < ::Dry::Struct
-    constructor_type :schema
```
orenbm, line 16: why was this removed?

app/controllers/authenticate_controller.rb
liavyona: ```https://conjur-domain/authn/cucumber/admin/authenticate?encoded=<whatever>```
If encoded is mention the response will be encoded.

app/domain/authentication/webservice.rb
liavyona: https://dry-rb.org/gems/dry-struct/0.7/constructor-types/#constructor-types-were-removed-in-v0-5-0-use-code-transform_types-code-and-code-transform_keys-code-as-a-replacement

app/domain/authentication/authenticator_input.rb
liavyona: API changes. The class variables in this current implementation are nill.

app/domain/authentication/authenticator_input.rb
liavyona: API changes. The class variables in this current implementation are nill.

app/controllers/authenticate_controller.rb
liavyona: Also that's why I prefer to keep the `params[:encoded].nil?` even though the codeclimate says that it's redundant. In my opinion, it clarifies that encoded attribute depends on whether encoded parameter has been submitted

orenbm: well, in that case it seems that it is better to use just `params[:encoded]` because in your implementation a user might set `?encoded=false` and you will encode it.

if yo use `encoded_token: params[:encoded]` then in case it is there you will get the given value and if it is not present then it will be nil (so false)

app/domain/authentication/authenticator_input.rb
orenbm: which api changes? is this because you upgraded `dry`?

app/domain/authentication/webservice.rb
orenbm: don't we need somthing to replace it?

orenbm:
@liavyona looks like the build is failing. the are also codeclimate issues

app/controllers/authenticate_controller.rb
orenbm: > that's why I prefer to keep the params[:encoded].nil? even though the codeclimate says that it's redundant

but it is redundant. it is nil in case it is nil so it is false

liavyona:
> @liavyona looks like the build is failing. the are also codeclimate issues

About the build - I'm working on to fix the issues that popped up from the version upgrade.
About the codeclimate - I left you a comment about the `.nil?`  usage. What do you say? @orenbm

app/domain/authentication/authenticate.rb
liavyona: Done

app/controllers/authenticate_controller.rb
orenbm: @liavyona i answered your question here

app/controllers/authenticate_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'wannabe_bool'
```
orenbm, line 3: what does this gem give us?

liavyona: convert string to bool.
"true" -> true
"yes" -> true
"1" -> true

"false"  -> false
"no" -> false
"0" -> false


<string which has no boolean meaning> -> false

In ruby there is no `to_boolean` to String and we need to convert the `encoded` parameter from the request (a string) to bool

orenbm: why can't we have:
"true" -> true
anything else (including nil) -> false

liavyona: This library does the same thing. including nil. But it converts to true other logical true strings such as yes... 

I think that it is a good place to use this 3 party in order to avoid:
1. `params[:encoded] == "true"` using true as a literal string seems wrong to me.
2. extending String:
``` 
Class String
    def true?
        return self == "true"
 ```

The 3 party extends the `.to_bool` method which is very clear and also take care of nil.

Comment (micahlee) [reply app/controllers/authenticate_controller.rb:3]: I'll second that:
1. We should only add new dependencies (gems) when the value is clear and obvious, and after it's been discussed by the team as a whole. Dependencies come with future maintenance costs, that I think in this case outweighs the value of boolean parsing. [See here for more info](https://github.com/conjurinc/docs/blob/master/reference/debt_free_rules.md#9-new-tools-require-documented-public-discussion-avoid-unnecessary-tools).

2. Keeping the interface simple is preferred than accepting all options for boolean values ([see the discussion here](https://github.com/cyberark/conjur-service-broker/pull/76)).

Comment (jonahx) [reply app/controllers/authenticate_controller.rb:3]: I'm going to third this just emphasize the point, because I think it's so critical.  I also want to spread this information among team members -- not just the information itself but the reasons for it.

When I wrote the ["no new tools without public disucssion"](https://github.com/conjurinc/docs/blob/master/reference/debt_free_rules.md#9-new-tools-require-documented-public-discussion-avoid-unnecessary-tools) rule that Micah referenced, ruby gems were one of the things I had in mind. 

We want a very high bar for adding new gem dependencies.  In the majority of cases, the better decision -- for clarity of code, long term maintenance, and avoiding security vulns -- will be to use vanilla ruby rather than a gem.

In this particular case, there is especially no reason for it because the logic we want can be written in a line or a few lines.

> extending String:

As a separate point, this is something that should _never_ be done, under any circumstances.  Monkey patching built-in classes is extremely brittle, and just asking for trouble.  The standard library classes represent an important invariant that every experienced ruby programmer can be assumed to understand.  Invisibly changing their behavior breaks that invariant.

If you do want a special String or Array, etc object with new methods, the preferred method is to wrap the object in a decorator that adds them.  This makes what you're doing explicit.

File: app/domain/authentication/authenticator_input.rb [hunk 9]
```diff
@@ -6,6 +6,7 @@ module Authentication
 
   # :reek:InstanceVariableAssumption
   class AuthenticatorInput < ::Dry::Struct
+    constructor_type :schema
```
orenbm, line 9: what's this? why is it needed to be added?

cucumber/api/features/authenticate.feature
```diff
@@ -21,6 +21,19 @@ Feature: Exchange a role's API key for a signed authentication token
       cucumber:user:alice successfully authenticated with authenticator authn
     """
 
+  Scenario: Authenticate response should be encoded if asked
+    Then I can POST "/authn/cucumber/alice/authenticate?encoded=true" with plain text body ":cucumber:user:alice_api_key"
```
orenbm, line 25: this should be `When I POST` as this is not what you are testing. you want to verify that `WHEN` you send the request `THEN` the token is encoded

cucumber/api/features/authenticate.feature
```diff
@@ -21,6 +21,19 @@ Feature: Exchange a role's API key for a signed authentication token
...
+    And the HTTP response content type is "text/plain"
```
orenbm, line 26: Then

app/domain/authentication/authenticator_input.rb
liavyona: In order to let the "optional" values to be optionals. The default schema of struct doesn't allow optional keys. reference : https://dry-rb.org/gems/dry-struct/0.7/constructor-types/#constructor-types-were-removed-in-v0-5-0-use-code-transform_types-code-and-code-transform_keys-code-as-a-replacement

cucumber/api/features/step_definitions/request_steps.rb
```diff
@@ -162,6 +162,11 @@
   expect(@content_type).to match(content_type)
 end
 
+Then(/^the HTTP response content type is base64 encoded$/) do
```
orenbm, line 165: this seems incorrect. you don't verify that the content type is base64 encoded.

cucumber/api/features/authenticate.feature
```diff
@@ -21,6 +21,19 @@ Feature: Exchange a role's API key for a signed authentication token
...
+    And the HTTP response content type is base64 encoded
```
orenbm, line 27: i would add also a step in the manner of `Then user "alice" has been authorized by Conjur`, and move the logic behind the step above to be similar to what we do there. 

please see the authn-azure tests for reference.

cucumber/api/features/authenticate.feature
liavyona: ding
I agree. But I used to main authentication test above as an example.. Should I change it too?

cucumber/api/features/authenticate.feature
liavyona: I agree. But I used to main authentication test above as an example.. Should I change it too?

app/domain/authentication/authenticator_input.rb
orenbm: i see, thanks

cucumber/api/features/authenticate.feature
orenbm: it is a different test. in the test above we verify that we can use the api-key to authenticate so the `Then` is correct. here you test that the response is encoded.

cucumber/api/features/step_definitions/request_steps.rb
liavyona: `  decoded_body = JSON.parse Base64.strict_decode64(@result.body)` This is decoding and json parsing. In my opinion that verifies the encoding. I can add keys validation but again I used the main authentication test as an example. I cant change it in both of them

orenbm: yes but it says `the... content type is base64`

orenbm: i think that it is stronger to not only validate that the returned object is a hash but that it is a token for `alice`. please see authn-azure (or any other) for reference. even if the test above doesn't implement it i think it is best that you do.

liavyona: Ok thanks

cucumber/api/features/authenticate.feature
liavyona: Discussed

cucumber/api/features/step_definitions/request_steps.rb
liavyona: The audit isn't enough? @orenbm

orenbm: let me ask you this - if you change the line of code that returns the token (`render content_type => authn_token`) to `render content_type => "liav is the kingggg`, will your current test still pass?

cucumber/api/features/authenticate.feature
```diff
@@ -21,6 +21,20 @@ Feature: Exchange a role's API key for a signed authentication token
       cucumber:user:alice successfully authenticated with authenticator authn
     """
 
+  Scenario: Authenticate response should be encoded if asked
+    When I can POST "/authn/cucumber/alice/authenticate?encoded=true" with plain text body ":cucumber:user:alice_api_key"
```
orenbm, line 25: should be `When I POST`. we use `can` only in `Then`.

cucumber/api/features/step_definitions/request_steps.rb
```diff
@@ -162,6 +162,16 @@
...
 
+Then(/^the HTTP response is base64 encoded$/) do
+  @result = Base64.strict_decode64(@result)
+  expect(JSON.parse(@result).is_a?(Hash)).to be true
+end
+
+Then(/^the user "([^"]*)" is successfully authenticated to Conjur$/) do |username|
+  token = ConjurToken.new(@result)
+  expect(token.username).to eq(username)
+end
```
orenbm, line 173: we already have something similar here: https://github.com/cyberark/conjur/blob/master/cucumber/_authenticators_common/features/step_definitions/authn_common_steps.rb#L1

can we use it?

cucumber/api/features/support/conjur_token.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
```
orenbm, line 1: please use the one that is already implemented

cucumber/api/features/support/conjur_token.rb
```diff
@@ -0,0 +1,19 @@
...
+  def username
+    payload['sub']
+  end
+
+  private
+
+  def payload
+    @payload ||= JSON.parse(Base64.decode64(@token['payload']))
+  end
+end
```
jonahx, line 19: This is a general comment, not anything you need to change:

This is a nice abstraction, and probably belongs in the code itself, and not merely as a test helper.  I imagine (I haven't looked though) that this `Base64` logic is spread out in multiple places across the code base, a perfect example of [connascense of algorithm](https://practicingruby.com/articles/connascence) code smell.

The fix would be to have this object (or one like it) which is used everywhere.

micahlee: @liavyona @jonahx Regarding...

> The fix would be to have this object (or one like it) which is used everywhere.

We do have this token class that we use to create the token: https://github.com/cyberark/slosilo/blob/master/lib/slosilo/jwt.rb

Could this be reused here? Example:
```
Slosilo::JWT.parse_json(raw_token)
```

cucumber/api/features/support/conjur_token.rb
liavyona: Yes there are two already implemented classes such as ConjurToken. But there are in other cucumber profiles' support directory - we can move this into the code itself or find a shared support folder to all cucumber tests

cucumber/api/features/step_definitions/request_steps.rb
liavyona: these step is not accessible from api profile. I'll look for shared steps file solution

app/controllers/authenticate_controller.rb
```diff
@@ -70,7 +70,8 @@ def authenticate(input = authenticator_input)
       authenticators: installed_authenticators,
       enabled_authenticators: Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV)
     )
-    render json: authn_token
+    content_type = input[:encoded_token] ? :plain : :json
```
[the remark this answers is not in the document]
liavyona: I agree with the guideline about the endpoint. But the encoded response is not a json so a separate endpoint sounds reasonable.

cucumber/api/features/support/conjur_token.rb
liavyona: @micahlee  It works great. Thanks.

app/controllers/authenticate_controller.rb
```diff
@@ -95,7 +96,8 @@ def authenticator_input
       username:           params[:id],
       credentials:        request.body.read,
       client_ip:          request.ip,
-      request:            request
+      request:            request,
+      encoded_token:      params[:encoded] == "true"
```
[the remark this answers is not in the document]
liavyona: Fine with it. What do you say @orenbm ?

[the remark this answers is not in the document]
liavyona: How about ..../authenticate/encoded

cucumber/api/features/step_definitions/request_steps.rb
liavyona: So both

cucumber/api/features/step_definitions/request_steps.rb
orenbm: you can load these steps in `cucumber.yml` under the `api` profile. you can see how we do it in other profiles.

cucumber/api/features/support/conjur_token.rb
orenbm: see other comment on how to load these files

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
orenbm: @liavyona can you please investigate on a best practice for this?

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
orenbm: i strongly agree with @uCatu that it should be `.../authenticate?encoded`.
regarding the second endpoint please further investigate

cucumber/api/features/step_definitions/request_steps.rb
liavyona: yes already found and updated

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
liavyona: Which brings us to previous comments here... What's the behaviour when '.../authenticate?encoded=false` or any value to encoded?

app/domain/authentication/authenticator_input.rb
```diff
@@ -14,6 +15,7 @@ class AuthenticatorInput < ::Dry::Struct
     attribute :credentials, ::Types::String.optional
     attribute :client_ip, ::Types::NonEmptyString
     attribute :request, ::Types::Any
+    attribute :encoded_token, ::Types::Bool.optional.default(false)
```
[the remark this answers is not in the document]
liavyona: I can see why (nil is same as false), but isn't it much cleaner to refer encoded as true and false? instead of true and nil?

app/domain/authentication/authenticate.rb
```diff
@@ -106,10 +106,11 @@ def audit_role_id
     end
 
     def new_token
-      @token_factory.signed_token(
+      token = @token_factory.signed_token(
         account:  account,
         username: username
       )
+      encoded_token ? Base64.strict_encode64(token.to_json) : token
```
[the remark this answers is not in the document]
liavyona: why encoded64 isn't enough? The client should support the `urlsafe_decode` which is not supported by base64 bin

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
liavyona: endpoint should stick with its content-type. btw, requests also ( requests and response type should be the same). But this is out of this PR's scope

cucumber/api/features/authenticate.feature
liavyona: fixed

cucumber/api/features/support/conjur_token.rb
liavyona: fixed

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
orenbm: i think it's ok to document it just with `?encoded` and assume that if it used incorrectly (e.g `?encoded=false`) then it will not behave as expected.

i think it's ok to not support `encoded=false`

app/domain/authentication/authenticator_input.rb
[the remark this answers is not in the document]
orenbm: not if the only valid value is that it's there (e.g `?encoded`)

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
orenbm: why is it out of the scope of this PR? this PR adds a new type of response and should be handled in the proper way. if the best practice is to have just one content type then the correct implementation of this PR is to add a new endpoint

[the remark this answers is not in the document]
liavyona: I talked about one content-type to requests and response throughout the entire project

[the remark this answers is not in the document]
orenbm: so i will ask again - is it considered a bad practice to have one endpoint return different types of response depending on the input?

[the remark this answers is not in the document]
liavyona: Yes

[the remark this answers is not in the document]
orenbm: well in that case, let's create a new endpoint and not introduce a bad practice. Sorry that we got to this point after you already implemented this but it's better to catch it now rather than when it's in master.

let me know if you need any assistance in implementing a new endpoint (e.g `.../authenticate/encoded`)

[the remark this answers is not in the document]
uCatu: We can open it for discussion (with architact?) since it's best practice, and not an error to implement it as such it can go either way - so we can hold on with implementing another endpoint, @liavyona do we have another endpoint that act the same?

cucumber/api/features/step_definitions/request_steps.rb
```diff
@@ -162,6 +162,12 @@
   expect(@content_type).to match(content_type)
 end
 
+Then(/^the HTTP response is base64 encoded$/) do
+  # Override encoded response with decode one to use other helpers
+  @result = @response_body = Base64.strict_decode64(@result)
+  expect(JSON.parse(@result).is_a?(Hash)).to be true
+end
```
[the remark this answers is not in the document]
liavyona: 1. Some steps use `@result` as request_steps.rb,  and some use `@response_body` as authenticator_helpers.rb. Same happens with `@status` and `@http_status`. In order to prevent code reuse (of steps), I needed to use both `@result` and `@respone_body`. About the naming - valid base64 should be decode successfully. I need to override the response with decoded response in order to use other steps as the comments suggests.

2. A valid decoded token should be parsed by JSON. So if it fails during parsing, the test does its job. Do you prefer custom exception instead?

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
liavyona: > well in that case, let's create a new endpoint and not introduce a bad practice. Sorry that we got to this point after you already implemented this but it's better to catch it now rather than when it's in master.
> 
> let me know if you need any assistance in implementing a new endpoint (e.g `.../authenticate/encoded`)

It's Ok. Just adding it to route.rb . Thanks.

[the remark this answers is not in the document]
liavyona: > We can open it for discussion (with architact?) since it's best practice, and not an error to implement it as such it can go either way - so we can hold on with implementing another endpoint, @liavyona do we have another endpoint that act the same?

I'll look for this kind of behaviour. But I'm pretty sure there are no such cases (from search "render" in the project)

app/domain/authentication/authenticate.rb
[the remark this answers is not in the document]
liavyona: I'll change

app/controllers/authenticate_controller.rb
```diff
@@ -70,7 +70,15 @@ def authenticate(input = authenticator_input)
       authenticators: installed_authenticators,
       enabled_authenticators: Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV)
     )
-    render json: authn_token
+    content_type = :json
+    if encoded_response?
+      # The response will be a base64 encoded version of the JWT.
+      Rails.logger.debug("Responding with base64 encoded JWT")
```
orenbm, line 76: can you move this message to `logs.rb` as we do with other log messages?

orenbm: @InbalZilberman @shulifink please approve message

app/controllers/authenticate_controller.rb
```diff
@@ -70,7 +70,15 @@ def authenticate(input = authenticator_input)
...
+      content_type = :plain
+      authn_token = ::Base64.encode64(authn_token.to_json).delete("\n").strip
```
orenbm, line 78: extract to method?

cucumber/api/features/step_definitions/request_steps.rb
```diff
@@ -103,6 +103,14 @@
   end
 end
 
+When(/^I authenticate Alice with Accept-Encoding header "([^"]*)"(?: with plain text body "([^"]*)")?$/) do |header, body|
```
orenbm, line 106: can we use the existing step for this? looks like you can add there an optional header

orenbm (changes requested):
left some comments

app/controllers/authenticate_controller.rb
```diff
@@ -70,7 +70,15 @@ def authenticate(input = authenticator_input)
       authenticators: installed_authenticators,
       enabled_authenticators: Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV)
     )
-    render json: authn_token
+    content_type = :json
+    if encoded_response?
+      # The response will be a base64 encoded version of the JWT.
```
sigalsax, line 75: IMO this comment is a bit redundant

app/controllers/authenticate_controller.rb
uCatu: What do you think on https://apidock.com/rails/String/squish ? is it solving the issue?

uCatu: the `to_json` is mendetory?

app/controllers/authenticate_controller.rb
```diff
@@ -214,4 +222,8 @@ def installed_authenticators
   def enabled_authenticators
     Authentication::InstalledAuthenticators.enabled_authenticators(ENV)
   end
+
+  def encoded_response?
+    request.accept_encoding&.include?("base64")
```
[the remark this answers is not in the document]
rafis3:
```suggestion
    request.accept_encoding&.casecmp?("base64")
```
I think it should use `casecmp` because you want to check specifically that it is equal to `base64` (case is insensitive), not that it contains it, among other characters.

app/controllers/authenticate_controller.rb
liavyona: > What do you think on https://apidock.com/rails/String/squish ? is it solving the issue?

Cool! we need to add `require 'active_support/core_ext/string/filters'`. Is that OK? @uCatu

liavyona: > the `to_json` is mendetory?

Yes.  We encode the json format.

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -99,6 +99,10 @@ def strip_rest_params_from_path path
   def set_result result
     @response_api_key = nil
     @status = result.code
+
+    # Authenticators helpers using http_status
+    @http_status = result.code
```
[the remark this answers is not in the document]
liavyona: Some helpers use `@status` some use `@http_status`. Same with `@result` and `@response_content`. I need to use helpers in order to avoid code dups and this is why I need to align with this.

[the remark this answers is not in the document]
orenbm: can't we align the usage between all helpers? so they all use only one?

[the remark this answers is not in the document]
liavyona: I I think that it is out of scope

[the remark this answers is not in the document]
orenbm: it really shouldn't be hard to do. just another small commit that will not require you to add a new variable.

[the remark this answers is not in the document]
liavyona: This is only 2 variables which are relevant to this PR. I think they might be more. This might affect all the integration tests and won't be in the scope of just one commit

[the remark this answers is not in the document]
orenbm: i really don't understand why we can't just replace these occurences from `@http_status` to `@status`.
image

cucumber/api/features/step_definitions/request_steps.rb
liavyona: I'm doing manipulations over the body and using `try_request false` to test the `401`case. Think that it's an overhead. Especially when we have `with parameters`, `with body`, `with username`, `with no Content-Header

cucumber/api/features/support/rest_helpers.rb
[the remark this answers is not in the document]
liavyona: In the common helpers it is being used without `@` 
image
and so as `result`

app/controllers/authenticate_controller.rb
shulifink: fine

cucumber/api/features/support/rest_helpers.rb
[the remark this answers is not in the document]
orenbm: that still the same. it's a small find and replace in a separate commit.

cucumber/api/features/step_definitions/request_steps.rb
orenbm: i will not block this PR for this but still think it can be generalized

app/domain/logs.rb
```diff
@@ -44,6 +44,11 @@ module Authentication
       code: "CONJ00030D"
     )
 
+    EncodedJWTResponse = ::Util::TrackableLogMessageClass.new(
+      msg:  "Responding with base64 encoded JWT",
```
orenbm, line 48: @InbalZilberman is this message ok? do we want something like `Returning a base64 encoded access token`?

cucumber/api/features/authenticate.feature
```diff
@@ -21,6 +21,58 @@ Feature: Exchange a role's API key for a signed authentication token
...
+    And user "alice" has been authorized by Conjur
+    And there is an audit record matching:
+    """
+      <86>1 * * conjur * authn
+      [auth@43868 authenticator="authn"]
+      [subject@43868 role="cucumber:user:alice"]
+      [action@43868 operation="authenticate" result="success"]
+      cucumber:user:alice successfully authenticated with authenticator authn
+    """
+  Scenario: Authenticate response should be encoded if Accept-Encoding includes base64
```
orenbm, line 37: add line between scenarios

cucumber/api/features/authenticate.feature
```diff
@@ -21,6 +21,58 @@ Feature: Exchange a role's API key for a signed authentication token
...
+    And user "alice" has been authorized by Conjur
+    And there is an audit record matching:
+    """
+      <86>1 * * conjur * authn
+      [auth@43868 authenticator="authn"]
+      [subject@43868 role="cucumber:user:alice"]
+      [action@43868 operation="authenticate" result="success"]
+      cucumber:user:alice successfully authenticated with authenticator authn
+    """
+  Scenario: Authenticate response should be encoded if Accept-Encoding includes base64 with mixed case and spaces
```
orenbm, line 50: add line between scenarios

cucumber/api/features/authenticate.feature
```diff
@@ -37,6 +89,17 @@ Feature: Exchange a role's API key for a signed authentication token
       [action@43868 operation="authenticate" result="failure"]
       cucumber:user:alice failed to authenticate with authenticator authn
     """
+  Scenario: Attempting to use an invalid API key to authenticate with Accept-Encoding base64 result in 401 error
```
orenbm, line 92: add line between scenarios

cucumber/api/features/authenticate.feature
```diff
@@ -21,6 +21,58 @@ Feature: Exchange a role's API key for a signed authentication token
...
+    And there is an audit record matching:
+    """
+      <86>1 * * conjur * authn
+      [auth@43868 authenticator="authn"]
+      [subject@43868 role="cucumber:user:alice"]
+      [action@43868 operation="authenticate" result="success"]
+      cucumber:user:alice successfully authenticated with authenticator authn
+    """
+
+  Scenario: Authenticate response should be json if Accept-Encoding doesn't include base64
```
orenbm, line 64: what about a test where `Accept-Encoding` is not present that validates the content-type is `application/json`?

liavyona: Other test do that also we have a test that sends `Accept-Encoding` without `base64` that verifies `application/json` returned

cucumber/api/features/authenticate.feature
liavyona: Done

cucumber/api/features/authenticate.feature
liavyona: Done

cucumber/api/features/authenticate.feature
liavyona: Done

cucumber/api/features/authenticate.feature
orenbm: thanks, i thought that you added this step.

cucumber/api/features/support/rest_helpers.rb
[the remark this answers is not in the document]
liavyona: I agree that find and replace will do the work. But the main problem will be tests failing due to the change and if we change it. Also we should find other examples of dups like that  and it can be a good valid new. Again - out of scope of the PR.

cucumber/api/features/authenticate.feature
liavyona: I did add it.... authenticate.feature:66

app/domain/logs.rb
liavyona: https://github.com/cyberark/conjur/pull/1710#discussion_r477150181
BTW, @shulifink said it's fine...

cucumber/api/features/support/rest_helpers.rb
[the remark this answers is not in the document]
uCatu: @orenbm If it's OK with you, lets fix it in a "tinny" PR, so we don't hold this feature before code freeze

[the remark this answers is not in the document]
orenbm: i'm fine with the use-case and against the idea in general. go for it.

app/controllers/authenticate_controller.rb
```diff
@@ -214,4 +221,10 @@ def installed_authenticators
   def enabled_authenticators
     Authentication::InstalledAuthenticators.enabled_authenticators(ENV)
   end
+
+  def encoded_response?
+    return false unless request.accept_encoding.is_a?(String)
```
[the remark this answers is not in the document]
uCatu: - Why are we checking if it's String, what is the alternative?
- Is it possible usecase to not having `accept_encoding` header leading us to NPE?

uCatu (commented):
See comments please

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
liavyona: Returns the Base64-encoded version of bin. This method complies with RFC 4648. No line feeds are added.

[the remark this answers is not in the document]
liavyona: In rspec flow the request has `nil` value for `request.accept_encoding`.
In integration and realtime when not supplied the value will be `""`.

app/controllers/authenticate_controller.rb
```diff
@@ -70,7 +70,14 @@ def authenticate(input = authenticator_input)
       authenticators: installed_authenticators,
       enabled_authenticators: Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV)
     )
-    render json: authn_token
+    content_type = :json
+    if encoded_response?
+      Rails.logger.debug(LogMessages::Authentication::EncodedJWTResponse.new)
```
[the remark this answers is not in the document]
orenbm: you are correct, we can use `logger`

app/controllers/authenticate_controller.rb
```diff
@@ -70,7 +70,14 @@ def authenticate(input = authenticator_input)
...
+      content_type = :plain
+      authn_token = ::Base64.encode64(authn_token.to_json).squish
```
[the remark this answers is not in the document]
liavyona: It works but we use token = token && JSON.parse(Base64.decode64(token)) to load the given token

cucumber/api/features/support/rest_helpers.rb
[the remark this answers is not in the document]
liavyona: I already change to `http_status` in all api profile tests..... @uCatu

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
liavyona: Changed

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
liavyona: And still I think youre right. Changed

orenbm (approved):
Looks good, thanks @liavyona !

please wait for an approval from @InbalZilberman on the log message before merging.

and merge using squash please.

app/domain/logs.rb
InbalZilberman: Im with @orenbm it is not regarding english it is regarding what we are returning. We call it access token in all other places.

orenbm: @liavyona can you change it to `Responding with a base64 encoded access token` and get @shulifink 's approval please?

cucumber/api/features/step_definitions/request_steps.rb
```diff
@@ -103,6 +103,31 @@
...
+  body ||= ":cucumber:user:alice_api_key"
+  authn_request(
+    url: "/authn/cucumber/alice/authenticate",
+    api_key: body,
+    encoding: header.downcase.include?("base64"),
+    can: false
+  )
+end
+
+When(/^I authenticate Alice "([^"]*)" times in "([^"]*)" threads with Accept-Encoding: base64 header?$/) do |requests_num, threads_num|
```
orenbm, line 116: these two methods can be merged into one. please see this for reference: https://github.com/cyberark/conjur/blob/master/cucumber/authenticators_azure/features/step_definitions/authn_azure_steps.rb#L22

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -144,6 +145,13 @@ def authn_local_request command
     end
   end
 
+  def authn_request(url:, api_key:, encoding:, can:)
+    headers["Accept-Encoding"] = encoding ? "base64" : ""
```
orenbm, line 149: shouldn't this be nil in case the encoding is not `base64`? something like:
```
if encoding
  headers["Accept-Encoding"] = "base64"
end
```

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -144,6 +145,13 @@ def authn_local_request command
     end
   end
 
+  def authn_request(url:, api_key:, encoding:, can:)
```
orenbm, line 148: looks like `encoding` is actually a boolean and not an encoding value. i think that `accept_base64_encoding` is better.

cucumber/api/features/step_definitions/request_steps.rb
```diff
@@ -103,6 +103,31 @@
...
+  body ||= ":cucumber:user:alice_api_key"
+  authenticate_with_performance(
+    requests_num,
+    threads_num,
+    authentication_func: :authn_request,
+    authentication_func_params: {
+      url: "/authn/cucumber/alice/authenticate",
+      api_key: body,
+      encoding: true,
+      can: true
```
orenbm, line 126: what is the usage of `can`? i don't understand it.

orenbm (changes requested):
left some comments

cucumber/api/features/support/rest_helpers.rb
liavyona: Tried to avoid codeclimate message by put a value for `else`. I will revert this change.

cucumber/api/features/step_definitions/request_steps.rb
liavyona: `try_request` uses `can` boolean parameter to  whether to raise exception if the http call failed or keep the test going (ended with error code)

cucumber/api/features/step_definitions/request_steps.rb
liavyona: In the second case, the Accept-Encoding is not optional ...Still. will change

app/domain/logs.rb
liavyona: Changed. WDYS @shulifink ?

cucumber/api/features/step_definitions/request_steps.rb
liavyona: CHanged

cucumber/api/features/support/rest_helpers.rb
liavyona: Not a boolean anymore

orenbm (approved):
LGTM, thanks!

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
orenbm: i agree with @uCatu here. the `.is_a?(String)` is redundant and should be removed

orenbm (changes requested):
left a small comment

app/controllers/authenticate_controller.rb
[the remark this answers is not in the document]
liavyona: Changed

orenbm (dismissed):

uCatu (approved):
LGTM

orenbm (approved):

uCatu (approved):
LGTM

uCatu (dismissed):
LGTM * 2

orenbm (approved):

PR #1723, opened 2020-08-03
GCP authenticator solution design

nessiLahav:
Adding a GCP authenticator solution design 

Readable version can be found [here](https://github.com/cyberark/conjur/blob/gcp-solution-design/design/authenticators/authn_gcp/authn_gcp_solution_design.md)


design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+## Issue description
+We want enable GCE hosts to authenticate Conjur with their google identity in order to retrieve secrets from Conjur.
+
+## Out of scope
+The only authentication strategy we are supporting is Environment-provided service account in google Compute engine.
+Other strategies are not supported: Service account key, OAuth 2.0 client, API key
+Other google components are not supported:  App Engine, GKE, Cloud Run, Cloud Functions
+
+## Solution
+We will add a new GCP authenticator to Conjur, which GCE hosts will can authenticate with by delivering their host Conjur identity and google identity as JWT token.
```
oburstein-hub, line 51: hat is the structure of JWT Token ? Is it string - if yes please explain its structure.
If it is struct - please tell what fields does it contains

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+
+As mentioned in the [feature doc](https://ca-il-confluence.il.cyber-ark.com/display/rndp/Conjur+GCP+Authenticator#/), before any authentication 
+request is sent to Conjur, the admin will load the authenticator policy: 
+
+```yaml
+# policy id needs to match the convention `conjur/authn-azure/<service ID>`
+- !policy
+  id: conjur/authn-gcp/mygcp
```
oburstein-hub, line 59: please explain the logic behind this id

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+  - !group apps
+  
+  - !permit
+    role: !group apps
+    privilege: [ read, authenticate ]
+    resource: !webservice
+```
+
+All the following configurations will have constants values and will not require any action from the customer:
+
```
oburstein-hub, line 72: where are they going to be stored ?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+* **provider-uri** = `https://www.googleapis.com/oauth2/v1/certs` #represnts the address to fetch the public key in order to validate the JWT token
+* **mandatory-claims** = `iss, exp, aud, iat, nbf` #the list of the mandatory claims to validate in the JWT token
+* **iss** = `https://accounts.google.com` #the issuer claim we should validate in the JWT token
+* **aud** = `conjur` #the audience claim we should validate in the JWT token
+
+### GCP Resource Restrictions
+To authenticate with Conjur, GCP-specific fields will need to be provided in the host annotations of the Conjur host identity with the following options:
+
+* **project_id** - A customizable unique identifier for your project.
```
oburstein-hub, line 81: is it string ? what are its limitations

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+* **instance_name** - The name of the GCE machine
+* **service_account_id** - The google instance identity   
+* **service_account_email** - The email of the instance identity
+
```
oburstein-hub, line 85: what are validation of above fields ?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+At least one of the above should be provided, if more than 1 is provided, we will authenticate them with AND logic.
+
```
oburstein-hub, line 87: validation should be added for this

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+### Environment-provided service account google identity
+In order to authenticate against Conjur GCP authenticator, a GCE instance first need to fetch its environment-provided identity.
+
+**Example:** 
+Run the following from the GCE machine:
+
+`curl \ --header Metadata-Flavor: Google" \ --get \ --data-urlencode "audience=conjur" \ --data-urlencode "format=full" \ "http://metadata/computeMetadata/v1/instance/service-accounts/default/identity`
+
+The audience value must to be `conjur`
```
oburstein-hub, line 122: please, explain why 'conjur'

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+    }
+  },
+  "iat": 1595157038,
+  "iss": "https://accounts.google.com",
+  "sub": "110987294251917851298"
+}
+```
+
+The following is a mapping of Host annotations with the values we will be extracting in the google JWT token:
+
```
oburstein-hub, line 150: do we have such mappings in other authenticators ?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+|  | |
+|----------|-----------------|
+| **Header**          |  Content-Type: application/x-www-form-urlencoded               |
+| **Body**         |  The body must include the GCP jwt token for GCE instance. jwt=eyJhbGciOiJSUzI1NiIs......uTonCA               |
+
+### Design
+When we got the requirement to implement GCP authenticator the first thought was: if we had a generic JWT authenticator we could support it, 
+maybe not we the best UX but sure we could present a proof of concept.
+
```
oburstein-hub, line 171: Agree

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+![Authn gcp class diagram](authn-gcp-class-diagram.png)
+
+GCP & JWT will have 3 dependencies which will change their behavior by dependency injection  
+
+1. **Fetch Authenticator configuration**
+
+    1.1 **Fetch provider uri**
+    
+    * GCP implementation - return an hardcoded
+    * Authenticator policy implementation - fetch from provider uri value in the policy, this logic is relevant for: OIDC, Azure, JWT
```
oburstein-hub, line 206: where is it going to be kept in policy ?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+      * Azure - not relevant
+      * JWT design phase - return the values of optional claims from variable in policy
+      
+      1.2.4 **Fetch optional claims values**
+      * JWT design phase - return the values of optional claims from variable in policy
+      
+      1.2.5 **authn prefix parameter** - the authenticator prefix which will later be use the fetch a the host annotations 
+
+2. **Validate and decode** - same code a today with small refactoring of sending also the claims values if necessary 
+
```
oburstein-hub, line 235: At which stage is going to be executed the above validation ?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+2. **Validate and decode** - same code a today with small refactoring of sending also the claims values if necessary 
+    
+    * Fetch public key
+    * Validate signature
+    * Validate claims
+    * Save key in cache 
+
+3. **Validate resource restrictions** - Responsible to validate the resource restrictions which defined in conjur, the idea is to refactor Azure and K8S to work with this pattern 
+
```
oburstein-hub, line 242: It the Azure refactoring going to be a part of current project ?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,313 @@
...
+## Issue description
+We want enable GCE hosts to authenticate Conjur with their google identity in order to retrieve secrets from Conjur.
+
+## Out of scope
+The only authentication strategy we are supporting is Environment-provided service account in google Compute engine.
+Other strategies are not supported: Service account key, OAuth 2.0 client, API key
+Other google components are not supported:  App Engine, GKE, Cloud Run, Cloud Functions
+
+## Solution
+We will add a new GCP authenticator to Conjur, which GCE hosts will can authenticate with by delivering their host Conjur identity and google identity as JWT token.
```
eranha, line 51: `can` `will`
by presenting their...

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,313 @@
...
+  - !permit
+    role: !group apps
+    privilege: [ read, authenticate ]
+    resource: !webservice
+```
+
+All the following configurations will have constants values and will not require any action from the customer:
+
+* **provider-uri** = `https://www.googleapis.com/oauth2/v1/certs` #represnts the address to fetch the public key in order to validate the JWT token
+* **mandatory-claims** = `iss, exp, aud, iat, nbf` #the list of the mandatory claims to validate in the JWT token
```
eranha, line 74: Google identity token does not contain `nbf` claim.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,313 @@
...
+* **iss** = `https://accounts.google.com` #the issuer claim we should validate in the JWT token
+* **aud** = `conjur` #the audience claim we should validate in the JWT token
```
eranha, line 76: We may ask to inclide Conjur host in the `aud` claim.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: Projcet id is string, what do yo mean by limitations?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: The fileds should match the host annotaitons.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: Again I suggest to add the host id in Conjur to the audience claim.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,313 @@
...
+  },
+  "iat": 1595157038,
+  "iss": "https://accounts.google.com",
+  "sub": "110987294251917851298"
+}
+```
+
+The following is a mapping of Host annotations with the values we will be extracting in the google JWT token:
+
+| Host annotation                                                      | Azure access token             |
```
eranha, line 151: Azure?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: Where the mapping is stored?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,313 @@
...
+    1.1 **Fetch provider uri**
+    
+    * GCP implementation - return an hardcoded
+    * Authenticator policy implementation - fetch from provider uri value in the policy, this logic is relevant for: OIDC, Azure, JWT
+    
+    1.2 **Fetch claims**
+        
+      1.2.1 **Fetch mandatory claims keys to validate** - list of claims that must exist in token
+        
+      * GCP implementation - return an hardcoded list (Iss, exp, aud, iat, nbf)
```
eranha, line 212: `nbf` is not part of th identity token.

jonahx:
@uCatu I haven't finished reviewing the class diagram, but I wanted to say: This is a model writeup and I really appreciate how easy it was to read.  I especially like how you took the time to explain the context before jumping into the technical details.  I plan to finish reviewing the technical proposal in the morning.

uCatu:
> @uCatu I haven't finished reviewing the class diagram, but I wanted to say: This is a model writeup and I really appreciate how easy it was to read. I especially like how you took the time to explain the context before jumping into the technical details. I plan to finish reviewing the technical proposal in the morning.

Thanks @jonahx, @nessiLahav has been my protege for a long time and it's nice to finally get recognision for all the hard work and the time I'm putting up with him. 

But all the kudos should go to @nessiLahav 🥇

design/authenticators/authn_gcp/authn_gcp_solution_design.md
oburstein-hub: I meant if the string should have any specific structure

design/authenticators/authn_gcp/authn_gcp_solution_design.md
oburstein-hub: I dont know - I though you would know..

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: The is an example to JWT token under section `Environment-provided service account google identity`

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: This example is from the feature doc and its alligned with all other authneticators, see the comment above:
`conjur/authn-gcp/<service ID>` 
authn-gcp = name of authenticator
mygcp = service id

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: As constants in the code

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: Yes in azure authenticator we have a similar logic, the `KeyAdapter` class from the design diagrams is responsible to perform the mapping logic.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: Yes

design/authenticators/authn_gcp/authn_gcp_solution_design.md
oburstein-hub: @nessiLahav can you please share design diagrams with me via slack ? I cannot open them from here..

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: See the sequence diagram (i didnt elaborate about this step becasue i do not going to change it, but it is written there)

oburstein-hub: @nessiLahav - please send me link to the sequence diagram in slack

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: No, i want to suggest to do it at the moment we will finish this feature

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: thanks

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: thanks i will change it

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: you should take it with Inbal it is not part of the requierments

design/authenticators/authn_gcp/authn_gcp_solution_design.md
oburstein-hub: @nessiLahav - but you mean here that resource restriction validation for GCP will be a part of this feature, right ?

oburstein-hub:
Comments regarding static diagram:
1. Please, explain what does FetchClaims class do (along with other FetchClaims classes)
2. Please GCP Authenticator and JWT Authenticator connection lines to others not to overlap in drawing - otherwise its hard to understand

oburstein-hub:
Regarding sequence diagram:
1. It is not a good practice to create diagram that cannot be observed without zoom-in. In order to solve it I would suggest to separate it to 2 different diagrams, or to skip number of classes in it
2. I think that configuration validation should be a part of policy creation step - which should be described in a sequence diagram as well

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: `service id` is the GCP Service Accont ID?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: Project ID: A customizable unique identifier for your project.

The project ID is a unique, user-assigned ID that can be used by Google APIs. If you do not specify a project ID during project creation, a project ID will be generated automatically.

**The project ID must be a unique string of 6 to 30 lowercase letters, digits, or hyphens. It must start with a letter, and cannot have a trailing hyphen. You cannot change a project ID once it has been created.**

Some words are restricted from use in project IDs. If you use restricted words in the project name, such as google or ssl, the generated project ID will not include these words.
[Link](https://cloud.google.com/resource-manager/docs/creating-managing-projects)

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: The filed gets validated against an annotaitons.

jonahx:
Apologies, @nessiLahav!  Missed that you were a main contributor here.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,582 @@
...
+|           |           |          |          |             |
+|           |           |          |          |             |
+
+## Logs
+
+### Error Log level
+
+|    | Scenario                                                                | Log message                                                                            | Comment            |
+|--- |-----------------------------------------------------------------------  |----------------------------------------------------------------------------------------|------------------------|
+| 1  | Authenticator is not enabled (in DB/ENV)                                | Authenticator '{0-authenticator-name}' is not enabled                                  | |
```
InbalZilberman, line 314: @nessiLahav can you please expand the scenario saying that one called an authenticator that is disabled?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,582 @@
...
+| 2  | Webservice is not defined in a Conjur policy                            | Webservice '{0-webservice-name}' wasn't found                                          | |
+| 3  | Host is not permitted to authenticate with the webservice               | '{0-role-name}' does not have 'authenticate' privilege on {1-service-name}             | |
+| 4  | Host is not defined in Conjur                                           | '{0-role-name}' wasn't found                                                           | |
+| 5  | Couldn't make connection with Google Identity Provider in time          | Google Identity Provider failed with timeout error (Provider URI: '{0}'). Reason: '{1}'| |
```
InbalZilberman, line 318: @nessiLahav can we specify what is the timeout - BTW is it configurable somewhere? do we log the uri we called?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,582 @@
...
+| 6  | Failed to confirm Google token signature                                | Failed to confirm signature of '{0-token}' issued by (Provider URI: '{1}'              | |
+| 7  | Failed to validate Google token claims                                  | Will be decided in Impl phase              | |
```
InbalZilberman, line 320: @nessiLahav please let me review when times come

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,582 @@
...
+| 8  | Authentication request body is missing a field (e.g `jwt_token`)        | Field '{0-field-name}' is missing or empty in request body                             | |
+| 9  | A required annotation is missing in the Conjur host                     | Will be decided in Impl phase                                                          | Annotation is missing for authentication |
```
InbalZilberman, line 322: @nessiLahav please let me review when times come. why is it different from other authenticators?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,582 @@
...
+| 10 | Resource Restrictions includes an illegal constraints                   | Will be decided in Impl phase                                                          | Host contains an un permitted annotation |
+| 11 | Resource Restrictions defined in Conjur host doesn't match JWT token    | Resource Restrictions field '{0-field-name}' does not match JWT token                  | |
```
InbalZilberman, line 324: @nessiLahav I dont understand the meaning of "Resource restriction field" you mean annotation? if so lets write "application identity  annotation"

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: Those message are aligned with Azure, at the development stage you will review them again

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: Those message are aligned with Azure, at the development stage you will review them again
Yes we log the privder uri
the timeout is not configurable

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: Those message are aligned with Azure, at the development stage you will review them again

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: i think it is overkill to review the log messages now, we will aligned them to other authenticators and will review them again at the development phase

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: Yes

nessiLahav:
> Please, explain what does FetchClaims class do (along with other FetchClaims classes)
> Please GCP Authenticator and JWT Authenticator connection lines to others not to overlap in drawing - otherwise its hard to understand

@oburstein-hub i think all the answers are already in the design, please lets speak today

nessiLahav:
> What you can do instead is make the interface more general (essentially just the function I mentioned or something similar), and still provide the other classes listed in the subsections of 3 for the existing implementations to use. That is: make our interface more abstract while still providing utilities for existing implementations that can be shared. Currently, I think, the interface is being too tightly specified.

@jonahx  So if I understand your suggestion:
You suggest to change `ValidateResourceRestrictions` class (we can focus on the name later, just want to understand the idea) :
**From:** being shared will all authenticators with tightly logic implementation 
**To:**  duplicate this class `ValidateResourceRestrictions`  for each authenticator specific needs and expose one general function (valid?) that get as inputs all the 3 dependencies that I mentioned in design

An example to the code change:
```

            def validate_resource_restrictions
              @validate_gcp_resource_restrictions.valid(
                 authenticator_input:                @authenticator_input,
                 authenticator_configuration:        @authenticator_configuration,
                 extract_source_restrictions:        ExtractHostAnnotationRestrictions.new,
                 validate_source_restrictions:       ValidateGCPSourceRestrictions.new,
                 validate_destination_restrictions:  ValidateJWTTokenRestrictions.new,
                 key_adapter:                        GCPKeyAdapter
              )
            end
```

So if the I understand you correctly this approach maybe will duplicate a little bit code but will gain us a lot of flexibility for future changes

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: No, service id is for crating multiple authn gcp authenticators in conjur

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,312 @@
...
+|--------------------------------------------------------------------|-------------------------------------------|-------------------------------------------------------------------------------------|
+|     api                                                            |     authn-jwt/                            |     authn-gcp/                                                                      |
+|     Mandatory base claims to validate                              |     Iss, exp, nbf (Configurable)          |     iss, exp, aud, iat, nbf (Constants)                                           |
+|     Optional base claims to validate (if exists in the token)      |     aud,sub,iat (Configurable)            |     -                                                                               |
+|     provider-uri                                                   |     Configurable   (list?)                |     Constant                                                                        |
```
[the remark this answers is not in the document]
nessiLahav: Since it is just a suggestion for JWT design that should be as generic and flexible as possible, i tought that it should be able to handle in a special scenrios where for some reason you will need more than 1 value.
But you always be able to confgiure it with only 1 value

jonahx:
> > What you can do instead is make the interface more general (essentially just the function I mentioned or something similar), and still provide the other classes listed in the subsections of 3 for the existing implementations to use. That is: make our interface more abstract while still providing utilities for existing implementations that can be shared. Currently, I think, the interface is being too tightly specified.
> 
> @jonahx So if I understand your suggestion:
> You suggest to change `ValidateResourceRestrictions` class (we can focus on the name later, just want to understand the idea) :
> **From:** being shared will all authenticators with tightly logic implementation
> **To:** duplicate this class `ValidateResourceRestrictions` for each authenticator specific needs and expose one general function (valid?) that get as inputs all the 3 dependencies that I mentioned in design
> 
> An example to the code change:
> 
> ```
> 
>             def validate_resource_restrictions
>               @validate_gcp_resource_restrictions.valid(
>                  authenticator_input:                @authenticator_input,
>                  authenticator_configuration:        @authenticator_configuration,
>                  extract_source_restrictions:        ExtractHostAnnotationRestrictions.new,
>                  validate_source_restrictions:       ValidateGCPSourceRestrictions.new,
>                  validate_destination_restrictions:  ValidateJWTTokenRestrictions.new,
>                  key_adapter:                        GCPKeyAdapter
>               )
>             end
> ```
> 
> So if the I understand you correctly this approach maybe will duplicate a little bit code but will gain us a lot of flexibility for future changes
> 
> Note: Also possible to move the dependencies inside to the Class definition

@nessiLahav 

First, a point of clarification:

This might be partly confusing because ruby doesn't have explicit interfaces like Java or most other static langs.  So when we're drawing a class diagram with "interfaces" those interfaces only exist implicitly, as duck-typed objects which implement the interface methods in question.  Also, in most (or all) of your examples the interfaces are just single method interfaces (ie, function objects with a `call` method).

design/authenticators/authn_gcp/authn_gcp_solution_design.md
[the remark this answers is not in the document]
InbalZilberman: @nessiLahav I understand that you would like to build a state of the art framework. but this authenticator is also user facing.
Having several uris, several iss and several auds is confusing... can you give a real use case for that?
I see here also a security issue if they are not tupled.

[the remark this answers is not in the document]
InbalZilberman: @sashaCher @oburstein-hub WDYT?
@Tovli @nessiLahav as we are not releasing any JWT authenticator to users anytime soon can we have a list supported in our FW but not provided to end users?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
sashaCher: As long as the authenticator has no policy defined configuration what can drive a customer to define more than one GCP authenticator.
In other words can we suppress `service-id` for this authn method?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,573 @@
...
+The audience value must to be `conjur` 
+
+JWT token:
+
+```json
+{
+  "aud": "http://vault/conjur",
+  "azp": "110987294251917851298",
+  "email": "716149158341-compute@developer.gserviceaccount.com",
+  "email_verified": true,
```
sashaCher, line 155: @InbalZilberman @nessiLahav 
Is it something that we need to be aware on?
In which cases is it `false`?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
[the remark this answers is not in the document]
sashaCher: The [JWT reference](https://tools.ietf.org/html/rfc7519#section-4.1) stands that `iss` and `sub` are `value`. While `aud` `is an array`.
As far as `iss` is a single value `provider-uri` should be the same.
In any case defining one big `authn-jwt` supports all possible variations is pretty far from to be a best practice...

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,573 @@
...
+|     api                                                            |     authn-jwt/                            |     authn-gcp/                                                                      |
+|     Mandatory base claims to validate                              |     Iss, exp, nbf (Configurable)          |     iss, exp, aud, iat (Constants)                                           |
+|     Mandatory base claims to validate                              |     Iss, exp, nbf (Configurable)          |     iss, exp, aud, iat, nbf (Constants)                                           |
+|     Optional base claims to validate (if exists in the token)      |     aud,sub,iat (Configurable)            |     -                                                                               |
+|     provider-uri                                                   |     Configurable   (list?)                |     Constant                                                                        |
```
sashaCher, line 210: @nessiLahav Along with `provider-uri` we should have optional parameter `provider-uri-certificate` for provider's public certificate in case and it's a 3rd party or self-signed.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
Tovli: @oburstein-hub - would appreciate your input here - do we want to have a standard audience structure across our services?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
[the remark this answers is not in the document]
Tovli: @sashaCher Totally agree, but that depends on how do you define `big` authn-jwt. The goal is to have it generic in a way that we can support more scenarios by using configuration rather then re-implementing the same flow for different providers. Same as [JWTBearer](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.jwtbearerextensions.addjwtbearer?view=aspnetcore-3.0) works in .net

design/authenticators/authn_gcp/authn_gcp_solution_design.md
Tovli: @InbalZilberman - please see this excellent comment from @sashaCher . It is different UX as instead of provider-uri we can just ask for the key itself, thus reducing the chance of failure from getting it from external resource

design/authenticators/authn_gcp/authn_gcp_solution_design.md
Tovli: @oburstein-hub - you can see a readable version with the charts when pressing the link in the description of this PR.
[Here is the link for ease of use ](https://github.com/cyberark/conjur/blob/gcp-solution-design/design/authenticators/authn_gcp/authn_gcp_solution_design.md)

design/authenticators/authn_gcp/authn_gcp_solution_design.md
Tovli: @oburstein-hub - you can see a readable version with the charts when pressing the link in the description of this PR.
[Here is the link for ease of use ](https://github.com/cyberark/conjur/blob/gcp-solution-design/design/authenticators/authn_gcp/authn_gcp_solution_design.md)

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: good point i will add a card to investigate it

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: Like the idea thanks

design/authenticators/authn_gcp/authn_gcp_solution_design.md
InbalZilberman: That's a good question!! maybe we can have only one. @nessiLahav @eranha @Tovli wdyt? 
I also think we need to rename it to authn-gcp-metadata or authn-gcp-gce - wdyt?\

design/authenticators/authn_gcp/authn_gcp_solution_design.md
InbalZilberman: @eranha above you gave strict instructions to validate project id maybe there are for other fields?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
InbalZilberman: @eranha I am not sure i see an additional value here. can you elaborate how that that help?
If it is for security reasons I suggest to involve also Sasha

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,576 @@
...
+|     Public-certificate                                             |     Configurable (In case of self sign)   |     -                                                                        |
```
sashaCher, line 211: List??
If we're accepting multiple URIs we should accept self-signed certificate for each one of them...
Actually I think that `provider-uri` should not be a list.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
InbalZilberman: I agree, if this stand the security review Isee only advantages of performance and ux for this option.
in gcp as the provider uri is not user facing the cert should not be mandatory. if it doesn't exists we use the hardcoded uri.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
InbalZilberman: me too :)

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: @InbalZilberman i like the authn-gcp-gce 

> In other words can we suppress service-id for this authn method?

Good point i think we should suppress it in this use case
@InbalZilberman Update me if you are changing the feature doc

design/authenticators/authn_gcp/authn_gcp_solution_design.md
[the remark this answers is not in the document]
oburstein-hub: @sashaCher / @InbalZilberman - I agree with you. I think we should maybe consider to split JWT authenticator to number of specific ones

design/authenticators/authn_gcp/authn_gcp_solution_design.md
oburstein-hub: Great idea @sashaCher

design/authenticators/authn_gcp/authn_gcp_solution_design.md
oburstein-hub: yes - I prefer to have standard audience structure for all our services, and its values should be 'conjur'
@InbalZilberman  - please approve

design/authenticators/authn_gcp/authn_gcp_solution_design.md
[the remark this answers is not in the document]
sashaCher: I feel that we need to avoid providers enumeration.
Each `authn` instance should accept tokens only form one provider.
It simplifies implementation - invalid signature is always treated as a fail, rather let's try the next one.
It also less vulnerable to time based staff. Reaching each provider takes time obviously.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
oburstein-hub: thanks @Tovli . I have already got diagrams from Nessi and reviewed them

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,591 @@
...
+    role: !group apps
+    privilege: [ read, authenticate ]
+    resource: !webservice
+```
+
+All the following configurations will have constants values and will not require any action from the customer:
+
+* **provider-uri** = `https://www.googleapis.com/oauth2/v1/certs` #represnts the address to fetch the public key in order to validate the JWT token
+* **mandatory-claims** = `iss, exp, aud, iat` #the list of the mandatory claims to validate in the JWT token
+* **mandatory-claims** = `iss, exp, aud, iat, nbf` #the list of the mandatory claims to validate in the JWT token
```
sigalsax, line 102: which one is correct? there are two here?

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: @inbal it was in response to Ofira's question.

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: @InbalZilberman @Tovli @sashaCher  @oburstein-hub 
By specifying the `http://conjur/account/host/my-policy/my-host/authenticate` in the audience claim same as you do in Conjure's authenticate URL `http://conjur/account/host/my-policy/my-host/authenticate` you are binding the token to its destination url, that is also what Google [recommends](https://cloud.google.com/compute/docs/instances/verifying-instance-identity).
```
[AUDIENCE] is the unique URI agreed upon by both the instance and the system verifying the instance's identity. For example, the audience could be a URL for the connection between the two systems.
```

design/authenticators/authn_gcp/authn_gcp_solution_design.md
eranha: @InbalZilberman @Tovli @sashaCher  @oburstein-hub 
By specifying the `http://conjur/account/host/my-policy/my-host/authenticate` in the audience claim same as you do in Conjure's authenticate URL `http://conjur/account/host/my-policy/my-host/authenticate` you are binding the token to its destination url, that is also what Google [recommends](https://cloud.google.com/compute/docs/instances/verifying-instance-identity).

`[AUDIENCE] is the unique URI agreed upon by both the instance and the system verifying the instance's identity. For example, the audience could be a URL for the connection between the two systems.`

design/authenticators/authn_gcp/authn_gcp_solution_design.md
oburstein-hub: @eran - please elaborate what is the advantage of having url-formatted audience. As per my understand - it means that audience  will be coupled with specific conjur operation URL, right

design/authenticators/authn_gcp/authn_gcp_solution_design.md
nessiLahav: Fixed, thanks

design/authenticators/authn_gcp/authn_gcp_solution_design.md
```diff
@@ -0,0 +1,600 @@
...
+We will add a new GCP authenticator to Conjur, which GCE hosts can will authenticate with by presenting their host Conjur identity and google identity as JWT token.
+
+As mentioned in the [feature doc](https://ca-il-confluence.il.cyber-ark.com/display/rndp/Conjur+GCP+Authenticator#/), before any authentication 
+request is sent to Conjur, the admin will load the authenticator policy: 
+
+```yaml
+# policy id needs to match the convention `conjur/authn-gcp/<service ID>`
+- !policy
+  id: conjur/authn-gcp/mygcp
+  body:
```
orenbm, line 87: @nessiLahav where did the `provider-uri` go? do we not have this variable anymore?

Issue #1725, opened 2020-08-03
Only `127.0.0.1` is a trusted proxy by default


micahlee:
Currently, if the `TRUSTED_PROXIES` environment variable is not set, Conjur falls back to the default Rack behavior for trusted proxy IP address. This causes it to trust all non-routable IP address according to the regex: https://github.com/rack/rack/blob/master/lib/rack/request.rb#L19

This prevents a spoof risk from intranet IP addresses when trusted proxies are not configured. This also may cause it to appear like reverse proxies and load balancers are working correctly, when the may not work if the IP address don't match the regex. Further, this makes the rules for how Conjur selects the client IP address more complicated than they need to be.

To simplify this, Conjur should only trust `127.0.0.1` and those IP addresses (or ranges in CIDR notation) explicitly configured in the `TRUSTED_PROXIES`, with no fallback to the Rack regex.

DoD:
  - [ ] SPIKE: To determine what needs to change in code.
  - [ ] SPIKE: Does anything else in the product depend on this.
  - [ ] TEST: when no TRUSTED PROXY is set and IP is non-routable

PR #1726, opened 2020-08-03
Only trust `127.0.0.1` by default for sending the `X-Forwarded-For` header

micahlee:
### What does this PR do?

Previously, if `TRUSTED_PROXIES` was not configured, Conjur would fallback to
the default Rack behavior for which IP addresses to trust to send the
`X-Forwarded-For` HTTP header. This used a regular expression to determine
whether an IP address was routable or not. Any IP address considered non-routable
according to this regex was trusted.

This commit modifies this behavior to only trust `127.0.0.1` by default, effectively
locking down the `X-Forwarded-For` header, unless explicitly trusted with the
`TRUSTED_PROXIES` environment variable.

Connected to #1725, #1727 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Add `/whoami` API endpoint for improved supportability and debugging for access
   tokens and client IP address determination. [cyberark/conjur#1697](https://github.com/cyberark/conjur/issues/1697)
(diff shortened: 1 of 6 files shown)


micahlee:
> Add a check when creating the Trusted Proxy filter to ensure TRUSTED_PROXIES is valid

Thanks @h-artzi! Here's the follow-up issue to add start up validation for `TRUSTED_PROXIES`: https://github.com/cyberark/conjur/issues/1727

PR #1745, opened 2020-08-11
Gcp extract source restrictions

nessiLahav:
### What does this PR do?
Add resource restrictions extraction logic, which is the first dependency of the `ValidateResourceRestriction` command class 

### Logs example from a valid flow
```
CONJ00039D Extracting restrictions from resource 'usertest1@nessitest' with prefix 'authn-gcp/'
  Sequel::Postgres::Database (1.5ms)  SELECT * FROM "resources" WHERE "resource_id" = 'cucumber:user:usertest1@nessitest'
  Sequel::Postgres::Database (1.3ms)  SELECT * FROM "annotations" WHERE ("annotations"."resource_id" = 'cucumber:user:usertest1@nessitest')
CONJ00024D Retrieved value of annotation authn-gcp/instance-name
CONJ00024D Retrieved value of annotation authn-gcp/project-id
CONJ00024D Retrieved value of annotation authn-gcp/service-account-id
CONJ00024D Retrieved value of annotation authn-gcp/service-account-email
CONJ00024D Retrieved value of annotation authn-gcp/null
CONJ00024D Retrieved value of annotation authn-gcp/empty
CONJ00024D Retrieved value of annotation authn-gcp/
CONJ00040D 7 restrictions were extracted

CONJ00039D Extracting restrictions from resource 'host/nessitest/hosttest1' with prefix 'authn-gcp/'
  Sequel::Postgres::Database (0.6ms)  SELECT * FROM "resources" WHERE "resource_id" = 'cucumber:host:nessitest/hosttest1'
  Sequel::Postgres::Database (0.5ms)  SELECT * FROM "annotations" WHERE ("annotations"."resource_id" = 'cucumber:host:nessitest/hosttest1')
CONJ00024D Retrieved value of annotation authn-gcp/instance-name
CONJ00024D Retrieved value of annotation authn-gcp/project-id
CONJ00024D Retrieved value of annotation authn-gcp/service-account-id
CONJ00024D Retrieved value of annotation authn-gcp/service-account-email
CONJ00024D Retrieved value of annotation authn-gcp/null
CONJ00024D Retrieved value of annotation authn-gcp/empty
CONJ00024D Retrieved value of annotation authn-gcp/
CONJ00040D 7 restrictions were extracted
```

### DIFF
--- a/app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
+++ b/app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
@@ -6,31 +6,42 @@ module Authentication
     # This class is responsible of restrictions extraction that are set on a Conjur host or user as annotations.
     ExtractResourceRestrictions = CommandClass.new(
       dependencies: {
-        role_class:     ::Role,
-        resource_class: ::Resource,
-        logger:         Rails.logger
+        role_class:              ::Role,
+        resource_class:          ::Resource,
+        validate_account_exists: ::Authentication::Security::ValidateAccountExists.new,
+        logger:                  Rails.logger
       },
       inputs:       %i(account username extraction_prefix)
     ) do
 
       def call
+        validate_account_exists
         extract_resource_restrictions
-        restrictions_list
+        resource_restrictions
       end
 
       private
 
-      def extract_resource_restrictions
-        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractingRestrictionsFromResource.new(@username, @extraction_prefix))
-        prefixed_resource_annotations
-        init_restrictions_list
-        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractedResourceRestrictions.new(restrictions_list.length()))
+      def validate_account_exists
+        @validate_account_exists.(
+          account: @account
+        )
       end
 
-      def restrictions_list
-        return @restrictions_list if @restrictions_list
-
-        @restrictions_list = Array.new
+      def extract_resource_restrictions
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractingRestrictionsFromResource.new(@username, @extraction_prefix))
+        prefixed_resource_annotations.select do |a|
+          annotation_name = a.values[:name]
+          resource_value = annotation_value(annotation_name)
+          next unless resource_value
+          resource_restrictions.push(
+            ResourceRestriction.new(
+              type: annotation_name,
+              value: resource_value
+            )
+          )
+        end
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractedResourceRestrictions.new(resource_restrictions.length))
       end
 
       def prefixed_resource_annotations
@@ -40,12 +51,12 @@ module Authentication
         end
       end
 
-      def init_restrictions_list
+      def init_resource_restrictions
         prefixed_resource_annotations.select do |a|
           annotation_name = a.values[:name]
           resource_value = annotation_value(annotation_name)
           next unless resource_value
-          restrictions_list.push(
+          resource_restrictions.push(
             ResourceRestriction.new(
               type: annotation_name,
               value: resource_value
@@ -54,15 +65,20 @@ module Authentication
         end
       end
 
-      def resource_annotations
-        @resource_annotations ||= role.annotations
+      def resource_restrictions
+        @resource_restrictions ||= Array.new
+      end
 
+      def resource_annotations
+        return @resource_annotations if @resource_annotations
+
+        @resource_annotations ||= role.annotations
         # This is an illegal state, because if the resource doesnt have annotations we should get an empty array
         raise Errors::Conjur::FetchAnnotationsFailed.new(role_id) unless @resource_annotations
         @resource_annotations
       end
 
-      def annotation_value name
+      def annotation_value(name)
         annotation = prefixed_resource_annotations.find {|a| a.values[:name] == name}
 
         # return the value of the annotation if it exists, nil otherwise
--- a/app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
+++ b/app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
@@ -1,0 +1,84 @@
+require 'command_class'
+
+module Authentication
+  module AuthnGCP
+
+    # This class is responsible of restrictions extraction that are set on a Conjur host or user as annotations.
+    ExtractResourceRestrictions = CommandClass.new(
+      dependencies: {
+        role_class: ::Role,
+        resource_class: ::Resource,
+        logger: Rails.logger
+      },
+      inputs: %i(extraction_prefix account username)
+    ) do
+
+      def call
+        extract_resource_restrictions
+        restrictions_list
+      end
+
+      private
+
+      def extract_resource_restrictions
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractingRestrictionsFromResource.new(@username, @extraction_prefix))
+        prefixed_resource_annotations
+        init_restrictions_list
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractedResourceRestrictions.new(restrictions_list.length()))
+      end
+
+      def restrictions_list
+        return @restrictions_list if @restrictions_list
+
+        @restrictions_list = Array.new
+      end
+
+      def prefixed_resource_annotations
+        @prefixed_resource_annotations ||= resource_annotations.select do |a|
+          annotation_name = a.values[:name]
+          annotation_name.start_with?(@extraction_prefix)
+        end
+      end
+
+      def init_restrictions_list
+        prefixed_resource_annotations.select do |a|
+          annotation_name = a.values[:name]
+          resource_value = annotation_value(annotation_name)
+          next unless resource_value
+          restrictions_list.push(
+            ResourceRestriction.new(
+              type: annotation_name,
+              value: resource_value
+            )
+          )
+        end
+      end
+
+      def resource_annotations
+        @resource_annotations ||= role.annotations
+      end
+
+      def annotation_value name
+        annotation = prefixed_resource_annotations.find {|a| a.values[:name] == name}
+
+        # return the value of the annotation if it exists, nil otherwise
+        if annotation
+          @logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
+          annotation[:value]
+        end
+      end
+
+      def role
+        return @role if @role
+
+        @role = @resource_class[role_id]
+        raise Errors::Authentication::Security::RoleNotFound, role_id unless @role
+        @role
+      end
+
+      def role_id
+        @role_id ||= @role_class.roleid_from_username(@account, @username)
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
+++ b/app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
@@ -6,11 +6,11 @@ module Authentication
     # This class is responsible of restrictions extraction that are set on a Conjur host or user as annotations.
     ExtractResourceRestrictions = CommandClass.new(
       dependencies: {
-        role_class: ::Role,
+        role_class:     ::Role,
         resource_class: ::Resource,
-        logger: Rails.logger
+        logger:         Rails.logger
       },
-      inputs: %i(extraction_prefix account username)
+      inputs:       %i(account username extraction_prefix)
     ) do
 
       def call
@@ -56,6 +56,10 @@ module Authentication
 
       def resource_annotations
         @resource_annotations ||= role.annotations
+
+        # This is an illegal state, because if the resource doesnt have annotations we should get an empty array
+        raise Errors::Conjur::FetchAnnotationsFailed.new(role_id) unless @resource_annotations
+        @resource_annotations
       end
 
       def annotation_value name
--- a/app/domain/authentication/authn_gcp/resource_restriction.rb
+++ b/app/domain/authentication/authn_gcp/resource_restriction.rb
@@ -1,0 +1,17 @@
+module Authentication
+  module AuthnGCP
+
+    class ResourceRestriction
+      attr_reader :type, :value
+
+      def initialize(type:, value:)
+        @type = type
+        @value = value
+      end
+
+      def ==(other)
+        @type == other.type && @value == other.value
+      end
+    end
+  end
+end
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -25,6 +25,11 @@ module Errors
       code: "CONJ00065E"
     )
 
+    FetchAnnotationsFailed = ::Util::TrackableErrorClass.new(
+      msg:  "Failed to fetch annotations for resource '{0-resource-id}'",
+      code: "CONJ00066E"
+    )
+
   end
 
   module Authentication
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -192,12 +192,12 @@ module LogMessages
     module AuthnGCP
 
       ExtractingRestrictionsFromResource = ::Util::TrackableLogMessageClass.new(
-          msg: "Extracting restrictions from resource '{0-resource-id}' with prefix '{1-prefix}'",
+          msg: "Extracting resource restrictions from role '{0-role-id}' with prefix '{1-prefix}'",
           code: "CONJ00039D"
       )
 
       ExtractedResourceRestrictions = ::Util::TrackableLogMessageClass.new(
-          msg: "{0-amount} restrictions were extracted",
+          msg: "{0-amount} resource restrictions were extracted",
           code: "CONJ00040D"
       )
 
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -188,6 +188,20 @@ module LogMessages
       )
 
     end
+
+    module AuthnGCP
+
+      ExtractingRestrictionsFromResource = ::Util::TrackableLogMessageClass.new(
+          msg: "Extracting restrictions from resource '{0-resource-id}' with prefix '{1-prefix}'",
+          code: "CONJ00039D"
+      )
+
+      ExtractedResourceRestrictions = ::Util::TrackableLogMessageClass.new(
+          msg: "{0-amount} restrictions were extracted",
+          code: "CONJ00040D"
+      )
+
+    end
   end
 
   module Util
--- a/spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
+++ b/spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
@@ -1,0 +1,440 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'Authentication::AuthnGCP::ExtractResourceRestrictions' do
+  let(:valid_account) {'valid-account'}
+  let(:invalid_account) {'invalid-account'}
+
+  let(:valid_host) {'valid-host'}
+  let(:invalid_host) {'invalid-host'}
+
+  let(:valid_prefix) {'authn-gcp/'}
+  let(:empty_prefix) {''}
+  let(:case_sensitive_prefix) {'authn-GCP/'}
+  let(:without_slash_prefix) {'authn-gcp'}
+
+  let(:gcp_instance_name_annotation_key) {'authn-gcp/instance-name'}
+  let(:gcp_instance_name_annotation_value) {'instance-name'}
+  let(:gcp_project_id_annotation_key) {'authn-gcp/project-id'}
+  let(:gcp_project_id_annotation_value) {'project-id'}
+  let(:gcp_service_account_id_annotation_key) {'authn-gcp/service-account-id'}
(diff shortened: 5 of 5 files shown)


app/domain/errors.rb
```diff
@@ -25,6 +25,11 @@ module Conjur
       code: "CONJ00065E"
     )
 
+    FetchAnnotationsFailed = ::Util::TrackableErrorClass.new(
+      msg:  "Failed to fetch annotations for resource '{0-resource-id}'",
```
nessiLahav, line 29: @shulifink please review

app/domain/logs.rb
```diff
@@ -188,6 +188,20 @@ module AuthnAzure
       )
 
     end
+
+    module AuthnGCP
+
+      ExtractingRestrictionsFromResource = ::Util::TrackableLogMessageClass.new(
+          msg: "Extracting restrictions from resource '{0-resource-id}' with prefix '{1-prefix}'",
```
nessiLahav, line 195: @shulifink please review

app/domain/logs.rb
```diff
@@ -188,6 +188,20 @@ module AuthnAzure
...
+          code: "CONJ00039D"
+      )
+
+      ExtractedResourceRestrictions = ::Util::TrackableLogMessageClass.new(
+          msg: "{0-amount} restrictions were extracted",
```
nessiLahav, line 200: @shulifink please review

spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,440 @@
+# frozen_string_literal: true
+
```
nessiLahav, line 2: @eladkug UTs to review :)

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,88 @@
...
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractingRestrictionsFromResource.new(@username, @extraction_prefix))
+        prefixed_resource_annotations
+        init_restrictions_list
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractedResourceRestrictions.new(restrictions_list.length()))
+      end
+
+      def restrictions_list
+        return @restrictions_list if @restrictions_list
+
+        @restrictions_list = Array.new
```
orenbm, line 33: how about:
```
@restrictions_list ||= Array.new
```

seems that you're not doing anything else so you can use memoization

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,88 @@
...
+      private
+
+      def extract_resource_restrictions
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractingRestrictionsFromResource.new(@username, @extraction_prefix))
+        prefixed_resource_annotations
+        init_restrictions_list
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractedResourceRestrictions.new(restrictions_list.length()))
+      end
+
+      def restrictions_list
```
orenbm, line 30: i think it will be better placed further down in the file. i like defining methods after we call them so it's easier to read through the file

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,88 @@
...
+        return @restrictions_list if @restrictions_list
+
+        @restrictions_list = Array.new
+      end
+
+      def prefixed_resource_annotations
+        @prefixed_resource_annotations ||= resource_annotations.select do |a|
```
orenbm, line 37: i think that `resource_annotation` or `annotation` is better than `a`

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,88 @@
...
+          annotation_name = a.values[:name]
+          annotation_name.start_with?(@extraction_prefix)
+        end
+      end
+
+      def init_restrictions_list
```
orenbm, line 43: this method is redundant imo. you can move its logic to `extract_resource_restrictions`.

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,88 @@
...
+        prefixed_resource_annotations.select do |a|
+          annotation_name = a.values[:name]
+          resource_value = annotation_value(annotation_name)
+          next unless resource_value
+          restrictions_list.push(
+            ResourceRestriction.new(
+              type: annotation_name,
```
orenbm, line 50: `annotation_name` will hold also the prefix. do you want it to be that way?

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,88 @@
...
+            )
+          )
+        end
+      end
+
+      def resource_annotations
+        @resource_annotations ||= role.annotations
+
+        # This is an illegal state, because if the resource doesnt have annotations we should get an empty array
+        raise Errors::Conjur::FetchAnnotationsFailed.new(role_id) unless @resource_annotations
```
orenbm, line 61: this check will happen every time we call `resource_restrictions`. how about changing the memoization into 
```
return @resource_annotations if @resource_annotations

@resource_annotations ||= role.annotations
# This is an illegal state, because if the resource doesnt have annotations we should get an empty array
raise Errors::Conjur::FetchAnnotationsFailed.new(role_id) unless @resource_annotations
@resource_annotations
```

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,88 @@
...
+        role_class:     ::Role,
+        resource_class: ::Resource,
+        logger:         Rails.logger
+      },
+      inputs:       %i(account username extraction_prefix)
+    ) do
+
+      def call
+        extract_resource_restrictions
+        restrictions_list
```
orenbm, line 18: optional name change: `resource_restrictions`.

The class name is `ExtractResourceRestrictions`. seems logical to return `resource_restrictions`.

app/domain/authentication/authn_gcp/resource_restriction.rb
```diff
@@ -0,0 +1,17 @@
+module Authentication
+  module AuthnGCP
+
+    class ResourceRestriction
```
orenbm, line 4: how about implementing this class under the `Authentication` module? you don't need to use it yet in other authenticators (though you can) but it seems that this will be used in other places as well.

app/domain/logs.rb
orenbm: i think that this is confusing. we mostly refer to hosts as a role and not as a resource when it's customer facing. 
i think that this is better:
```
ExtractingResourceRestrictionsFromRole = ::Util::TrackableLogMessageClass.new(
          msg: "Extracting Resource Restrictions from role '{0-role-id}' with prefix '{1-prefix}'",
          code: "CONJ00039D"
      )

File: app/domain/logs.rb (hunk 3 above)
Comment (orenbm) [reply app/domain/logs.rb:200]: i would write `"{0-amount} Redsource Restrictions were extracted"`

File: spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb [hunk 13]
```diff
@@ -0,0 +1,440 @@
...
+
+  end
+
+  #  ____  _   _  ____    ____  ____  ___  ____  ___
+  # (_  _)( )_( )( ___)  (_  _)( ___)/ __)(_  _)/ __)
+  #   )(   ) _ (  )__)     )(   )__) \__ \  )(  \__ \
+  #  (__) (_) (_)(____)   (__) (____)(___/ (__) (___/
+
+  context "An invalid input parameters to extract resource restrictions" do
```
orenbm, line 271: we usually start with vanilla and then go to invalid tests

spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,440 @@
...
+    context "when account not exists" do
```
orenbm, line 272: does not

spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,440 @@
...
+          resource_class: mocked_resource_class_return_nil
+        ).call(
+          account:           invalid_account,
+          username:          valid_host,
+          extraction_prefix: valid_prefix
+        )
+      end
+
+      it "raises an error" do
+        expect {subject}.to raise_error(Errors::Authentication::Security::RoleNotFound)
```
orenbm, line 284: this is not a healthy behaviour. if the account does not exist i would expect an `AccountNotDefined` error

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,88 @@
+require 'command_class'
+
+module Authentication
+  module AuthnGCP
+
+    # This class is responsible of restrictions extraction that are set on a Conjur host or user as annotations.
+    ExtractResourceRestrictions = CommandClass.new(
```
orenbm, line 7: this class can also be implemented under the `Authentication` module as it's not `authn-gcp` specific. Do we want to implement it that way or start like this and have a leftover to extract it (among other places)?

spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,440 @@
...
+          )
+        end
+
+        it "returns expected list" do
+          expect(subject).to eq(expected_resource_restrictions_list_with_duplications_for_valid_prefix)
+        end
+      end
+    end
+  end
+end
```
orenbm, line 440: missing newline

app/domain/logs.rb
nessiLahav: Did you meant without the upper case?
`{0-amount} resource restrictions were extracted`

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
nessiLahav: correct

app/domain/logs.rb
orenbm: *with* the upper case

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
nessiLahav: Yes, for now I don’t see a good reason to delete the prefix

orenbm: i think you will need to once you validate them but i guess it's ok. it's quite odd to have a `ResourceRestriction`:
```
type: authn-gcp/project_id
value: some_project_id`
```
but i guess we can live with it.

by the way, in this implementation we may have a `ResourceRestriction`:
```
type: authn-gcp/blah
value: some_blah`
```

this will be removed in the `ValidateResourceRestrictions` stage? we first populate all the values and only then validate them? another approach can be to populate a value only if it's validated

app/domain/authentication/authn_gcp/resource_restriction.rb
nessiLahav: I think it is part of the effort we should re thinking about post GCP authenticator, because `ExtractResourceRestrictions` also should be in more general module, but right now i want to focus on the GCP and therefore all the logic around the resource restrictions will reside under authGCP module

spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
nessiLahav: good point i will add an account validation to the logic

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
nessiLahav: the other option :) since we may need some modifications to use it in other places

orenbm: np

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
nessiLahav: I think it is more readable when it divided into two methods

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,95 @@
...
+          resource_value = annotation_value(annotation_name)
+          next unless resource_value
+          resource_restrictions.push(
+            ResourceRestriction.new(
+              type: annotation_name,
+              value: resource_value
+            )
+          )
+        end
+      end
```
orenbm, line 57: consider changing this to:
```
        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractingRestrictionsFromResource.new(@username, @extraction_prefix))
        prefixed_resource_annotations.select do |a|
          resource_value = annotation_value(annotation_name)
          next unless resource_value
          resource_restrictions.push(
            ResourceRestriction.new(
              type: annotation_name,
              value: resource_value
            )
          )
        end
        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractedResourceRestrictions.new(resource_restrictions.length()))
      end

      def prefixed_resource_annotations
        @prefixed_resource_annotations ||= resource_annotations.select do |a|
          annotation_name.start_with?(@extraction_prefix)
        end
      end
```

it is less complex imo

app/domain/logs.rb
```diff
@@ -188,6 +188,20 @@ module AuthnAzure
       )
 
     end
+
+    module AuthnGCP
+
+      ExtractingRestrictionsFromResource = ::Util::TrackableLogMessageClass.new(
+          msg: "Extracting resource restrictions from role '{0-role-id}' with prefix '{1-prefix}'",
```
orenbm, line 195: @shulifink please review

app/domain/logs.rb
```diff
@@ -188,6 +188,20 @@ module AuthnAzure
...
+          code: "CONJ00039D"
+      )
+
+      ExtractedResourceRestrictions = ::Util::TrackableLogMessageClass.new(
+          msg: "{0-amount} resource restrictions were extracted",
```
orenbm, line 200: @shulifink please review

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
orenbm: and it's more ruby style

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,95 @@
...
+        @validate_account_exists.(
+          account: @account
+        )
+      end
+
+      def extract_resource_restrictions
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractingRestrictionsFromResource.new(@username, @extraction_prefix))
+        prefixed_resource_annotations
+        init_resource_restrictions
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractedResourceRestrictions.new(resource_restrictions.length()))
```
orenbm, line 35: redundant parentheses. should be `length`

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
nessiLahav: Yes my idea was first to populate all the values without any validation, so this class will only one responsibility and may be useful later in other places.
And in the `ValidateResourceRestrictions` stage we will validate the values.

orenbm: i'm ok with that approach. we can challenge it once we implement `ValidateResourceRestrictions`

app/domain/errors.rb
shulifink: Fine

app/domain/logs.rb
shulifink: why with upper case?

app/domain/logs.rb
shulifink: earlier you mentioned that you think these should be upper case. Same here, no?
But why should they be uppercase?

shulifink: I think this is fine as it is

app/domain/logs.rb
shulifink: Fine - see comment in line 200 about caps

app/domain/logs.rb
orenbm: i thought of doing that because Resource Restrictions is a thing (it actually replaces Application Identity) but i then changed my mind and prefer it lower-cased

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,104 @@
...
+      private
+
+      def validate_account_exists
+        @validate_account_exists.(
+          account: @account
+        )
+      end
+
+      def extract_resource_restrictions
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ExtractingRestrictionsFromResource.new(@username, @extraction_prefix))
```
eranha, line 32: Couldn't we shorten it using alias or alike, it clutters the code.

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      def resource_restrictions
+        @resource_restrictions ||= Array.new
+      end
+
+      def resource_annotations
+        return @resource_annotations if @resource_annotations
+
+        @resource_annotations ||= role.annotations
+        # This is an illegal state, because if the resource doesnt have annotations we should get an empty array
```
eranha, line 76: if its an illegal state the error should reflect it.

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,104 @@
...
+          @logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
+          annotation[:value]
+        end
+      end
+
+      def role
+        return @role if @role
+
+        @role = @resource_class[role_id]
+        raise Errors::Authentication::Security::RoleNotFound, role_id unless @role
```
eranha, line 95: inconsistent with `raise Errors::Conjur::FetchAnnotationsFailed.new(role_id)` missing ()

app/domain/authentication/authn_gcp/extract_resource_restrictions.rb
orenbm: we decided not to do that.
however, i woulld change the code to:
```
@logger.debug(
  LogMessages::Authentication::AuthnGCP::ExtractingRestrictionsFromResource.new(
    @username, 
    @extraction_prefix
  )
)
```

spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,440 @@
...
+            resource_class: mocked_resource_class_return_nil
+          ).call(
+            account:           valid_account,
+            username:          invalid_host,
+            extraction_prefix: valid_prefix
+          )
+        end
+
+        it "raises an error" do
+          expect {subject}.to raise_error(Errors::Authentication::Security::RoleNotFound)
```
eladkug: I think better validation will be `InvalidHostId` error

spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,440 @@
...
+          Authentication::AuthnGCP::ExtractResourceRestrictions.new(
+            resource_class: mocked_resource_class_return_annotations_empty_list
+          ).call(
+            account:           valid_account,
+            username:          valid_host,
+            extraction_prefix: valid_prefix
+          )
+        end
+
+        it "returns expected list" do
```
eladkug, line 375: we suppose to fail if we dont have annotations , no?

nessiLahav: Not it this class, here we only fetch the data

spec/app/domain/authentication/authn-gcp/extract_resource_restrictions_spec.rb
orenbm: why? the user gave us a hostname and it is not found in the db

eladkug: username give invalid_host , i think the error InvalidHostId is the relevant error for this. RoleNotFound its too general imo.

orenbm: that's just a naming decision. @nessiLahav could have just called it `non_existing_host`. in any case, what we verify here is that the host doesn't exist.

PR #1747, opened 2020-08-13
Gcp validate restrictions combination

nessiLahav:
### What does this PR do?
Add constraints combination validate logic for resource restrictions, which is the 2nd dependency of the `ValidateResourceRestriction` command class


app/domain/authentication/authn_gcp/restrictions_validator.rb
```diff
@@ -0,0 +1,105 @@
+module Authentication
+  module AuthnGCP
+
+    # This class is responsible of validating resource restrictions combination.
+    # the allowed combination is: Role should have at least 1 permitted constraints, and non of illegal constraints
+    class RestrictionsValidator
```
orenbm, line 6: @nessiLahav why is this class not a command class `ValidateResourceRestrictions`?

app/domain/authentication/authn_gcp/restrictions_validator.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      private
+
+      def validate
+        input_validation
+        validate_constraints_are_permitted
+        validate_restrictions_values_exist
+      end
+
+      def input_validation
```
orenbm, line 31: validate_input

app/domain/authentication/authn_gcp/restrictions_validator.rb
```diff
@@ -0,0 +1,105 @@
...
+        validate
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ValidatedRestrictionsConstraintCombination.new)
+      end
+
+      private
+
+      def validate
+        input_validation
+        validate_constraints_are_permitted
+        validate_restrictions_values_exist
```
orenbm, line 28: why are these methods not in `input_validation`? how are they different than the validations performed there?

btw, i think that we can get rid of the `input_validation` method and move its validations to `validate`.

app/domain/authentication/authn_gcp/restrictions_validator.rb
```diff
@@ -0,0 +1,105 @@
...
+  module AuthnGCP
+
+    # This class is responsible of validating resource restrictions combination.
+    # the allowed combination is: Role should have at least 1 permitted constraints, and non of illegal constraints
+    class RestrictionsValidator
+
+      ILLEGAL_STATE_MISSING_CONSTRAINTS_MESSAGE = "constraints are missing or empty"
+      ILLEGAL_STATE_MISSING_RESOURCE_TYPE_MESSAGE = "resource restriction type is missing or empty"
+      ILLEGAL_STATE_DUPLICATE_CONSTRAINTS_MESSAGE = "constraints contain duplications"
+      ILLEGAL_STATE_DUPLICATE_RESTRICTIONS_MESSAGE = "resource restrictions contain duplications"
```
orenbm, line 11: Is there a reason why these error messages are not defined as errors in `errors.rb`?

app/controllers/authenticate_controller.rb
```diff
@@ -181,6 +181,9 @@ def handle_authentication_error(err)
       Errors::Authentication::AuthnK8s::CertMissingCNEntry
       raise ArgumentError
 
+    when Errors::Authentication::IllegalStateMissingConstraints
+      raise InternalServerError
```
orenbm, line 185: why are you returning 500 in this case? is there a reason we do not return 401 like we do with other validation failures?

app/domain/authentication/authn_gcp/restrictions_validator.rb
nessiLahav: OK i will move them, i though it is too noisy to mix the logic with the parameters validation

app/domain/authentication/authn_gcp/restrictions_validator.rb
nessiLahav: I have here a new error which indicate on illegal state that should not happened to the user, but due to the zero trust rule i validate those parameters that are valid before executing the logic.
So i created a general message of illegal state which is similar to all 4 use cases, BYTW i returned here 500 error instead of 401

app/controllers/authenticate_controller.rb
nessiLahav: See comment above

app/domain/authentication/authn_gcp/restrictions_validator.rb
orenbm: i don't think that this is different from any other configuration error we have. for example, if the user didn't define the `provider-uri` variable in the oidc authenticator policy we return 401. would you return 500 in that case as well?

a 500 response is not very good imo as it `indicates that the server encountered an unexpected condition that prevented it from fulfilling the request`. so it's some kind of "i don't know what happened but i can't fulfill your request". But here we know what happened. specifically, what happened is that the user that is trying to authenticate doesn't have the sufficient info in order to authenticate so i think that a 401 response code will be good.

orenbm: regarding the location of the errors - i still think that they should go in the `errors.rb` class and that we should have 4 different objects, instead of the general `IllegalStateResourceRestrictionsValidation` object that you created.

app/domain/authentication/authn_gcp/restrictions_validator.rb
```diff
@@ -0,0 +1,101 @@
...
+        end
+      end
+
+      def resource_restrictions_types
+        return @resource_restrictions_types if @resource_restrictions_types
+
+        @resource_restrictions_types = Array.new
+        @resource_restrictions.each do |r|
+          resource_type = r.type
+          next unless resource_type
```
orenbm, line 93: will this happen? when will `r` won't have a `type` method?

app/domain/authentication/authn_gcp/restrictions_validator.rb
nessiLahav: It is not the same example as the `provider-uri` wrong configuration, if those errors (illegal state errors) will happened it is not the user fault.
For example if we will have an error in the DB which will return a nil annotation name or duplicate values (maybe as a result of a bug in a previous conjur upgrade), so it will create a restriction list (which is the input of this class) with un processable values, and to an error that it is a defiantly not 401 error

app/domain/authentication/authn_gcp/restrictions_validator.rb
orenbm: so i'd like to take this example to learn from as a group. @nessiLahav says that he didn't implement this class as a command-class because it doesn't have any dependencies (apart from `logger` which is a somewhat special case). However, in the `README` that @jonahx wrote recently he separated the classes into 3 types:
- Command objects
- Value objects
- Objects with side effects

it is clear that this class is a Command Object as it performs some command (it validates resource restrictions). So from this, it seems that it should be a command class. However, i do remember a rule of thumb that if there are no dependencies then it shouldn't be a command class. 

So i'm quite confused. @jonahx care to jump in?

app/domain/authentication/authn_gcp/restrictions_validator.rb
orenbm: so i don't understand the case. can you describe a use case in which this will happen? for example, how will you test it?

app/domain/authentication/authn_gcp/restrictions_validator.rb
nessiLahav: In case of a `r=nil` is already validated before this action,
About getting a different object or a code changing in `resource_restriction` class i think the UT is protecting us.
Do you think we need an additional protection here?

orenbm: i don't think we need to check it at all. you're getting a list of `ResourceRestrictions` objects and all of them have a `type` attribute.

orenbm: i think that you can do this instead:
```
def resource_restrictions_types
  @resource_restrictions_types ||= @resource_restrictions.collect {|r| r.type }
end
```

app/domain/authentication/authn_gcp/restrictions_validator.rb
nessiLahav: It can be tested just with UT

app/domain/authentication/authn_gcp/restrictions_validator.rb
nessiLahav: Ohh i am testing here that type is not empty, it is an illegal state error that shouldn't happened.
But i think this class responsibility is also to validate the input parameters

orenbm: why would type be empty? you create the `ResourceRestriction` object and give it value. in which scenario will the type be empty?

app/domain/authentication/authn_gcp/restrictions_validator.rb
orenbm: i don't think we need to make up issues that can be reproduced

app/domain/authentication/authn_gcp/restrictions_validator.rb
```diff
@@ -0,0 +1,101 @@
...
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ValidatedRestrictionsConstraintCombination.new)
+      end
+
+      private
+
+      def validate
+        validate_constraints_exist
+        validate_restrictions_exist
+        validate_restrictions_types_exist
+        validate_there_are_no_duplicated_constraints
```
eranha, line 29: I suggest `validate_there_are_no_restriction_duplication`?

app/domain/errors.rb
```diff
@@ -75,6 +75,21 @@ module Authentication
...
+    )
+
+    MissingResourceRestrictionsValue = ::Util::TrackableErrorClass.new(
+      msg:  "Resource restriction '{0-resource-restriction-name}' value is missing or empty",
+      code: "CONJ000669E"
+    )
+
     module AuthenticatorClass
 
       DoesntStartWithAuthn = ::Util::TrackableErrorClass.new(
```
eranha, line 95: I would rephrase to `IllegalAuthnPrefix`

nessiLahav: You can create for this a PR, this is not part of this change

app/controllers/authenticate_controller.rb
orenbm: is this still used?

app/domain/authentication/authn_gcp/validate_resource_restrictions_configuration.rb
```diff
@@ -0,0 +1,49 @@
+module Authentication
+  module AuthnGCP
+
+    # This class is responsible of validating resource restrictions combination.
```
orenbm, line 4: combination -> configuration

app/domain/authentication/authn_gcp/validate_resource_restrictions_configuration.rb
```diff
@@ -0,0 +1,49 @@
...
+        @logger.debug(LogMessages::Authentication::AuthnGCP::ValidatedResourceRestrictionsConfiguration.new)
+      end
+
+      private
+
+      def validate_restrictions_exist
+        if @resource_restrictions.nil? || @resource_restrictions.empty?
+          raise Errors::Authentication::RoleMissingRequiredConstraints.new(@permitted_constraints)
+        end
+      end
```
orenbm, line 27: so here's something. when you read this code you see that if `@resource_restrictions` is empty then the role doesn't have the required constraints. But we don't see here any logic of reading the restrictions from the role. i am not saying that you should change the logic so that the validations happen in the extraction class but just wanted to point to this thing so you can consider it.

app/domain/authentication/authn_gcp/validate_resource_restrictions_configuration.rb
```diff
@@ -0,0 +1,49 @@
...
+            raise Errors::Authentication::ConstraintNotSupported.new(resource_type, @permitted_constraints)
+          end 
+        end
+      end
+
+      def validate_restrictions_values_exist
+        @resource_restrictions.each do |r|
+          resource_type = r.type
+          resource_value = r.value
+          if resource_value.nil? || resource_value.empty?
```
orenbm, line 42: when would this happen? can you share a host policy definition that would yield this error?

app/domain/errors.rb
```diff
@@ -75,6 +75,16 @@ module Authentication
       code: "CONJ00017E"
     )
 
+    RoleMissingRequiredConstraints = ::Util::TrackableErrorClass.new(
+      msg:  "Role should have at least 1 constraints: {0-constraints}",
+      code: "CONJ00067E"
+    )
```
orenbm, line 81: i think that this is a GCP authenticator restriction, so it should go under its module

orenbm: also - i would change the text to `Role should have at least one of the following constraints: {0-constraints}`

spec/app/domain/authentication/authn-gcp/validate_resource_restrictions_configuration_spec.rb
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  #  ____  _   _  ____    ____  ____  ___  ____  ___
+  # (_  _)( )_( )( ___)  (_  _)( ___)/ __)(_  _)/ __)
+  #   )(   ) _ (  )__)     )(   )__) \__ \  )(  \__ \
+  #  (__) (_) (_)(____)   (__) (____)(___/ (__) (___/
+
+  context "A valid resource restrictions list" do
+    context "permitted constraints are 4 gcp values" do
+      context "when resource restrictions lis contains 1 valid restriction" do
```
orenbm, line 221: list

orenbm: throughout the file

app/controllers/authenticate_controller.rb
nessiLahav: No, i will delete it

app/domain/authentication/authn_gcp/validate_resource_restrictions_configuration.rb
nessiLahav: Yes this is the downside of separating those logics, but i sill preferred it separated

app/domain/authentication/authn_gcp/validate_resource_restrictions_configuration.rb
nessiLahav: Yes, the host policy will be like:
```
  - !host
    id: hosttest1
    annotations:
      description: nessi temp host
      authn-gcp/instance-name: 
```
The logs will looks like:
```
CONJ00039D Extracting resource restrictions from role 'host/nessitest/hosttest1' with prefix 'authn-gcp/'
  Sequel::Postgres::Database (0.4ms)  SELECT * FROM "resources" WHERE "resource_id" = 'cucumber:host:nessitest/hosttest1'
  Sequel::Postgres::Database (0.5ms)  SELECT * FROM "annotations" WHERE ("annotations"."resource_id" = 'cucumber:host:nessitest/hosttest1')
CONJ00024D Retrieved value of annotation authn-gcp/instance-name
CONJ00040D 1 resource restrictions were extracted
CONJ00041D Validating resource restrictions configuration
  Sequel::Postgres::Database (0.4ms)  SELECT * FROM "roles" WHERE ("role_id" = 'cucumber:host:nessitest/hosttest1') LIMIT 1
Authentication Error: #<Errors::Authentication::MissingResourceRestrictionsValue: CONJ000068E Resource restriction 'authn-gcp/instance-name' value  is missing or empty>
```

orenbm: thanks for clarifying

app/domain/errors.rb
```diff
@@ -349,13 +354,18 @@ module AuthnGcp
...
 
       TokenClaimNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
         msg:  "Claim '{0-claim-name}' not found or empty in token. " \
               "Verify you request the token using 'format=full'",
-        code: "CONJ00067E"
+        code: "CONJ00068E"
+      )
+
+      RoleMissingRequiredConstraints = ::Util::TrackableErrorClass.new(
+        msg:  "Role should have at least one of the following constraints: {0-constraints}",
```
orenbm, line 367: will constraints have the `authn-gcp` prefix? do we want them to be present?

orenbm:
@nessiLahav please squash before you merge

app/domain/errors.rb
nessiLahav: Yes they will be

PR #1749, opened 2020-08-13
Run authn-azure tests in an azure node

orenbm:
Prior to this commit we allocated an Azure node that just exported its
IP to the environment and then waited to the actual tests to finish.

This commit changes the behaviour so that the tests will run on the Azure
node itself and we have one less node running.

It also simplifies the Jenkinsfile.

Note: we can still run the tests locally the same way.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -83,13 +83,7 @@ pipeline {
         stage('OIDC Authenticator') {
           steps { sh 'ci/test cucumber_authenticators_oidc' }
         }
-        // We have 2 stages for the Azure Authenticator tests. One is
-        // responsible for allocating an Azure VM, from which we will get an
-        // Azure access token for Conjur authentication. It sets the Azure VM IP
-        // in the env and waits until the authn-azure tests are finished. We use
-        // this mechanism as we need a live Azure VM for getting the Azure token
-        // and we don't want to have a long running-machine deployed in Azure.
-        stage('Azure Authenticator preparation - Allocate Azure Authenticator Instance') {
+        stage('Azure Authenticator') {
           steps {
             script {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -83,43 +83,16 @@ pipeline {
...
           steps { sh 'ci/test cucumber_authenticators_oidc' }
         }
-        // We have 2 stages for the Azure Authenticator tests. One is
-        // responsible for allocating an Azure VM, from which we will get an
-        // Azure access token for Conjur authentication. It sets the Azure VM IP
-        // in the env and waits until the authn-azure tests are finished. We use
-        // this mechanism as we need a live Azure VM for getting the Azure token
-        // and we don't want to have a long running-machine deployed in Azure.
-        stage('Azure Authenticator preparation - Allocate Azure Authenticator Instance') {
+        stage('Azure Authenticator') {
```
sigalsax, line 86: @orenbm can you explain why this still works and still gives us the behavior we expect and need for our tests? I remember we needed this node up to get a live token for the tests in case of timeout. Do we not need this extra node because this process only happens once?

orenbm: this still works because instead of having an azure node that is just used for getting an Azure AD token, and another one that runs the tests using that token, we now run the tests inside the azure node. this node SSHs into itself and retrieves the token.

This way, we have one less node alive as we don't need the linux node that ran the tests.

orenbm: > I remember we needed this node up to get a live token for the tests in case of timeout

this is incorrect. we had this node to get a live token for the tests. there was nothing to do with a timeout. We just didn't think of the idea to run the actual tests on this node at the time.

PR #1751, opened 2020-08-16
Gcp validate resource restrictions match jwt

nessiLahav:
### What does this PR do?
Implement `ValidateResourceRestrictionsMatchJWT` command class, which is the 3rd dependency of the `ValidateResourceRestriction` command class

### DIFF
--- a/app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
+++ b/app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
@@ -32,14 +32,11 @@ module Authentication
       end
 
       def restriction_type_without_prefix(restriction_type)
-        restriction_type_without_prefix = restriction_type.sub(@restriction_prefix, "")
-        restriction_type_without_prefix
+        restriction_type.sub(@restriction_prefix, "")
       end
 
-      def restriction_value_from_token(restriction_type_without_prefix)
-        restriction_value_from_token = nil
-
-        case restriction_type_without_prefix
+      def restriction_value_from_token(restriction_type)
+        case restriction_type
         when PROJECT_ID_RESTRICTION_NAME
           restriction_value_from_token = @decoded_token.project_id
         when INSTANCE_NAME_RESTRICTION_NAME
--- a/app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
+++ b/app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
@@ -1,0 +1,57 @@
+module Authentication
+  module AuthnGcp
+
+    # This class is responsible of validating resource restrictions values against JWT token.
+    # the assumption is that resource restrictions list contains only permitted types which already validated
+    # in previous steps
+    ValidateResourceRestrictionsMatchJWT = CommandClass.new(
+      dependencies: {
+        logger: Rails.logger
+      },
+      inputs:       %i(resource_restrictions decoded_token restriction_prefix)
+    ) do
+
+      def call
+        @logger.debug(LogMessages::Authentication::AuthnGcp::ValidatingResourceRestrictionsValues.new)
+        validate
+        @logger.debug(LogMessages::Authentication::AuthnGcp::ValidatedResourceRestrictionsValues.new)
+      end
+
+      private
+
+      def validate
+        @resource_restrictions.each do |r|
+          restriction_type = r.type
+          restriction_value = r.value
+          restriction_value_from_token = restriction_value_from_token(restriction_type_without_prefix(restriction_type))
+          unless restriction_value == restriction_value_from_token
+            raise Errors::Authentication::Jwt::InvalidResourceRestrictions.new(restriction_type)
+          end
+          @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictionValue.new(restriction_type))
+        end
+      end
+
+      def restriction_type_without_prefix(restriction_type)
+        restriction_type_without_prefix = restriction_type.sub(@restriction_prefix, "")
+        restriction_type_without_prefix
+      end
+
+      def restriction_value_from_token(restriction_type_without_prefix)
+        restriction_value_from_token = nil
+
+        case restriction_type_without_prefix
+        when PROJECT_ID_RESTRICTION_NAME
+          restriction_value_from_token = @decoded_token.project_id
+        when INSTANCE_NAME_RESTRICTION_NAME
+          restriction_value_from_token = @decoded_token.instance_name
+        when SERVICE_ACCOUNT_ID_RESTRICTION_NAME
+          restriction_value_from_token = @decoded_token.service_account_id
+        when SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME
+          restriction_value_from_token = @decoded_token.service_account_email
+        end
+
+        restriction_value_from_token
+      end
+    end
+  end
+end
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -181,6 +181,12 @@ module Errors
         code: "CONJ00051E"
       )
 
+      InvalidResourceRestrictions = ::Util::TrackableErrorClass.new(
+        msg:  "Resource restriction '{0-resource-restriction-name}' does not match " \
+            "resource in JWT token",
+        code: "CONJ00049E"
+      )
+
     end
 
     module AuthnOidc
@@ -324,12 +330,6 @@ module Errors
     end
 
     module AuthnAzure
-
-      InvalidResourceRestrictions = ::Util::TrackableErrorClass.new(
-        msg:  "Resource restriction '{0-resource-restriction-name}' does not match " \
-            "resource in Azure token",
-        code: "CONJ00049E"
-      )
 
       XmsMiridParseError = ::Util::TrackableErrorClass.new(
         msg:  "Failed to parse xms_mirid {0}. Reason: {1}",
--- a/app/domain/authentication/authn_azure/validate_resource_restrictions.rb
+++ b/app/domain/authentication/authn_azure/validate_resource_restrictions.rb
@@ -36,7 +36,7 @@ module Authentication
         resource_restrictions.resources.each do |resource_from_role|
           resource_from_request = resources_from_request.find { |resource| resource == resource_from_role }
           unless resource_from_request
-            raise Errors::Authentication::AuthnAzure::InvalidResourceRestrictions, resource_from_role.type
+            raise Errors::Authentication::Jwt::InvalidResourceRestrictions, resource_from_role.type
           end
         end
         @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictions.new)
--- a/app/domain/authentication/authn_gcp/consts.rb
+++ b/app/domain/authentication/authn_gcp/consts.rb
@@ -3,6 +3,10 @@ module Authentication
   module AuthnGcp
 
     PROVIDER_URI = "https://accounts.google.com"
+    PROJECT_ID_RESTRICTION_NAME = "project-id"
+    INSTANCE_NAME_RESTRICTION_NAME = "instance-name"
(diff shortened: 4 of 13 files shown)


app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+
+      private
+
+      def validate
+        @resource_restrictions.each do |r|
+          resource_type = r.type
+          resource_value = r.value
+
+          case remove_prefix(resource_type)
```
orenbm, line 27: i would add a line before this `resource_type = remove_prefix(resource_type)` for readability

app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
```diff
@@ -0,0 +1,60 @@
...
+              raise Errors::Authentication::AuthnGcp::InvalidResourceRestrictions.new(resource_type)
+            end
+            @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictionValue.new(resource_type))
+
+          when SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME
+            unless resource_value == @decoded_token.service_account_email
+              raise Errors::Authentication::AuthnGcp::InvalidResourceRestrictions.new(resource_type)
+            end
+            @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictionValue.new(resource_type))
+          end
```
orenbm, line 51: i'd like to suggest another solution, similar to what we have in `authn-azure`. Instead of having a switch case that runs over all the options and performs a similar block of code, we can add another method called `jwt_token_resources` that will create a list of `ResourceRestrictions` objects. Then, the `validate` method will verify that it contains all of the objects of `resource_restrictions`.

in `authn-azure` it looks like this:
```
def validate_resource_restrictions_match_request
        resource_restrictions.resources.each do |resource_from_role|
          resource_from_request = resources_from_request.find { |resource| resource == resource_from_role }
          unless resource_from_request
            raise Errors::Authentication::AuthnAzure::InvalidResourceRestrictions, resource_from_role.type
          end
        end
        @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictions.new)
      end

      # xms_mirid is a term in Azure to define a claim that describes the resource
      # that holds the encoding of the instance's among other details the subscription_id,
      # resource group, and provider identity needed for authorization.
      # xms_mirid is one of the fields in the JWT token. This function will extract the relevant information from
      # xms_mirid claim and populate a representative hash with the appropriate fields.
      def resources_from_request
        return @resources_from_request if @resources_from_request

        @resources_from_request = [
          AzureResource.new(
            type: "subscription-id",
            value: xms_mirid.subscriptions
          ),
          AzureResource.new(
            type: "resource-group",
            value: xms_mirid.resource_groups
          )
        ]

        # determine which identity is provided in the token. If the token is
        # issued to a user-assigned identity then we take the identity name.
        # If the token is issued to a system-assigned identity then we take the
        # Object ID of the token.
        if xms_mirid.providers.include? "Microsoft.ManagedIdentity"
          @resources_from_request.push(
            AzureResource.new(
              type: "user-assigned-identity",
              value: xms_mirid.providers.last
            )
          )
        else
          @resources_from_request.push(
            AzureResource.new(
              type: "system-assigned-identity",
              value: @oid_token_field
            )
          )
        end
        @logger.debug(LogMessages::Authentication::AuthnAzure::ExtractedResourceRestrictionsFromToken.new)
        @resources_from_request
      end
```

One issue that rises from this is that having a `ResourceRestriction` objects list of the JWT is quite weird because the JWT doesn't hold restrictions. it holds resources. To that i would say that the `ResourceRestriction` object does not really represent a resource restriction, rather than a resource (we restrict on it but the object itself reperesents a resource). So i would change the code to the above and refactor `ResourceRestriction` to `Resource`.

wdyt?

app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
```diff
@@ -0,0 +1,60 @@
...
+        end
+      end
+
+      def remove_prefix(restriction_type)
```
orenbm, line 55: when methods return an object we give it the name of the returned object. so i would change this to `restriction_type_without_prefix` or `non-prefixed-restriction_type`.

app/domain/errors.rb
```diff
@@ -367,6 +367,12 @@ module AuthnGcp
         msg:  "Role should have at least one of the following constraints: {0-constraints}",
         code: "CONJ00069E"
       )
+
+      InvalidResourceRestrictions = ::Util::TrackableErrorClass.new(
+        msg:  "Resource restriction '{0-resource-restriction-name}' does not match " \
+            "resource in GCP token",
+        code: "CONJ00072E"
+      )
```
orenbm, line 375: i suggest that we move error `CONJ00049E` to the JWT module and replace `Azure` with `JWT`. this way we can use it also here (and it will make sense)

app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
nessiLahav: Thanks 🙏  for the elaborated suggestion, i liked your approach but it still will have a duplicate code block, for example in your solution the duplication will be in the Array creation:
```
      def resources_from_request
        return @resources_from_request if @resources_from_request

        @resources_from_request = [
          ResourceRestriction.new(
            type: PROJECT_ID_RESTRICTION_NAME,
            value: @decoded_token.project_id
          ),
          ResourceRestriction.new(
            type: INSTANCE_NAME_RESTRICTION_NAME,
            value: @decoded_token.instance_name
          ),
          ResourceRestriction.new(
            type: SERVICE_ACCOUNT_ID_RESTRICTION_NAME,
            value: @decoded_token.service_account_id
          ),
          ResourceRestriction.new(
            type: SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME,
            value: @decoded_token.service_account_email
          )
        ]
      end
```

What do you think about the following alternative:
```
      def validate
        @resource_restrictions.each do |r|
          restriction_type = r.type
          restriction_value = r.value
          restriction_type_without_prefix = remove_prefix(restriction_type)
          restriction_value_from_token = get_value_from_token(restriction_type_without_prefix)
          unless restriction_value == restriction_value_from_token
            raise Errors::Authentication::AuthnGcp::InvalidResourceRestrictions.new(restriction_type)
          end
          @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictionValue.new(restriction_type))
        end
      end

      def remove_prefix(restriction_type)
        restriction_type_without_prefix = restriction_type.sub(@restriction_prefix, "")
        restriction_type_without_prefix
      end

      def get_value_from_token(restriction_type_without_prefix)
        restriction_value_from_token = nil

        case restriction_type_without_prefix
        when PROJECT_ID_RESTRICTION_NAME
          restriction_value_from_token = @decoded_token.project_id
        when INSTANCE_NAME_RESTRICTION_NAME
          restriction_value_from_token = @decoded_token.instance_name
        when SERVICE_ACCOUNT_ID_RESTRICTION_NAME
          restriction_value_from_token = @decoded_token.service_account_id
        when SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME
          restriction_value_from_token = @decoded_token.service_account_email
        end

        restriction_value_from_token
      end
```

I think it is more readable, and the code duplication here is minimal.

WDYT?

orenbm: yes, i'm ok with your suggestion. if you do go with it then you don't need the `==` method implementation in `ResourceRestriction` class (i added that to compare between `ResourceRestriction` objects but you don't do it here).

also please change `remove_prefix` to `restriction_type_without_prefix` or `non_prefixed_restriction_type`

app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      private
+
+      def validate
+        @resource_restrictions.each do |r|
+          restriction_type = r.type
+          restriction_value = r.value
+          restriction_value_from_token = restriction_value_from_token(restriction_type_without_prefix(restriction_type))
+          unless restriction_value == restriction_value_from_token
+            raise Errors::Authentication::Jwt::InvalidResourceRestrictions.new(restriction_type)
```
orenbm, line 28: do we want to print also the value in the host? do we want to print also the value in the token? 

are these considered sensitive information?

app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
```diff
@@ -0,0 +1,57 @@
...
+          end
+          @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictionValue.new(restriction_type))
+        end
+      end
+
+      def restriction_type_without_prefix(restriction_type)
+        restriction_type_without_prefix = restriction_type.sub(@restriction_prefix, "")
+        restriction_type_without_prefix
+      end
```
orenbm, line 37: this can be changed to 
```
      def restriction_type_without_prefix(restriction_type)
        restriction_type.sub(@restriction_prefix, "")
      end
```

app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      def restriction_value_from_token(restriction_type_without_prefix)
```
orenbm, line 39: the `_without_prefix` is redundant in the context of this function

app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
```diff
@@ -0,0 +1,57 @@
...
+        restriction_value_from_token = nil
```
orenbm, line 40: i don't think you need to explicitly set this

app/domain/authentication/authn_gcp/validate_resource_restrictions_match_jwt.rb
nessiLahav: IMO both values (in the host and in the token) are sensitive, WDYT?

orenbm: first, let's add @sashaCher to the mix

orenbm: the thing is that we assume that whoever reads the logs has access to the Conjur container. So they already have some privileges.

Now, the annotations are part of the configuration and thus are not sensitive (or else they would have been set as secrets).

nessiLahav: As i see at:
1. The token is equivalent to password, so in my opinion we should not print its values to the logs
2. Regarding the annotations you are right they are exists in the policy so they already exposed, but i do not like it because it is not a really annotations and more like a secrets, but we choose to implement it as annotations, so the decision here should take into consideration that those restrictions values may will be implemented in a different way in the future.

orenbm: when in doubt, do not print. let's continue without printing.

if we'll see a case where it would have helped us then we can revisit this decision.

sashaCher: Annotations are not secret in general but in any case they can be accused only by authenticated Conjur user with relevant permissions only. The person that reads logs can have significantly lower permissions means may be it has only monitoring access on relevant environment.
So I think that we should no print values in logs.

@orenbm 
> when in doubt, do not print. let's continue without printing.

Good approach!

PR #1752, opened 2020-08-17
Gcp validate resource restrictions

nessiLahav:
### What does this PR do?
Implement the last piece of GCP authenticator by adding resource restrictions validation logic

### DIFF
--- a/app/domain/authentication/authn_gcp/authenticator.rb
+++ b/app/domain/authentication/authn_gcp/authenticator.rb
@@ -5,22 +5,23 @@ module Authentication
 
     Authenticator = CommandClass.new(
       dependencies: {
-        logger: Rails.logger
+        validate_resource_restrictions: ValidateResourceRestrictions.new,
+        logger:                         Rails.logger
       },
       inputs:       [:authenticator_input]
     ) do
 
-      extend Forwardable
-      def_delegators :@authenticator_input, :account, :credentials, :username
-
       def call
-        # TODO: validate resource restrictions
-        # Note: 'credentials' is now the decoded token. you can access its
-        # fields with an argument reader (e.g credentials.project_id)
+        validate_resource_restrictions
       end
 
       private
 
+      def validate_resource_restrictions
+        @validate_resource_restrictions.call(
+          authenticator_input: @authenticator_input
+        )
+      end
     end
 
     class Authenticator
--- a/app/domain/authentication/authn_gcp/authenticator.rb
+++ b/app/domain/authentication/authn_gcp/authenticator.rb
@@ -11,6 +11,9 @@ module Authentication
       inputs:       [:authenticator_input]
     ) do
 
+      extend Forwardable
+      def_delegators :@authenticator_input, :account, :username, :credentials
+
       def call
         validate_resource_restrictions
       end
@@ -19,7 +22,9 @@ module Authentication
 
       def validate_resource_restrictions
         @validate_resource_restrictions.call(
-          authenticator_input: @authenticator_input
+          account:     account,
+          username:    username,
+          credentials: credentials
         )
       end
     end
--- a/app/domain/authentication/authn_gcp/consts.rb
+++ b/app/domain/authentication/authn_gcp/consts.rb
@@ -7,6 +7,8 @@ module Authentication
     INSTANCE_NAME_RESTRICTION_NAME = "instance-name"
     SERVICE_ACCOUNT_ID_RESTRICTION_NAME = "service-account-id"
     SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME = "service-account-email"
+    PERMITTED_CONSTRAINTS = %w(authn-gcp/instance-name authn-gcp/project-id authn-gcp/service-account-id authn-gcp/service-account-email).freeze
+    AUTHN_PREFIX = "authn-gcp/"
 
   end
 end
--- a/app/domain/authentication/authn_gcp/consts.rb
+++ b/app/domain/authentication/authn_gcp/consts.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 
 module Authentication
   module AuthnGcp
@@ -7,7 +8,7 @@ module Authentication
     INSTANCE_NAME_RESTRICTION_NAME = "instance-name"
     SERVICE_ACCOUNT_ID_RESTRICTION_NAME = "service-account-id"
     SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME = "service-account-email"
-    PERMITTED_CONSTRAINTS = %w(authn-gcp/instance-name authn-gcp/project-id authn-gcp/service-account-id authn-gcp/service-account-email).freeze
+    PERMITTED_CONSTRAINTS = %w(instance-name project-id service-account-id service-account-email)
     AUTHN_PREFIX = "authn-gcp/"
 
   end
--- a/app/domain/authentication/authn_gcp/validate_resource_restrictions.rb
+++ b/app/domain/authentication/authn_gcp/validate_resource_restrictions.rb
@@ -1,0 +1,53 @@
+module Authentication
+  module AuthnGcp
+
+    # This class is responsible of validating resource restrictions that configured on a Conjur host or user
+    # against google JWT token.
+    ValidateResourceRestrictions = CommandClass.new(
+      dependencies: {
+        extract_resource_restrictions:                ExtractResourceRestrictions.new,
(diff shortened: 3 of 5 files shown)


app/domain/authentication/authn_gcp/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,53 @@
+module Authentication
+  module AuthnGcp
+
+    # This class is responsible of validating resource restrictions that configured on a Conjur host or user
```
eranha, line 4: `that` doesn't seem ion place, maybe `which are` or just omit it.

app/domain/authentication/authn_gcp/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,53 @@
...
+        logger:                                       Rails.logger
+      },
+      inputs:       [:authenticator_input]
+    ) do
+
+      extend Forwardable
+      def_delegators :@authenticator_input, :account, :username, :credentials
+
+      def call
+        @logger.debug(LogMessages::Authentication::ValidatingResourceRestrictions.new)
```
eranha, line 20: is that enter message?

app/domain/authentication/authn_gcp/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,53 @@
...
+        extract_resource_restrictions
+        validate_resource_restrictions_configuration
+        validate_resource_restrictions_match_jwt
+        @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictions.new)
```
eranha, line 24: is that exit message?

spec/app/domain/authentication/authn-gcp/validate_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+  #  ____  _   _  ____    ____  ____  ___  ____  ___
+  # (_  _)( )_( )( ___)  (_  _)( ___)/ __)(_  _)/ __)
+  #   )(   ) _ (  )__)     )(   )__) \__ \  )(  \__ \
+  #  (__) (_) (_)(____)   (__) (____)(___/ (__) (___/
+
+  context "A valid resource restrictions configuration" do
+    context "when all resource restrictions values exists in GCP token" do
+      subject do
+        ::Authentication::AuthnGcp::ValidateResourceRestrictions.new(
+          extract_resource_restrictions: mocked_resource_restrictions_return_4_valid
```
eranha, line 134: name not clear, don't couple to number it may change

app/domain/authentication/authn_gcp/validate_resource_restrictions.rb
nessiLahav: yes

app/domain/authentication/authn_gcp/validate_resource_restrictions.rb
nessiLahav: yes

app/domain/authentication/authn_gcp/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,53 @@
...
+    # This class is responsible of validating resource restrictions that configured on a Conjur host or user
+    # against google JWT token.
+    ValidateResourceRestrictions = CommandClass.new(
+      dependencies: {
+        extract_resource_restrictions:                ExtractResourceRestrictions.new,
+        validate_resource_restrictions_configuration: ValidateResourceRestrictionsConfiguration.new,
+        validate_resource_restrictions_match_jwt:     ValidateResourceRestrictionsMatchJWT.new,
+        logger:                                       Rails.logger
+      },
+      inputs:       [:authenticator_input]
```
orenbm, line 13: i think that it's not very good to pass around the `authenticator_input` object. instead, it will be better to state `:account, :username, :credentials` so that it's clear from the class signature what the input is.

wdyt?

nessiLahav: OK

app/domain/authentication/authn_gcp/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      private
+
+      def extract_resource_restrictions
+        @resource_restrictions ||= @extract_resource_restrictions.call(
+          account:           account,
+          username:          username,
+          extraction_prefix: AUTHN_PREFIX
+        )
+      end
```
orenbm, line 35: when you use memoization you need the memoized object name to match the method name. what you could do is:
```
        resource_restrictions
      end

      def resource_restrictions
        @resource_restrictions ||= @extract_resource_restrictions.call(
          account:           account,
          username:          username,
          extraction_prefix: AUTHN_PREFIX
        )
      end
```

app/domain/authentication/authn_gcp/consts.rb
```diff
@@ -7,6 +7,8 @@ module AuthnGcp
     INSTANCE_NAME_RESTRICTION_NAME = "instance-name"
     SERVICE_ACCOUNT_ID_RESTRICTION_NAME = "service-account-id"
     SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME = "service-account-email"
+    PERMITTED_CONSTRAINTS = %w(authn-gcp/instance-name authn-gcp/project-id authn-gcp/service-account-id authn-gcp/service-account-email).freeze
```
orenbm, line 10: you can add `# frozen_string_literal: true` to the head of the file and remove the `.freeze` here

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -5,22 +5,25 @@ module AuthnGcp
...
       dependencies: {
-        logger: Rails.logger
+        validate_resource_restrictions: ValidateResourceRestrictions.new,
+        logger:                         Rails.logger
       },
       inputs:       [:authenticator_input]
     ) do
 
-      extend Forwardable
-      def_delegators :@authenticator_input, :account, :credentials, :username
```
orenbm, line 14: i would leave this

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -5,22 +5,25 @@ module AuthnGcp
...
+        validate_resource_restrictions
       end
 
       private
 
+      def validate_resource_restrictions
+        @validate_resource_restrictions.call(
+          account:     @authenticator_input.account,
+          username:    @authenticator_input.username,
+          credentials: @authenticator_input.credentials
```
orenbm, line 24: let's use the def_delegators

app/domain/authentication/authn_gcp/consts.rb
```diff
@@ -7,6 +8,8 @@ module AuthnGcp
     INSTANCE_NAME_RESTRICTION_NAME = "instance-name"
     SERVICE_ACCOUNT_ID_RESTRICTION_NAME = "service-account-id"
     SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME = "service-account-email"
+    PERMITTED_CONSTRAINTS = %w(authn-gcp/instance-name authn-gcp/project-id authn-gcp/service-account-id authn-gcp/service-account-email)
```
orenbm, line 11: do we really need the `authn-gcp/` prefix here? can't we add it later so this const looks nicer?

orenbm: something like this:
```
PERMITTED_CONSTRAINTS = %w(instance-name project-id service-account-id service-account-email)

prefixed_permitted_constraints = PERMITTED_CONSTRAINTS.map { |k| "authn-gcp/#{k}" }
```

or even
```
PERMITTED_CONSTRAINTS = %w(instance-name project-id service-account-id service-account-email)

prefixed_permitted_constraints = PERMITTED_CONSTRAINTS.map { |k| "#{AUTHN_PREFIX}#{k}" }
```

btw, you can use `authenticator_name` instead of `AUTHN_PREFIX`.
```
PERMITTED_CONSTRAINTS = %w(instance-name project-id service-account-id service-account-email)

prefixed_permitted_constraints = PERMITTED_CONSTRAINTS.map { |k| "#{authenticator_name}/#{k}" }
```

nessiLahav: Done :)

spec/app/domain/authentication/authn-gcp/validate_resource_restrictions_spec.rb
nessiLahav: done

PR #1755, opened 2020-08-18
move parser to conjur working phase I

oburstein-hub:
### What does this PR do?
This PR adds code from conjur-policy-parser git repo to conjur repo
The motivation behind it to make code more easily maintainable / debugable

### Checklists
Automatic test running on merge will raise issues if there will be

#### Test coverage
Test coverage does not change, because this change does not influence functionalty


oburstein-hub:
Hi @jvanderhoof and @micahlee 
Do you mean that after your changes conjur-policy-parser will be placed in the same github repo as conjur ?
Does it mean that for example print addiion in conjur-policy-parser will appear at runtime immediately after rerunning conjur ?
Please, elaborate

Issue #1761, opened 2020-08-19
Automated Test Verify Behavior of Invalid CIDR string in `restricted_to`


micahlee:
We are currently missing a test that verifies the policy load behavior (failure) when attempting to load a policy that includes an invalid CIDR notation string in the `restricted_to` attribute for roles and host factories.

PR #1762, opened 2020-08-19
Add test for invalid CIDR string in policy load

micahlee:
### What does this PR do?

This PR adds tests for the existing behavior of attempting to load policy with an invalid `restricted_to` attribute on a host.

Connected to #1761 
Connected to #1763

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/cucumber/policy/features/cidr_restrictions.feature
+++ b/cucumber/policy/features/cidr_restrictions.feature
@@ -37,3 +37,41 @@ from a particular network, defined by a CIDR in the policy
     """
        ["192.168.0.1/32", "192.168.1.10/32"]
     """
+
+  Scenario: Invalid CIDR restriction string
+
+    When I load a policy:
+    """
+    - !host
+      id: serviceA
+      restricted_to: an_invalid_cidr_string
+    """
+    Then there is an error
+    And the error code is "validation_failed"
+    And the error message includes "Invalid IP address or CIDR range 'an_invalid_cidr_string'"
+    
+
+  Scenario: Domain name as CIDR restriction string
+
+    When I load a policy:
+    """
+    - !host
+      id: serviceA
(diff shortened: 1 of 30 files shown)


cucumber/policy/features/cidr_restrictions.feature
```diff
@@ -37,3 +37,28 @@ from a particular network, defined by a CIDR in the policy
...
+
+    When I load a policy:
+    """
+    - !host
+      id: serviceA
+      restricted_to: an_invalid_cidr_string
+    """
+    Then there is an error
+    And the error code is "validation_failed"
+    And the error message is "policy_text Error at line 2, column 39 in  : Invalid IP address or CIDR range 'an_invalid_cidr_string'"
```
jtuttle, line 51: Might be good to make this less specific so that it's something more like:

```
    And the error message includes "Invalid IP address or CIDR range 'an_invalid_cidr_string'"
```

Otherwise if you add a line to the policy for some reason you're going to have to go through and update all these error messages for no good reason.

cucumber/policy/features/step_definitions/policy_steps.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-Given(/^a policy:$/) do |policy|
+Given(/^(?:I load )?a policy:$/) do |policy|
```
jtuttle, line 3: I would prefer to make the "I load" mandatory here and just update all the places this is used, but I imagine that requires changing A LOT of files so if that's a pain forget about it for now. Definitely better things to do.

cucumber/policy/features/cidr_restrictions.feature
micahlee: Good call!

PR #1778, opened 2020-08-25
Add GCE Authenticator

orenbm:
### What does this PR do?
Adds the GCE Authenticator to Conjur. It also adds integrations tests for it that run on Jenkins.

@shulifink @InbalZilberman please review the messages that were added to [`logs.rb`](https://github.com/cyberark/conjur/pull/1778/files#diff-17dce2adc91ec6547779c7023aef1f95) and [`errors.rb`](https://github.com/cyberark/conjur/pull/1778/files#diff-3f3e09f89ed6ec49603919967e554686).

@ismarc plese review the changes in [`Jenkinsfile`](https://github.com/cyberark/conjur/pull/1778/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3).

#### Change log

#### Test coverage
- [x] This PR includes new unit and integration tests to go with the code changes

#### Documentation
- [x] Docs will be updated in a [separate effort](https://github.com/conjurinc/appliance/issues/1372)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Hosts can authenticate from Google Compute Engines (GCE) using a GCE instance 
+  identity token. See [design](design/authenticators/authn_gce/authn_gce_solution_design.md) 
+  for details ([cyberark/conjur#1711](https://github.com/cyberark/conjur/issues/1711)).
 - Add `/whoami` API endpoint for improved supportability and debugging for access
   tokens and client IP address determination. [cyberark/conjur#1697](https://github.com/cyberark/conjur/issues/1697)
 - `TRUSTED_PROXIES` is validated at Conjur startup to ensure that it contains
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -97,6 +97,60 @@ pipeline {
             }
           }
         }
+        // We have 2 stages for GCE Authenticator tests. The first one runs inside
+        // a GCE instance and retrieves all the tokens that will be used in the tests.
+        // It then stashes the tokens, which are unstashed in the stage that runs the
+        // GCE Authenticator tests using the tokens.
+        // This way we can have a light-weight GCE instance that has no need for conjurops
+        // or git identities and is not open for SSH
+        stage('GCE Authenticator preparation - Allocate GCE Instance') {
+          steps {
+            script {
+              node('executor-v2-gcp-small') {
+                sh '''
+                  retrieve_token() {
+                    local token_format="$1"
+                    local audience="$2"
+
+                    curl \
+                      -s \
+                      -H 'Metadata-Flavor: Google' \
+                      "http://metadata/computeMetadata/v1/instance/service-accounts/default/identity?format=${token_format}&audience=${audience}"
+                  }
+
+                  echo "$(retrieve_token "full" "conjur/cucumber/host/test-app")" > "gce_token_valid"
+                  echo "$(retrieve_token "full" "conjur/cucumber/host/non-existing")" > "gce_token_non_existing_host"
+                  echo "$(retrieve_token "full" "conjur/cucumber/host/non-rooted/test-app")" > "gce_token_non_rooted_host"
+                  echo "$(retrieve_token "full" "conjur/cucumber/test-app")" > "gce_token_user"
+                  echo "$(retrieve_token "full" "conjur/non-existing/host/test-app")" > "gce_token_non_existing_account"
+                  echo "$(retrieve_token "full" "invalid_audience")" > "gce_token_invalid_audience"
+                  echo "$(retrieve_token "standard" "conjur/cucumber/host/test-app")" > "gce_token_standard_format"
+                '''
+
+                stash name: 'authnGceTokens', includes: 'gce_token_valid,gce_token_invalid_audience,gce_token_standard_format,gce_token_user,gce_token_non_existing_host,gce_token_non_existing_account,gce_token_non_rooted_host', allowEmpty:false
+                env.GCE_TOKENS_FETCHED = "true"
+              }
+            }
+          }
+        }
+        stage('GCE Authenticator') {
+          steps {
+            timeout(time: 10, unit: 'MINUTES') {
+              waitUntil {
+                script {
+                  return (env.GCE_TOKENS_FETCHED == "true")
+                }
+              }
+            }
+            script {
+              dir('ci/authn-gce/tokens') {
+                unstash 'authnGceTokens'
+              }
+
+              sh 'ci/test cucumber_authenticators_gce'
+            }
+          }
+        }
         stage('Policy') {
           steps { sh 'ci/test cucumber_policy' }
         }
--- a/app/domain/authentication/authn_gce/extract_resource_restrictions.rb
+++ b/app/domain/authentication/authn_gce/extract_resource_restrictions.rb
@@ -1,0 +1,86 @@
+require 'command_class'
+
+module Authentication
+  module AuthnGce
+
+    # This class is responsible of restrictions extraction that are set on a Conjur host or user as annotations.
+    ExtractResourceRestrictions = CommandClass.new(
+      dependencies: {
+        role_class:              ::Role,
+        resource_class:          ::Resource,
+        validate_account_exists: ::Authentication::Security::ValidateAccountExists.new,
+        logger:                  Rails.logger
+      },
+      inputs:       %i(account username extraction_prefix)
+    ) do
+
+      def call
+        validate_account_exists
+        extract_resource_restrictions
+        resource_restrictions
+      end
+
+      private
+
+      def validate_account_exists
+        @validate_account_exists.(
+          account: @account
+        )
+      end
+
+      def extract_resource_restrictions
+        @logger.debug(
+          LogMessages::Authentication::AuthnGce::ExtractingRestrictionsFromResource.new(
+            @extraction_prefix,
+            @username
+          )
+        )
+        prefixed_resource_annotations.select do |a|
+          annotation_name = a.values[:name]
+          resource_value = annotation_value(annotation_name)
+          next unless resource_value
+          resource_restrictions.push(
+            ResourceRestriction.new(
+              type: annotation_name,
+              value: resource_value
+            )
+          )
+        end
+        @logger.debug(LogMessages::Authentication::AuthnGce::ExtractedResourceRestrictions.new(resource_restrictions.length))
+      end
+
+      def prefixed_resource_annotations
+        @prefixed_resource_annotations ||= role.annotations.select do |a|
+          annotation_name = a.values[:name]
+          annotation_name.start_with?(@extraction_prefix)
+        end
+      end
+
+      def resource_restrictions
+        @resource_restrictions ||= []
+      end
+
+      def annotation_value(name)
+        annotation = prefixed_resource_annotations.find {|a| a.values[:name] == name}
+
+        # return the value of the annotation if it exists, nil otherwise
+        if annotation
+          @logger.debug(LogMessages::Authentication::RetrievedAnnotationValue.new(name))
+          annotation[:value]
+        end
+      end
+
+      def role
+        return @role if @role
+
+        @role = @resource_class[role_id]
+        raise Errors::Authentication::Security::RoleNotFound, role_id unless @role
+        @role
+      end
+
+      def role_id
+        @role_id ||= @role_class.roleid_from_username(@account, @username)
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_gce/resource_restriction.rb
+++ b/app/domain/authentication/authn_gce/resource_restriction.rb
@@ -1,0 +1,17 @@
+module Authentication
+  module AuthnGce
+
+    class ResourceRestriction
+      attr_reader :type, :value
+
+      def initialize(type:, value:)
+        @type = type
+        @value = value
+      end
+
+      def ==(other)
+        @type == other.type && @value == other.value
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_gce/update_authenticator_input.rb
+++ b/app/domain/authentication/authn_gce/update_authenticator_input.rb
@@ -1,0 +1,93 @@
+require 'command_class'
+
+module Authentication
+  module AuthnGce
+
+    # Update the 'authenticator_input' object to include the username and the decoded token
+    # The general 'Authenticate' command class performs validations on the user
+    # so we need to first extract the username from the given token.
+    # We also update the 'credentials' field in the 'authenticator_input' object
+    # to include the decoded token instead of the original one so we don't need
+    # to decode it again
+    UpdateAuthenticatorInput = CommandClass.new(
+      dependencies: {
+        verify_and_decode_token: Authentication::OAuth::VerifyAndDecodeToken.new,
+        decoded_token_class:     DecodedToken,
+        logger:                  Rails.logger
+      },
+      inputs:       [:authenticator_input]
+    ) do
+
+      extend Forwardable
+      def_delegators :@authenticator_input, :authenticator_name, :account, :credentials, :username
+
+      def call
+        validate_token
+        updated_input
+      end
+
+      private
+
+      def validate_token
+        decoded_token
+        validate_audience
+      end
+
+      def decoded_token
+        @decoded_token ||= @decoded_token_class.new(
+          decoded_token_hash: @verify_and_decode_token.call(
+            provider_uri:     PROVIDER_URI,
+            token_jwt:        decoded_credentials.jwt,
+            claims_to_verify: {
+              verify_iss:        true,
+              iss:               PROVIDER_URI,
+              verify_iat:        true,
+              verify_expiration: true
+            }
+          ),
+          logger:             @logger
+        )
+      end
+
+      def decoded_credentials
+        @decoded_credentials ||= Authentication::Jwt::DecodedCredentials.new(credentials)
+      end
+
+      def validate_audience
+        if audience_parts.length != 3 ||
+            audience_parts[0] != "conjur" ||
+            audience_parts[1] != account
+          raise Errors::Authentication::AuthnGce::InvalidAudience, audience
+        end
+      end
+
+      def audience_parts
+        @audience_parts ||= audience.split('/', 3)
+      end
+
+      def audience
+        @audience ||= @decoded_token.audience
+      end
+
+      def conjur_username
+        return @conjur_username if @conjur_username
+
+        @conjur_username = audience_parts[2]
+
+        @logger.debug(
+          LogMessages::Authentication::Jwt::ExtractedUsernameFromToken.new(
+            conjur_username
+          )
+        )
+        @conjur_username
+      end
+
+      def updated_input
+        @authenticator_input.update(
+          username: conjur_username,
+          credentials: decoded_token
+        )
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_gce/validate_resource_restrictions.rb
+++ b/app/domain/authentication/authn_gce/validate_resource_restrictions.rb
@@ -1,0 +1,66 @@
+module Authentication
+  module AuthnGce
+
+    # This class is responsible for validating resource restrictions which are configured on a Conjur host or user
+    # against google JWT token.
+    ValidateResourceRestrictions = CommandClass.new(
+      dependencies: {
+        extract_resource_restrictions:                ExtractResourceRestrictions.new,
+        validate_resource_restrictions_configuration: ValidateResourceRestrictionsConfiguration.new,
+        validate_resource_restrictions_match_jwt:     ValidateResourceRestrictionsMatchJWT.new,
+        logger:                                       Rails.logger
+      },
+      inputs:       %i(authenticator_name account username credentials)
+    ) do
+
+      def call
+        @logger.debug(
+          LogMessages::Authentication::ValidatingResourceRestrictions.new(
+            @username
+          )
+        )
+        extract_resource_restrictions
+        validate_resource_restrictions_configuration
+        validate_resource_restrictions_match_jwt
+        @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictions.new)
+      end
+
+      private
+
+      def extract_resource_restrictions
+        resource_restrictions
+      end
+
+      def resource_restrictions
+        @resource_restrictions ||= @extract_resource_restrictions.call(
+          account:           @account,
+          username:          @username,
+          extraction_prefix: restriction_prefix
+        )
+      end
+
+      def validate_resource_restrictions_configuration
+        @validate_resource_restrictions_configuration.call(
+          resource_restrictions: resource_restrictions,
+          permitted_constraints: prefixed_permitted_constraints
+        )
+      end
+
+      def validate_resource_restrictions_match_jwt
+        @validate_resource_restrictions_match_jwt.call(
+          resource_restrictions: resource_restrictions,
+          decoded_token:         @credentials,
+          annotation_prefix:     restriction_prefix
+        )
+      end
+
+      def prefixed_permitted_constraints
+        @prefixed_permitted_constraints ||= PERMITTED_CONSTRAINTS.map { |c| "#{restriction_prefix}#{c}" }
+      end
+
+      def restriction_prefix
+        @restriction_prefix ||= "#{@authenticator_name}/"
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_gce/validate_resource_restrictions_configuration.rb
+++ b/app/domain/authentication/authn_gce/validate_resource_restrictions_configuration.rb
@@ -1,0 +1,49 @@
+module Authentication
+  module AuthnGce
+
+    # This class is responsible for validating resource restrictions configuration.
+    # the allowed configuration is: Role should have at least 1 permitted constraints, and non of illegal constraints
+    ValidateResourceRestrictionsConfiguration = CommandClass.new(
+      dependencies: {
+        logger: Rails.logger
+      },
+      inputs:       %i(resource_restrictions permitted_constraints)
+    ) do
+
+      def call
+        @logger.debug(LogMessages::Authentication::AuthnGce::ValidatingResourceRestrictionsConfiguration.new)
+        validate_required_restrictions_exist
+        validate_constraints_are_permitted
+        validate_restrictions_values_exist
+        @logger.debug(LogMessages::Authentication::AuthnGce::ValidatedResourceRestrictionsConfiguration.new)
+      end
+
+      private
+
+      def validate_required_restrictions_exist
+        if @resource_restrictions.nil? || @resource_restrictions.empty?
+          raise Errors::Authentication::AuthnGce::RoleMissingRequiredConstraints.new(@permitted_constraints)
+        end
+      end
+
+      def validate_constraints_are_permitted
+        @resource_restrictions.each do |r|
+          resource_type = r.type
+          unless @permitted_constraints.include?(resource_type)
+            raise Errors::Authentication::ConstraintNotSupported.new(resource_type, @permitted_constraints)
+          end 
+        end
+      end
+
+      def validate_restrictions_values_exist
+        @resource_restrictions.each do |r|
+          resource_type = r.type
+          resource_value = r.value
+          if resource_value.nil? || resource_value.empty?
+            raise Errors::Authentication::MissingResourceRestrictionsValue.new(resource_type)
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_gce/validate_resource_restrictions_match_jwt.rb
+++ b/app/domain/authentication/authn_gce/validate_resource_restrictions_match_jwt.rb
@@ -1,0 +1,52 @@
+module Authentication
+  module AuthnGce
+
+    # This class is responsible for validating resource restrictions values against JWT token.
+    # the assumption is that resource restrictions list contains only permitted types which already validated
+    # in previous steps
+    ValidateResourceRestrictionsMatchJWT = CommandClass.new(
+      dependencies: {
+        logger: Rails.logger
+      },
+      inputs:       %i(resource_restrictions decoded_token annotation_prefix)
+    ) do
+
+      def call
+        @logger.debug(LogMessages::Authentication::AuthnGce::ValidatingResourceRestrictionsValues.new)
+        validate
+        @logger.debug(LogMessages::Authentication::AuthnGce::ValidatedResourceRestrictionsValues.new)
+      end
+
+      private
+
+      def validate
+        @resource_restrictions.each do |r|
+          resource_type = r.type
+          resource_value = r.value
+          resource_value_from_token = resource_value_from_token(resource_type_without_prefix(resource_type))
+          unless resource_value == resource_value_from_token
+            raise Errors::Authentication::Jwt::InvalidResourceRestrictions.new(resource_type)
+          end
+          @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictionValue.new(resource_type))
+        end
+      end
+
+      def resource_type_without_prefix(resource_type)
+        resource_type.sub(@annotation_prefix, "")
+      end
+
+      def resource_value_from_token(resource_type)
+        case resource_type
+        when PROJECT_ID_RESTRICTION_NAME
+          @decoded_token.project_id
+        when INSTANCE_NAME_RESTRICTION_NAME
+          @decoded_token.instance_name
+        when SERVICE_ACCOUNT_ID_RESTRICTION_NAME
+          @decoded_token.service_account_id
+        when SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME
+          @decoded_token.service_account_email
+        end
+      end
+    end
+  end
+end
--- a/app/domain/authentication/authn_gce/validate_status.rb
+++ b/app/domain/authentication/authn_gce/validate_status.rb
@@ -1,0 +1,24 @@
+module Authentication
+  module AuthnGce
+
+    ValidateStatus = CommandClass.new(
+      dependencies: {
+        discover_identity_provider:  Authentication::OAuth::DiscoverIdentityProvider.new
+      },
+      inputs:       %i()
+    ) do
+
+      def call
+        validate_provider_is_responsive
+      end
+
+      private
+
+      def validate_provider_is_responsive
+        @discover_identity_provider.(
+          provider_uri: PROVIDER_URI
+        )
+      end
+    end
+  end
+end
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -70,6 +70,11 @@ module Errors
       code: "CONJ00017E"
     )
 
+    MissingResourceRestrictionsValue = ::Util::TrackableErrorClass.new(
+      msg:  "Resource restriction '{0-resource-restriction-name}' value is missing or empty",
+      code: "CONJ00070E"
+    )
+
     module AuthenticatorClass
 
       DoesntStartWithAuthn = ::Util::TrackableErrorClass.new(
@@ -336,6 +341,26 @@ module Errors
                 "xms_mirid must contain the resource provider namespace, the " \
                 "resource type, and the resource name",
         code: "CONJ00054E"
+      )
+    end
+
+    module AuthnGce
+
+      InvalidAudience = ::Util::TrackableErrorClass.new(
+        msg:  "'audience' token claim {0} is invalid. The format should be " \
+              "'conjur/<account_name>/<host_id>'",
+        code: "CONJ00067E"
+      )
+
+      TokenClaimNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
+        msg:  "Claim '{0-claim-name}' not found or empty in token. " \
+              "Verify you request the token using 'format=full'",
+        code: "CONJ00068E"
+      )
+
+      RoleMissingRequiredConstraints = ::Util::TrackableErrorClass.new(
+        msg:  "Role should have at least one of the following constraints: {0-constraints}",
+        code: "CONJ00069E"
       )
     end
   end
--- a/app/domain/logs.rb
+++ b/app/domain/logs.rb
@@ -39,6 +39,16 @@ module LogMessages
       code: "CONJ00033D"
     )
 
+    ValidatedResourceRestrictionValue = ::Util::TrackableLogMessageClass.new(
+      msg:  "Resource restriction '{0-resource-restriction-name}' value validated",
+      code: "CONJ00043D"
+    )
+
+    ValidatingResourceRestrictions = ::Util::TrackableLogMessageClass.new(
+      msg:  "Validating resource restriction for role '{0-role-id}'",
+      code: "CONJ00046D"
+    )
+
     ValidatedResourceRestrictions = ::Util::TrackableLogMessageClass.new(
       msg:  "Resource restrictions validated",
       code: "CONJ00030D"
@@ -185,6 +195,40 @@ module LogMessages
       ExtractedResourceRestrictionsFromToken = ::Util::TrackableLogMessageClass.new(
         msg:  "Extracted resource restrictions from token",
         code: "CONJ00029D"
+      )
+
+    end
+
+    module AuthnGce
+
+      ExtractingRestrictionsFromResource = ::Util::TrackableLogMessageClass.new(
+        msg: "Extracting resource restrictions with prefix '{0-prefix}' from role '{1-role-id}'",
+        code: "CONJ00039D"
+      )
+
+      ExtractedResourceRestrictions = ::Util::TrackableLogMessageClass.new(
+        msg: "{0-amount} resource restrictions were extracted",
+        code: "CONJ00040D"
+      )
+
+      ValidatingResourceRestrictionsConfiguration = ::Util::TrackableLogMessageClass.new(
+        msg: "Validating resource restrictions configuration",
+        code: "CONJ00041D"
+      )
+
+      ValidatedResourceRestrictionsConfiguration = ::Util::TrackableLogMessageClass.new(
+        msg: "Resource restrictions configuration is validated",
+        code: "CONJ00042D"
+      )
+
+      ValidatingResourceRestrictionsValues = ::Util::TrackableLogMessageClass.new(
+        msg: "Validating resource restrictions values against Google ID token",
+        code: "CONJ00044D"
+      )
+
+      ValidatedResourceRestrictionsValues = ::Util::TrackableLogMessageClass.new(
+        msg: "Resource restrictions values are validated",
+        code: "CONJ00045D"
       )
 
     end
--- a/ci/test
+++ b/ci/test
@@ -63,6 +63,12 @@ cucumber_authenticators_azure() {
   _run_cucumber_tests authenticators_azure "" "$(_get_azure_env_args)"
 }
 
+cucumber_authenticators_gce() {
+  : DOC - Runs Cucumber GCE Authenticator features
+
+  _run_cucumber_tests authenticators_gce "" "$(_get_gce_env_args)"
+}
+
 cucumber_authenticators_oidc() {
   : DOC - Runs Cucumber OIDC Authenticator features
 
@@ -250,6 +256,41 @@ _get_azure_env_args() {
   azure_env_args="$azure_env_args -e USER_ASSIGNED_IDENTITY_CLIENT_ID=$USER_ASSIGNED_IDENTITY_CLIENT_ID"
   azure_env_args="$azure_env_args -e SYSTEM_ASSIGNED_IDENTITY=$SYSTEM_ASSIGNED_IDENTITY"
   echo "${azure_env_args}"
+}
+
+_get_gce_env_args() {
+  gce_token_file="authn-gce/tokens/gce_token_valid"
+  if ! [ -f "$gce_token_file" ]; then
+    echo "GCE token file $gce_token_file does not exist."
+    exit 1
+  fi
+
+  decoded_gce_token_payload=$(_get_gce_token_payload "$gce_token_file")
+
+  GCE_PROJECT_ID="$(echo "$decoded_gce_token_payload" | jq -r '.google.compute_engine.project_id')"
+  GCE_INSTANCE_NAME="$(echo "$decoded_gce_token_payload" | jq -r '.google.compute_engine.instance_name')"
+  GCE_SERVICE_ACCOUNT_ID="$(echo "$decoded_gce_token_payload" | jq -r '.sub')"
+  GCE_SERVICE_ACCOUNT_EMAIL="$(echo "$decoded_gce_token_payload" | jq -r '.email')"
+
+  local gce_env_args="-e GCE_INSTANCE_NAME=$GCE_INSTANCE_NAME"
+  gce_env_args="$gce_env_args -e GCE_SERVICE_ACCOUNT_ID=$GCE_SERVICE_ACCOUNT_ID"
+  gce_env_args="$gce_env_args -e GCE_SERVICE_ACCOUNT_EMAIL=$GCE_SERVICE_ACCOUNT_EMAIL"
+  gce_env_args="$gce_env_args -e GCE_PROJECT_ID=$GCE_PROJECT_ID"
+
+  echo "${gce_env_args}"
+}
+
+_get_gce_token_payload() {
+  local gce_token_file="$1"
+
+  # read token and set env vars
+  gce_token=$(cat "$gce_token_file")
+
+  source jwt/decode_token.sh
+
+  decoded_gce_token_payload=$(decode_jwt_payload "$gce_token")
+
+  echo "$decoded_gce_token_payload"
 }
 
 _wait_for_keycloak() {
--- a/ci/test
+++ b/ci/test
@@ -237,24 +237,25 @@ _find_cucumber_network() {
 
 _get_oidc_env_args() {
   local oidc_env_args=''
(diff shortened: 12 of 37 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Hosts can authenticate from Google Compute Engines (GCE) using a Google ID token. See
```
InbalZilberman, line 9: @orenbm I dont think the term Google ID token is the right term... maybe GCE instance identity token https://cloud.google.com/compute/docs/instances/verifying-instance-identity#request_signature

app/domain/errors.rb
```diff
@@ -338,6 +343,26 @@ module AuthnAzure
...
+
+      InvalidAudience = ::Util::TrackableErrorClass.new(
+        msg:  "'audience' token claim {0} is invalid. The format should be " \
+              "'conjur/<account_name>/<host_id>'",
+        code: "CONJ00067E"
+      )
+
+      TokenClaimNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
+        msg:  "Claim '{0-claim-name}' not found or empty in token. " \
+              "Verify you request the token using 'format=full'",
```
InbalZilberman, line 357: How about Verify the token is requested using 'format=full'?
We should assume no human is involved.

InbalZilberman (commented):
@shulifink can we please start create from this feature a "error messages" page?

spec/app/domain/authentication/authn-gce/decoded_token_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+        "\"sub\": \"#{sub_value}\", " \
+        "\"email\": \"#{email_value}\", " \
+        "\"google\": { " \
+          "\"compute_engine\": {" \
+            "\"instance_name\": \"#{instance_name_value}\", " \
+            "\"project_id\": \"#{project_id_value}\"" \
+          "}" \
+        "}" \
+      "}"
+    )
```
jonahx, line 30: Fly by comment: For all of these examples, you can avoid the explosion of `"` characters by using either a heredoc like `<<~EOS` or by using `%Q`.

app/domain/authentication/authn_gce/validate_status.rb
```diff
@@ -0,0 +1,24 @@
...
+      private
+
+      def validate_provider_is_responsive
+        @discover_identity_provider.(
+          provider_uri: PROVIDER_URI
+        )
+      end
+    end
+  end
+end
```
jonahx, line 24: Fly by comment: This class seems unnecessary.  Just use the `discovery_identity_provider` directly where needed.

app/domain/authentication/authn_gce/validate_resource_restrictions_match_jwt.rb
```diff
@@ -0,0 +1,54 @@
...
+        when INSTANCE_NAME_RESTRICTION_NAME
+          resource_value_from_token = @decoded_token.instance_name
+        when SERVICE_ACCOUNT_ID_RESTRICTION_NAME
+          resource_value_from_token = @decoded_token.service_account_id
+        when SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME
+          resource_value_from_token = @decoded_token.service_account_email
+        end
+
+        resource_value_from_token
+      end
```
jonahx, line 51: The ruby case statement is an expression, so you should be able to rewrite the above as:
```ruby
def resource_value_from_token(resource_type)
  case resource_type
  when PROJECT_ID_RESTRICTION_NAME
    @decoded_token.project_id
  when INSTANCE_NAME_RESTRICTION_NAME
    @decoded_token.instance_name
  when SERVICE_ACCOUNT_ID_RESTRICTION_NAME
    @decoded_token.service_account_id
  when SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME
    @decoded_token.service_account_email
  end
end
```

Jenkinsfile
```diff
@@ -97,6 +97,59 @@ pipeline {
...
+                sh '''
+                  get_token_into_file() {
+                    local token_format="$1"
+                    local audience="$2"
+                    local filename="$3"
+
+                    curl \
+                      -s \
+                      -H 'Metadata-Flavor: Google' \
+                      "http://metadata/computeMetadata/v1/instance/service-accounts/default/identity?format=${token_format}&audience=${audience}" \
```
nessiLahav, line 119: I think we should change `default` to the name of the service account, so if something will be changed in GCE configuration we will fail here and not later in the tests

CHANGELOG.md
orenbm: thanks, updated.

app/domain/errors.rb
orenbm: well, a human wrote the `curl` to get the token so they should also add `format=full`. 

@shulifink wdyt?

app/domain/authentication/authn_gce/validate_status.rb
orenbm: this file is needed because it handles the specific status validation. It. is called from the [GCE Authenticator class](https://github.com/cyberark/conjur/blob/gcp-feature-branch/app/domain/authentication/authn_gce/authenticator.rb#L47) that is called from the generic [`validate_status`](https://github.com/cyberark/conjur/blob/gcp-feature-branch/app/domain/authentication/validate_status.rb#L70).

one can argue that the `status` method in the GCE Authenticator class will be defined as:
```
      def status(authenticator_status_input:)
        Authentication::OAuth::DiscoverIdentityProvider.new.(
          provider_uri: PROVIDER_URI
        )
      end
```

but i think that it's better to call the `validate_status` class and handle the logic there, for clarity. It is also open for addition in the future this way and it conforms with the other status validation classes.

Jenkinsfile
orenbm: @nessiLahav can you elaborate? what is the difference between running this request with `default` and running it with the service-account name?

how will a change in GCE config will break this?

i am not keen on using the service account name as we don't have it here (as far as i know) and we will need to store it in `conjurops`, which is something that we are trying to avoid.

orenbm: is service account name the same as service-account id?

orenbm: @eranha ^

nessiLahav: An example is for the following service account: `nessi-service-account-test@refreshing-mark-284016.iam.gserviceaccount.com`
`curl --header "Metadata-Flavor: Google" --data-urlencode "audience=conjur/cucumber/host/nessitest/hosttest1" --get --data-urlencode "format=full" "http://metadata/computeMetadata/v1/instance/service-accounts/nessi-service-account-test@refreshing-mark-284016.iam.gserviceaccount.com/identity"`

Since we are relying on this service account id later in the resource restrictions tests, i prefer to found out here that something change, instead of later in the tests.

spec/app/domain/authentication/authn-gce/decoded_token_spec.rb
orenbm: but then won't it need to be 
```
%Q({"aud": "#{sub_value}","sub": "#{sub_value}", "email": "#{email_value}","google": {"compute_engine": {"instance_name": "#{instance_name_value}","project_id": "#{project_id_value}"}}})
``` 

to avoid printing `\n`?

orenbm: same goes for `EOS`. 
this:
```
<<-EOS
{
  "aud": "#{sub_value}",
  "sub": "#{sub_value}",
  "email": "#{email_value}",
  "google": {
    "compute_engine": {
      "instance_name": "#{instance_name_value}",
      "project_id": "#{project_id_value}"
    }
  }
}
EOS
```

outputs this:
```
"{\n  \"aud\": \"sub\",\n  \"sub\": \"sub\",\n  \"email\": \"email\",\n  \"google\": {\n    \"compute_engine\": {\n      \"instance_name\": \"ins\",\n      \"project_id\": \"proj\"\n    }\n  }\n}\n"
```

Jenkinsfile
orenbm: oh you're using the whole id. got it.

anyway - i think we need a good reason to use this and not `default` as the service account email will need to be saved in conjurops. this requires the instance to be conjurized and it will need to be changed in case the SA is replaced.

so unless we have a good reason to use it, i would stay with default.

what do you think?

nessiLahav: OK, this is too much effort, i agree to stay with the default

Jenkinsfile
```diff
@@ -97,6 +97,59 @@ pipeline {
...
+                      > $filename
+                  }
+
+                  get_token_into_file "full" "conjur/cucumber/host/test-app" "gce_token_valid"
```
sigalsax, line 123: `get_token_into-file` naming is a bit weird. `copy_token_into_file`?

config/routes.rb
```diff
@@ -32,6 +32,7 @@ def matches?(request)
 
         # authn-oidc has to be first as it can be ambgiuous with the optional :service_id & :id
         post '/authn-oidc(/:service_id)/:account/authenticate' => 'authenticate#authenticate_oidc'
+        post '/authn-gce/:account/authenticate' => 'authenticate#authenticate_gce'
```
sigalsax, line 35: we decided against a service_id here? even if it's optional?

orenbm: note that we also don't have an id so we need an explicit route for `authn-gce`

Jenkinsfile
orenbm: but this function also gets it from the service

app/domain/authentication/authn_gce/update_authenticator_input.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        @logger.debug(
+          LogMessages::Authentication::Jwt::ExtractedUsernameFromToken.new(
+            conjur_username
+          )
+        )
+        @conjur_username
+      end
+
+      def updated_input
```
sigalsax, line 79: `updated_input` a couple of notes here
1. purposeful past tense?
2. doesn't explain what we are updating and for what. Why are we updating the authenticator object with items from the token? @orenbm Is this like xms_mirid for authn-azure to identify the machine?
3. I would add a sentence or two at the top of CommandClass explaining why this update is necessary

app/domain/authentication/authn_gce/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,66 @@
+module Authentication
+  module AuthnGce
+
+    # This class is responsible of validating resource restrictions which are configured on a Conjur host or user
```
sigalsax, line 4: we decided to support user as well? was this checked manually?

app/domain/authentication/authn_gce/update_authenticator_input.rb
orenbm: 1. it's not past tense. it's a noun as this method returns an updated input.
2. it is more like `authn-oidc`. the reason for this is that in the general `Authenticate` command class we perform validations on the username, but the username is retrieved from the token. So we first decode the token and retrieve the username and update the authenticator input with it. The credentials is also updated to have the decoded token instead of the original so we don't need to decode it again.
3. will do, thanks!

app/domain/authentication/authn_gce/validate_resource_restrictions.rb
sigalsax: This class is responsible for *

orenbm: we won't document this but it is tested in an automated test.

app/domain/authentication/authn_gce/resource_restriction.rb
```diff
@@ -0,0 +1,17 @@
+module Authentication
+  module AuthnGce
+
+    class ResourceRestriction
```
sigalsax, line 4: IMO this class is missing some comments like we have for OIDC and Azure. I think we can benefit with giving examples like we did with Azure
```
    # For example, if `resources` includes the AzureResource:
    #   - type: "subscription-id"
    #   - value: "some-subscription-id"
```

app/domain/authentication/authn_gce/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,86 @@
...
+            @extraction_prefix,
+            @username
+          )
+        )
+        prefixed_resource_annotations.select do |a|
+          annotation_name = a.values[:name]
+          resource_value = annotation_value(annotation_name)
+          next unless resource_value
+          resource_restrictions.push(
+            ResourceRestriction.new(
```
sigalsax, line 43: Reason why this is called ResourceRestriction and not `GCEResource`?

sigalsax: if we want to make it common, then let's take it out of the module AuthnGce, otherwise IMO it is cleaner to label it as gce

app/domain/authentication/authn_gce/resource_restriction.rb
orenbm: just to calrify, you mean the comment should be in `validate_resource_restriction`, correct?

app/domain/authentication/authn_gce/validate_resource_restrictions_configuration.rb
```diff
@@ -0,0 +1,49 @@
...
+    # the allowed configuration is: Role should have at least 1 permitted constraints, and non of illegal constraints
+    ValidateResourceRestrictionsConfiguration = CommandClass.new(
+      dependencies: {
+        logger: Rails.logger
+      },
+      inputs:       %i(resource_restrictions permitted_constraints)
+    ) do
+
+      def call
+        @logger.debug(LogMessages::Authentication::AuthnGce::ValidatingResourceRestrictionsConfiguration.new)
```
sigalsax, line 14: not a huge fan of this in the call since IMO call is a checklist, a quick summary of the code to follow but I can live with it

app/domain/authentication/authn_gce/extract_resource_restrictions.rb
orenbm: it may go outside the module in the future but we didn't want to do it now. i think that `GCEResource` is redundant as it's under the `AuthnGce` module (so it's `AuthnGCE::GCEResource`).

However, i do prefer `AuthnGce::Resource` rather than `AuthnGce::ResourceRestriction` as this describes the resource itself (the restriction is on the host).

@nessiLahav what do you think?

app/domain/authentication/authn_gce/validate_resource_restrictions_configuration.rb
orenbm: i will add another question - do we even want this kind of log? we don't do it in other places. if we don't have something useful to log (e.g "working with provider uri {provider-uri}") then do we want to say that we are starting some procedure?

if the procedure will fail, an error will be raised. If not, why bother with this line?

again, i'd like to get @nessiLahav 's idea because he wrote this.

spec/app/domain/authentication/authn-gce/validate_resource_restrictions_match_jwt_spec.rb
```diff
@@ -0,0 +1,311 @@
...
+  #  ____  _   _  ____    ____  ____  ___  ____  ___
+  # (_  _)( )_( )( ___)  (_  _)( ___)/ __)(_  _)/ __)
+  #   )(   ) _ (  )__)     )(   )__) \__ \  )(  \__ \
+  #  (__) (_) (_)(____)   (__) (____)(___/ (__) (___/
+
+  context "A valid resource restrictions values" do
+    context "when all resource restrictions are valid" do
+      subject do
+        Authentication::AuthnGce::ValidateResourceRestrictionsMatchJWT.new.call(
+          resource_restrictions: resource_restrictions_with_4_valid,
```
sigalsax, line 186: spacing. It is nicer if all keys and values are on the same lines.
Like
```
decoded_token:         mocked_decoded_token_class_return_valid_values,
annotation_prefix:     gce_restriction_prefix
```

sigalsax (commented):
left some comments

app/domain/authentication/authn_gce/validate_resource_restrictions_configuration.rb
```diff
@@ -0,0 +1,49 @@
...
+        validate_required_restrictions_exist
+        validate_constraints_are_permitted
+        validate_restrictions_values_exist
```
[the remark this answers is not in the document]
orenbm: well, you can define a host as follows:
```
- !host
  id: test-app
  annotations:
    authn-gce/project-id:
```

in this case we want to raise a proper error. if we won't perform the check above, the user will get an `InvalidResourceRestrictions- Resource restriction '{0-resource-restriction-name}' does not match resource in JWT token` error. Although it is true, it is also less upfront than saying `Resource restriction '{0-resource-restriction-name}' value is missing or empty`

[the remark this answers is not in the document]
orenbm: the order of the validations doesn't really matter. we perform different validations on the annotations.

app/domain/authentication/authn_gce/validate_status.rb
```diff
@@ -0,0 +1,24 @@
...
+    ) do
+
+      def call
+        validate_provider_is_responsive
+      end
+
+      private
+
+      def validate_provider_is_responsive
+        @discover_identity_provider.(
```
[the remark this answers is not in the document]
orenbm: because it's static and won't change between GCE authenticators so we don't want to bother the user with setting it.

app/domain/authentication/authn_gce/extract_resource_restrictions.rb
nessiLahav: I like `ResourceRestriction` name more, because i think it is make the code more understandable as this resource is an actually a restriction.

But we can challenge it again at the moment we will refactor it to be outside of the module

ci/test
```diff
@@ -252,6 +258,41 @@ _get_azure_env_args() {
...
+
+  decoded_gce_token_payload=$(_get_gce_token_payload "$gce_token_file")
+
+  GCE_PROJECT_ID="$(echo "$decoded_gce_token_payload" | jq -r '.google.compute_engine.project_id')"
+  GCE_INSTANCE_NAME="$(echo "$decoded_gce_token_payload" | jq -r '.google.compute_engine.instance_name')"
+  GCE_SERVICE_ACCOUNT_ID="$(echo "$decoded_gce_token_payload" | jq -r '.sub')"
+  GCE_SERVICE_ACCOUNT_EMAIL="$(echo "$decoded_gce_token_payload" | jq -r '.email')"
+
+  local gce_env_args="-e GCE_INSTANCE_NAME=$GCE_INSTANCE_NAME"
+  gce_env_args="$gce_env_args -e GCE_SERVICE_ACCOUNT_ID=$GCE_SERVICE_ACCOUNT_ID"
```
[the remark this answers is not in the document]
orenbm: i actually prefer not to do it in one-line but any way i have another fix for this from appliance. i'll push it soon.

spec/app/domain/authentication/authn-gce/decoded_token_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+  end
+
+  let(:decoded_token_hash_missing_aud) do
+    decoded_token_hash(
+      "{" \
```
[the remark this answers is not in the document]
orenbm: what do you mean? i don't understand your suggestion

spec/app/domain/authentication/authn-gce/decoded_token_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+    context "that is missing token claims" do
+      context "missing aud claim" do
+        subject(:decoded_token) do
+          ::Authentication::AuthnGce::DecodedToken.new(
+            decoded_token_hash: decoded_token_hash_missing_aud,
+            logger:             Rails.logger
+          )
+        end
+
+        it "raises an error" do
```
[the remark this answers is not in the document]
orenbm: it seems that this issue is spread along the repo so let's do it in a separate effort

spec/app/domain/authentication/authn-gce/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,436 @@
...
+
+  let(:valid_host) {'valid-host'}
+  let(:invalid_host) {'invalid-host'}
+
+  let(:valid_prefix) {'authn-gce/'}
+  let(:empty_prefix) {''}
+  let(:case_sensitive_prefix) {'authn-GCE/'}
+  let(:without_slash_prefix) {'authn-gce'}
+
+  let(:gce_instance_name_annotation_key) {'authn-gce/instance-name'}
```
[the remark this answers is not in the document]
orenbm: it's not the same. the value is the value of the annotation. i don't think there's an issue here.

spec/app/domain/authentication/authn-gce/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,436 @@
...
+  let(:gce_instance_name_annotation_value) {'instance-name'}
+  let(:gce_project_id_annotation_key) {'authn-gce/project-id'}
+  let(:gce_project_id_annotation_value) {'project-id'}
+  let(:gce_service_account_id_annotation_key) {'authn-gce/service-account-id'}
+  let(:gce_service_account_id_annotation_value) {'service account id'}
```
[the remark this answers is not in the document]
orenbm: it's an annotation value, it can have spaces

app/domain/authentication/authn_gce/validate_resource_restrictions_configuration.rb
nessiLahav: I have a different opinion here, i really like that from reading of the logs you can identified the Big and small logic units. for example:
```
Authenticating host  
   Extracting resources
   .
   .
   Resources extracted
   Validating resources
   Resources are validated 
   .
   .
Host is authenticated

```

Also in cases of performance issues it will be more easy to identified where are the issues

spec/app/domain/authentication/authn-gce/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,436 @@
...
+        end
+
+        it "returns expected list" do
+          expect(subject).to eq(expected_resource_restrictions_list_with_duplications_for_valid_prefix)
+        end
+      end
+    end
+  end
+
+  context "An invalid input parameters to extract resource restrictions" do
```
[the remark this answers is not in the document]
orenbm: removed

spec/app/domain/authentication/authn-gce/update_authenticator_input_spec.rb
```diff
@@ -0,0 +1,253 @@
...
+    allow(mocked_verify_and_decode_token).to receive(:call) { |*args|
+      JSON.parse(args[0][:token_jwt]).to_hash
+    }
+  end
+
+  ####################################
+  # request mock
+  ####################################
+
+  def mock_authenticate_gce_token_request(request_body_data:)
```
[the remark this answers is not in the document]
orenbm: where else is it implemented?

app/domain/authentication/authn_gce/extract_resource_restrictions.rb
orenbm: it is not really a restriction but it's not important enough for me to challenge this.

ci/test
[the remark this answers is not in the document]
orenbm: please take a look now

spec/app/domain/authentication/authn-gce/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,436 @@
...
+
+
+  end
+
+  #  ____  _   _  ____    ____  ____  ___  ____  ___
+  # (_  _)( )_( )( ___)  (_  _)( ___)/ __)(_  _)/ __)
+  #   )(   ) _ (  )__)     )(   )__) \__ \  )(  \__ \
+  #  (__) (_) (_)(____)   (__) (____)(___/ (__) (___/
+
+  context "A valid resource restrictions configuration" do
```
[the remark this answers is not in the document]
orenbm: removed the word `valid`

app/domain/authentication/authn_gce/extract_resource_restrictions.rb
orenbm: `ResourceProperty`?

it will be used for `project-id`, `instance-name`, `service-account-id` and `service-account-email` which are all properties of a GCE resource.

But i can really live with `ResourceRestriction`.

Jenkinsfile
sigalsax: `copy_token_from_service`?
Just `get_token_into_file` is grammatically weird to me

app/domain/authentication/authn_gce/extract_resource_restrictions.rb
sigalsax: no problem 💯

app/domain/authentication/authn_gce/resource_restriction.rb
sigalsax: I see we have 3 classes ResourceRestriction in the codebase, for authn-k8s,authn-azure, and for authn-gce. So I think it would be helpful to give examples of type and value like we do in the other authenticators

app/domain/authentication/authn_gce/update_authenticator_input.rb
sigalsax: 1. What about `update_input_from_token`?
2. great! 
3. great! x2

app/domain/authentication/authn_gce/validate_resource_restrictions_configuration.rb
```diff
@@ -0,0 +1,49 @@
...
+        @logger.debug(LogMessages::Authentication::AuthnGce::ValidatedResourceRestrictionsConfiguration.new)
```
sigalsax, line 18: @orenbm were these logs approved by shuli?

app/domain/authentication/authn_gce/validate_status.rb
jonahx: Yeah, I'm arguing that the method is the better solution in this case.  The class isn't being re-used, and I don't think is adding any clarity beyond the method.  It's actually less clear in that it requires reading a lot more boilerplate before you realize all the content is just in `alidate_provider_is_responsive`.

spec/app/domain/authentication/authn-gce/update_authenticator_input_spec.rb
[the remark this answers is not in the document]
sigalsax: https://github.com/cyberark/conjur/pull/1778/files#diff-449804441d94db54fff17ace1db00518R39

sigalsax (approved):
LGTM! Please ensure Shuli has reviewed the logs

spec/app/domain/authentication/authn-gce/decoded_token_spec.rb
jonahx: If you want it without newlines you can just remove them.


```ruby
sub_value = 99
email_value = 'a@b.com'
instance_name_value = 'foo'
project_id_value = 100

x = <<~EOS
{
  "aud": "#{sub_value}",
  "sub": "#{sub_value}",
  "email": "#{email_value}",
  "google": {
    "compute_engine": {
      "instance_name": "#{instance_name_value}",
      "project_id": "#{project_id_value}"
    }
  }
}
EOS

puts x.gsub("\n", '')

```

[Try it online!][TIO-kebejxbs]

[Ruby]: https://www.ruby-lang.org/
[TIO-kebejxbs]: https://tio.run/##bY/BC4IwFMbv768Y67ACETsaBl06d@gajKlLDN1EXRiy/vW1udCKDoP33u/te9/XqvRhTKdSemeV4miP4hh4zcpqHhB2SMNM1gRK0fVMZJwKVvOFX6Uk0LTyxrOelvkMtlEEMNgiSZ7H0xlGQAgzleMdwqtxvqlx4IDt/4PJjUcfxt6wkLKouKVO3PbWZ6N6TrkoSrHMLfky7@X@5PGy04clkd/@TajxtKnBPQ0uIdjbHRrCwmZY44vAASJkA8a8AA "Ruby – Try It Online"

Jenkinsfile
orenbm: if this really bothers you, i can do this instead:
```
retrieve_token() {
  local token_format="$1"
  local audience="$2"

  curl \
    -s \
    -H 'Metadata-Flavor: Google' \
    "http://metadata/computeMetadata/v1/instance/service-accounts/default/identity?format=${token_format}&audience=${audience}"
}

$(get_token_into_file "full" "conjur/cucumber/host/test-app") > "gce_token_valid"
```

app/domain/authentication/authn_gce/update_authenticator_input.rb
orenbm: 1. the method returns an object so its name needs to be a noun. if it was performing an action then the name would be an action.

app/domain/authentication/authn_gce/resource_restriction.rb
orenbm: we don't have a `ResourceRestriction` class for `authn-azure` and `authn-k8s`. in those modules the corresponding class is `<something>Resource`. the comment that you referred to above is from `validate_resource_restrictions`. would you want that comment to be also in `authn-gce/validate_resource_restrictions`?

app/domain/authentication/authn_gce/validate_resource_restrictions_configuration.rb
orenbm: no, still waiting for an approval

app/domain/authentication/authn_gce/validate_status.rb
orenbm: ok, i can live with that.

orenbm: huh but what about specs? it's much easier to write the spec for the command class. or do you think that we don't need to write a spec for the `status` method of this class?
```
class Authenticator
      # This delegates to all the work to the call method created automatically
      # by CommandClass
      #
      # This is needed because we need `valid?` to exist on the Authenticator
      # class, but that class contains only a metaprogramming generated
      # `call(authenticator_input:)` method.  The methods we define in the
      # block passed to `CommandClass` exist only on the private internal
      # `Call` objects created each time `call` is run.
      def valid?(input)
        call(authenticator_input: input)
      end

      def status(authenticator_status_input:)
        Authentication::OAuth::DiscoverIdentityProvider.new.call(
          provider_uri: PROVIDER_URI
        )
      end
    end
```

spec/app/domain/authentication/authn-gce/decoded_token_spec.rb
orenbm: thanks!

btw, i'm a [`replit`](https://repl.it/repls/ElaborateSteelblueLight) fan :)

app/domain/authentication/authn_gce/validate_status.rb
jonahx: Presumably `Authentication::OAuth::DiscoverIdentityProvider` already has tests, so you should be good to go there.

As for the class where you're using it inside the proposed `validate_provider_is_responsive` method, you'd still be injecting `Authentication::OAuth::DiscoverIdentityProvider` as a dependency, so you'll be able to control that in your unit tests, which should be all you need.

Please lmk if I'm missing anything.  I just made a quick review so far, not a thorough one.  Will try to review in depth later if I have time.

nessiLahav (approved):

spec/app/domain/authentication/authn-gce/decoded_token_spec.rb
orenbm: works perfectly, thanks

app/domain/authentication/authn_gce/validate_status.rb
orenbm: i don't follow. are you saying that the code above can use a dependency injection?

jonahx: Currently you inject `discover_identity_provider:  Authentication::OAuth::DiscoverIdentityProvider.new` into `ValidateStatus`.  I'm proposing remove `ValidateStatus` and instead inject `discover_identity_provider:  Authentication::OAuth::DiscoverIdentityProvider.new` into whatever class is currently being injected with `ValidateStatus`.

orenbm: no class is being injected with it. it is used as is in the authenticator class. same as with [`authn-azure`](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_azure/authenticator.rb#L91) and [`authn-oidc`](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_oidc/authenticator.rb#L21)

orenbm: i still think that we should leave the `ValidateStatus` command class for uniformity with the other authenticators. this will be the case for each authenticator that will implement the status validation and i think the logic should go in the command class, however small it is.

app/domain/errors.rb
jvanderhoof: Are we combining empty and not found for security reasons, or is the result of an unfound claim an empty token?

Jenkinsfile
```diff
@@ -97,6 +97,57 @@ pipeline {
...
+              while (!env.GCE_TOKENS_FETCHED?.trim()) {
+                sleep(time: 15, unit: "SECONDS")
+              }
+
+              dir('ci/authn-gce/tokens') {
+                unstash 'authnGceTokens'
+              }
+
+              sh 'ci/test cucumber_authenticators_gce'
+            }
```
ismarc, line 148: There are a couple of constructs in Jenkins that make this a lot more straightforward to do.  `waitUntil` will continuously run with a backoff until it returns true, and can be wrapped in a timeout so that it will eventually exit even if there are issues with other nodes.  The suggestion below demonstrates the full usage:
```suggestion
            timeout(time: 10, unit: 'MINUTES') {
              waitUntil {
                script {
                  return (env.GCE_TOKENS_FETCHED == "true")
                }
              }
            }
            script {
              dir('ci/authn-gce/tokens') {
                unstash 'authnGceTokens'
              }

              sh 'ci/test cucumber_authenticators_gce'
            }
```

app/domain/authentication/authn_gce/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,66 @@
...
+        )
+        extract_resource_restrictions
+        validate_resource_restrictions_configuration
+        validate_resource_restrictions_match_jwt
+        @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictions.new)
+      end
+
+      private
+
+      def extract_resource_restrictions
```
jvanderhoof, line 30: Why not just call `resource_restrictions` directly? Is there another reason for this proxy method?

app/domain/authentication/authn_gce/validate_status.rb
jonahx: I will try to look at this more closely later and actually pull down your branch to explore.

spec/app/domain/authentication/authn-gce/decoded_token_spec.rb
jonahx: Yeah, I like TIO for sharing snippets because all the code is encoded in the URL itself.  But mainly it's used for code golf because it supports every esoteric language under the sun.  I don't use to actually code ruby with :).

app/domain/errors.rb
orenbm: it's not a security concern. it's just that in both cases the result will be the same.

app/domain/authentication/authn_gce/validate_resource_restrictions.rb
orenbm: yes. we like the `call` method to tell the story of the class. so in this case, the `call` method will be:
```
        extract_resource_restrictions
        validate_resource_restrictions_configuration
        validate_resource_restrictions_match_jwt
```

and it is clear what is happening in the class's logic. so the `extract_resource_restrictions` is there to tell the story and the `resource_restrictions` method is there for memoization.

Jenkinsfile
sigalsax: yes please

app/domain/authentication/authn_gce/resource_restriction.rb
sigalsax: Authn-Azure:
```
module Authentication
  module AuthnAzure
     Class ResourceRestrictions
```
I thought was just like we wrote it in ResourceRestrictions, given an exaimple for what `type` and `value` would be

ci/test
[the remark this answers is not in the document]
sigalsax: what was changed?
I still see
```
  local gce_env_args="-e GCE_INSTANCE_NAME=$GCE_INSTANCE_NAME"
  gce_env_args="$gce_env_args -e GCE_SERVICE_ACCOUNT_ID=$GCE_SERVICE_ACCOUNT_ID"
  gce_env_args="$gce_env_args -e GCE_SERVICE_ACCOUNT_EMAIL=$GCE_SERVICE_ACCOUNT_EMAIL"
  gce_env_args="$gce_env_args -e GCE_PROJECT_ID=$GCE_PROJECT_ID"

  env_args="$env_args $gce_env_args"
```

spec/app/domain/authentication/authn-gce/decoded_token_spec.rb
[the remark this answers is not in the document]
sigalsax: fixed with Jonah's suggestion. Looks more better

app/domain/authentication/authn_gce/resource_restriction.rb
orenbm: yes, but we don't really have a corresponding class in gce. we have a planned effort to try and generalize this so we can handle it there

ci/test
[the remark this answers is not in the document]
orenbm: correct, changed it now to 
```
local gce_env_args="-e GCE_INSTANCE_NAME=$GCE_INSTANCE_NAME \
                      -e GCE_SERVICE_ACCOUNT_ID=$GCE_SERVICE_ACCOUNT_ID \
                      -e GCE_SERVICE_ACCOUNT_EMAIL=$GCE_SERVICE_ACCOUNT_EMAIL \
                      -e GCE_PROJECT_ID=$GCE_PROJECT_ID"
```

sigalsax (approved):
Approved but please wait for Shuli log approval before merging

sigalsax (approved):
LGTM please wait for Shuli approval on logs before merging

app/domain/errors.rb
```diff
@@ -70,6 +70,11 @@ module Authentication
       code: "CONJ00017E"
     )
 
+    MissingResourceRestrictionsValue = ::Util::TrackableErrorClass.new(
+      msg:  "Resource restriction '{0-resource-restriction-name}' value is missing or empty",
+      code: "CONJ00070E"
+    )
+
```
shulifink, line 77: fine

app/domain/errors.rb
```diff
@@ -338,6 +343,26 @@ module AuthnAzure
         code: "CONJ00054E"
       )
     end
+
+    module AuthnGce
+
+      InvalidAudience = ::Util::TrackableErrorClass.new(
+        msg:  "'audience' token claim {0} is invalid. The format should be " \
```
shulifink, line 350: can you use hyphens instead of underscores? account-name. and host-id

app/domain/errors.rb
shulifink, line 357: Make sure that you requested the token using ...

or 
Make sure to request the token using...

depending on if it's already happened, or something that you are "instructing them to do next"

app/domain/errors.rb
```diff
@@ -338,6 +343,26 @@ module AuthnAzure
...
+
+      TokenClaimNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
+        msg:  "Claim '{0-claim-name}' not found or empty in token. " \
+              "Verify you request the token using 'format=full'",
+        code: "CONJ00068E"
+      )
+
+      RoleMissingRequiredConstraints = ::Util::TrackableErrorClass.new(
+        msg:  "Role should have at least one of the following constraints: {0-constraints}",
+        code: "CONJ00069E"
```
shulifink, line 363: "should" or "must"

app/domain/logs.rb
```diff
@@ -39,6 +39,16 @@ module Authentication
       code: "CONJ00033D"
     )
 
+    ValidatedResourceRestrictionValue = ::Util::TrackableLogMessageClass.new(
+      msg:  "Resource restriction '{0-resource-restriction-name}' value validated",
+      code: "CONJ00043D"
```
shulifink, line 44: fine

app/domain/logs.rb
```diff
@@ -39,6 +39,16 @@ module Authentication
...
+    )
+
+    ValidatingResourceRestrictions = ::Util::TrackableLogMessageClass.new(
+      msg:  "Validating resource restriction for role '{0-role-id}'",
+      code: "CONJ00046D"
```
shulifink, line 49: fine

app/domain/logs.rb
```diff
@@ -39,6 +39,16 @@ module Authentication
...
+    )
+
     ValidatedResourceRestrictions = ::Util::TrackableLogMessageClass.new(
       msg:  "Resource restrictions validated",
```
shulifink, line 53: fine

app/domain/logs.rb
```diff
@@ -188,6 +198,40 @@ module AuthnAzure
       )
 
     end
+
+    module AuthnGce
+
+      ExtractingRestrictionsFromResource = ::Util::TrackableLogMessageClass.new(
+        msg: "Extracting resource restrictions with prefix '{0-prefix}' from role '{1-role-id}'",
```
shulifink, line 205: fine

app/domain/logs.rb
```diff
@@ -188,6 +198,40 @@ module AuthnAzure
...
+        code: "CONJ00039D"
+      )
+
+      ExtractedResourceRestrictions = ::Util::TrackableLogMessageClass.new(
+        msg: "{0-amount} resource restrictions were extracted",
```
shulifink, line 210: fine

app/domain/logs.rb
```diff
@@ -188,6 +198,40 @@ module AuthnAzure
...
+        code: "CONJ00040D"
+      )
+
+      ValidatingResourceRestrictionsConfiguration = ::Util::TrackableLogMessageClass.new(
+        msg: "Validating resource restrictions configuration",
```
shulifink, line 215: fine

app/domain/logs.rb
```diff
@@ -188,6 +198,40 @@ module AuthnAzure
...
+        code: "CONJ00041D"
+      )
+
+      ValidatedResourceRestrictionsConfiguration = ::Util::TrackableLogMessageClass.new(
+        msg: "Resource restrictions configuration is validated",
```
shulifink, line 220: without the "is"
"Resource restrictions configuration validated",

app/domain/logs.rb
```diff
@@ -188,6 +198,40 @@ module AuthnAzure
...
+        code: "CONJ00042D"
+      )
+
+      ValidatingResourceRestrictionsValues = ::Util::TrackableLogMessageClass.new(
+        msg: "Validating resource restrictions values against Google ID token",
```
shulifink, line 225: Validating resource restriction values against Google ID token

( no "s" on "restriction")

app/domain/logs.rb
```diff
@@ -188,6 +198,40 @@ module AuthnAzure
...
+        code: "CONJ00044D"
+      )
+
+      ValidatedResourceRestrictionsValues = ::Util::TrackableLogMessageClass.new(
+        msg: "Resource restrictions values are validated",
+        code: "CONJ00045D"
```
shulifink, line 231: "Resource restriction values validated"

no "s" on "restriction", and removed "are"

app/domain/errors.rb
```diff
@@ -70,6 +70,11 @@ module Authentication
       code: "CONJ00017E"
     )
 
+    MissingResourceRestrictionsValue = ::Util::TrackableErrorClass.new(
```
shulifink, line 73: Line 19: In "    InsufficientPasswordComplexity" .... ", and 1 special character"

line 32 : In " AuthenticatorNotFound", is this Conjur only? What about DAP? Also, I would say "... not configured/installed/enabled" but not "not implemented..."

Lines 58-59: "Resource type '{0}' is not a supported resource restriction. " \
            "The supported resources are '{1}'",	            "The supported resources are '{1}'",
Do you mean "The supported resource restrictions are....", or "The supported restrictions are..."

Line 47: "Status check not implemented for authenticator"  What does "not implemented" mean? How about "Status of [authn-name] authenticator not verified."

Line 53 - what does this mean, in plain english? ""Resource restrictions include an illegal resource constraint combination - '{0-constraints}'","

app/domain/errors.rb
```diff
@@ -70,6 +70,11 @@ module Authentication
...
+      msg:  "Resource restriction '{0-resource-restriction-name}' value is missing or empty",
```
shulifink, line 74: fine

app/domain/errors.rb
```diff
@@ -338,6 +343,26 @@ module AuthnAzure
         code: "CONJ00054E"
       )
     end
+
+    module AuthnGce
+
+      InvalidAudience = ::Util::TrackableErrorClass.new(
+        msg:  "'audience' token claim {0} is invalid. The format should be " \
+              "'conjur/<account_name>/<host_id>'",
```
shulifink, line 351: fine (can we use hypens instead of underscores: account-name, host-id?)

app/domain/errors.rb
shulifink: Is this an instruction that needs to be carried out? Then  "Make sure to request the token using 'format=full'",

If something failed because it wasn;t done correctly, and now you are giving them a hint of why it fails:
"Make sure you requested the token using 'format=full'","

app/domain/errors.rb
```diff
@@ -338,6 +343,26 @@ module AuthnAzure
...
+      )
+
+      TokenClaimNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
+        msg:  "Claim '{0-claim-name}' not found or empty in token. " \
+              "Verify you request the token using 'format=full'",
+        code: "CONJ00068E"
+      )
+
+      RoleMissingRequiredConstraints = ::Util::TrackableErrorClass.new(
+        msg:  "Role should have at least one of the following constraints: {0-constraints}",
```
shulifink, line 362: fine

app/domain/logs.rb
```diff
@@ -39,6 +39,16 @@ module Authentication
       code: "CONJ00033D"
     )
 
+    ValidatedResourceRestrictionValue = ::Util::TrackableLogMessageClass.new(
+      msg:  "Resource restriction '{0-resource-restriction-name}' value validated",
```
shulifink, line 43: fine

app/domain/logs.rb
```diff
@@ -39,6 +39,16 @@ module Authentication
...
+      code: "CONJ00043D"
+    )
+
+    ValidatingResourceRestrictions = ::Util::TrackableLogMessageClass.new(
+      msg:  "Validating resource restriction for role '{0-role-id}'",
```
shulifink, line 48: fine

app/domain/logs.rb
shulifink, line 53: fine

app/domain/logs.rb
shulifink, line 205: fine

app/domain/logs.rb
shulifink, line 210: fine

shulifink (commented):
see smy comments

app/domain/errors.rb
orenbm: > line 32 : In " AuthenticatorNotFound", is this Conjur only? What about DAP? Also, I would say "... not configured/installed/enabled" but not "not implemented..."

this is for both conjur and dap but i think that it's ok to say `Conjur` as it is a component inside DAP. Also - it says "not implemented" because this is literally the case - the authenticator is not implemented in the code (e.g `authn-shuli`)

> Lines 58-59: "Resource type '{0}' is not a supported resource restriction. "
"The supported resources are '{1}'", "The supported resources are '{1}'",
Do you mean "The supported resource restrictions are....", or "The supported restrictions are..."

i mean "the supported resource". For example the resource "service-account".

> Line 47: "Status check not implemented for authenticator" What does "not implemented" mean? How about "Status of [authn-name] authenticator not verified."

It is literally not implemented in the code. not every authenticator has a status check.

> Line 53 - what does this mean, in plain english? ""Resource restrictions include an illegal resource constraint combination - '{0-constraints}'","

it means that, for example, a user set a host that has resource restrictions on both a system-assigned-identity and on a user-assigned-identity. this configuration doesn't make any sense and will never pass as no Azure AD token will have these characteristics. So we want to raise a configuration error.

app/domain/errors.rb
orenbm: it's the second one, thanks

orenbm: btw, i remember @yserota telling me to prefer `verify that` over `make sure`. did this change?

app/domain/errors.rb
orenbm: must. changed.

app/domain/logs.rb
orenbm: thanks

app/domain/errors.rb
shulifink: not implemnted (line 47: Then how about "There is not status check for ..." or Status check not supported for ...."

orenbm: i like it!

should it be `Status check not supported for authenticator` or `Status check is not supported for authenticator`?

shulifink: line 53 How about: Resource restrictions include an illegal combination of resource constraints - '{0-constraints}'"
It's easier to read (if it means the same thing...)

orenbm: good, thanks

PR #1780, opened 2020-08-26
Design proposal for Conjur variables content validation and enforcement

rafis3:
### What does this PR do?
**This PR does not make any code changes in the product. It only introduces a design proposal, to be added under the `design/proposals` directory.**
This proposal suggests an enhancement to how we store secrets in Conjur, while offering a way to validate the given structure and provide the user with useful error messages if values are missing or invalid.

Please share your thoughts about the idea, if you think it has value, if you like the proposed solution and user experience.
The outcome of the review should be either accepting the design (with any modifications based on the given feedback) or discarding it, if this proposal is not desired.

### DIFF
--- a/design/proposals/variable_schema_validation.md
+++ b/design/proposals/variable_schema_validation.md
@@ -1,0 +1,346 @@
+# Variable Schema Validation Proposal
+
+- [Variable Schema Validation Proposal](#variable-schema-validation-proposal)
+  - [Introduction](#introduction)
+  - [Current Limitations](#current-limitations)
+  - [Proposed Solution](#proposed-solution)
+    - [User Experience](#user-experience)
+    - [Functionality](#functionality)
+      - [JSON Schema Management](#json-schema-management)
+        - [Creating and Updating JSON Schemas](#creating-and-updating-json-schemas)
+        - [Deleting JSON Schemas](#deleting-json-schemas)
+      - [Validating Variable Values Using a JSON Schema](#validating-variable-values-using-a-json-schema)
+        - [Creating and Updating Variables](#creating-and-updating-variables)
+        - [Deleting Variables](#deleting-variables)
+      - [Secrets Encryption](#secrets-encryption)
+      - [Built-in Schemas](#built-in-schemas)
+      - [Automatic Permission Granting to Built-in Schemas](#automatic-permission-granting-to-built-in-schemas)
+    - [Out of Scope - Future Development](#out-of-scope---future-development)
+      - [Align Conjur Clients to a Single Variable Usage](#align-conjur-clients-to-a-single-variable-usage)
+      - [Align Conjur Rotators to a Single Variable Usage](#align-conjur-rotators-to-a-single-variable-usage)
+      - [Granular Updates of Variable Values](#granular-updates-of-variable-values)
+  - [Main Advantages and Expected Value From This Proposal](#main-advantages-and-expected-value-from-this-proposal)
+  - [Affected Areas](#affected-areas)
+  - [Backwards Compatibility](#backwards-compatibility)
+  - [Performance](#performance)
+  - [Security](#security)
+  - [Documentation](#documentation)
+  - [Version Update](#version-update)
+  - [Delivery Plan](#delivery-plan)
+    - [Minimal Functionality](#minimal-functionality)
+    - [Extended Functionality](#extended-functionality)
+  
+## Introduction
+
+Conjur variables are resources that allow a user or an application to define and store sensitive information.
+
+In many cases, such sensitive information is comprised of multiple pieces that are tightly coupled and together represent a complete object. A few examples:
+
+- In order to access a MySQL database, a user needs a username, a password and an address, URL or connection string.
+- In order to access a REST API exposed by a certain service, a user might need either a username and password, a private key and a public certificate, or a token. In addition, the user will need the server public certificate and its address.
+
+Currently, this is handled by defining multiple variables in Conjur. Each variable contains a single value, such as a username, a password, etc. To bundle these pieces together, we use a naming convention. For example:
+
+```yaml
+- !policy
+  id: my-policy
+  body:
+  - &vars
+    - !variable mysql-db-creds/password
+    - !variable mysql-db-creds/username
+    - !variable mysql-db-creds/address
+    - !variable mysql-db-creds/port
+
+  - !host my-app
+  
+  - !permit
+    role: !host my-app
+    privileges: [ read,execute ]
+    resource: *vars
+```
+
+## Current Limitations
+
+- Multiple variables cannot be updated together, as a single transaction. This might lead to a momentary inconsitency when for example, a username is updated, the password should probabaly be updated as well.
+- There is no enforcement on the content of a variable, which can lead to unintended invalid content.
+- There is no enforcement on the group of variables needed for a certain use case. For example, a database connection requires a username, password, connection URL or address and port. A user could accidentally forget to specify a certain needed variable.
+
+## Proposed Solution
+
+The proposed solution is to add the ability to use JSON schemas and link them to variables, for which we want validation of their content. This will allow a single variable to store multiple, related values. While it's already possible today, to store JSON structured information in a variable, this proposed feature adds two added values:
+
+- The JSON structure is validated and enforced. If the input is invalid, the user will get a proper error message and the variable update will be denied.
+- The JSON structure will not be encrypted completely, but only the sensitive information. This will allow future seach capabilities based on the non sensitive information.
+
+### User Experience
+
+Let's have a look at the following example:
+
+```yaml
+- !policy
+  id: my-policy
+  body:
+  - !variable
+    id: mysql-db-creds
+    annotations:
+      schema-reference: !variable /conjur/schemas/mysql-schema
+
+  - !host my-app
+  
+  - !permit
+    role: !host my-app
+    privileges: [ read,execute ]
+    resource: !variable mysql-db-creds
+```
+
+And
+
+```yaml
+- !policy
+  id: conjur/schemas
+  body:
+  - &schemas
+    - !variable
+      id: mysql-schema
+      annotations:
+        schema: true
+
+  - !permit
+    role: !group /conjur/all
+    privileges: [ read,execute ]
+    resource: *schemas
+```
+
+In the example above, the `/conjur/schemas/mysql-schema` variable will be updated with the following value:
+
+```json
+{
+  "$id": "https://cyberark.com/mysql.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "MySQL Creds",
+  "type": "object",
+  "properties": {
+    "username": {
+      "type": "string",
+      "description": "The MySQL username"
+    },
+    "password": {
+      "type": "string",
+      "description": "The MySQL password",
+      "minLength": 12,
+      "maxLength": 40,
+      "pattern": "[A-Za-z0-9]"
+    },
+    "address": {
+      "type": "string",
+      "description": "The MySQL address"
+    },
+    "port": {
+      "description": "The MySQL port number",
+      "type": "integer",
+      "minimum": 1,
+      "maximum": 65535
+    }
+  },
+  "required": ["username", "password"],
+  "additionalProperties": false,
+  "non-secrets": [
+      "username", "port", "address"
+    ]
+}
+```
+
+And the `mysql-db-creds` variable could be updated with the following value:
+
+```json
+{
+  "username": "my-user",
+  "password": "my-password-12345",
+  "address": "my-db.com",
+  "port": 3306
+}
+```
+
+### Functionality
+
+#### JSON Schema Management
+
+JSON schemas are variables that have the following annotation:
+
+```yaml
+schema: true
+```
+
+##### Creating and Updating JSON Schemas
+
+When creating or updating a JSON schema, the user will create a variable and annotated it as a JSON schema (schema: true). Then Conjur will validate the variable content, to ensure it's a valid JSON schema. A proper error message will be given if the input is not a valid JSON schema.
+
+##### Deleting JSON Schemas
+
+When deleting a JSON schema variable, if there are existing variables with references to that deleted JSON schema, the references will be automatically removed. Therefore these existing variables will be left as general variables with no schema validation.
+
+#### Validating Variable Values Using a JSON Schema
+
+Variables that their content is validated by a JSON schema have the following annotation:
+
+```yaml
+schema-reference: !variable full-path-of-schema-variable
+```
+
+##### Creating and Updating Variables
+
+If a variable was annotated with a schema reference (schema-reference: schema-variable-name), then Conjur will validate the variable content, using the JSON schema that is specified in the `schema` attribure of that variable.
+
+For the example above, Conjur will use the JSON schema specified in `conjur/schemas/mysql-schema` to verify the content of `mysql-db-creds`. This verification includes:
+
+- `username` and `password` are mandatory properties.
+- `address` and `port` are allowed properties.
+- `port` must contain a valid port number.
+- `password` must align to a sufficient complexity.
+
+If the variable content is invalid, the user will get an HTTP code of 422 (Unprocessable Entity) and the body of the response will also contain an elaborative error message that explains what part of the input was found to be invalid.
+
+For the example above, if the user would provide a string as the port value, the following error message will be returned:
+
+```text
+Message: Invalid type. Expected Integer but got String.
+Schema path: https://cyberark.com/mysql.schema.json#/properties/port/type
+```
+
+At first stage, a user that wants to update a specific field within the JSON structure, will read the variable, change its value appropriately and call Conjur to update the variable. In a later stage, we will add support for updating specific fields directly through the variable API.
+
+##### Deleting Variables
+
+No changes in deletion of variables that are referenced to JSON schemas. The behavior remains the same as regular variables.
+
+#### Secrets Encryption
+
+The JSON schema can contain a property called `non-secrets` which expects an array of JSON schema property names. The properties specified in this array will not be encrypted in the database, along with the JSON structure. The unspecified attributes, will by default get encrypted. This allows the protection of sensitive data while also allowing the non-sensitive values to be searchable. Meaning that the variable will not be encrypted as a whole, but only the sensitive attributes in the JSON structure.
+
+For the example above, the `mysql-db-creds` variable will be saved in the database as follows:
+
+```json
+{
+  "username": "my-user",
+  "password": "U2FsdGVkX19Ji6JpgnVHW3V47OtMJwuKi1Yf9nc0aP5QcuzdnIrpzZ2zMC90f24g",
+  "address": "my-db.com",
+  "port": 3306
+}
+```
+
+Before the variable value is returned to the client, the secret attributes are decrypted so that the given example above, would look like this:
+
+```json
+{
+  "username": "my-user",
+  "password": "my-password",
+  "address": "my-db.com",
+  "port": 3306
+}
+```
+
+#### Built-in Schemas
+
+To simplify the user experience, Conjur can come with prdefined JSON schemas for the most common secrets use cases. The list of predefined schmeas should include:
+
+- Database secrets - Oracle, MSSQL, MySQL, Postgres, MariaDB, DB2.
+- Cloud access keys - AWS, Azure, GCP.
+- X509 certificates.
+- JWTs
+
+Write about the advantages of atomicity, minimal changes in the database to implement this and so on.
+
+#### Automatic Permission Granting to Built-in Schemas
+
+In order to provide access for any role to the built-in schemas, a new built-in group should be introduced: `conjur/all`. All roles in the Conjur account should be added to this group automatically, thus keeping this group always up to date with all the roles that exist in the account.
+
+Usage example:
+
+```yaml
+- !permit
+  role: !group /conjur/all
+  privileges: [ read,execute ]
+  resource: !variable my-var
+```
+
+### Out of Scope - Future Development
+
+#### Align Conjur Clients to a Single Variable Usage
+
+Modify our examples and integrations, such as the Conjur Summon provider, to leverage this new single variable JSON structure, instead of multiple variables.
+
+#### Align Conjur Rotators to a Single Variable Usage
+
+Modify the existing Conjur rotators, to update a JSON structured variable.
+
+#### Granular Updates of Variable Values
+
+As a first step, updating a variable value means that the entire JSON structure should be written at once. In the future, for a granular update that would allow the user to update a specific field within the JSON structure. For example, if a variable contains a JSON with two fields: `username` and `password`, the variable API will be able to update one of the fields, without replacing the whole JSON.
+
+## Main Advantages and Expected Value From This Proposal
+
+- We leverage a standard way to enforce the content of the variable, with a well known JSON schema.
+- Tightly coupled values, such as username and password, are updated together in a single transaction. This will prevent momentary inconsistency in which each value was updated independently, one after the other.
+- The user will not have to look in the documentation or to understand the required variables, their structure or allowed input. Conjur will provide this feedback when an update attempt is made.
+- The feature does not require changes in the database schma, only in new built-in content.
+- The returned variable is in a JSON structure, same as the rest of our APIs responses.
+
+## Affected Areas
+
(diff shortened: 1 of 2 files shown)


design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+
+```yaml
+- !policy
+  id: conjur/schemas
+  body:
+  - &schemas
+    - !variable mysql-schema
+  
+  - !permit
+    role: !group /conjur/all
```
orenbm, line 88: is this a thing? there's a way to grant permissions for all conjur users?

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+  "address": "my-db.com",
+  "port": 3306
+}
+```
+
+### Functionality
+
+#### Input Validation
+
+Conjur will validate the variable content, using the JSON schema that is specified in the `schema` attribure of that variable.
```
orenbm, line 147: this will happen when we set a value to the variable, correct?

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+  - &schemas
+    - !variable mysql-schema
+  
+  - !permit
+    role: !group /conjur/all
+    privileges: [ read,execute ]
+    resource: !variable *schemas
+```
+
+In the example above, the `/conjur/schemas/mysql-schema` variable will be updated with the following value:
```
orenbm, line 93: this will be set by the user? how will they know the structure of this schema? is this something common?

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+Conjur will validate the variable content, using the JSON schema that is specified in the `schema` attribure of that variable.
+
+For the example above, Conjur will use the JSON schema specified in `conjur/schemas/mysql-schema` to verify the content of `mysql-db-creds`. This verification includes:
+
+- `username` and `password` are mandatory properties.
+- `address` and `port` are allowed properties.
+- `port` must contain a valid port number.
+- `password` must align to a sufficient complexity.
+
+If the varialbe content is invalid, the user will get an HTTP code of 422 (Unprocessable Entity) and the body of the response will also contain an elaborative error message that explains what part of the input was found to be invalid.
```
orenbm, line 156: is this schema considered sensitive? is it ok to return the error in the response body?

orenbm: also `varialbe` -> `variable`

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+
+#### Secrets Encryption
+
+The JSON schema can contain a property called `secrets` which expects an array of JSON schema property names. The properties specified in this array, will be encrypted in the database. All the rest of the properties and the JSON structure will remain in cleartext. This allows the protection of sensitive data while also allowing the non-sensitive values to be searchable.
```
orenbm, line 160: what happens if the `secrets` property is not set in the schema? will all the fields be encrypted or none of them?

jvanderhoof:
I don't feel like this document captures the problem we're hoping to solve, particularly considering it'll take about two months to complete. Can we dig into the problem a bit more? A crisp problem definition will help validate:

1. The problem is real for our customers
2. Solving the problem will bring our customer real value

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+    resource: !variable mysql-db-creds
+```
+
+And
+
+```yaml
+- !policy
+  id: conjur/schemas
+  body:
+  - &schemas
```
orenbm, line 84: this doesn't require any implementation, correct? this alias can be any word

design/proposals/variable_schema_validation.md
rafis3: No, it's part of the proposal

design/proposals/variable_schema_validation.md
rafis3: Yes

design/proposals/variable_schema_validation.md
rafis3: JSON schema is a known standard, but we'll document and provide some examples.
In addition, we'll give some builtin schemas for commonly useful structure types like common databases, cloud access keys and so on.

design/proposals/variable_schema_validation.md
rafis3: No, on the contrary, you want to know what the schema is, so you can know how your variable needs to be structured. It's like loading policy. Same as our policy syntax is documented and public.

design/proposals/variable_schema_validation.md
rafis3: That's a good point! My initial thought was that the default is not, but that's not secure. I think the default should be that all attributes are encrypted, unless the user will define an attribute as non-secret. I'll update the proposal.

orenbm: yes, that is what i was thinking too. thanks.

design/proposals/variable_schema_validation.md
rafis3: Correct

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+For the example above, Conjur will use the JSON schema specified in `conjur/schemas/mysql-schema` to verify the content of `mysql-db-creds`. This verification includes:
+
+- `username` and `password` are mandatory properties.
+- `address` and `port` are allowed properties.
+- `port` must contain a valid port number.
+- `password` must align to a sufficient complexity.
+
+If the varialbe content is invalid, the user will get an HTTP code of 422 (Unprocessable Entity) and the body of the response will also contain an elaborative error message that explains what part of the input was found to be invalid.
+
+#### Secrets Encryption
```
sgnn7, line 158: If I understand this section correctly, my perspective here is that per-field/attribute encryption of credentials here is unnecessary if the whole variable value is already encrypted in the database at-rest. I don't think we gain much by doing a per-field encryption and would increase code complexity by quite a bit for it.

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+  },
+  "required": ["username", "password"],
+  "additionalProperties": false,
+  "secrets": [
+      "password",
+    ]
+}
+```
+
+And the `mysql-db-creds` variable could be updated with the following value:
```
sigalsax, line 132: Something that I thought about here, is how easy is it for a user to change their schema? What will happen if they change the schema after variables are loaded. Will they get a list of current variables that do not abide by the newly generated/changed schema?

Can a user have multiple schema's attached to their variables? Something like
```
- !policy
  id: my-policy
  body:
  - !variable
    id: mysql-db-creds
    schema: !variable /conjur/schemas/mysql-schema
    schema:  !variable /conjur/schemas/another-mysql-schema
```

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+#### Input Validation
+
+Conjur will validate the variable content, using the JSON schema that is specified in the `schema` attribure of that variable.
+
+For the example above, Conjur will use the JSON schema specified in `conjur/schemas/mysql-schema` to verify the content of `mysql-db-creds`. This verification includes:
+
+- `username` and `password` are mandatory properties.
+- `address` and `port` are allowed properties.
+- `port` must contain a valid port number.
+- `password` must align to a sufficient complexity.
```
sigalsax, line 154: What about allowed characters?

design/proposals/variable_schema_validation.md
sigalsax, line 156: Can you provide an example of an error message that you anticipate to add?

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,264 @@
...
+
+#### Additional Required Functionality
+
+In order to provide access for any role to the builtin schemas, a new builtin group should be introduced: `conjur/all`. All roles in the Conjur account should be added to this group automatically, thus keeping this group always up to date with all the roles that exist in the account.
+
+Usage example:
+
+```yaml
+- !permit
+  role: !group /conjur/all
```
sigalsax, line 203: I really like this idea! Is this part of the core design? As in can we go forth with the design without implementing this or is it an enhancement?

design/proposals/variable_schema_validation.md
rafis3: You raise a good point, that I don't think I explained well enough. The proposal suggests that the variable will no longer be encrypted as a whole, but only the secret field would, while the JSON structure and the non sensitive data will remain clear. This will help support search capabilities, for example. Look for a password, based on the username. But also, this is not a minimal requirement, this proposal could also be implemented without this feature, and we could leave that for later and encrypt the entire payload for now.

design/proposals/variable_schema_validation.md
rafis3: Great questions, I'll clarify better in the design. My thoughts are:

1. No, you can't use more than one schema. I think it will over complicate our functionality, and will not provide a worthwhile value.
1. The schema is validated before the secret value is updated. I think that if the schema is changing, it will not affect the existing variables, but will only affect future updates.

design/proposals/variable_schema_validation.md
rafis3: Allowed and disallowed characters are supported in JSON schema. In my example, the port has to be a number between 1 and 65535.

design/proposals/variable_schema_validation.md
rafis3: This isn't a minimal requirement, because the admin could manage access to the builtin schema, but for a better user experience, it would be useful to just have access to them automatically

design/proposals/variable_schema_validation.md
rafis3: Done

rafis3:
@jvanderhoof I added a section about the current limitations in the beginning and after explaining about the solution, I elaborated more on the expected value. Let me know if you think it's still missing anything.

rafis3:
@sgnn7 regarding your second comment about built-in validations, can you please clarify? The purpose of these JSON schema validations is to allow the user to enforce the input, while allowing the user to easily set the enforcement rules. What did you have in mind?

design/proposals/variable_schema_validation.md
sgnn7: I see your point here but I feel like search for a username may be inadvertently looking for a solution to a problem that is not common among users as well as add confusion (and potential information leakage) to how we store variable values.

sgnn7:
> @sgnn7 regarding your second comment about built-in validations, can you please clarify? The purpose of these JSON schema validations is to allow the user to enforce the input, while allowing the user to easily set the enforcement rules. What did you have in mind?

@rafis3 What I think is going to be the most common scenario is simple validation of a password strength/charset for a specific backend credential (like a DB). What I would like to see first (before adding the full ability to specify a schema) is better UX for this workflow so I'm thinking something along the line of either:

More flexible:
```yaml
- !policy
  id: my-policy
  body:
  - !variable
    id: mysql-db-creds
    schema:
      - !pattern alphanumeric
      - !pattern length-min 12
      - !pattern length-max 40
```
Simpler:
```yaml
# `mysql_password` would have the default password complexity pattern for mysql passwords
- !policy
  id: my-policy
  body:
  - !variable
    id: mysql-db-creds
    annotations:
      schema: [ mysql_password ]
```

The examples here I gave aren't great but they should hopefully show enough of a pattern to explain it

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,293 @@
...
+- There is no enforcement on the group of variables needed for a certain use case. For example, a database connection requires a username, password, connection URL or address and port. A user could accidentally forget to specify a certain needed variable.
+
+## Proposed Solution
+
+The proposed solution is to add the ability to use JSON schemas and link them to variables, for which we want validation of their content. This will allow a single variable to store multiple, related values. While it's already possible today, to store JSON structured information in a variable, this proposed feature adds two added values:
+
+- The JSON structure is validated and enforced. If the input is invalid, the user will get a proper error message and the variable update will be denied.
+- The JSON structure will not be encrypted completely, but only the sensitive information. This will allow future seach capabilities based on the non sensitive information.
+
+### User Experience
```
nessiLahav, line 67: I think you should elaborate here on more scenarios, because we have here a new entity with a new relation to variables. For example:

1. Deleting schema - it is possible to delete a schema ? if it already defined to some variables ? if you want to return an error which contains all the variables under the schema you should save this info in the DB.
2. Creating new schema - i suppose we will have a validation on the schema itself 
3. Updating variable - what is the UX? for example first i performed get on the variable to find the schema id, and then i fetch the schema rules, and base on those rules i create a new variable object to update

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,293 @@
...
+- We leverage a standard way to enforce the content of the variable, with a well known JSON schema.
+- Tightly coupled values, such as username and password, are updated together in a single transaction. This will prevent momentary inconsistency in which each value was updated independently, one after the other.
+- The user will not have to look in the documentation or to understand the required variables, their structure or allowed input. Conjur will provide this feedback when an update attempt is made.
+- The feature does not require changes in the database schma, only in new built-in content.
+- The returned variable is in a JSON structure, same as the rest of our APIs responses.
+
+## Affected Areas
+
+- The new functionality will be developed in the Conjur server and the policy parser.
+- No new APIs are needed, since all the functionality will be given as part of the policy loading.
```
nessiLahav, line 235: I think we will need an api to fetch all the variables ids under a specific schema? one use case i can think of is if you will want to delete a schema

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,293 @@
...
+- No changes required in the database structure. But new built-in data should be added into the database tables, when they are initially created (clean install) or when an upgrade occurs from any prioir version.
```
nessiLahav, line 236: Se my UX comment, i think we will need to save the relation between the schemas to the variables in the DB

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Main Advantages and Expected Value From This Proposal
+
+- We leverage a standard way to enforce the content of the variable, with a well known JSON schema.
+- Tightly coupled values, such as username and password, are updated together in a single transaction. This will prevent momentary inconsistency in which each value was updated independently, one after the other.
+- The user will not have to look in the documentation or to understand the required variables, their structure or allowed input. Conjur will provide this feedback when an update attempt is made.
+- The feature does not require changes in the database schma, only in new built-in content.
+- The returned variable is in a JSON structure, same as the rest of our APIs responses.
+
+## Affected Areas
```
nessiLahav, line 232: What will be the complexity of 3rd parties to rotate those new complex variables password values ?

rafis3:
@sgnn7 would you rather see that kind of UX in addition or instead of my proposal?
Your suggestion solves the input validation but doesn't offer the user to save multiple values, which was a significant part of the purpose of my proposal. Do you think it makes sense to keep multiple values in a single variable or would you rather see a single value in a single variable? In any case, how would you support the coupling between different variables for search and update purposes?

sgnn7:
> @sgnn7 would you rather see that kind of UX in addition or instead of my proposal?
Your suggestion solves the input validation but doesn't offer the user to save multiple values, which was a significant part of the purpose of my proposal. Do you think it makes sense to keep multiple values in a single variable or would you rather see a single value in a single variable? In any case, how would you support the coupling between different variables for search and update purposes?

I think my preference would be not to co-mingle variable values together. If we need a grouping of variable values I believe that we should construct it at a higher abstraction level rather than a single variable since the variable itself is the smallest permission granularity we can have. By adding capability to mix sensitive and non-sensitive data, we then have to (by the nature of it) introduce not just new schema/format/validation primitives but also security primitives and that increases complexity and counterbalances the UX advantages of the all-in-one approach.

As a counterpoint example, there's nothing preventing us from having "composite" virtual secrets that return a value that is a combination of a few variables:
```yaml
- !policy
  id: my-policy
  body:
  - !variable
    id: mysql-db-creds
    composite_json:
      - username: mysql-ec2-24/username
      - password: mysql-ec2-24/password
      - address: mysql-ec2-24/address
      - port: mysql-ec2-24/port

File: design/proposals/variable_schema_validation.md [hunk 15]
```diff
@@ -0,0 +1,293 @@
...
+      "maximum": 65535
+    }
+  },
+  "required": ["username", "password"],
+  "additionalProperties": false,
+  "non-secrets": [
+      "username", "port", "address"
+    ]
+}
+```
```
oburstein-hub, line 139: I think we will have to define guidlenes for developers when to add this kind of complex variables and whet to add new entities in policy. I think entity in policy also has atomicity support

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,293 @@
...
+  - [Documentation](#documentation)
+  - [Version Update](#version-update)
+  - [Future Work](#future-work)
+  - [Delivery Plan](#delivery-plan)
+    - [Minimal Functionality](#minimal-functionality)
+    - [Extended Functionality](#extended-functionality)
+  
+## Introduction
+
+Conjur variables are resources that allow a user or an application to define and store sensitive information.
```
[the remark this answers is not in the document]
rafis3: Yes. I am differentiating between a variable content, which is based on the business logic of the user. As opposed to policy validation, which is based on the business logic of Conjur. I think each should have a different solution

design/proposals/variable_schema_validation.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Current Limitations
+
+- Multiple variables cannot be updated together, as a single transaction. This might lead to a momentary inconsitency when for example, a username is updated, the password should probabaly be updated as well.
+- There is no enforcement on the content of a variable, which can lead to unintended invalid content.
+- There is no enforcement on the group of variables needed for a certain use case. For example, a database connection requires a username, password, connection URL or address and port. A user could accidentally forget to specify a certain needed variable.
+
+## Proposed Solution
+
+The proposed solution is to add the ability to use JSON schemas and link them to variables, for which we want validation of their content. This will allow a single variable to store multiple, related values. While it's already possible today, to store JSON structured information in a variable, this proposed feature adds two added values:
```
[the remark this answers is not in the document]
rafis3: Yes. It's possible today as well, the structure is currently just not enforced in any way.

design/proposals/variable_schema_validation.md
rafis3: Regarding atomicity, I mentioned that in the `Main Advantages and Expected Value`
Regarding the guidelines, I agree, I will provide more content about that.

design/proposals/variable_schema_validation.md
rafis3: I added a section that explains about it. As a start, we will support only updating of the entire JSON structure. In the future, the variable API will support granular updates as well.

design/proposals/variable_schema_validation.md
rafis3: I added a section to discuss this point, thanks for asking about that!
I think the solution should be cascading removal of the annotations from the variables that had references to the deleted JSON schema. It should be also very easy to find them in the `annotations` table in the database and delete them.

design/proposals/variable_schema_validation.md
rafis3: Please see my response above.

design/proposals/variable_schema_validation.md
rafis3: I currently defined it as out of scope. If it will be important, we'll add support to that structure as well.

design/proposals/variable_schema_validation.md
rafis3: And @nessiLahav, regarding your questions, I added appropriate sections in the proposal to answer your questions. Thanks for raising them, they are really good points.

design/proposals/variable_schema_validation.md
rafis3: Actually the ability to search for a secret based on certain properties were requested by multiple users.

oburstein-hub: I think that for this kind of search we will have to use some third-party infa for searching in json. We can issue a research for this later

rafis3: Postgres can do that out of the box. We might need to change our column from text to JSON in order for that to work. But that would probably provide the best performance and capabilities.

rafis3:
> I think my preference would be not to co-mingle variable values together. If we need a grouping of variable values I believe that we should construct it at a higher abstraction level rather than a single variable since the variable itself is the smallest permission granularity we can have. By adding capability to mix sensitive and non-sensitive data, we then have to (by the nature of it) introduce not just new schema/format/validation primitives but also security primitives and that increases complexity and counterbalances the UX advantages of the all-in-one approach.

@sgnn7 regarding more granular permissions, I don't think we want that. The purpose of this feature is not to put data with different intention, under a single variable. The purpose is to put the sensitive data, together with its insensitive properties. Together it's supposed to be one whole data type. Similar to the concept of a struct in programming languages.

But this feature also allows the user to be creative and find new and interesting uses for this capability. Since we rely on JSON schema, which is a well known standard. The user could leverage it for a highly flexible, custom input validation. It won't need to support only the subset of validations that we'll provide, but it could easily support a vast variaty of validations that the JSON schema supports.

InbalZilberman:
Thank you Rafi for this design!
This is an important issue and it's great we address it.
I think from UX point of view the schemas should be an object of their own and we need to validate whether or not their definition should be reflected in the policy. I see it more as some sort of a configuration. But this should be validated with end-users.

rafis3:
@InbalZilberman regarding the name, I think `schema` is more explanatory than platform. It really doesn't have to be related to a platform at all, it's a JSON schema that validates and enforces the correct content of a variable.
Regarding making it a resource of its own, I'm open to that. Curious to hear what @oburstein-hub, @jvanderhoof, @alexkalish, @netacoral think about this, or anyone else reading this thread.

design/proposals/variable_schema_validation.md
oburstein-hub: As I understand, there is a way to grant permissions for all conjur users. Am I right ?

design/proposals/variable_schema_validation.md
oburstein-hub: I think it would be good to provide examples also for hosts, users and hostFactorTokens. WDYT ?

design/proposals/variable_schema_validation.md
oburstein-hub: I think the we should use entcrypted attributes  as a usual paradigm. The exception in non-secret attributes and maybe some more

design/proposals/variable_schema_validation.md
oburstein-hub: I think that search of a username will be looking for a solution problem

design/proposals/variable_schema_validation.md
oburstein-hub: I agree that REGEX would be a good solution here, by it is impossible in JSON

design/proposals/variable_schema_validation.md
oburstein-hub: I think this idea is very good because it betters user experience

design/proposals/variable_schema_validation.md
oburstein-hub: To summarize - there will be a an update of specific parts of json possible, and deleting them also, right

design/proposals/variable_schema_validation.md
oburstein-hub: I think API fetching is a great idea and it is very good it is elaborated in design

design/proposals/variable_schema_validation.md
oburstein-hub: please, elaborate about the complexity of 3rd parties rotation - maybe it worth to discuss it

oburstein-hub:
Rafi
I think this with this design we will make a great progress in improving our Conjur infrastructure and putting all DAP products in the first like of market products of this type. We will also address most of the subjects raised by SEs and the pre and post sales work will become much easier. This kind of changes will leas us forward both from organisational and architectural point of view.

izgeri:
This hasn't been updated in a month now; should this PR get closed, or should this proposal be added to the `design/proposals/` folder and refined in there until it's ready to move to `design/` as a fully considered, ready-to-implement design?

PR #1790, opened 2020-08-31
Policy infra

oburstein-hub:
Problems addressed by this PR: 
1. Currently most of policy configurations in system can be added without any verification process.
This creates situations when Conjur DB holds invalid policies, and their usage / handling is done at runtime when Conjur features chose to ignore / skip / handle invalid configurations in some creative ways (specific per case). According to best practices of configuration engines these situations should not happen and invalid configurations should be rejected on loading to the system.
2. In many cases it would be much simpler to user not to fill parameters with obvious values, and that the system would provide defaults to them

Solutions:
1. This PR adds infra to mandate verifications on policy entities 
2. This PR creates an ordered place to put fields defaults

Security:
Preventing wrong inputs improves security of the system

This PR does not change any existing functionality and is oriented to future features - thus no new tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,17 +6,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
-- Hosts can authenticate from Google Compute Engines (GCE) using a GCE instance
-  identity token. See [design](design/authenticators/authn_gce/authn_gce_solution_design.md)
+- Hosts can authenticate from Google Compute Engines (GCE) using a GCE instance 
+  identity token. See [design](design/authenticators/authn_gce/authn_gce_solution_design.md) 
   for details ([cyberark/conjur#1711](https://github.com/cyberark/conjur/issues/1711)).
 - Add `/whoami` API endpoint for improved supportability and debugging for access
   tokens and client IP address determination. [cyberark/conjur#1697](https://github.com/cyberark/conjur/issues/1697)
 - `TRUSTED_PROXIES` is validated at Conjur startup to ensure that it contains
   valid IP addresses and/or address ranges in CIDR notation.
   [cyberark/conjur#1727](https://github.com/cyberark/conjur/issues/1727)
-- The `/authenticate` endpoint now returns a text/plain base64 encoded access token
-  if the `Accept-Encoding` request header includes `base64`.
-  [cyberark/conjur#151](https://github.com/cyberark/conjur/issues/151)
 
 ### Changed
 - The Conjur server request logs now records the same IP address used by audit
--- a/app/models/loader/types.rb
+++ b/app/models/loader/types.rb
@@ -23,14 +23,14 @@ module Loader
       # Wraps a policy object with a corresponding +Loader::Types+ object.
       #
       # +external_handler+ should provide the methods +policy_id+, +handle_password+,
-      # +handle_public_key+. This argument is optional if the policy will not use 
+      # +handle_public_key+. This argument is optional if the policy will not use
       # that functionality.
       def wrap obj, external_handler = nil
         cls = Types.const_get obj.class.name.split("::")[-1]
         cls.new obj, external_handler
       end
     end
-    
+
     class Base
       extend Forwardable
 
@@ -77,7 +77,7 @@ module Loader
       def create_role!
         ::Role.create role_id: roleid
       end
-      
+
       def role
         ::Role[roleid]
       end
@@ -96,42 +96,57 @@ module Loader
           resource.annotations_dataset.import(%i(resource_id name value), records)
         end
       end
-      
+
       def resource
         ::Resource[resourceid]
       end
     end
-      
+
     class Record < Types::Base
       include CreateRole
       include CreateResource
-      
+
+      def verify
+        message = "Verify method for entity #{self} does not exist"
+        raise Exceptions::InvalidPolicyObject.new(self.id, message: message)
+      end
+
+      def calculate_defaults!; end
+
       def create!
+        verify
+        calculate_defaults!
         create_role! if policy_object.respond_to?(:roleid)
         create_resource! if policy_object.respond_to?(:resourceid)
       end
     end
 
     class Role < Record
+      def verify; end
     end
 
     class Resource < Record
+      def verify; end
     end
 
     class Layer < Record
+      def verify; end
     end
 
     class Host < Record
+      def verify; end
       def_delegators :@policy_object, :restricted_to
 
       def create!
-        self.handle_restricted_to(self.roleid, restricted_to)    
-        super  
+        self.handle_restricted_to(self.roleid, restricted_to)
+        super
       end
     end
 
     class HostFactory < Record
       def_delegators :@policy_object, :layers
+
+      def verify; end
 
       def create!
         super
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,17 +6,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+  identity token. See [design](design/authenticators/authn_gce/authn_gce_solution_design.md) 
   for details ([cyberark/conjur#1711](https://github.com/cyberark/conjur/issues/1711)).
 - Add `/whoami` API endpoint for improved supportability and debugging for access
   tokens and client IP address determination. [cyberark/conjur#1697](https://github.com/cyberark/conjur/issues/1697)
 - `TRUSTED_PROXIES` is validated at Conjur startup to ensure that it contains
   valid IP addresses and/or address ranges in CIDR notation.
   [cyberark/conjur#1727](https://github.com/cyberark/conjur/issues/1727)
-- The `/authenticate` endpoint now returns a text/plain base64 encoded access token
-  if the `Accept-Encoding` request header includes `base64`.
-  [cyberark/conjur#151](https://github.com/cyberark/conjur/issues/151)
```
micahlee, line 17: Cleanups that are not related to the primary code change should be in their own commit with a `cleanup: ` prefix in the commit message.

This makes it significantly easier to review the code changes versus formatting/whitespace cleanups.

Note, I'm not sure you mean to remove the `/authenticate` changelog entry though, do you?

app/models/loader/types.rb
```diff
@@ -23,14 +23,14 @@ def admin_role account
...
-      # +handle_public_key+. This argument is optional if the policy will not use 
+      # +handle_public_key+. This argument is optional if the policy will not use
       # that functionality.
       def wrap obj, external_handler = nil
         cls = Types.const_get obj.class.name.split("::")[-1]
         cls.new obj, external_handler
       end
     end
-    
+
```
micahlee, line 33: See note above about cleanup commits vs code change commits. All of these whitespace changes should go in a separate commit.

app/models/loader/types.rb
```diff
@@ -96,43 +96,58 @@ def create_resource!
...
     end
-      
+
     class Record < Types::Base
       include CreateRole
       include CreateResource
-      
+
+      def verify
+        message = "Unsupported verify for entity #{self}"
```
micahlee, line 110: I'm not sure what this means, should this be instead:
```
"#{self} doesn't implement the `#verify` method in `loader/types.rb`.
```

app/models/loader/types.rb
```diff
@@ -96,43 +96,58 @@ def create_resource!
...
       end
     end
-      
+
     class Record < Types::Base
       include CreateRole
       include CreateResource
-      
+
+      def verify
```
micahlee, line 109: Can we use a more descriptive method name? `validate_record_values` maybe?

app/models/loader/types.rb
```diff
@@ -170,16 +185,33 @@ def identifier
...
+        #  raise Exceptions::InvalidPolicyObject.new(self.id, message: message)
+        # end
+      end
+
+      # Below is a sample method filling defaults for User entity in policy
+      # def calculate_defaults!
+      #  if self.uidnumber == nil
+      #    self.annotations["conjur/uidnumber"] = 10
+      #  end
+      # end
```
micahlee, line 214: I don't think I'm in support of this PR if it's only adding this for future use. Do you have verifications and default values you are planning to add to a type here?

If so, can we include at least one of those with this PR?

CHANGELOG.md
oburstein-hub: Its not mine - came from merge
will try to separate commits

app/models/loader/types.rb
oburstein-hub: I was said to wait for Shuli for proper error message - I will suggest her yours

app/models/loader/types.rb
oburstein-hub: its not only values - it can be also match with other policy parts

app/models/loader/types.rb
oburstein-hub: I agree with you, but for such thing we need feature signoff and it would be problematic to have it in this stage

app/models/loader/types.rb
```diff
@@ -170,16 +185,33 @@ def identifier
...
         self.annotations["conjur/gidnumber"] ||= self.gidnumber if self.gidnumber
 
         super
       end
     end
-    
+
     class User < Record
-      def_delegators :@policy_object, :public_keys, :account, :role_kind, :uidnumber, :restricted_to
+      def_delegators :@policy_object, :public_keys, :account, :role_kind, :uidnumber, :uidnumberabc, :restricted_to
```
rafis3, line 199: what does `uidnumberabc` stand for?

app/models/loader/types.rb
jvanderhoof: Could we capture this interface somewhere else? It feels like it's just adding a bunch of code without any value. Can we remove the interface stubbing until we actually implement the feature?

app/models/loader/types.rb
jvanderhoof: The Ruby convention would be something like `uid_number_abc`

app/models/loader/types.rb
shulifink: This seems to be okay. 
I thought of something more simple:

'Verify' method does not exist for #[self]

nessiLahav (changes requested):
I think we should add to this PR a simple end to end flow and to implement UTs and integrations tests, to see how it is all get connected and to prevent a future surprises.

A possible use cases can be:
1. Validate user/host name 
2. I know we already have an input validation for CIDR restrictions as part of policy load - so maybe refactoring the code and move it to here

micahlee:
@nessiLahav 

A couple of questions:

> Validate user/host name

Can you elaborate what the user/host name should be validated against here?

> I know we already have an input validation for CIDR restrictions as part of policy load - so maybe refactoring the code and move it to here

The CIDR format check we perform should stay in the parser. By catching it during parsing, we find it sooner and we can more easily report exactly where the error is in the original policy file (e.g. line number). This is probably going to be the case for the host name validation you're referring to as well, if it has to do with the format, syntax, etc..

The orchestration step would be most appropriate for errors where the incoming policy has an error integrating with the policy objects that already exist in Conjur, or with the current configuration of Conjur.

app/models/loader/types.rb
oburstein-hub: The implementation is going to be done in the shortest future by Nessi

app/models/loader/types.rb
oburstein-hub: field removed

oburstein-hub:
> I think we should add to this PR a simple end to end flow and to implement UTs and integrations tests, to see how it is all get connected and to prevent a future surprises.
> 
> A possible use cases can be:
> 
> 1. Validate user/host name
> 2. I know we already have an input validation for CIDR restrictions as part of policy load - so maybe refactoring the code and move it to here

Agreed to postpone it to next merge

oburstein-hub:
> What's the value this PR brings? Is there anything other than adding an interface for future functionality?

No - it is only for future functionality. However, both Nessi and myself are going to add the actual verifications in the coming day or 2

oburstein-hub:
duplicate

PR #1794, opened 2020-09-02
adding verification and default calculation infra to policy

oburstein-hub:
Problems addressed by this PR:

Currently most of policy configurations in system can be added without any verification process.
This creates situations when Conjur DB holds invalid policies, and their usage / handling is done at runtime when Conjur features chose to ignore / skip / handle invalid configurations in some creative ways (specific per case). According to best practices of configuration engines these situations should not happen and invalid configurations should be rejected on loading to the system.
In many cases it would be much simpler to user not to fill parameters with obvious values, and that the system would provide defaults to them
Solutions:

This PR adds infra to mandate verifications on policy entities
This PR creates an ordered place to put fields defaults
Security:
Preventing wrong inputs improves security of the system

This PR does not change any existing functionality and is oriented to future features - thus no new tests added

### DIFF
--- a/app/models/loader/types.rb
+++ b/app/models/loader/types.rb
@@ -23,14 +23,14 @@ module Loader
       # Wraps a policy object with a corresponding +Loader::Types+ object.
       #
       # +external_handler+ should provide the methods +policy_id+, +handle_password+,
-      # +handle_public_key+. This argument is optional if the policy will not use 
+      # +handle_public_key+. This argument is optional if the policy will not use
       # that functionality.
       def wrap obj, external_handler = nil
         cls = Types.const_get obj.class.name.split("::")[-1]
         cls.new obj, external_handler
       end
     end
-    
+
     class Base
       extend Forwardable
 
@@ -77,7 +77,7 @@ module Loader
       def create_role!
         ::Role.create role_id: roleid
       end
-      
+
       def role
         ::Role[roleid]
       end
@@ -96,42 +96,57 @@ module Loader
           resource.annotations_dataset.import(%i(resource_id name value), records)
         end
       end
-      
+
       def resource
         ::Resource[resourceid]
       end
     end
-      
+
     class Record < Types::Base
       include CreateRole
       include CreateResource
-      
+
+      def verify
+        message = "Verify method for entity #{self} does not exist"
+        raise Exceptions::InvalidPolicyObject.new(self.id, message: message)
+      end
+
+      def calculate_defaults!; end
+
       def create!
+        verify
+        calculate_defaults!
         create_role! if policy_object.respond_to?(:roleid)
         create_resource! if policy_object.respond_to?(:resourceid)
       end
     end
 
     class Role < Record
+      def verify; end
     end
 
     class Resource < Record
+      def verify; end
     end
 
     class Layer < Record
+      def verify; end
     end
 
     class Host < Record
+      def verify; end
       def_delegators :@policy_object, :restricted_to
 
       def create!
-        self.handle_restricted_to(self.roleid, restricted_to)    
-        super  
+        self.handle_restricted_to(self.roleid, restricted_to)
+        super
       end
     end
 
     class HostFactory < Record
       def_delegators :@policy_object, :layers
+
+      def verify; end
 
       def create!
         super
@@ -170,6 +185,8 @@ module Loader
     class Group < Record
       def_delegators :@policy_object, :gidnumber
 
+      def verify; end
+
       def create!
         self.annotations ||= {}
         self.annotations["conjur/gidnumber"] ||= self.gidnumber if self.gidnumber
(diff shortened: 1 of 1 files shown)


micahlee (dismissed):
@oburstein-hub would you prefer I reply to comments on the prior PR (https://github.com/cyberark/conjur/pull/1790), or start new comments on this PR?

app/models/loader/types.rb
```diff
@@ -170,17 +185,34 @@ def identifier
...
         super
       end
     end
-    
+
     class User < Record
       def_delegators :@policy_object, :public_keys, :account, :role_kind, :uidnumber, :restricted_to
 
+      # Below is a sample method verifying policy data validity
+      def verify
```
rafis3, line 202: Why is `calculate_defaults!` completely commented out but method declaration of `verify` is not commented?
Just making sure you're not missing anything by mistake.

app/models/loader/types.rb
```diff
@@ -170,17 +185,34 @@ def identifier
     class Group < Record
       def_delegators :@policy_object, :gidnumber
 
+      def verify; end
```
rafis3, line 188: Shouldn't this also call `super` so that the verify of `Record` will run as well?

app/models/loader/types.rb
```diff
@@ -96,43 +96,58 @@ def create_resource!
...
     end
-      
+
     class Record < Types::Base
       include CreateRole
       include CreateResource
-      
+
+      def verify
+        message = "Verify method for entity #{self} does not exist"
```
rafis3, line 110: Isn't this going to cause failures on create? Looks like `create` is currently calling it.

rafis3 (commented):
@oburstein-hub please take a look at my questions

app/models/loader/types.rb
oburstein-hub: because I dont mandate calculate_defaults implementation, but do verify! implementation.
There can be entity without defaults. entity without verify will through exception, and thus verify method must exist

app/models/loader/types.rb
oburstein-hub: No - it will through an exception then

app/models/loader/types.rb
oburstein-hub: No - because it is executing the verify method of derived classes, and it is implemented  in all of them now

jvanderhoof (dismissed):
If this is a WIP PR, can you please re-open as a draft?  Also, is this a duplication of #1790?

app/models/loader/types.rb
```diff
@@ -96,43 +96,58 @@ def create_resource!
...
+        raise Exceptions::InvalidPolicyObject.new(self.id, message: message)
+      end
+
+      def calculate_defaults!; end
```
jonahx, line 114: Unlike in Java, where you explicitly define interfaces, we should never have an empty method like this in ruby.  Ruby uses duck-typing instead.

app/models/loader/types.rb
```diff
@@ -96,43 +96,58 @@ def create_resource!
           resource.annotations_dataset.import(%i(resource_id name value), records)
         end
       end
-      
+
```
jonahx, line 99: I'll add a plus one to @micahlee 's comment from the other PR that all formatting changes should be moved into their own commit.  You had mentioned:

> Its not mine - came from merge

If you do a rebase, instead of a merge (which you always should to bring your branch up to date with master) then this won't happen.  There are ways you can still fix this in git that won't require manual intervention on your part.  Let me know if you need any help with that.

Btw, is there a difference between this and #1790 ?

app/models/loader/types.rb
```diff
@@ -96,43 +96,58 @@ def create_resource!
...
     class Role < Record
+      def verify; end
     end
 
     class Resource < Record
+      def verify; end
     end
 
     class Layer < Record
+      def verify; end
```
jonahx, line 133: See my comment above re: empty methods for all 3 of these too.

micahlee:
Hey @oburstein-hub, have you considered starting with a design doc under https://github.com/cyberark/conjur/tree/master/design/proposals to first scope out the outcome you're wanting to achieve?Then it would likely be more clear how this PR fits into the bigger picture.

The problem statement in the description is a useful one, to ensure proper configuration values, but this PR on its own doesn't make the case that this is a problem, what the specific problem is you want to address, and then how this change addresses the problem.

This appears to be step 1 of many, but I don't know from this what the other steps you're thinking about are. It would help (at least) me a lot to see more of that context to make an informed review.

Thanks!

app/models/loader/types.rb
oburstein-hub: @jonahx thank you for your comments. I agree that empty methods is not the perfect way - but if I would not use them I could not have a method in Record class (verify method which throws exception). And it is essential to enforce people to implement verification. I also plan that all the verify methods that are currently empty will be filled soon by applicational logic (relevant verifications). I just could not do it here because it would change customer-facing functionality

app/models/loader/types.rb
oburstein-hub: There is not difference between these commits - only technical problems with merge. If we will use the old one - we will loose Rafi's comments..

izgeri:
This has been open for over a month - can we close it for now, to be reopened when work will continue? Thanks!

oburstein-hub:
@izgeri please dont close it. IOt is going to be merged to another PR shortly

oburstein-hub:
Reopening for merge

oburstein-hub:
This PR contains verification infra that will be used in the implementation of following requirements https://app.zenhub.com/workspaces/palmtree-5d99d900491c060001c85cba/issues/cyberark/conjur/1863

rafis3 (dismissed):

oburstein-hub:
@micahlee - to answer regarding the objective of this PR:
There is no orderly mechanism of verifications upon policy load, also there is no mechanism for implicit default addition on unconfigured fields. 
Current PR adds these mechanism with sample examples conmented out.

After having this PR we can add features to validate policy on loading, and prevent illegal policies to persist in DB

rafis3 (approved):

rafis3 (approved):

Issue #1559, opened 2020-05-19
authn_k8s:inject_client_cert: 


subodh-shrivastava:

micahlee:
Hi @subodh-shrivastava, thanks for the issue!

One question to better understand what you're describing...

The most common workflow is to use the [Kubernetes client sidecar or init container](https://github.com/cyberark/conjur-authn-k8s-client) to handle the Pod's authentication with Conjur, including the client certificate injection.

Are you using this container for authentication, and you want to configure the certificate path for that container? Or are you looking to perform the entire authentication workflow from within your own application container, and need the configurable path for that?

Thanks!

subodh-shrivastava:
@micahlee It's the later, we are looking to perform the entire authentication workflow from within our own application plugin and need the configurable path for that.

izgeri:
It looks like to implement these we'd need to
- Update the Conjur server to provide a configuration option for the `DefaultClientCertPath` for authn-k8s, with the default value if no config is specified set as `/etc/conjur/ssl/client.pem` - TBD what exactly this configuration will look like
- Update the Kubernetes authenticator to consume the new cert path configuration option in the `inject_client_cert` class here: https://github.com/cyberark/conjur/blob/cd0550a6ff282394325eac8a0ab681d698ffa176/app/domain/authentication/authn_k8s/inject_client_cert.rb#L71
- Potentially add an rspec test [here](https://github.com/cyberark/conjur/blob/master/spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb)

Note also that:
- the Jenkins pipeline runs an e2e test in GKE that we could consider augmenting with a custom cert dir flow: https://github.com/cyberark/conjur/blob/master/Jenkinsfile#L137
- there are a set of cucumber tests for authn-k8s that should be augmented with a non-standard cert directory flow: https://github.com/cyberark/conjur/tree/master/cucumber/kubernetes/features
- there's not an obvious place to update the documentation with this change; our Kubernetes documentation lists our supported client methods as the authn-k8s client, secretless, or the secrets provider (see reference [here](https://docs.conjur.org/Latest/en/Content/Integrations/Kubernetes_deployApplicationApplication.htm)) - we don't officially support other flows for authn-k8s, and all of the referenced flows use the default certificate directory.

Given all of that, I guess I'm wondering if @subodh-shrivastava could give us some more information about this line:
> This will come in handy with native spring boot plug-in which can be used by existing applications without making code changes or creating new docker image.

[...]

PR #1802, opened 2020-09-05
Configurable Cert Path for Auth K8s

JakeQuilty:
Connected to https://github.com/cyberark/conjur/issues/1559

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/conjur#151](https://github.com/cyberark/conjur/issues/151)
 - [Documentation](./UPGRADING.md) explaining how to upgrade a Conjur server deployed in a
   Docker Compose environment. [cyberark/conjur#1528](https://github.com/cyberark/conjur/issues/1528), [cyberark/conjur#1584](https://github.com/cyberark/conjur/issues/1584)
+- Add ability to specify cert installation path on `/authn-k8s/:service_id/inject_client_cert`.
+  [cyberark/conjur#1802](https://github.com/cyberark/conjur/pull/1802)
 
 ### Changed
 - The Conjur server request logs now records the same IP address used by audit
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -120,10 +120,11 @@ class AuthenticateController < ApplicationController
   def k8s_inject_client_cert
     # TODO: add this to initializer
     Authentication::AuthnK8s::InjectClientCert.new.(
-      conjur_account:   ENV['CONJUR_ACCOUNT'],
-      service_id:       params[:service_id],
-      client_ip:        request.ip,
-      csr:              request.body.read,
+      conjur_account:                ENV['CONJUR_ACCOUNT'],
+      service_id:                    params[:service_id],
+      client_ip:                     request.ip,
+      csr:                           request.body.read,
+      cert_installation_path:        request.headers["Certificate-Install-Path"],
 
       # The host-id is split in the client where the suffix is in the CSR
       # and the prefix is in the header. This is done to maintain backwards-compatibility
--- a/app/domain/authentication/authn_k8s/inject_client_cert.rb
+++ b/app/domain/authentication/authn_k8s/inject_client_cert.rb
@@ -6,6 +6,7 @@ module Authentication
   module AuthnK8s
 
     KUBERNETES_AUTHENTICATOR_NAME = 'authn-k8s'
+    DEFAULT_CERT_PATH = "/etc/conjur/ssl/client.pem"
 
     InjectClientCert ||= CommandClass.new(
       dependencies: {
@@ -17,7 +18,7 @@ module Authentication
         extract_container_name: ExtractContainerName.new,
         audit_log:              ::Audit.logger
       },
-      inputs: %i(conjur_account service_id csr host_id_prefix client_ip)
+      inputs: %i(conjur_account service_id csr host_id_prefix client_ip cert_installation_path)
     ) do
 
       # :reek:TooManyStatements
@@ -68,10 +69,14 @@ module Authentication
         @validate_pod_request.(pod_request: pod_request)
       end
 
+      def cert_installation_path
+        @cert_installation_path.nil? || @cert_installation_path.empty? ? DEFAULT_CERT_PATH : @cert_installation_path
+      end
+
       def install_signed_cert
         pod_namespace = spiffe_id.namespace
         pod_name = spiffe_id.name
-        cert_file_path = "/etc/conjur/ssl/client.pem"
+        cert_file_path = cert_installation_path
         @logger.debug(LogMessages::Authentication::AuthnK8s::CopySSLToPod.new(
           container_name,
           cert_file_path,
--- a/spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb
+++ b/spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb
@@ -16,6 +16,9 @@ RSpec.describe Authentication::AuthnK8s::InjectClientCert do
   let(:nil_host_id_prefix) { nil }
   let(:csr) { "CSR" }
 
+  let(:default_cert_file_path) { "/etc/conjur/ssl/client.pem" }
+  let(:expected_cert_file_path) { "/etc/conjur/ssl/client.pem" }
+
   let(:host_id) { "HostId" }
   let(:host_role) { double("HostRole", id: host_id) }
 
@@ -182,6 +185,7 @@ RSpec.describe Authentication::AuthnK8s::InjectClientCert do
           expect { injector.(conjur_account: account,
                             service_id: service_id,
                             csr: csr,
+                            cert_installation_path: nil,
                             host_id_prefix: host_id_prefix,
                             client_ip: client_ip) }.to raise_error(error_type, missing_spiffe_id_error)
         end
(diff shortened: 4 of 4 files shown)


app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -71,7 +72,7 @@ def validate
       def install_signed_cert
         pod_namespace = spiffe_id.namespace
         pod_name = spiffe_id.name
-        cert_file_path = "/etc/conjur/ssl/client.pem"
+        cert_file_path = @cert_path.nil? || @cert_path.empty? ? DEFAULT_CERT_PATH : @cert_installation_path
```
izgeri, line 75: this logic should probably be in its own method and can be unit tested

you can see how we do this for the `host_id_prefix` (which can also be set in a header) here:
https://github.com/cyberark/conjur/blob/73c197037a3bce4122d9ae97654344e5e265e470/app/domain/authentication/authn_k8s/inject_client_cert.rb#L57-L59

maybe we can add a unit test for the logic here: https://github.com/cyberark/conjur/blob/master/spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb?

izgeri (commented):
@JakeQuilty you might find it useful to search for how the value for `host_id_prefix` is passed / parsed in a header and how its unit / integration tests are implemented - the client cert path should be similar
https://github.com/cyberark/conjur/search?q=host_id_prefix&unscoped_q=host_id_prefix

app/domain/authentication/authn_k8s/inject_client_cert.rb
orenbm, line 75: i may be missing something but where is `@cert_path` defined?

orenbm: oh i guess it should be `@cert_installation_path`, correct?

config/routes.rb
```diff
@@ -43,7 +43,7 @@ def matches?(request)
         # this to be accessible when using other authenticators to login.
         put  '/:authenticator/:account/api_key'  => 'credentials#rotate_api_key'
 
-        post '/authn-k8s/:service_id/inject_client_cert' => 'authenticate#k8s_inject_client_cert'
+        post '/authn-k8s/:service_id/inject_client_cert' => 'authenticate#k8s_inject_client_cert'  # <-----------------------------------
```
orenbm, line 46: what are all these `# <--------------`?

app/domain/authentication/authn_k8s/inject_client_cert.rb
JakeQuilty: 👍

config/routes.rb
JakeQuilty: Yeah, I was using these just to mark important things for myself. I'm deleting these when I submit the PR as a non-WIP

spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb
```diff
@@ -182,6 +187,7 @@
           expect { injector.(conjur_account: account,
                             service_id: service_id,
                             csr: csr,
+                            cert_installation_path: nil_cert_installation_path,
```
jonahx, line 190: I'd just get rid of `nil_cert_installation_path` and use the literal `nil`.  Seeing:
```
cert_installation_path: nil
```
will be clearer.

JakeQuilty: 👍

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -68,10 +69,14 @@ def validate
         @validate_pod_request.(pod_request: pod_request)
       end
 
+      def set_cert_installation_path
```
izgeri, line 72: I'll let the Ruby experts give their recommendations on this one, but my sense is that the method name shouldn't be `set_X` as that's not standard for Ruby

looking at our [style guide](https://github.com/cyberark/community/blob/master/Conjur/conventions/ruby-style-guide.md), there is indeed a link to this helpful snippet: https://github.com/rubocop-hq/ruby-style-guide#accessor_mutator_method_names

orenbm: yes, i would change this method to:
```
def set_cert_installation_path
  @cert_installation_path ||= @cert_installation_path.empty? ? DEFAULT_CERT_PATH : @cert_installation_path
end
```

in the implementation above you don't need to explicitly check if `@cert_installation_path` is `nil` as the memoization will do it for you. please add UTs for this logic if you haven't.

orenbm: btw, when will `request.headers["Certificate-Install-Path"]` be empty? i also don't see any UTs for this use-case?

JakeQuilty: I was copying the same logic as https://github.com/cyberark/conjur/blob/f61ebd54b946d4210cb537049fcfcaab90a40d93/app/domain/authentication/authn_k8s/inject_client_cert.rb#L56 since they are both headers that are doing the same thing. I could see a use case as someone adding a `Certificate-Install-Path` header to the request with no value

JakeQuilty: I'm still having trouble creating a UT for this. In the rspec tests we haven't been able to find a good way to do this. Do you have any suggestions?

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/conjur#151](https://github.com/cyberark/conjur/issues/151)
 - [Documentation](./UPGRADING.md) explaining how to upgrade a Conjur server deployed in a
   Docker Compose environment. [cyberark/conjur#1528](https://github.com/cyberark/conjur/issues/1528), [cyberark/conjur#1584](https://github.com/cyberark/conjur/issues/1584)
+- Add ability to specify cert installation path on `/authn-k8s/:service_id/inject_client_cert`.
+  [cyberark/conjur#1802](https://github.com/cyberark/conjur/pull/1802)
```
izgeri, line 23: this should be linked to the issue that was filed with this request, both here and in the PR description. can you please update the changelog entry and PR description to include this link (and drop the PR link here), and also add some text to the PR description to explain what this PR does?

spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb
```diff
@@ -16,6 +16,9 @@
   let(:nil_host_id_prefix) { nil }
   let(:csr) { "CSR" }
 
+  let(:default_cert_file_path) { "/etc/conjur/ssl/client.pem" }
+  let(:expected_cert_file_path) { "/etc/conjur/ssl/client.pem" }
```
izgeri, line 20: why is this needed? I would guess that not all tests have this as the expected cert path (you must have some tests that use a non-default entry, no?) - and when they don't, the `default_cert_file_path` can be referred to

spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb
```diff
@@ -371,6 +384,60 @@
...
+            allow(copy_response).to receive(:[])
+              .with(:error)
+              .and_return(expected_error_text)
+  
+            expect { injector.(conjur_account: account,
+                              service_id: service_id,
+                              csr: csr,
+                              cert_installation_path: "not_valid_path",
+                              host_id_prefix: host_id_prefix,
+                              client_ip: client_ip) }.to raise_error(error_type, expected_full_error_text)
```
izgeri, line 438: these lines use different indentation than the rest of the file

izgeri (changes requested):

app/controllers/authenticate_controller.rb
```diff
@@ -120,10 +120,11 @@ def authenticator_input
...
     Authentication::AuthnK8s::InjectClientCert.new.(
-      conjur_account:   ENV['CONJUR_ACCOUNT'],
-      service_id:       params[:service_id],
-      client_ip:        request.ip,
-      csr:              request.body.read,
+      conjur_account:                ENV['CONJUR_ACCOUNT'],
+      service_id:                    params[:service_id],
+      client_ip:                     request.ip,
+      csr:                           request.body.read,
+      cert_installation_path:        request.headers["Certificate-Install-Path"],
```
orenbm, line 127: should this be `Certificate-Installation-Path`?

@shulifink can you review this? it is the path in which the Conjur server will install the certificate inside the client container.

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -68,10 +69,14 @@ def validate
         @validate_pod_request.(pod_request: pod_request)
       end
 
+      def cert_installation_path
+        @cert_installation_path.nil? || @cert_installation_path.empty? ? DEFAULT_CERT_PATH : @cert_installation_path
+      end
```
orenbm, line 74: Let's memoize this

```suggestion
      def cert_installation_path
        @cert_installation_path ||= @cert_installation_path.nil? || @cert_installation_path.empty? ? DEFAULT_CERT_PATH : @cert_installation_path
      end
```

orenbm, line 74: Are you not using memoization on purpose? i think that if you'd use memoization then in case it not nil but is empty it will not work because the logic will not run. If this is the case please add a comment on why we're not using memoization so a future developer will not change it.

app/controllers/authenticate_controller.rb
shulifink: It's funny how each one has half abbreviated and half a full word.
Why not just use "Cert_Installation_Path" line the option?

orenbm: but that's funny in the same way because now `Cert` is half abbreviated and `Installation` is a full word. I would use either `Cert-Install-Path` or `Certificate-Installation-Path`.

Regarding `cert_installation_path` - that's not customer facing but just a code variable so i wouldn't bother with its naming.

app/domain/authentication/authn_k8s/inject_client_cert.rb
orenbm: what if you path `cert_install_path: ""`?

PR #1821, opened 2020-09-10
Move email claim to be an optional claim

nessiLahav:
### What does this PR do?
Move email claim to be an optional claim

In the decoded token stage we will fail the request only if the following claims are missing:
1. Aud - contains the host-id
2. Sub - is the only cliam which shared with GCE and GCF

All other token are optional, and will fail the request only the JWT token validation stage

### DIFF
--- a/app/domain/authentication/authn_gcp/decoded_token.rb
+++ b/app/domain/authentication/authn_gcp/decoded_token.rb
@@ -26,10 +26,10 @@ module Authentication
       def initialize_required_claims
         @audience = required_token_claim_value(AUDIENCE_TOKEN_CLAIM_NAME)
         @service_account_id = required_token_claim_value(SUB_TOKEN_CLAIM_NAME)
-        @service_account_email = required_token_claim_value(EMAIL_TOKEN_CLAIM_NAME)
       end
 
       def initialize_optional_claims
+        @service_account_email = optional_token_claim_value(EMAIL_TOKEN_CLAIM_NAME)
         @project_id = optional_token_claim_value(PROJECT_ID_TOKEN_CLAIM_NAME)
         @instance_name = optional_token_claim_value(INSTANCE_NAME_TOKEN_CLAIM_NAME)
       end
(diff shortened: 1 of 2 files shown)


app/domain/authentication/authn_gcp/decoded_token.rb
```diff
@@ -50,6 +50,7 @@ def optional_token_claim_value(optional_token_claim)
         optional_token_claim_value = token_claim_value(optional_token_claim)
 
         if optional_token_claim_value.nil? || optional_token_claim_value.empty?
+          optional_token_claim_value = nil
```
eranha, line 53: Why does it have to be `nil`?

spec/app/domain/authentication/authn-gcp/decoded_token_spec.rb
```diff
@@ -290,8 +296,15 @@ def decoded_token_hash(token_str)
           )
         end
 
-        it "raises an error" do
-          expect { decoded_token }.to raise_error(Errors::Authentication::Jwt::TokenClaimNotFoundOrEmpty)
+        it "does not raise an error" do
+          expect { decoded_token }.to_not raise_error
+        end
+
+        it "parses the token expectedly" do
```
eranha, line 303: as expected

spec/app/domain/authentication/authn-gcp/decoded_token_spec.rb
```diff
@@ -303,13 +316,18 @@ def decoded_token_hash(token_str)
...
```
eranha, line 323: as expected (it is valid but IMHO not common)

eranha (changes requested):

spec/app/domain/authentication/authn-gcp/decoded_token_spec.rb
nessiLahav: It is aligned with the whole file

app/domain/authentication/authn_gcp/decoded_token.rb
nessiLahav: i prefer to have 1  value which indicate not exists, so i choose between nil and empty i choose nil

eranha (approved):

PR #1789, opened 2020-08-30
Handle cert injection properly for authn-k8s

orenbm:
Connected to #1807 

Currently, we close the WebSocket connection after we send
the injection request to kubectl. This is not correct as this
way we do not wait for kubectl to send its response.

This has 2 implications:
1. If the injection request has succeeded we will return a 200 OK response
   to the 'inject_client_cert' request even if the cert has not finished to
   be written to the pod. This can lead to a race-condition where the client
   will try to read the cert before it is there.
2. If the injection request has failed in the k8s API, we will not
   write the error to the Conjur logs, and we will falsely return a 200
   OK response to the 'inject_client_cert' request

There is also [an issue](https://github.com/kubernetes/kubernetes/issues/89899)
in k8s where messages that are sent via WebSockets to Kubectl Exec are
left hanging without returning a response when using STDIN. This happens because
the `tar` command that we use to copy the file into the container is waiting for the STDIN
to end. 

The solution to the issue above is to inject the cert by running a bash script that exits explicitly.
This way we don't have the hanging issue, and the K8s API will close the socket
upon cert injection success/failure.

Additionally, to overcome the issue that we don't log the error if it occurs, we now redirect
the stdout and stderr into a file inside the client container. This will help is troubleshooting
errors in the cert injection. 

*Logs*
Let's see the Conjur server logs, and the client container logs, in cases where the cert injection succeeds and fails. 
Note: we show the logs of the secrets-provider that uses the authn-client as it's easier to deploy.

**Success**

Conjur server:
```
[origin=10.1.3.190] [request_id=f80c9d90-6100-4ca4-9e7b-c1c54d9d2e60] [tid=45] Started POST "/authn-k8s/authn-dev-env/inject_client_cert" for 10.1.3.190 at 2020-09-30 09:37:54 +0000
[origin=10.1.3.190] [request_id=f80c9d90-6100-4ca4-9e7b-c1c54d9d2e60] [tid=45] CONJ00028D Setting common name to host.conjur.authn-k8s.authn-dev-env.apps.local-secrets-provider.*.*
[origin=10.1.3.190] [request_id=f80c9d90-6100-4ca4-9e7b-c1c54d9d2e60] [tid=45] CONJ00027D Host id cucumber:host:conjur/authn-k8s/authn-dev-env/apps/local-secrets-provider/*/* extracted from CSR common name
[...]

PR #1825, opened 2020-09-15
Add solution design for inject client cert fix

orenbm:
### What does this PR do?
Adds a solution design for a fix to the "inject-client-cert" issue. More info can be found in the design itself.

It is best to give inline comments where needed and read the full design [here](https://github.com/cyberark/conjur/blob/kubectl-exec-solution-design/design/authenticators/authn_k8s/inject_client_cert_solution_design.md).


design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,235 @@
...
+   OK response to the 'inject_client_cert' request
+
+We would like to find a solution that will indicate whether the cert was successfully
+injected and log the error otherwise.
+
+## Solution
+[//]: # (Elaborate on the solution you are suggesting in this page. Address the functional requirements and the non functional requirements that this solution is addressing. If there are a few options considered for the solution, mention them and explain why the actual solution was chosen over them. Add an execution plan when relevant. It doesn't have to be a full breakdown of the feature, but just a recommendation to how the solution should be approached.)
+
+We would like to inject the cert and get an indication that the cert was injected 
+or the error if it was not. Let's explore the different options that we have to inject the cert.
```
abrahamko, line 47: @orenbm Did you consider replacing the websocket solution with POST HTTP request?
In the same way `kubectl` does it.
If it's good for `kubectl` it should be good for us, no?
It uses SDPY protocol, not WebSocket.
Read here for more info: https://erkanerol.github.io/post/how-kubectl-exec-works/

orenbm: Thanks @abrahamko !

It looks like the SPDY protocol is being [deprecated](https://github.com/kubernetes/enhancements/issues/384) so it's best not to use it. As written [here](https://www.openshift.com/blog/executing-commands-in-pods-using-k8s-api):
```
kubectl and oc use the SPDY protocol at the moment, which is being deprecated . The second option is to use Websockets, which seems to be the best way. Anyway one of these two protocols, SPDY or WebSockets, is required for communication with this endpoint, and the API will refuse requests without Upgrade headers.
```

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+- 2 people can work on this solution but it's ok to have only 1.
+
+### Designs 
+[//]: # "Note and estimate the tasks which are related to designs which needed for the solution"
+
+- Add a diagram to the solution design once we decide on the solution
+  - EE: 1 day
+
+### Spikes and researches
+
```
Tovli, line 162: Do you plan to spike all the options? Isn't there a preffered option that worth testing first?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+- [Existing PR with solution](https://github.com/cyberark/conjur/pull/1789)
+  - In this solution we wait for the timeout to get an answer and if we didn't
+    get one we create another WebSocket with an "ls" command to verify that the
+    cert is injected.
+- [Known issue in k8s exec](https://github.com/kubernetes/kubernetes/issues/89899)
+
+## Background
+[//]: # "Give relevant background for the designed feature. What is the motivation for this solution?"
+We have an issue with the "inject client cert" process where we don't always return a correct response to the client regarding the injection success. The injection is done via a WebSocket to Kubectl Exec where we use a "tar" command and the cert content in STDIN. There is [an issue](https://github.com/kubernetes/kubernetes/issues/89899) in k8s where messages that are sent via WebSockets to Kubectl Exec are left hanging without returning a response when using STDIN.
```
rafis3, line 18: When you say kubectl, you actually mean the Kubernetes API server or the Ruby kubeclient?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+The current behaviour is that we close the WebSocket connection after we send the injection request to kubectl. This is not correct as this way we do not wait for kubectl to send its response.
+
+This has 2 implications:
+1. If the injection request has succeeded we will return a 200 OK response to the 'inject_client_cert' request even if the cert has not finished to be written to the pod. This can lead to a race-condition where the client will try to read the cert before it is there.
+1. If the injection request has failed in the k8s API, we will not write the error to the Conjur logs, and we will falsely return a 200 OK response to the 'inject_client_cert' request
+
+## Solution
+[//]: # "Elaborate on the solution you are suggesting in this page. Address the functional requirements and the non functional requirements that this solution is addressing. If there are a few options considered for the solution, mention them and explain why the actual solution was chosen over them. Add an execution plan when relevant. It doesn't have to be a full breakdown of the feature, but just a recommendation to how the solution should be approached."
```
Tovli, line 29: I appreciate the multiple options you've presented but it seems to me that options 1 and 3 are inferior to option 2. So why not just verifying the URL length issue?
This should be easy by sending a 8k cert, no?

If I understand it correctly - if the URL length is not an issue then option 2 will be chosen, no?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+
+However, we may have a limit on the size of the URL and we will have an issue  if the certificate will grow in the future. To overcome this, we can split the  request. The first one will create the file. The following requests will populate it with the cert content. Then the last request will verify that the whole content has passed (with one of the methods described above). We can decide on the number  of requests dynamically according to the size of the cert.
+
+#### Option 3 - Client sets its State once cert is injected
+This option is relevant only if the "inject-client-cert" request will return a 202 Accepted response instead of 200 OK (more on that in [this section](#inject-client-cert-response-code).
+
+In this case, we respond with 202 Accepted to the "inject-client-cert" request and start injecting the cert in a separate thread. The server then polls on the state of the authn-client container and waits until its State changes to some  agreed State (e.g "CertInjected"). This state will be set by the authn-client once it sees that the cert is installed.
+
+This option is not fully backwards-compatible and is quite complex so I wouldn't suggest to implement it.
```
rafis3, line 50: Completely agree

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+- conjur-authn-k8s-client
+- secrets-provider-for-k8s
+
+### Cross-team dependencies
+[//]: # "Elaborate about your solution's requirements or needs from other teams"
+We would like to get a review on the solution from @cyberark/conjur-core-team  and @cyberark/community-and-integrations-team. There are no requirements from  the Infra team.
+
+## Security
+[//]: # "Are there any security issues with your solution? Even if you mentioned them somewhere in the doc it may be convenient for the security architect review to have them centralized here"
+1. Is it ok to pass the base64 encoded cert in the URL? Is it less secure than sending it in the STDIN?
```
rafis3, line 105: While the traffic is fully encrypted when using https, including the URL. It's still possible that the URL will be visible in certain locations, while the request travels to its destination. See this [page](https://owasp.org/www-community/vulnerabilities/Information_exposure_through_query_strings_in_url) for more details.

It is possible that the Kubernetes API server will not leak any information from the URL, but if the Conjur instance runs outside the cluster, it will surely pass some proxies and load balancers before it reaches Kubernetes.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+
+ Will be decided upon investigation.
+
+### Design
+[//]: # "Add any diagrams, charts and explanations about the design aspect of the solution. Elaborate also about the expected user experience for the feature"
+TODO: Add diagram once we decide on a solution
+
+### Backwards compatibility
+[//]: # "Address how you are going to handle backwards compatibility, if necessary"
+At current, the authn-client sends the "inject-client-cert" request and then  tries to read the cert. As this solution will just fix the way that we inject the  cert, we should have any backwards-compatibility issues.
```
rafis3, line 84:
```suggestion
At current, the authn-client sends the "inject-client-cert" request and then  tries to read the cert. As this solution will just fix the way that we inject the  cert, we shouldn't have any backwards-compatibility issues.
```

rafis3: Also, what about forward compatibility? In option 3, if a user takes the latest authn client that expects either 202 or 200 but also uses an older Conjur version, the new authn client would always get 200 and expect the file to exists. Right?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
rafis3: So while I agree with @Tovli that option 2 looks the most promising technically. Let's make sure that it's safe enough to use.

rafis3 (commented):
@orenbm added some comments.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
Tovli: @sashaCher - would appreciate your response here. Mainly to understand the severity of having option 2 with follower outside the cluster

sashaCher: I'm totally agree with @rafis3.
Despite HTTPS standard guarantees that url is encrypted (except its FQDN part) we have no guarantee that it's not logged in any of upstream systems. For example in conjur case nginx logs all urls.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+
+We would like to inject the cert and get an indication that the cert was injected  or the error if it was not. Let's explore the different options that we have to inject the cert. 
+
+**Note**: The SPDY protocol [is being deprecated](https://github.com/kubernetes/enhancements/issues/384)  so it's not an option.
+
+#### Option 1 - Run a parallel request to verify cert injection
+In this option, we send the tar request as we do today, but also create a parallel WebSocket in which we will send another request to verify that the cert was successfully injected. We have several options for the verification command:
+1. Run an "ls" command on the cert path. This way we get an immediate feedback on the  cert presence. However, this may return a positive answer even if the cert is not fully injected (this assumption will be investigated as part of this effort).
+1. Calculate the md5 hash of the cert content before sending it and then sending  a request to get the calculated md5 hash of the cert in the container. Once the two hashes are equal then we know that the cert was fully installed.
```
sashaCher, line 38: `MD5` is not FIPS combined hash algorithm and MD5 calculation will throw an exception.
I would recommend to use `SHA-1` or higher.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+
+However, we may have a limit on the size of the URL and we will have an issue  if the certificate will grow in the future. To overcome this, we can split the  request. The first one will create the file. The following requests will populate it with the cert content. Then the last request will verify that the whole content has passed (with one of the methods described above). We can decide on the number  of requests dynamically according to the size of the cert.
+
+#### Option 3 - Client sets its State once cert is injected
+This option is relevant only if the "inject-client-cert" request will return a 202 Accepted response instead of 200 OK (more on that in [this section](#inject-client-cert-response-code).
+
+In this case, we respond with 202 Accepted to the "inject-client-cert" request and start injecting the cert in a separate thread. The server then polls on the state of the authn-client container and waits until its State changes to some  agreed State (e.g "CertInjected"). This state will be set by the authn-client once it sees that the cert is installed.
+
+This option is not fully backwards-compatible and is quite complex so I wouldn't suggest to implement it.
+
```
moticless, line 51: **Option 4 - Combination of strategies**
Obviously waiting to write completion ack is very time consuming. Such approach might severely overload the server and its workers. Alternatively, for the common case, we can offloading of the work to the client such that:
1. Like today, the client will send a request, get response and polling the cert up to X seconds, before timeout. The server will inject CERT without waiting.
2. If timeout, send again inject_client_cert but this time with ack_on_write indication. Only this time the server will wait on inject CERT completion/error.

oburstein-hub: I think timeout is a good idea, but an asynchronous retry after it might bring the system again to the condition of missing cert file

moticless: Ofira, I am not sure i understand your comment. On retry client shift its strategy and send the request again with ack_on_write indication to the server- which is what suggested on the other paper.

oburstein-hub: In this case it is a good idea - it was not clear to me originally

nessiLahav: @moticless in your solution how do you prevent form the client to fetch a cert which is not fully injected ? this will be really hard to troubleshoot

nessiLahav: @moticless the client could get the hash from the server and verify it, so when i rethinking about it this is not a problem

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+This has 2 implications:
+1. If the injection request has succeeded we will return a 200 OK response to the 'inject_client_cert' request even if the cert has not finished to be written to the pod. This can lead to a race-condition where the client will try to read the cert before it is there.
+1. If the injection request has failed in the k8s API, we will not write the error to the Conjur logs, and we will falsely return a 200 OK response to the 'inject_client_cert' request
+
+## Solution
+[//]: # "Elaborate on the solution you are suggesting in this page. Address the functional requirements and the non functional requirements that this solution is addressing. If there are a few options considered for the solution, mention them and explain why the actual solution was chosen over them. Add an execution plan when relevant. It doesn't have to be a full breakdown of the feature, but just a recommendation to how the solution should be approached."
+
+We would like to inject the cert and get an indication that the cert was injected  or the error if it was not. Let's explore the different options that we have to inject the cert.
```
abrahamko, line 31: @orenbm I want to suggest another option.
Proposed solution: Run `sh` command instead of `tar`, and in the STDIN pass it a full script that will be executed.

The script will contain the whole content of the certificate, and just write it to the file.
It will also include any error handling, and the `exit` command closes the `sh` process which closes the websocket.
The script could be similar to:
```sh
#!/bin/sh

cd /etc/conjur/ssl

cat > conjur.pem <<EOF
#<SSL CLIENT CERTIFICATE>
EOF

exit
```
Benefits:
1. Short URL. Certificate is in the script. (avoid issues raised from option 2)
1. Deterministic operation. You know exactly when it is done, because the connection is closed once it reaches `exit`.
1. Custom error handling. You can get any error you wish for, handle it properly, and send it back.
1. Versatility. Script can be customized later if needed to handle more issues in this area.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,195 @@
...
+| ------------------------------------------------------------ | --------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| An authn-client container with permissions on the "ssl" directory | An "inject-client-cert" request is sent | - The response code is 200 OK<br />- The cert is in the container |                                                              |
+| An authn-client container without permissions on the "ssl" directory | An "inject-client-cert" request is sent | - The response code is 401<br />- The error is written to the log |                                                              |
+| A service-account that has no sufficient permissions         | An "inject-client-cert" request is sent | - The response code is 401<br />- The error is written to the log | This test is different than the one above because the error type is different. This request will fail today as well while the previous one would have succeeded (although he cert wouldn't be installed). |
+
```
eladkug, line 118: Another test which is relevant for all suggestion is to verify injer-client-cert request is sent to a server which is down/not exist. suppose to return failure with 404 error.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: the former. i updated the doc to say `via a WebSocket to the Kubernetes API server`

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: @sashaCher @rafis3 so do you see a security concern in this option? Do you suggest not to use it or do we have something to further investigate?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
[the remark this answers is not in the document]
orenbm: this will be decided as part of this effort. we may want to return a response to the client to not block it.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
sashaCher: My suggestion is not to use this option due to security concerns.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,235 @@
...
+WebSocket in which we will send another request to verify that the cert was successfully
+injected. We have several options for the verification command:
+1. Run an "ls" command on the cert path. This way we get an immediate feedback on the 
+   cert presence. However, this may return a positive answer even if the cert is not fully
+   injected (TODO: verify this)
+1. Calculate the md5 hash of the cert content before sending it and then sending 
+   a request to get the calculated md5 hash of the cert in the container. Once the
+   two hashes are equal then we know that the cert was fully installed.
+
+#### Send the cert content as part of the URL
```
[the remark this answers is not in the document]
orenbm: i may be missing something but isn't sending data via STDIN the way to send a "body" in a websocket?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
oburstein-hub: Sounds the best option to me either

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
oburstein-hub: Maybe it is possible to pass certificate as a body of POST request instead of URL ? It will not have a security problem then.
@sashaCher - WDYT ? @orenbm - is it possible ?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: @moticless so basically you're saying that we will add a header to the `inject_client_cert` request (e.g `Wait-On-Injection`). If the header is present then we will wait until the injection is done and return 200 OK if succeeded. If the header is not present we will not wait and return 202 Accepted. 

I think that changing the response code upon a header presence is a bit troublesome, did i miss something?

also - i don't think we have enough value in this suggestion to add this complexity. with this suggestion i would just wait for the timeout every time.

Last point - i don't think that this is really a solution suggestion and fits more in the `Inject Client Cert response code` section. wdyt?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: thanks @abrahamko ! adding this to the doc.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: why do we need to verify this? do we do this check for every REST endpoint that we have?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: i ordered the spikes and added the comment `The spikes are ordered by priority. If a spike assesses that the desired solution is feasible then we will not perform the following spikes. `

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: i may be missing something but isn't sending data via STDIN the way to send a "body" in a websocket?

can you elaborate on passing a body in a websocket?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: This option also adds backwards-compatibility issues as an old server will not do anything with this header.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
abrahamko: @oburstein-hub, this connection is using websocket.
This means there's no "body".
The server treats as STDIN any bytes received on the socket after it is opened.
@orenbm suggests passing the certificate in the URL instead of sending it on the socket.

oburstein-hub: I found syntax like this: WebSocket.send("Hello server!");
Instead of "Hello server!" can we send certificate ?

orenbm: yes, this is [what we do today](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/kubectl_exec.rb#L60).

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
abrahamko: This solution does require the container to have the `sh` executable.
I know that in secrets provider and authn k8s we have it.
Still, this requires that the customer must make sure the caller container has the `sh` executable.
@rafis3 , @sashaCher Can we require it from the customer? Is it risky?

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: i like presenting all the options that rose in the process so that the reader knows that they were considered. also - sometimes people have other perspective that adds to the doc and changes the decisions.

I added a sentence describing the order in which we will spike the options.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: thanks @sashaCher and @rafis3 . I added this paragraph to Option 2:
```
We have a security concern with this option. While the traffic is fully encrypted when using https, including the URL, [it's still possible that the URL will be visible in certain locations](https://owasp.org/www-community/vulnerabilities/Information_exposure_through_query_strings_in_url), while the request travels to its destination. It is possible that the Kubernetes API server will not leak any information from the URL, but if the Conjur instance runs outside the cluster, it will surely pass some proxies and load balancers before it reaches Kubernetes.

Thus, we will not use this option due to security concerns.
```

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: yes, option 3 has backwards-compatibility issues. we will not go with it.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
rafis3: @abrahamko since the command is running on our sidecar container, then it could use the `sh` that is already available in that container. You don't need to expect the user to have it their app container.

abrahamko: @rafis3 You're right, but this API is in the server, and so every client using it must be aware that this API requires `sh` executable available in its environment.
Our clients do, but other clients might not have it.

rafis3: That's true, but also true for `tar`. I don't know of anyone that chose to implement the authn client logic by themselves, so I'm relatively optimistic that this is not our biggest concern.

orenbm: i agree with the @rafis3 . the authn-client is a main part of the authentication process and the server side can rely on it.

orenbm:
@Tovli @rafis3 @oburstein-hub @cyberark/palmtree please take another look at the design. 
Changes:
1. Another solution option (thanks @abrahamko ) that i suggest that we go with (ran also [a POC](https://github.com/cyberark/conjur/pull/1789) on it).
2. Removed option 2 due to security concern
3. Updated delivery plan

Thanks :)

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,233 @@
...
+
+This approach has the following benefits:
+1. Short URL. Certificate is in the script. (avoid issues raised from option 2)
+1. Deterministic operation. You know exactly when it is done, because the connection is closed once it reaches exit.
+1. Custom error handling. You can get any error you wish for, handle it properly, and send it back.
+1. Versatility. Script can be customized later if needed to handle more issues in this area.
+1. Simple implementation
+
+This solution is implemented in [this PR](https://github.com/cyberark/conjur/pull/1789).
+
```
izgeri, line 82: I actually don't understand these solutions in relation to the problem

The problem seems to me to be:
- The server sends a 200 to the client once the kube exec command is off, without waiting for a response to see if the command actually succeeds

Could we frame the solutions in relation to the problem? I'd be interested to know, for each potential solution:
- Does the server still return 200 to the client? At which point? If not, what alternate response codes does it send, and in which scenarios?
- Does the server wait to know if the kube exec / cert injection succeeds before returning a response to the client?

I would like to understand better the practical implications of each solution - what impact does it have on the server operations? What impact does it have on the client? It would be great if we could be crystal clear about those pros/cons at this stage, because we'll have to deal with those practical impacts once we get to implementation.

oburstein-hub:
@orenbm - the option proposed by @abrahamko looks good to me. I think we should proceed with it - just verify with @sashaCher that we don't have security concern about it

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
eladkug: Its important to see how we act in case of no response from server side. We dont verify the same for every REST endpoint today.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+This approach has the following benefits:
+1. Short URL. Certificate is in the script. (avoid issues raised from option 2)
+1. Deterministic operation. You know exactly when it is done, because the connection is closed once it reaches exit.
+1. Custom error handling. You can get any error you wish for, handle it properly, and send it back.
+1. Versatility. Script can be customized later if needed to handle more issues in this area.
+1. Simple implementation
+
+This solution is implemented in [this PR](https://github.com/cyberark/conjur/pull/1789).
```
orenbm, line 81: @rafis3 @sashaCher do you see any security concern with option 4?

rafis3: I don't see any security concerns. Does it technically work better than using the old tar method?

orenbm: what do you mean by "better"? It does work for both successful and unsuccessful injections. You can see the logs i shared in [this PR](https://github.com/cyberark/conjur/pull/1789)'s description.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: Well ,the solution is separated into 2, unrelated, parts. The first one is to properly inject the certificate, in a way that we will log the failure if it occurred. The second one is to decide whether we wait for the cert to be injected before returning a response to the "inject_client_cert" request or not. The first part is answered in this section. The second part is answered in the following section - `Inject Client Cert response code`.

So in conclusion, this solution will answer both questions.

izgeri: I was actually suggesting it would be interesting to see how _each_ proposed solution would address the two parts of this problem. I do see you add more detail to that below for the solution that appears to be the winner, but the doc is missing context about how the other solutions fall short.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+This approach has the following benefits:
+1. Short URL. Certificate is in the script. (avoid issues raised from option 2)
+1. Deterministic operation. You know exactly when it is done, because the connection is closed once it reaches exit.
+1. Custom error handling. You can get any error you wish for, handle it properly, and send it back.
+1. Versatility. Script can be customized later if needed to handle more issues in this area.
+1. Simple implementation
+
+This solution is implemented in [this PR](https://github.com/cyberark/conjur/pull/1789).
```
sigalsax, line 105: From meeting:
We need to understand how to link the response of script errors with inject client failure

sigalsax: @oburstein-hub to check how many threads a worker can have. Is it from the same thread pool or does it take from other workers?

@moticless recommends that client sends request "wait on write" so it doesn't chock the server. Server doesn't wait and if didn't get cert with notification in header (ack on write). Must be asynchonrous to avoid overloading on the server

sigalsax: @rafis3 suggests not wait for response (common case). @moticless Act like we do today for the common case (immediately close thread) so that we don't off the bat chock the server. If it fails, we will show a send a new response with an additional header (ack on write) and the server will write the error to log. How should we perform the wait? This is a separate discussion

@Tovli all logs should be in the same template

@nessiLahav what happens if the file exists but the cert is half written. @Tovli says we can check the normal size of the file and compare

sigalsax: Dekel says place a circuit breaker pattern to avoid the accumulation of requests. Stop requests after a certain amount until the other requests are handled

sigalsax: Rafi suggests to Moti's idea but that can be activated through _troubleshooting_ with a flag
@nessiLahav suggests guiding the SDE error to another file to host the file instead of Conjur. Might be worth a POC

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,229 @@
...
+This option is relevant only if the "inject-client-cert" request will return a 202 Accepted response instead of 200 OK (more on that in [this section](#inject-client-cert-response-code).
+
+In this case, we respond with 202 Accepted to the "inject-client-cert" request and start injecting the cert in a separate thread. The server then polls on the state of the authn-client container and waits until its State changes to some  agreed State (e.g "CertInjected"). This state will be set by the authn-client once it sees that the cert is installed.
+
+This option is not fully backwards-compatible and is quite complex so I wouldn't suggest to implement it.
+
+#### Option 4 - Send cert as part of bash script
+
+The issue with the STDIN that is not closing _may_ be an issue of the `tar` command that waits for more input and doesn't close the connection. Another option that we have is to change the command from `tar` to `sh` and send a bash script in the STDIN.
+The script can be something like:
```
orenbm, line 85: can we redirect stderr to the client container

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,229 @@
...
+
+EE: 11-16 days
+
+### Parallelism
+
+[//]: # "Suggest how many team members can work on the solution in parallel"
+
+- 2 people can work on this solution but it's ok to have only 1.
+
+### Spikes and researches
```
orenbm, line 203: add researches:
- Moti's workaround: at first the server doesn't wait for he injection. if the cert is not found in the client we send another request with some header to tell the server to wait (so the error is written to the log)
- Block the server from communicating with the k8s API by some [circuitbox](https://github.com/yammer/circuitbox) or out concurrency limit (from OIDC)

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
oburstein-hub: Number of threads per worker is defined in the environment variable WEB_CONCURRENCy which can be different between conjur installations

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,233 @@
+# Solution Design - Inject Client Cert fix
+[//]: # "Change the title above from "Template" to your design's title"
+
+## Table of Contents
+[//]: # "You can use this tool to generate a TOC - https://ecotrust-canada.github.io/markdown-toc/"
+
+## Useful links
+[//]: # "Add links that may be useful for the reader"
```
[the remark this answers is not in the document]
orenbm: No. they are not visible in the UI.

design/authenticators/authn-k8s/inject_client_cert_solution_design.md
orenbm: i don't think that it is needed as the way that we are injecting the cert is independent to the question if we wait for the cert to be injected before returning the response. 

I re-arranged this section a bit to make it clearer.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+    + [Demo](#demo)
+    
+## Useful links
+[//]: # "Add links that may be useful for the reader"
+
+- [Known issue in k8s exec](https://github.com/kubernetes/kubernetes/issues/89899)
+
+## Background
+[//]: # "Give relevant background for the designed feature. What is the motivation for this solution?"
+We have an issue with the "inject client cert" process where we don't always return a correct response to the client regarding the injection success. The injection is done via a WebSocket to the Kubernetes API server where we use a "tar" command and the cert content in STDIN. There is [an issue](https://github.com/kubernetes/kubernetes/issues/89899) in k8s where messages that are sent via WebSockets to Kubectl Exec are left hanging without returning a response when using STDIN.
```
izgeri, line 43: in particular, we return a 200 OK as long as the request _should_ be processed successfully, without waiting to see if the certification injection is able to be completed

fwiw, I found the original design document and this was the intended flow for `login`:
Screen Shot 2020-09-23 at 9 16 48 PM

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+
+We have a security concern with this option. While the traffic is fully encrypted when using https, including the URL, [it's still possible that the URL will be visible in certain locations](https://owasp.org/www-community/vulnerabilities/Information_exposure_through_query_strings_in_url), while the request travels to its destination. It is possible that the Kubernetes API server will not leak any information from the URL, but if the Conjur instance runs outside the cluster, it will surely pass some proxies and load balancers before it reaches Kubernetes.
+
+Thus, we will not use this option due to security concerns.
+
+#### Option 3 - Client sets its State once cert is injected
+This option is relevant only if the "inject-client-cert" request will return a 202 Accepted response instead of 200 OK (more on that in [this section](#inject-client-cert-response-code).
```
izgeri, line 101:
```suggestion
This option is relevant only if the "inject-client-cert" request will return a 202 Accepted response instead of 200 OK (more on that in [this section](#inject-client-cert-response-code)).
```

izgeri: also, has this been updated since the earlier section was updated? I felt when reading to this point that this was already something you were setting out to do

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+```
+#!/bin/sh
+
+cat > /etc/conjur/ssl/client.pem <<EOF
+#{cert-content}
+EOF
+
+chmod 644 /etc/conjur/ssl/client.pem
+exit
+```
```
izgeri, line 121: should this be updated to what you shared today?
```
#!/bin/sh
copy_file() {
  cat > #{path} <<EOF
#{content}
EOF
  chmod #{mode} #{path}
}
copy_file 2>&1 | tee -a #{COPY_FILE_LOG_FILE}
exit
```

also, this looks to be connected to the open PR #1802 - I'd like to make sure we factor in that the cert may be injected at a configurable path in the near future (cc @JakeQuilty)

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+1. Short URL. Certificate is in the script. (avoid issues raised from option 2)
+1. Deterministic operation. You know exactly when it is done, because the connection is closed once it reaches exit.
+1. Custom error handling. You can get any error you wish for, handle it properly, and send it back.
+1. Versatility. Script can be customized later if needed to handle more issues in this area.
+1. Simple implementation
+
+This solution is implemented in [this PR](https://github.com/cyberark/conjur/pull/1789).
+
+#### Conclusion
+
```
izgeri, line 133: there's really no pros/cons listed for option 1 - is there anything more to say about that one?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+   1. The answer is yes and thus we are not going with option 2.
+
+## Affected Components
+
+[//]: # "Address the components that will be affected by your solution [Conjur, DAP, clients, integrations, etc.]"
+- Conjur
+- DAP
+- conjur-authn-k8s-client
+- seed-fetcher
+- secrets-provider-for-k8s
```
izgeri, line 171: if secrets provider is impacted, isn't secretless also impacted?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+Option 3 is off the table due to redundant complexity and backwards-compatibiliy issues.
+
+Option 4 has the most benefits of the options and is better than option 1.
+
+<u>Decision</u>: Option 4
+
+### Design
+[//]: # "Add any diagrams, charts and explanations about the design aspect of the solution. Elaborate also about the expected user experience for the feature"
+The authentication process is depicted in the following diagram:
+![Authn-k8s authenti](authn-k8s-process.png)
```
izgeri, line 145: in the process diagram, it seems to show that the server validates the request, kicks off the cert injection, and closes the socket before sending the 202 response. this isn't clear from the text of the doc, however - can a note about this be added?

also, is it possible to use plantuml and include the source for this process diagram in this PR? we could perhaps create a `design/diagrams` folder in this project, and put both the .txt source and the .png diagram in that folder. for an example, please see the [Puppet diagrams folder](https://github.com/cyberark/conjur-puppet/tree/master/diagrams)

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+### Performance
+[//]: # "Elaborate on whether your solution will affect the product's performance, and how"
+The performance of the "inject-client-cert" request will not be affected as the Websocket is still being closed immediately after we inject the cert.
+
+### Security
+
+[//]: # "Are there any security issues with your solution? Even if you mentioned them somewhere in the doc it may be convenient for the security architect review to have them centralized here"
+
+1. Is it ok to pass the base64 encoded cert in the URL? Is it less secure than sending it in the STDIN?
+   1. The answer is yes and thus we are not going with option 2.
```
izgeri, line 162: it'd be great to have a security review of this solution design too - I know there's not a whole lot to it, but it's a good opportunity to look at the authn-k8s client cert injection process with fresh eyes. can @sashaCher or @andytinkham help with this?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+
+| **Given**                                                    | **When**                                | **Then**                                                     | **Comment**                                                  |
+| ------------------------------------------------------------ | --------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| An authn-client container with permissions on the "ssl" directory | An "inject-client-cert" request is sent | - The response code is 200 OK<br />- The cert is in the container |                                                              |
+| An authn-client container without permissions on the "ssl" directory | An "inject-client-cert" request is sent | - The response code is 401<br />- The error is written to the log |                                                              |
```
izgeri, line 186: how would the server determine the response should be 401? is this something that's already implemented? if this is new, it would be good to talk about this a bit more elsewhere

is 401 even the right response in this case, when a valid client sends a request but the directory in the client container doesn't exist or doesn't have the right permissions? the status code definition for 401 says
> specifically for use when authentication is required and has failed or has not yet been provided.

I don't think that matches this case. I'm not sure what the right response would be - maybe 403 Forbidden (`The request contained valid data and was understood by the server, but the server is refusing action`)?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+| A service-account that has no sufficient permissions         | An "inject-client-cert" request is sent | - The response code is 401<br />- The error is written to the log | This test is different than the one above because the error type is different. This request will fail today as well while the previous one would have succeeded (although he cert wouldn't be installed). |
+
+## Logs
+[//]: # "Fill in the table below to depict the log messages that can enhance the supportability of your solution"
+[//]: # "You can use this tool to generate a table - https://www.tablesgenerator.com/markdown_tables#"
+
+| **Scenario**                                             | **Log message**                                              |
+| -------------------------------------------------------- | ------------------------------------------------------------ |
+| Timeout reached for injection kubectl exec command       | Command timed out after {0} seconds in container '{1}' of pod '{2}' |
```
izgeri, line 195: is this actually going to be added in the implementation of this design? I thought the chosen design wouldn't have the server wait for the command to time out?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+| ------------------------------------------------------------ | --------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| An authn-client container with permissions on the "ssl" directory | An "inject-client-cert" request is sent | - The response code is 200 OK<br />- The cert is in the container |                                                              |
+| An authn-client container without permissions on the "ssl" directory | An "inject-client-cert" request is sent | - The response code is 401<br />- The error is written to the log |                                                              |
+| A service-account that has no sufficient permissions         | An "inject-client-cert" request is sent | - The response code is 401<br />- The error is written to the log | This test is different than the one above because the error type is different. This request will fail today as well while the previous one would have succeeded (although he cert wouldn't be installed). |
+
```
izgeri, line 188: do we have a good sense of the errors that will happen on the server side when the `kubectl exec` command is run after it's changed to a bash script?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+[//]: # "Suggest how many team members can work on the solution in parallel"
+
+- 2 people can work on this solution but it's ok to have only 1.
+
+### Spikes and researches
+
+[//]: # "Note and estimate the tasks which are related to spikes and researches which needed for the solution"
+
+This spike will be done only if we decide to go with option 1 instead of option 4.
```
izgeri, line 228: is this still a possibility?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+- Print error to authn-client log
+  - Do we want to delete the log file after we log its content?
+  - EE: 2 days
+
+### Testing 
+
+[//]: # "Note and estimate the tasks which are related to writing/performing tests [automation, manual tests, etc.]"
+
+- Write tests according to the test plan
+  - EE: 4 days
```
izgeri, line 249: do you have a clear sense of the tests you'll be adding? are there unit/rspec tests we can add, or are you planning primarily to add cucumber or e2e tests?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+[//]: # "Note and estimate the tasks which are related to spikes and researches which needed for the solution"
+
+This spike will be done only if we decide to go with option 1 instead of option 4.
+
+- What happens if we send an "ls" command on a file that is being populated by a "tar" command? Will we get a response that the file exist although it is not fully populated?
+  - EE:  1 day
+  
+### Implementation 
+[//]: # "Note and estimate the tasks which are related to the solution's code implementation"
+
```
izgeri, line 235: I'd suggest highlighting which projects each of the items below should be implemented in - eg preface each line with repo name (eg `cyberark/conjur: `)

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+- Implement solution according to the design
+  - EE: 5-10 days (depending on solution - 4 will be faster)
+- Fix issue - [Kubectl exec Websocket logs do not have tags](https://github.com/cyberark/conjur/issues/1840)
+  - EE: 2 days
+- Print error to authn-client log
```
izgeri, line 240: is there any more detail for what changes are needed in the authn-k8s client?

I _think_ you'll need to implement some checking of the cert injection errors, logging any cert injection errors that are detected, and ensure the cert injection log is deleted after the cert is read

you might also want to add a task (maybe a spike?) for determining what the success / failure modes of the bash script look like, so it's clear what the output will look like in all cases before you implement the functionality in the authn-k8s client

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+  - Do we want to delete the log file after we log its content?
+  - EE: 2 days
+
```
izgeri, line 243: should it also be highlighted here that changes will be needed in other downstream projects to benefit from this change? we'll probably need to bump the versions of the authn-k8s client in secrets provider, seed fetcher, and secretless, and cut new releases of those projects. that wouldn't necessarily be fully on the team implementing this solution, but it is work that needs to be done and should be called out here.

izgeri (changes requested):
I read through the proposal and it's made some really nice progress - I left quite a few comments, but most of them are pretty minor. I'm looking forward to seeing this progress, and we'll be glad to help propagate these changes e.g. to Secretless, especially as I expect these will also improve the resilience and stability of that project. Great work!

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
+# Solution Design - Inject Client Cert fix
+[//]: # "Change the title above from 'Template' to your design's title"
```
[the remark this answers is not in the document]
orenbm: I'm not dropping these as they aren't visible in the UI. These comments can help the design writer and are no harm so i don't expect every writer to also remove them.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: yes, that's correct. we may fail and not return 200 OK in cases such as Conjur host validation error or handshake error with the k8s api

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: this wasn't updated but i left this as an option so people know it was considered. I'll add a tldr in the beginning of the section that option 4 was chosen in case someone wants to skip.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: yes, thanks. i updated the script.
I saw @JakeQuilty 's work and it looks like they can both live together. there will be some merge conflicts so whoever finishes first can help the other one with the merge. @JakeQuilty - let's be in contact once one of us merges their work so we can perform the merge together if needed.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: Not really. it is just less convenient than option 4.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: that's a good question. i assumed it is not affected because it doesn't use the authn-client code (as far as i know). am i wrong?

orenbm: coming to think of it, it will be affected anyway because the conjur server will change. but first can you tell me if it uses the authn-client?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: this is described in [this section](https://github.com/cyberark/conjur/blob/kubectl-exec-solution-design/design/authenticators/authn_k8s/inject_client_cert_solution_design.md#waiting-for-the-injection-to-finish).

The source for the diagram is not accesible as it's done in Lucidchart under my license. i'll consider using `plantuml` for this.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: you are correct, it's a leftover. it should be 202 in this case.

orenbm: i updated the test plan.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: it is not added, just modified. I added the number of seconds that we waited

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: the server side will not have the errors as the socket is closed. the only errors that will be present in the server are handshake errors (same as today)

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: nope. i'll remove it.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: i'm not adding any integration tests, just fixing them to verify 202 instead of 200.
I am not sure that we need any new UTs but i'll check that. it is part of the 4 days.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: i don't think we need to add anything there apart from logging the error and deleting the file.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: that's a good point, thanks

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
orenbm: you are correct, adding it.

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+  * [Testing](#testing)
+  * [Demo](#demo)
+    
+## Useful links
+[//]: # "Add links that may be useful for the reader"
+
+- [Known issue in k8s exec](https://github.com/kubernetes/kubernetes/issues/89899)
+
+## Background
+[//]: # "Give relevant background for the designed feature. What is the motivation for this solution?"
```
sigalsax, line 41: From meeting:
@oburstein-hub wanted clarification on error types. Clarify server-side and client-side errors for communication conjur server -> client
1. conjur server -> k8s api
2. k8s api -> authn client

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+#### Option 4 - Send cert as part of bash script
+
+The issue with the STDIN that is not closing is an issue of the `tar` command that waits for more input and doesn't close the connection. Another option that we have is to change the command from `tar` to `sh` and send a bash script in the STDIN.
+The script can be something like:
+
+```
+#!/bin/sh
+
+copy_file() {
+  cat > #{path} <<EOF
```
orenbm, line 117: add temp file so `ls` works always

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Cross-team dependencies
+[//]: # "Elaborate about your solution's requirements or needs from other teams"
+We would like to get a review on the solution from @cyberark/conjur-core-team  and @cyberark/community-and-integrations-team. There are no requirements from  the Infra team.
+
+## Test Plan
+[//]: # "Fill in the table below to depict the tests that should run to validate your solution"
+[//]: # "You can use this tool to generate a table - https://www.tablesgenerator.com/markdown_tables#"
+
+The current integration tests of the "inject-client-cert" request will be sufficient for this feature. we will just change them so they verify the response is 202 instead of 200.
```
orenbm, line 187: what about tests in authn-client?

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Logs
+[//]: # "Fill in the table below to depict the log messages that can enhance the supportability of your solution"
+[//]: # "You can use this tool to generate a table - https://www.tablesgenerator.com/markdown_tables#"
+
+| **Scenario**                                             | **Log message**                                              |
```
orenbm, line 193: add comments on what changed

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+| -------------------------------------------------------- | ------------------------------------------------------------ |
+| Timeout reached for injection kubectl exec command       | Command timed out after {0} seconds in container '{1}' of pod '{2}' |
```
sigalsax, line 195: add log level (:

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+| Before copying the file to the container                 | Copying file to {0-container-name}:{1-cert-file-path} in {2-pod-namespace}/{3-pod-name}. stdout and stderr will be written to {4-log-file-path}" |
```
sigalsax, line 196: from meeting:
might delete log path so it might not be relevant

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+- conjur-authn-k8s-client
+- seed-fetcher
+- secrets-provider-for-k8s
+- secretless-broker
+
+## Cross-team dependencies
+[//]: # "Elaborate about your solution's requirements or needs from other teams"
+We would like to get a review on the solution from @cyberark/conjur-core-team  and @cyberark/community-and-integrations-team. There are no requirements from  the Infra team.
+
+## Test Plan
```
sigalsax, line 183: From meeting:

@eladkug recommended to think about tests in client/other repos

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
sigalsax, line 187: Moti, suggested we invest in performance tests because we don't really know the numbers and don't want to worsen the performance experience with the changes in the PR

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Implementation 
+[//]: # "Note and estimate the tasks which are related to the solution's code implementation"
+
+- `cyberark/conjur`: Implement solution according to the design
+  - EE: 7 days
+- `cyberark/conjur`: Fix issue - [Kubectl exec Websocket logs do not have tags](https://github.com/cyberark/conjur/issues/1840)
+  - EE: 2 days
+- `cyberark/conjur-authn-k8s-client`: Print error to authn-client log
+  - Do we want to delete the log file after we log its content?
```
sigalsax, line 232: From meeting:
Is there something sensitive in errors?

nessiLahav (dismissed):

design/authenticators/authn_k8s/inject_client_cert_solution_design.md
sashaCher: @izgeri thank you, I'm following the PR.
You are right regarding fresh eyes and there's some background job around it, but it does not related to the bug fix.

Tovli (approved):
LGTM
Nice work !!

abrahamko:
@orenbm I know this is already been merged but something in the the integration tests of this design didn't look good enough to me and I didn't know why exactly.
Now I know; the tests do not cover in any way the `exec` part, which is a major part of this API.
This part is so delicate and can easily break for so many reasons, that I think it must be tested by integration tests.
We can test it manually, but I think this is bad practice.
Alternatively, this can be tested in `secrets-provider`, but AFAIK we don't do it on pre-released versions of Conjur.
Do you feel comfortable delivering it without testing this delicate part in integration tests?
@eladkug @Tovli @izgeri FYI.

orenbm:
> @orenbm I know this is already been merged but something in the the integration tests of this design didn't look good enough to me and I didn't know why exactly.
> Now I know; the tests do not cover in any way the `exec` part, which is a major part of this API.
> This part is so delicate and can easily break for so many reasons, that I think it must be tested by integration tests.
> We can test it manually, but I think this is bad practice.
> Alternatively, this can be tested in `secrets-provider`, but AFAIK we don't do it on pre-released versions of Conjur.
> Do you feel comfortable delivering it without testing this delicate part in integration tests?
> @eladkug @Tovli @izgeri FYI.

@abrahamko integration tests should test use-cases, not code. Is there a use-case that you think is missing from the integration tests in Conjur? As we don't change the behaviour, and the Conjur server will not fail a request that that will fail in the kubectl exec, then we can test it in an integration test.

It _can_ be tested in the secrets-provider (ideally in the client) because it will be impacted by the change but i don't see how we can test such cases in the server.

abrahamko:
@orenbm This API should inject certificate to the client, so we should test that it injects the certificate to the client.
Other implementations of 202 response code reply with a callback endpoint to check on the status of this request.
Since we do not do it, what do we expect this API to do?
From my understanding, this API should inject the certificate to the client in some later time, and write the error to a log in case there is any.
If it fails to do it, this API doesn't work.
This is not testing of code, this is behavior testing, but it is missing.

orenbm:
> @orenbm This API should inject certificate to the client, so we should test that it injects the certificate to the client.
> Other implementations of 202 response code reply with a callback endpoint to check on the status of this request.
> Since we do not do it, what do we expect this API to do?
> From my understanding, this API should inject the certificate to the client at some later time, and write the error to a log in case there is any.
> If it fails to do it, this API doesn't work.
> This is not testing of code, this is behavior testing, but it is missing.

Yes, the integration tests verify that the cert is injected to the container.

abrahamko:
Thanks, @orenbm , I see it now.
I also see that you added to the test to validate the error log in the client, so this covers anything I had in mind.
Good Job!
Sorry if I was nitpicking at you, I didn't mean to. 😃 
Thanks for your patience.

orenbm:
Sure @abrahamko, nitpick as much as you can :)

Issue #1812, opened 2020-09-08
GCP authenticator 2nd phase - Support Google cloud functions in dev script


nessiLahav:
Add support to `conjur/dev/start` script to be able to fetch cloud functions tokens in order to debug GCP authenticator in Conjur

**DoD**
- [x] We are able to debug GCP authenticator with cloud function use case

PR #1835, opened 2020-09-22
Add support of GCF to `dev/start` script

nessiLahav:
### What does this PR do?
Add support of generating google cloud function tokens as part of `dev/start` script, 
in order to be able to run the automation tests in a local environmnet 

Connected to #1812

### DIFF
--- a/dev/cli
+++ b/dev/cli
@@ -36,14 +36,89 @@ EOF
 exit
 }
 
-function enable_gcp() {
-  local gce_instance_name="$1"
+function print_gcp_help() {
+  cat << EOF
+NAME
+    --authn-gcp - Enables GCP features, in order to run local tests.
 
-  if [ -z "$gce_instance_name" ]; then
-    echo "--authn-gcp mode requires a GCE instance name argument."
-    echo "Usage ./cli exec --authn-gcp [GCE_INSTANCE_NAME]"
+SYNOPSIS
+    --authn-gcp [command options] [arguments...]
+GLOBAL OPTIONS
+    --help                                    - Show this message
+COMMANDS
+    -gce     Google compute engine instance name
+    -gcf     Google cloud function URL (Predefine ci/authn-gcp/function/main.py code as GCF)
+USAGE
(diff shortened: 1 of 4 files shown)


dev/cli
```diff
@@ -83,6 +161,37 @@ _get_gcp_token_payload() {
...
+  local audience="dummy_host"
+  local status_code=$(curl -o /dev/null -s -w "%{http_code}\n" \
+    -H 'Metadata-Flavor: Google' "$gcp_func_url?audience=$audience")
+
+  if [ ! "$status_code" = "200" ]; then
+    echo "Failed to access function, HTTP Status: $status_code"
+    exit 1
+  fi
+}
+
```
eranha, line 194: source the original script and call the function.

dev/cli
```diff
@@ -36,15 +36,93 @@ EOF
...
+  local option_arg=$2
+  if [ -z "$option_arg" ]; then
+    echo "argument is missing for $option option"
+    exit 1
+  fi
+}
+
+function print_gcp_option_error() {
+  local option=$1
+  if [ -z "$option" ]; then echo "required command options are missing";
```
eranha, line 101: Could you break the if?
```
if [ ]; then
else
fi
```

eranha (changes requested):

dev/cli
```diff
@@ -83,6 +161,37 @@ _get_gcp_token_payload() {
...
+  local gcp_func_url=$1
+  local token_prefix="tokens/gcf_"
+
+  validate_gcf_url_accessible $gcp_func_url
+
+  cd ../ci/authn-gcp/
+  sh ./get_tokens_to_files.sh "$gcp_func_url" "$token_prefix" || exit 1
+  cd -
+
+  echo "Finished obtaining and writing tokens to file"
```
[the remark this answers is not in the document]
nessiLahav: Yes, for development env is good enough

dev/cli
nessiLahav: source is actually running the script, i extracted the logic to another script

eranha: because there's  a call to a function at the bottom of the script.

eranha (approved):

PR #1836, opened 2020-09-22
Align subtitles with Typora

orenbm:
While using Typora to edit MD files, it changes subtitles from
`[//]: # (<subtitle>)` to `[//]: # "<subtitle>"`.

This commit aligns the subtitles in the solution design template so that
it won't change each time we use it.

For those who don't use Typora to edit MD file, there won't be any difference. Furthermore,
the outcome will still be the same and the subtitles are not present in the UI. You can see it here:
`master` - https://github.com/cyberark/conjur/blob/master/design/templates/solution_design.md
`typora-subtitle` - https://github.com/cyberark/conjur/blob/typora-subtitle/design/templates/solution_design.md

### DIFF
--- a/design/templates/solution_design.md
+++ b/design/templates/solution_design.md
@@ -1,12 +1,12 @@
 # Solution Design - Template
-[//]: # (Change the title above from "Template" to your design's title) 
+[//]: # "Change the title above from 'Template' to your design's title"
 
 ## Table of Contents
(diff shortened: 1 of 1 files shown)


sigalsax (dismissed):
LGTM

sigalsax (approved):
lgtm!

PR #1847, opened 2020-09-27
Remove contents of ssl dir but not dir

orenbm:
Before each cucumber test of authn-k8s, we delete the cert file
from `/etc/conjur/ssl` so that each test starts with a fresh folder.

However, the current implementation deletes the directory itself instead
of deleting only its content. The directory needs to be present as its permissions
are defined upon its initial creation.


sigalsax (approved):
LGTM

Issue #1855, opened 2020-10-01
Rename `kubectl` to `kubeClient` across the code


abrahamko:
Currently our code mentions many times the `kubectl` as a k8s client.
Although it is correct that `kubectl` **is** a k8s client, this is not the client we use in our code.
[kubectl](https://kubernetes.io/docs/reference/kubectl/kubectl/) is a k8s client used as a command-line tool and written in `go`.
We connect to k8s using a ruby gem called [kubeclient](https://github.com/abonas/kubeclient), which is a privately developed client that is unrelated to `kubectl`.

To prevent misunderstanding, we better replace all occurrences of `kubectl` with `kubeClient`.

orenbm:
Thanks @abrahamko .

Just a small correction - the occurrences of `kubectl` in our code does does use the ruby gem (in these cases we actually call the object `kube_client`). In these cases we use a websocket to communicate with the K8s API.

How would you distinguish between the two?

abrahamko:
> Thanks @abrahamko .
> 
> Just a small correction - the occurrences of `kubectl` in our code does does use the ruby gem (in these cases we actually call the object `kube_client`). In these cases we use a websocket to communicate with the K8s API.
> 
> How would you distinguish between the two?

So you actually don't use a client, so you can drop the `client`.
I would call it `k8s_exec` or `kube_exec`.
The fact that it uses websocket is an implementation detail, so it souldn't be in the class name.

PR #1856, opened 2020-10-01
Refactor KubectlExec to KubeExec

orenbm:
Connected to #1855

This class performs exec commands to the K8s API but it is not
using kubectl, which is the cli client of the K8s API.
Thus, the name is misleading and we should refactor it.

### DIFF
--- a/app/domain/authentication/authn_k8s/kube_exec.rb
+++ b/app/domain/authentication/authn_k8s/kube_exec.rb
@@ -11,7 +11,7 @@ require 'websocket-client-simple'
 module Authentication
   module AuthnK8s
 
-    KubectlExec ||= CommandClass.new(
+    KubeExec ||= CommandClass.new(
       dependencies: {
         env:    ENV,
         logger: Rails.logger
@@ -28,7 +28,7 @@ module Authentication
       extend Forwardable
       def_delegators :@k8s_object_lookup, :kube_client
 
-      DEFAULT_KUBECTL_EXEC_COMMAND_TIMEOUT = 5
+      DEFAULT_KUBE_EXEC_COMMAND_TIMEOUT = 5
 
       def call
         @message_log = MessageLog.new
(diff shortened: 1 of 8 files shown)


app/domain/authentication/authn_k8s/kube_exec.rb
```diff
@@ -126,12 +126,12 @@ def on_error(err)
       private
 
       def add_websocket_event_handlers(ws_client, body, stdin)
-        kubectl = self
+        kube = self
```
abrahamko, line 129: Why is this needed? this looks weird.
Can't you just call `on_open`,  `on_message` etc.?

spec/app/domain/authentication/authn_k8s/inject_client_cert_spec.rb
```diff
@@ -215,15 +215,15 @@
     end
 
     context "when cert is being installed" do
-      let(:kubectl_exec_response) { double("MockKubeCtlExecResponse") }
+      let(:kube_exec_response) { double("MockKubeCtlExecResponse") }
```
abrahamko, line 218: leftover... :)

abrahamko (approved):
Looks good!
Just two small things and you're good to go.

app/domain/authentication/authn_k8s/kube_exec.rb
abrahamko: I mean why this line is needed, not the change.
Why you need to set `self` to a var, that's my question.

orenbm: no as the brackets open a new context

orenbm: well, this was here before me but i guess we can use `self` instead. is this what you'd like?

abrahamko: fine

abrahamko (approved):

abrahamko (approved):

PR #1857, opened 2020-10-01
Align KubeExec with CommandClass style format

orenbm:
Align KubeExec with CommandClass style format by doing the following:
- Rename class to the action it performs
- Use dependencies so we can easily mock them
- call function tells the class's story

In addition we now handle WebSocket errors in this class instead of letting the caller
handle them.


orenbm:
> is improved unit testing something you'll do in a follow-on effort? I would LOVE to see more UTs for the baseline k8s authn functionality

Yeah, i'm working on it now but i have something that i don't understand. I created the PR so i can send it to @jonahx to assist.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -11,10 +11,13 @@
 module Authentication
   module AuthnK8s
 
-    KubeExec ||= CommandClass.new(
+    ExecuteCommandInContainer ||= CommandClass.new(
```
jonahx, line 14: Side note that is not part of this PR:
```rb
extend Forwardable
def_delegators :@k8s_object_lookup, :kube_client
```
We should remove this, since it all add another public method to the `CommandClass`, which is supposed to have only a `call` method.  Instead, we should either define `kube_client` privately, or simply use the long form `@k8s_object_lookup.kube_client` (there are only 2 references to `kube_client` so this is fine).

jonahx: Another existing code comment:

1. We should be able to remove all arguments from `kube.on_open(ws_client, body, stdin)`, because the kube instance already has access to all of them.
2. We should be able to remove `ws_client` arg from `kube.on_message(msg, ws_client)` for the same reason.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -11,10 +11,13 @@
...
       dependencies: {
-        env:    ENV,
-        logger: Rails.logger
+        env:               ENV,
+        websocket_client:  WebSocket::Client::Simple,
+        message_log_class: MessageLog,
+        websocket_message: WebSocketMessage,
```
jonahx, line 19: `WebSocketMessage` shouldn't  be a dependecy, since it's a value object.

jonahx: `MessageLog` is a bit trickier, because I made what I now consider a mistake in its design.  As it is currently, you do need to inject it because it has mutation and raises errors.  However, it should have been designed with an immutable interface (returning copies of itself when altered) and the error raising code should be moved into this class.  

This will make it a simpler value object, and will simplify testing of this class by reducing the number of doubles by one.

I will try to fix this class for you today on a separate branch you can pull in.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -25,43 +28,30 @@ module AuthnK8s
...
       def on_open(ws_client, body, stdin)
         hs       = ws_client.handshake
         hs_error = hs.error
 
         if hs_error
-          ws_client.emit(:error, "Websocket handshake error: #{hs_error.inspect}")
+          ws_client.emit(
+            :error,
+            Errors::Authentication::AuthnK8s::WebSocketHandshakeError.new(hs_error.inspect)
+          )
```
jonahx, line 47: Is this correct?  It looks like we're sending the error back through the websocket connection rather than raising it ourselves.  Is this intentional?  If so, can you explain what's happening?

Also I was confused because `emit` didn't seem to be documented for this websocket client:

https://rdoc.info/github/imanel/websocket-eventmachine-client/master/frames#simple-client-example

jonahx (commented):
To answer your main question,  the basic idea is to create a mock/double that allows you to control the events that are happening from the test code, and then make assertions on things that you expect to happen (messages sent, errors raised, and log messages written).

For example:

- You initialize the mock with a handshake error - an error is raised.
- If the command class is called with stdin, that results in the expected received_messages in
the mock.
- When you trigger a message event, that results in expected log method calls.
- When you trigger a close event, that results in expected log method calls.
- etc

Just go through all the code in your command class to create a comprehensive list.

I've written the custom mock/double (it serves both roles) for you.  This is untested, but basically what you'll need and well commented to explain the ideas.  Lmk if you have any questions.

```ruby
      class WsClientMock
        attr_accessor :received_messages

        def initialize(
          handshake_error: nil
        )
          @handshake_error = handshake_error

          # State saved and managed by the mock
          @received_messages = []
          @registered_events = {}
        end

        def connect(server_url, headers:)
          @connect_args = [server_url, headers]

          # In the production class, "connect" is defined on Client::Simple, and
          # calling "connect" returns an instance of Client::Simple::Client. For
          # the mock, there's not reason to make that distinction, and we can
          # use the same object to play both roles.
          self
        end

        # "on" is used by the implementation code, not the test code.
        def on(event, &blk)
          @registered_events[event] = blk
        end

        # The "trigger_X" methods are not part of the public API we're mocking.
        # They're used only by the test code, not the implementation code.
        def trigger_open
          @registered_events[:open].call
        end

        def trigger_close
          @registered_events[:close].call
        end

        # Note: The test code will need to construct the messages it's sending
        # in the correct format.  See MessageLog code for help with that.
        def trigger_message(msg)
          @registered_events[:message].call(msg)
        end

        def trigger_error(err)
          @registered_events[:error].call(err)
        end

        # "send" saves received messages to assert on later.  "send" is used by
        # implementation code.  "received_messages" is used by test code.
        def send(msg, type: nil)
          @received_messages << [msg, type]
        end
      end
```

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -119,11 +109,41 @@ def on_error(err)
...
+      def add_websocket_event_handlers(ws_client)
         # We need to set this so the handlers will call this class's methods.
         # If we use 'self' inside the curly brackets it will be try to use methods
-        # of the class WebSocket::Client::Simple::Client
-        kube = self
+        # of the class WebSocket::Client::Simple::Client.
+        # We need to have method params for the same reason.
+        kube  = self
+        body  = @body
+        stdin = @stdin
```
jonahx, line 146: It's not necessary to make do `kube = self` and make the `on_XXX` methods of this CommandClass public.  It can and should have the single public `call` method like all CommandClass's.  Ruby can create a closure over a private method with no issues:

[Demo of closing over a private method](https://tio.run/##VZBBCoMwEEX3nmJIoejGAxQKQi/QRfclJmMbGqMk0dqKZ09johY3YWb@48@f6K78OMckNQYuUqCyCQDHChp1x9636bGUr8wPAQpfwRn861tUPFlQRqXcgHzpZj0w0fuGZnUWSlhBpfhiysLGxX3w3sQ0NUKrRU8tAqeWkiBGMF9Dwbgy9xrts@Ew/SMtyppuDwa3trMGyFULZYV6QKWbetsZsRMcxmKYyO4Qny9@Ua7wncwXzUXKsoSFq537AQ "Ruby – Try It Online")

jonahx: I have added this fix to my branch.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
jonahx: I have added this fix to my branch.

jonahx (commented):
I made a number of changes on the branch `jonah/kube-exec-improvements` today, implementing my own comments as well as adding some new changes.

I have left the completion of the testing work to you though :)

app/domain/authentication/authn_k8s/execute_command_in_container.rb
orenbm: after this change we get: `undefined method `on_error' for #<WebSocket::Client::Simple::Client:0x00005589935f0060> (NoMethodError)`

app/domain/authentication/authn_k8s/execute_command_in_container.rb
orenbm: i added this comment:
```
          # Add the error to the stream instead of raising it so we continue to
          # receive other messages. At the end of this class run we will verify
          # the stream to see if any errors were written
```

app/domain/authentication/authn_k8s/web_socket_client_event_handler.rb
```diff
@@ -0,0 +1,109 @@
...
+      end
+
+      def on_open
+        hs       = @ws_client.handshake
+        hs_error = hs.error
+
+        if hs_error
+          # Add the error to the stream instead of raising it so we continue to
+          # receive other messages. At the end of this class run we will verify
+          # the stream to see if any errors were written
```
jonahx, line 38: I would think an error during the handshake is a fatal error.  What's the reason it's not?

app/domain/authentication/authn_k8s/web_socket_client_event_handler.rb
```diff
@@ -0,0 +1,109 @@
...
+        # stdin was provided. We send it to the client.
+
+        data = WebSocketMessage.channel_byte('stdin') + @body
+        @ws_client.send(data)
+
+        # We close the socket and don't wait for the other side to close it
+        # so that we can finish handling the request quickly and don't leave the
+        # Conjur server hanging. If an error occurred it will be written to
+        # the client container logs.
+        @ws_client.send(nil, type: :close)
```
jonahx, line 63: I'm confused by this comment.  It looks like this logic always runs, so we're sending data on line 57 and then immediately the server is hanging and we send a close?  I'm sure I'm missing something but some clarification would be helpful.

ci/authn-k8s/dev/Dockerfile.inventory_no_permissions
```diff
@@ -0,0 +1,10 @@
+FROM ruby:2.3
```
jonahx, line 1: Is there a reason we're using such an old version?  Conjur uses 2.5 and the current stable version is 2.7 I believe.

spec/app/domain/authentication/authn_k8s/copy_text_to_file_in_container_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+                    "cleanup() {\n" \
+                    "  rm -f \"path/to/file.tmp\"\n" \
+                    "}\n" \
+                    "trap cleanup EXIT\n\n" \
+                    "set_file_content() {\n" \
+                    "  cat > \"path/to/file.tmp\" <<EOF\nContent\nEOF\n\n" \
+                    "  chmod \"Mode\" \"path/to/file.tmp\"\n" \
+                    "  mv \"path/to/file.tmp\" \"path/to/file\"\n" \
+                    "}\n\n" \
+                    "set_file_content > \"${TMPDIR:-/tmp}/conjur_set_file_content.log\" 2>&1\n"
```
jonahx, line 62: Use squiggly here doc:

```rb
expected_body = <<~BODY
...
BODY
```

app/domain/authentication/authn_k8s/web_socket_client_event_handler.rb
orenbm: what do you mean by fatal? from what i know, fatal error should cause the server to crash which is not the case here. In the current implementation, the authentication request will responds with 401. Is this ok?

app/domain/authentication/authn_k8s/web_socket_client_event_handler.rb
orenbm: yes, you described exactly what happens and it is explained in the comment. what are you missing? i can enhance the comment according to your misunderstanding.

ci/authn-k8s/dev/Dockerfile.inventory_no_permissions
orenbm: ummm not that i know of, i just took the existing Dockerfile

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
```diff
@@ -42,29 +42,28 @@ def copy_text_to_file_in_container
       #
       # We redirect the output to a log file on the authn-client container
       # that will be written in its logs for supportability.
-      def set_file_content_script(path, content, mode)
+      def bash_script(path, content, mode)
```
eranha, line 45: I agree with @Tovli that the script should be in external file.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -0,0 +1,141 @@
...
+        raise Errors::Authentication::AuthnK8s::ExecCommandError.new(
+          @container,
+          @pod_name,
+          websocket_error(error_stream)
+        )
+      end
+
+      def server_url
+        api_uri  = kube_client.api_endpoint
+        base_url = "wss://#{api_uri.host}:#{api_uri.port}"
```
eranha, line 104: Take the values from configuration?

app/domain/authentication/authn_k8s/web_socket_client_event_handler.rb
```diff
@@ -0,0 +1,109 @@
...
+        @validate_message = validate_message
+        @message_log_class = message_log_class
+        @logger = logger
+
+        @message_log = @message_log_class.new
+        @channel_closed = false
+      end
+
+      def on_open
+        hs       = @ws_client.handshake
```
eranha, line 32: I would use a clear var name `handshake`

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
orenbm: i disagree on this. reading a script from an external file adds redundant complexity. Having it in the same file has better readability

app/domain/authentication/authn_k8s/execute_command_in_container.rb
orenbm: why not from the kubeclient? why add more configuration?

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
jonahx: A key question: Is the script intended to be used any place _other_ than in the ruby file?  If the answer is no, keeping it here seems convenient, since it avoids some boilerplate around reading the file from disk and then doing the cert interpolation on that read-in string.

You lose syntax highlighting when reading it, but since the script is so short and simple this seems like a reasonable price to pay.

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
```diff
@@ -42,29 +42,28 @@ def copy_text_to_file_in_container
...
         tmp_cert = "#{path}.tmp"
+        <<~BASH_SCRIPT
```
jonahx, line 47: We should use `mktemp` as a thread safe and standard way to make the temp file (avoiding possible collisions, eg)

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
orenbm: it is only used as part of this ruby file. it is a key part of this class and this class only.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -0,0 +1,141 @@
...
+require 'command_class'
+require 'uri'
+require 'websocket'
+require 'rubygems/package'
+
+require 'active_support/time'
+require 'websocket-client-simple'
+
+module Authentication
+  module AuthnK8s
```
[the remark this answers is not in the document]
jonahx: This is a CC suggestion I've been following almost every time lately.  Aside from providing useful info to someone unfamiliar with the code, I find the process of forcing myself to succinctly describe the class is a good exercise, and it often leads to insights that then lead to refactoring or a name change.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -0,0 +1,141 @@
...
+          message_log:      @message_log_class.new,
+          logger:           @logger
+        )
+      end
+
+      def add_websocket_event_handlers
+        # We need to set this as if we'll use @ws_client_event_handler inside
+        # the curly brackets it will look for such a member in the websocket_client
+        # class
+        ws_client_event_handler = @ws_client_event_handler
```
jonahx, line 67: I think I tested this the other night and it was not needed.  I'll check again today.

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
orenbm: if i understand correctly, this command will create the temporary file in a temp directory instead of the current one. This is a security concern as we want the temp cert to be in the same directory as the regular one so it has the same permissions.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -0,0 +1,141 @@
...
+        ws_client.on(:close) { ws_client_event_handler.on_close }
+        ws_client.on(:error) { |err| ws_client_event_handler.on_error(err) }
+      end
+
+      def wait_for_close_message
+        (timeout / 0.1).to_i.times do
+          break if ws_client_event_handler.channel_closed
+          sleep 0.1
+        end
+      end
```
jonahx, line 80: I'm going to see if I can rewrite this with a proper threading mechanism.

app/domain/authentication/authn_k8s/web_socket_client_event_handler.rb
jonahx: I meant that if we can't complete the handshake, isn't the whole prospect of communication dead?

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -0,0 +1,141 @@
...
+        # We need to set this as if we'll use @ws_client_event_handler inside
+        # the curly brackets it will look for such a member in the websocket_client
+        # class
+        ws_client_event_handler = @ws_client_event_handler
+
+        ws_client.on(:open) { ws_client_event_handler.on_open }
+        ws_client.on(:message) { |msg| ws_client_event_handler.on_message(msg) }
+        ws_client.on(:close) { ws_client_event_handler.on_close }
+        ws_client.on(:error) { |err| ws_client_event_handler.on_error(err) }
+      end
```
jonahx, line 73: I thought a lot about this code, and breaking out `WebSocketClientEventHandler`.  I think your instinct that this class was doing a bit too much was correct, but there is a better way to slice it up.  

A few things to notice that `WebSocketClientEventHandler` is not the ideal abstraction:

1. It duplicates almost all the arguments of this CommandClass.  
2. The logic that it performs really is the heart of this CommandClass.  It feels like we're separating something that should be a holistic thing.
3. The name is bulky.

As far as trimming the class down, I think the better candidates are:

1. `server_url` -> value object.
2. `timeout` is a purely functional, re-usable logic that doesn't have anything to do with this class per se.  It can be pulled out into something like:
```ruby
      class EnvVarInteger
        def initialize(env:, key:, default: nil)
          @env = env
          @key = key
          @default = default
        end
        
        def value
          return @value if @value
          
          val = @env[key]
          not_provided = val.to_s.strip.empty?
          # If the value is not an integer (eg, "blah"), it will be zero.
          @value = not_provided ? @default : val.to_i
        end
      end
```
and then `timeout` becomes:
```ruby
      def timeout
        @timeout ||= EnvVarInteger.new(
          env: @env,
          key: "KUBE_EXEC_COMMAND_TIMEOUT",
          default: DEFAULT_KUBE_EXEC_COMMAND_TIMEOUT
        ).value
      end
```
and I'm not sure if we need `DEFAULT_KUBE_EXEC_COMMAND_TIMEOUT` to be defined as a class constant versus just being a constant in this method.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
jonahx: So this can't be done with threads and mutexes, it turns out.  The mechanism we're looking for is the queue.  Here's a POC:

[Try it online!](https://tio.run/##dZHNTsQgFIX3fYqbrjobNm7HxOgTGE1m2WC5ba8iVH5K1PjslZ@pwYmuyD1wDh8H45/et22Q3Fo4cXJoGgCBI5AiR1zSB0YB4ObNo0e4hvu0MoWhyIPUFkXURy5tOopKNOeIvFe7j0dQJC@MzvgLX4gc/ahNn8/0r2gtn0qQQeeNAhr3gCbLi3cW2lucSGU3qQliQCFoKwS26KVy3KV9wBWVi9ED0oqiPcNkoFLMA5r1j2K6kAs7lPwyxAeFvca6Ci5ldW8JBKETaNDmhTFWMK1EXLqrX5lsL/IHK8Rrym/ln7BxLJFp7MKhcXOUHmeDXCQJPsGyxABfTWD/9Otm9hyBtu0b "Ruby – Try It Online")

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
jonahx: I see.  Let's create a random filename then?

app/domain/authentication/authn_k8s/execute_command_in_container.rb
jonahx: Oh one more thing I realized:  Rather than make the entire `ENV` struct a dependency, we should pass in just the timeout value and use the default `ENV["KUBE_EXEC_COMMAND_TIMEOUT"]` in the CommandClass initializer.  We can still keep the above logic, we just won't have to deal with the `@env[key]` part anymore.

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
eranha: Doesn't Ruby has the equivalent to [c#](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?view=netcore-3.1)?

app/domain/authentication/authn_k8s/execute_command_in_container.rb
eranha: @jonahx  I would add a warning if the value is not an int.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
eranha: If this pattern can change it should be configurable.

app/domain/authentication/authn_k8s/web_socket_client_event_handler.rb
eranha: It's fatal to the this session and to the client.

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
orenbm: > Let's create a random filename then?
yeah, that's a good idea.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
jonahx: @eranha That's a good idea.  Oren and I had previously discussed throwing an error, which was we thought was too heavy-headed in this case.  A warning is a good compromise.

jonahx:
I just finished and pushed my work from Friday.  There are some good improvements but I only had time for a small fraction of the changes that, imo, we still need to make.  Apologies for quick explanation but I'm about to fall asleep:

- Remove duplication in UTs
- Simplify UTs further
- Replace sleep in UTs with a proper queue, similar to how I did in the implementation class in these commits.  This will require a change to the mock I'll explain later.
- A few more simplifications to impl class (I mentioned a couple in previous comments, and completed one in this PR).

app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb
```diff
@@ -42,29 +42,28 @@ def copy_text_to_file_in_container
...
+          #!/bin/sh
+          set -e
 
-        "
-#!/bin/sh
-set -e
+          cleanup() {
+            rm -f \"#{tmp_cert}\"
```
abrahamko, line 52: you can remove the double quotes escaping, throughout all the script, since it is not part of a double-quoted string anymore.
```suggestion
            rm -f "#{tmp_cert}"
```

app/domain/authentication/authn_k8s/execute_command_in_container.rb
orenbm: removed this, please see in commits

spec/app/domain/authentication/authn_k8s/execute_command_in_container_spec.rb
```diff
@@ -0,0 +1,325 @@
...
+    x_k8s_object_lookup = k8s_object_lookup
+    x_pod_namespace     = pod_namespace
+    x_pod_name          = pod_name
+    x_container         = container
+    x_cmds              = cmds
+    x_body              = body
+    x_stdin             = stdin
+
+    Thread.new do
+      Thread.current[:output] =
```
jonahx, line 141: Why is this needed?  We should add an explanatory comment.

spec/app/domain/authentication/authn_k8s/execute_command_in_container_spec.rb
```diff
@@ -0,0 +1,325 @@
...
+          body:              x_body,
+          stdin:             x_stdin
+        )
+    end.tap do
+      # Verify all 4 listeners are ready before resuming the UT
+      Timeout.timeout(ready_listeners_timeout) {
+        4.times do
+          ws_client.ready_listeners_queue.pop
+        end
+      }
```
jonahx, line 160: This code should be moved into the mock itself under a method like `wait_until_ready` with a comment like:

```
We need to wait until the implementation code has set all of the event
handlers.  This allows us to block until they have all been called.
```
Also, more importantly, since the logic is now "we want to wait until _every handler has been set_", each event handler should add a string event with its own event name to the queue.  Then the "wait" method would pop until all 4 uniq events have been received.  The way the code is currently, it's coupled to the existing implementation.  If someone changed the code to, say, call a single event handler 4 times, the code would be convinced everything was set.

app/domain/authentication/authn_k8s/execute_command_in_container.rb
```diff
@@ -0,0 +1,158 @@
...
+          logger.formatter.current_tags.replace sub_thread_tags
+          ws_client_event_handler.on_open
+        end
+        ws_client.on(:message) { |msg| ws_client_event_handler.on_message(msg) }
+        ws_client.on(:close) { ws_client_event_handler.on_close }
+        ws_client.on(:error) { |err| ws_client_event_handler.on_error(err) }
+      end
+
+      def wait_for_close_message
+        Timeout.timeout(timeout) { @close_event_queue.pop }
```
jonahx, line 90: Something I didn't have time to do over the weekend but is definitely worth doing: Inject `Timeout` as a dependency.  Among other things, this allows us to play God with time in our unit tests, so that the UTs themselves can still execute instantly.  That is, in our test double, we can happily interpret "5 seconds" as "5 milliseconds".

jonahx (dismissed):

sigalsax (approved):
Great work!

PR #1864, opened 2020-10-05
Erase cert from every authenticator container

orenbm:
Before each cucumber test of authn-k8s, we delete the cert file
from `/etc/conjur/ssl` so that each test starts with a fresh folder.

However, we currently do this only for pods with the name 'inventory-deployment'.
This is not enough as we have different types of pods with the authenticator container
in it. They all have names that start with 'inventory-' so we can use that as the regex.


eladkug (approved):

izgeri:
I just updated this to include "in tests" in the title, because it seemed like a user-impacting feature with just the old title as it was

PR #1899, opened 2020-10-22
Added containerless CLI research doc

sigalsax:
### What does this PR do?
Adds a [document for research on containerless CLI](https://github.com/cyberark/conjur/blob/cli-spike-doc/design/proposals/spike-cli-dockerless.md)

Resolves [296](https://github.com/cyberark/conjur-cli/issues/296)

#### Change log

#### Test coverage

#### Documentation


design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+
+|            | Quality                                                 |
+| ---------- | ------------------------------------------------------- |
+| Ruby CLI   | Widely used<br />***Status***: Stable                   |
+| Go CLI     | Not "community" level certified <br />***Status***: POC |
+| Python CLI | ***Status***: Community/alpha level                     |
+
+|            | Deployment                                                   |
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | Options:<br />1. Deploy to a Docker container<br />2. Deploy using a very specific version of Ruby<br />Which means that the CLI is not plug-and-play and has prerequisites of either Docker or Ruby. |
```
InbalZilberman: @sigalsax the requirement is the not met with this option. Is there a way to address that?

InbalZilberman: What is the specific version of Ruby?
Can you please suggest how to solve this prerequisite? is there a risk to install it in RHEL?

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+| Go CLI     | The CLI is compiled down to a binary. A different binary is need to be compiled for each platform but no other requirements or installers are needed to launch the Go CLI. |
```
InbalZilberman, line 79: @sigalsax can you please elaborate what effort is needed to compile it to RHEL? is this done today?

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+
+[//]: # "Elaborate on the issue you are writing a solution for"
+
+Currently, we provide 3 CLIs that can be used to interface with Conjur, Ruby, Golang, Python. We will explore and differentiate each in terms of their functionality, quality, UX, testing, security and what it is required to deploy them.
+
+|            | Functionality                                                |
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | Most expressive coverage                                     |
+| Golang CLI | Implementation is limited with poor architectural design. We could salvage what is currently in-place but it was recommended by the contributor to rethink the architecture and even perform a complete rewrite. |
+| Python CLI | More or less at level of Golang CLI in terms of functionality coverage but with more intuitive commands. For example, in Golang `conjur-cli variable value <variable_id>` and in Python `conjur-cli variable get <variable_id>` . |
```
InbalZilberman, line 60: @sigalsax can you please specify all the UX differences?
@alexkalish was this UX been reviewed by PO ? by PM?

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+| ---------- | ------------------------------------------------------- |
+| Ruby CLI   | Widely used<br />***Status***: Stable                   |
+| Go CLI     | Not "community" level certified <br />***Status***: POC |
+| Python CLI | ***Status***: Community/alpha level                     |
+
+|            | Deployment                                                   |
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | Options:<br />1. Deploy to a Docker container (not applicable for our use-case)<br />2. Deploy using a very specific version of Ruby<br />Which means that the CLI is not plug-and-play and has prerequisites of either Docker or Ruby. |
+| Go CLI     | The CLI is compiled down to a binary. A different binary is need to be compiled for each platform but no other requirements or installers are needed to launch the Go CLI. |
+| Python CLI | Compiled down to a single binary without the need for Python or other prerequisites to be installed on machine. |
```
rafis3, line 80: Is it complied to a single binary? I was under the impression that it comes with the required Python runtime.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+
+Python CLI - TBD
+
+
+
+#### Takeways and recommendation
+
+Golang CLI would require a full rewrite due to it's limited functionality, test coverage, and architectural flaws. 
+
+Ruby requires the least amount of work but its UX is not intuitive and as part of its deployment, requires the installation of Docker (which is not valid for our use-case) or a specific version of Ruby to run the CLI.
```
rafis3, line 108: Can you please specify a little bit more about what's not intuitive? Otherwise, if the reader doesn't understand it, it's hard to agree or disagree with the statement.
One thing that comes to my mind, that I would want to improve, is being consistent with the use of verbs and nouns. For example, if I use `conjur variable` in a certain command but `conjur show` in a different command, it doesn't make sense to me. It should be either `conjur variable show` or `conjur show variable` and it should be consistent throughout the usage of it.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Takeways and recommendation
+
+Golang CLI would require a full rewrite due to it's limited functionality, test coverage, and architectural flaws. 
+
+Ruby requires the least amount of work but its UX is not intuitive and as part of its deployment, requires the installation of Docker (which is not valid for our use-case) or a specific version of Ruby to run the CLI. 
+
+Python has most of the basic actions needed to interface with the Conjur server. We will need to fill in the gaps but is architecturally more stable and well tested. The Python CLI is compiled down into a single binary so the user will not need install any additional dependencies.
+
+*Recommendation*: Python CLI because of its better design and it acts as a great point of entry.
```
rafis3, line 112: What is the criteria on which we are comparing between the options? Can you please specify them and rank the solutions based on them?

Also, I am missing the discussion about the effort size. What is the effort that each requires to bring it to GA quality and robustness?

rafis3 (commented):
@sigalsax left a few comments. Can you please take a look? Thanks!

design/proposals/spike-cli-dockerless.md
sgnn7: > Can you please specify a little bit more about what's not intuitive?

This is about the inconsistent and unintuitive suggested use of the CLI. Simple example would be:
```
# Completely unintuitive mapping between these two
conjur variable values add myapp/database/username "someusername"
conjur variable value myapp/database/username
...
# Duplicate arg & parameter names. Required switch that is useless
conjur host rotate_api_key --host myhost
conjur user rotate_api_key --user myhost
...
conjur policy load --delete --replace foo foo.yml
...
# No way to force a `yes` to owerwrite config if one is present
conjur init -a ... -u ...
```

The list is much longer than that but you can see how the current ruby CLI is in much need of UX improvements

design/proposals/spike-cli-dockerless.md
sgnn7: It's a compiled [single executable binary](https://github.com/cyberark/conjur-api-python3/releases/tag/v0.0.5) that contains the runtime and the package files. Python is not required on the machine (but you can also install this CLI via PIP and for that one you need Python).

design/proposals/spike-cli-dockerless.md
sgnn7: > I was under the impression that any CLI we develop, we'd keep the same commands as they are. Otherwise, we could potentially break functionality of existing automation that uses the Ruby CLI.

I agree about your point here but see my other response on Ruby CLI un-intuitiveness as to why the persistent deficiencies of our Ruby CLI have been a good catalyst to have this improved.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+
+*Core functionality*
+
+1. For each platform, we should support *init*, *login*, *authenticate*, *load policy*, and *fetch secrets* functionalities. These functionalities are must-haves but we should not be limited to providing just these.
+
+*Deployment*
+
+1. Provide a manual and automated way to deploy the CLI
+2. Require a maximum of 2 steps
+3. ***Recommended:*** cross-platform deployment (Linux, Windows) should all have the same flow
```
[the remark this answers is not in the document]
sgnn7: That's a good point. I think we need to be very clear about what OS targets we want to support as not all solutions work equally well in them.

design/proposals/spike-cli-dockerless.md
sigalsax: > the requirement is the not met with this option. Is there a way to address that?

My goal in listing the two options for how to deploy the RubyCLI was to give some background because the CLI world is a very unfamiliar one for our team in IMO so it was important for me to show the possibilities and not offer it as a valid option. That being said, I understand how it can be confusing here. In response, I will change to:
```
The Ruby CLI can be deployed in the following ways:
1. Deploy to a Docker container (not applicable for our use-case)
```

Is this better?

sigalsax: > What is the specific version of Ruby?
Can you please suggest how to solve this prerequisite? is there a risk to install it in RHEL?

Great questions! I will clarify this

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+
+In this document we will discuss possible options for a containerless CLI and present a decision for which one we will pursue.
+
+#### Requirements
+
+The CLI design should answer the following requirements:
+
+*Core functionality*
+
+1. For each platform, we should support *init*, *login*, *authenticate*, *load policy*, and *fetch secrets* functionalities. These functionalities are must-haves but we should not be limited to providing just these.
```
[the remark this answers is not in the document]
sigalsax: This list was provided by @InbalZilberman as *must haves* but from our meeting I understand that just supporting what is detailed on that list is not good enough for a first release. 
In addition to that list we need to make a total list of CLI functionalities that is good enough for our first release. I will add a section in the doc but from all the capabilities our CLI what is your list of good enough commands for an initial release? @rafis3

design/proposals/spike-cli-dockerless.md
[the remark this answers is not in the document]
sigalsax: @InbalZilberman do you have insights here regarding Rafi's questions on "Linux distro" and which macOS should be supported? 

> we want to support as not all solutions work equally well in them

& 

>  it might not work the same on both Debian and Fedora based systems

@sgnn7 @rafis3 Why not? isn't it just about compiling down the code to a binary that can run on each OS?

design/proposals/spike-cli-dockerless.md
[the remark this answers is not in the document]
rafis3: I think that at least: authn, check, help, host, init, list, policy, resource, role, show, user, variable.

design/proposals/spike-cli-dockerless.md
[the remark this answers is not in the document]
rafis3: A compiled binary should be compatible in both. But the installers are different - deb packages in Debian and rpm packages in Fedora. Also, for our tests, we need to know exactly where we need it to work, so we test accordingly and provide guidelines and assurance to our users, that it was tested and that it works on the stated platforms.

design/proposals/spike-cli-dockerless.md
rafis3: Ok, that's exactly what I thought. Just wanted to make sure it is not actually one single binary file.

design/proposals/spike-cli-dockerless.md
sigalsax: @rafis3 the UX table was not added/accidentally deleted so it makes sense that you cannot take a stance on it without some background info 🙃 Good catch! I will add one and include both of your pain points to the table

rafis3: Thanks @sgnn7 and @sigalsax that is really helpful, to better understand the topic and keeping us aligned on the understanding of these gaps.
@InbalZilberman I don't know if we can do it now, but maybe we should at least consider it with the timelines.

sigalsax: @sgnn7 
```
# Duplicate arg & parameter names. Required switch that is useless
conjur host rotate_api_key --host myhost
conjur user rotate_api_key --user myhost
```
is your concern here that host and user object are treated the same way in the conjur world so there is no need to support both of them `host` and `user` in the CLI?

```
# What exactly is happening here? It's not pretty obvious why you need both replace _and_ remove.
conjur policy load --delete --replace foo foo.yml
...
```
Not sure what you mean here. Why would someone write --delete and --replace in the same command?

design/proposals/spike-cli-dockerless.md
sigalsax: > What is the criteria on which we are comparing between the options?

We are comparing them in terms of functionalities, quality, ease of use (ux), deployment, testing, security.

> Can you please specify them and rank the solutions based on them?

Yes!

> Also, I am missing the discussion about the effort size. What is the effort that each requires to bring it to GA quality and robustness?

Will get back to you on that once I have researched a bit more but adding it to my open questions

design/proposals/spike-cli-dockerless.md
rafis3: I think the point here is that this command shouldn't be allowed, but it is. We should have blocked this with an error on the CLI side, asking the user to provide just one of the options.
`--patch` translates to a PUT request and `--replace` translates to a PUT request. You can do both.

design/proposals/spike-cli-dockerless.md
sigalsax: @sgnn7 Correct me if I am wrong but in the Gemfile I see the required ruby version is `2.2.5`

sigalsax: > Can you please suggest how to solve this prerequisite? is there a risk to install it in RHEL?

To solve this prerequisite would be just installation the Ruby version before deployment of the CLI. For installation on REHL 8, I saw [here](https://www.tecmint.com/install-ruby-on-centos-rhel-8/#:~:text=To%20install%20Ruby%20using%20the,invoking%20the%20following%20dnf%20command.&text=Next%2C%20ensure%20the%20below%2Dmentioned,before%20going%20ahead%20with%20Ruby.&text=Finally%2C%20install%20Ruby%20from%20the%20Appstream%20repositories.) that installation is pretty straight forward either via Appstream repositories or rvm (manages app envs by allowing us to switch between ruby versions). 

@rafis3 correct me if I am wrong but from what I read, Appstream repositories are more native to Centos/REHL and comes out of both and rvm is something that needs to be installed on the machine

design/proposals/spike-cli-dockerless.md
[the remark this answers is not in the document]
sigalsax: and I assume support flags for each of those commands. For example, policy --replace / --delete / etc?

design/proposals/spike-cli-dockerless.md
sigalsax: @rafis3 
> Just wanted to make sure it is not actually one single binary file.
What am I missing from the above explanation? It seems that the runtime and package files _are_ compiled down into a _single_ binary

rafis3: Ok, I should have just tried it from the start, instead of asking here. Now that I did, I understand. The files are executables, it's not an installation package that puts the runtime and binary in a folder, it's all inside the binary file.
Thanks for the clarification :)

design/proposals/spike-cli-dockerless.md
rafis3: Since we can't assume that the user will have a connection to the internet on the machine, I think we need to bring the Ruby and all dependencies in our installation package. So in case of RHEL and CentOS, that would be an rpm package. However, Ruby 2.2.5 has gone end of life, so looks like we need to upgrade it so we'll be using a supported version but also that it would be supported in the target platforms. Usually, versions that reached their end of life, are not supported in newer OS versions.

design/proposals/spike-cli-dockerless.md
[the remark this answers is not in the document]
sigalsax: Aren't we just supporting REHL (so that means Fedora) so why would also be supporting Debian?

What would testing look like for the different packages? Would it just be to the extent of running the package or would it be also testing all the major flows? @rafis3 @eladkug

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+| Go CLI     | Not "community" level certified <br />***Status***: POC |
+| Python CLI | ***Status***: Community/alpha level                     |
+
+|            | Deployment                                                   |
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | Options:<br />1. Deploy to a Docker container (not applicable for our use-case)<br />2. Deploy using a very specific version of Ruby<br />Which means that the CLI is not plug-and-play and has prerequisites of either Docker or Ruby. |
+| Go CLI     | The CLI is compiled down to a binary. A different binary is need to be compiled for each platform but no other requirements or installers are needed to launch the Go CLI. |
+| Python CLI | Compiled down to a single binary without the need for Python or other prerequisites to be installed on machine. |
+
+|            | Testing                                                      |
```
eladkug, line 82: Need to understand how we are going to test it, and on which platforms?
Do we have basic automation on each api today?

design/proposals/spike-cli-dockerless.md
sigalsax: So for REHL/Centos we would package the CLI + all dependencies + Ruby as a RPM package. In other words in terms of UX, the user would get a RPM package to install the full CLI?

design/proposals/spike-cli-dockerless.md
sigalsax: for all CLIs or just Go? @InbalZilberman

design/proposals/spike-cli-dockerless.md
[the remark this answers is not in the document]
rafis3: The CLI needs to run in the user's endpoint machine. That would mostly likely be Windows or macOS. Not as likely, Linux in my opinion.

eranha:
@rafis3 WE can create CLI versioning, keeping the old commands and adding new more intuitive ones.
e.g.
`conjur-cli alpha variable get`

rafis3:
@eranha we could, but let's be careful not to make the scope larger than what we can deliver. We first need to be able to provide the minimal set of commands commands in GA level, then we can provide additional enhancements.

design/proposals/spike-cli-dockerless.md
InbalZilberman: The requirement is that all our CLI will be RHEL compatible -  so we need to compare all options in order to choose one

InbalZilberman (commented):
I cant somehow comment on the thread regarding the UX of the python but Yes we have to have backward compatibility. So @sigalsax I guess the question is in high level what will it take to have the same UX as Ruby CLI in Pyhton.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,128 @@
...
+
+|            | Quality                                                 |
+| ---------- | ------------------------------------------------------- |
+| Ruby CLI   | Widely used<br />***Status***: Stable                   |
+| Go CLI     | Not "community" level certified <br />***Status***: POC |
+| Python CLI | ***Status***: Community/alpha level                     |
+
+|            | Deployment                                                   |
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | Options:<br />1. Deploy to a Docker container (not applicable for our use-case)<br />2. Deploy using a very specific version of Ruby<br />Which means that the CLI is not plug-and-play and has prerequisites of either Docker or Ruby. |
```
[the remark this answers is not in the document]
sigalsax: Does the machine have access to internet? If not, creating some sort of installer that is prepacked with ruby + dependencies that can be delivered as a package that can run on REHL/Windows environments

If they do have access to internet, then just running `gem install conjur cli` but that will need to be tested because we currently only support CLIs running on Docker.

@rafis3 any other directions come to mind?

sigalsax:
> I guess the question is in high level what will it take to have the same UX as Ruby CLI in Pyhton.

@InbalZilberman 
I am not sure we want that to happen. We don't want to put poor UX (Ruby) and overwrite a CLI that has good UX and good modeling (Python)

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,209 @@
...
+| **Term**   | **Description**                                |
+| ---------- | ---------------------------------------------- |
+| Epic       | https://github.com/cyberark/conjur/issues/1892 |
+| Ruby CLI   | https://github.com/cyberark/conjur-cli         |
+| Go CLI     | https://github.com/cyberark/conjur-cli-go      |
+| Python CLI | https://github.com/cyberark/conjur-api-python3 |
+
+## Background
+
+[//]: # "Give relevant background for the designed feature. What is the motivation for this solution?"
```
abrahamko, line 30: needed?

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,209 @@
...
+
+*Deployment*
+
+1. Provide a manual and automated way to deploy the CLI
+2. Require a maximum of 2 steps
+3. ***Recommended:*** cross-platform deployment (Linux, Windows, macOS) should all have the same flow
+
+## Research
+
+[//]: # "Elaborate on the issue you are writing a solution for"
```
abrahamko, line 52: needed?

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,209 @@
...
+
+Currently, we provide 3 CLIs that can be used to interface with Conjur, Ruby, Go, Python. We will explore and differentiate each in terms of their functionality, quality, UX, testing, security and what it is required to deploy them.
+
+|            | Functionality                                                |
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | Most expressive coverage                                     |
+| Go CLI     | Implementation is limited with poor architectural design. We could salvage what is currently in-place but it was recommended by the core contributor to rethink the architecture and even perform a complete rewrite. |
+| Python CLI | More or less at level of Go CLI in terms of functionality coverage. |
```
abrahamko, line 60: I would rephrase it, something like, `Implements most of the commands`.
Don't underestimate the maturity of the python CLI. I think it's significantly higher than go CLI.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,209 @@
...
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | Unintuitive and inconsistent CLI command construction. See list below for thorough explanations. |
+| Go CLI     | Similar to Ruby CLI                                          |
+| Python CLI | More straightforward approach. See list below. For example, when setting a Conjur secret,  `conjur-cli variable set my_id supersecretvarid` |
+
+#### Ruby/Go CLI UX
+
+The following are examples of flaws in our current Ruby/Go CLI UX
+
+*Verb/noun placement*
```
abrahamko, line 84: suggestion: create a bullet-list of it.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,209 @@
...
+
+*Unintuitive command construction*
+
+When running `conjur variable value`, a secret value will be returned to you but when looking at just the command, it is unclear what the purpose of the command is. For a more straightforward approach, the command for display a secret value could be `conjur get <variable_id>` or `conjur show <variable_id>`.  
+
+Another example of uninuitive command construction would be trying to understand what the difference between `conjur variable value`  is and `conjur variable values` without using the `--help` flag.
+
+*Permitted edge cases*
+
+We are currently not blocking the chaining of illogical flagging. For example, `conjur policy load --delete --replace foo foo.yml` does not lead to an error.
```
abrahamko, line 96: good point!

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,209 @@
...
+conjur-cli variable set my_var_id mysupersecretvar
+conjur-cli variable get my_var_id
+```
+
+For policy maintaince the outcome of the command is very clear. Unlike the Ruby/Go CLI, there is no chance of an unpredictable outcome because running `replace` and `delete`  would lead to an error.
+
+```bash
+conjur-cli policy apply root ./path_to_policy.yml
+conjur-cli policy replace root ./path_to_policy.yml
+conjur-cli policy delete root ./path_to_policy.yml
```
abrahamko, line 112: Add a comment about the `policy delete` does **not** mean deleting the policy.
This is a severe UX flaw.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,209 @@
...
+
+Python has most of the basic actions needed to interface with the Conjur server. We will need to fill in the gaps but is architecturally more stable and well tested. The Python CLI we deliver an executable so the user will not need install additional dependencies.
+
+*Recommendation*: Python CLI because of its better design and it acts as a great point of entry. In terms of a model and architectural design, it is where we want to be so it is a great first step into reforming our APIs.
+
+## Documentation
+
+[//]: # "Add notes on what should be documented in this solution. Elaborate on where this should be documented. If the change is in open-source projects, we may need to update the docs in github too. If it's in Conjur and/or DAP mention which products are affected by it"
+
+The CLI that we choose will require documentation.
```
abrahamko, line 198: Add to here the fact that Ruby CLI _has_ documentation, and it will apply for go CLI if it maintains the Ruby UX.
So only Python missing documentation, or if any UX changes.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,209 @@
...
+| Go CLI     | 3             | 3       | 3    | 2          | 3       | TBD      |
+| Python CLI | 2             | 2       | 1    | 1          | 2       | TBD      |
+
+Ruby requires the least amount of work in terms of what is already fuctionally supported but its UX is not intuitive. As part of its deployment, the CLI requires a specific version of Ruby (`2.2.5`) installed on the machine to run the CLI. Machines may not have internet access so we will need to pre-package Ruby and other dependencies and deliver as with an installer.
+
+Go CLI would require a full rewrite due to it's limited functionality, test coverage, and architectural flaws. 
+
+Python has most of the basic actions needed to interface with the Conjur server. We will need to fill in the gaps but is architecturally more stable and well tested. The Python CLI we deliver an executable so the user will not need install additional dependencies.
+
+*Recommendation*: Python CLI because of its better design and it acts as a great point of entry. In terms of a model and architectural design, it is where we want to be so it is a great first step into reforming our APIs.
```
abrahamko, line 192: suggestion: Since this sums your whole doc, put it under a header.

abrahamko (commented):
This is an amazing doc!
It's obvious you did great job to investigate all the options and bring up the relevant points, although you were unfamiliar with all of it.
👸

design/proposals/spike-cli-dockerless.md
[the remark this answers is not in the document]
rafis3: As we spoke, these are the main options I am thinking about, based on the fact that we can't rely on internet access:

- Providing a single runnable binary that contains inside all the dependencies. That can be run immediately without any installation. For Ruby specifically, I don't know if such capability exists.
- Providing OS native packages (rpm, pkg, etc.) that will install the binary + dependencies on the machine
- Providing a zip that will extract all the relevant content into a target folder, binary + dependencies

I also wrote them ranked from what I think is best UX to worst UX.

design/proposals/spike-cli-dockerless.md
sigalsax: > Implements most of the commands

This is not true either. It probably implements half of what we want but it is definitely more mature

design/proposals/spike-cli-dockerless.md
sigalsax: those are comments that don't appear in doc if rendered so I left them because they aren't visible. See here https://github.com/cyberark/conjur/blob/e14435868badec5ca5cce3091a16cd1dc3bbb55e/design/proposals/spike-cli-dockerless.md

design/proposals/spike-cli-dockerless.md
sigalsax: See below comment [here](https://github.com/cyberark/conjur/pull/1899#discussion_r511948329)

design/proposals/spike-cli-dockerless.md
sgnn7: This is somewhat of a complex question but I'll try to answer:
- I've added some time ago ability to compile to a dynamically-linked single binary for Linux platforms
- I have very high confidence that it should work on RHEL _but_ it's not a fully-statically-linked binary with `CGO_ENABLED=1` and `-extldflags "-static"`
- This may make it not work on certain Linux platform combination that do not have the same baseline libraries installed (e.g. `glibc`). Alpine is one of those OS permutation that may not work since it uses `musl` instead.
- We can make this a fully static build with some work as we do that with `secretless-broker` but it's a much more convoluted process

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,254 @@
...
+  variable get foo/bar
+```
+
+|            | Deployment                                                   |
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | The Ruby CLI can be deployed in the following ways:<br />1. Deploy to Docker container (**not** applicable for our use-case)<br />2. Deploy with specific Ruby version (`gem install conjur-cli`)<br />Which means that the CLI is not plug-and-play and has prerequisites |
+| Go CLI     | The CLI is compiled down to a binary. A different binary is need to be compiled for each platform but no other requirements or installers are needed to launch the Go CLI. |
+| Python CLI | The Python CLI can deployed in the following ways:<br />1. Install CLI via PIP (Python **is** required on machine)<br />2. Runnable executable (Python **is** **not** required on the machine)<br />The CLI is delivered as a [standalone executable](https://github.com/cyberark/conjur-api-python3/releases/tag/v0.0.5), packed with the required runtime and dependencies. For this option there is no need for Python or other prerequisites to be installed on machine. |
```
sgnn7, line 130: There may be dynamic library requirements for Linux but they should be minimal - I have not explored the support matrix much other than the most common Win/macOS/Linux permutations for those binaries.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,254 @@
...
+[//]: # "Add links that may be useful for the reader"
+
+| **Term**                                      | **Description**                                              |
+| --------------------------------------------- | ------------------------------------------------------------ |
+| Epic                                          | https://github.com/cyberark/conjur/issues/1892               |
+| Ruby CLI                                      | https://github.com/cyberark/conjur-cli                       |
+| Go CLI                                        | https://github.com/cyberark/conjur-cli-go                    |
+| Python CLI                                    | https://github.com/cyberark/conjur-api-python3               |
+| [FPM](https://github.com/jordansissel/fpm)    | For building packages for multiple platforms (RPM, DEB, OS X) |
+| [Debify](https://github.com/conjurinc/debify) | Build a package from a Ruby gem                              |
```
izgeri, line 28: are you aware of [Goreleaser's tooling](https://carlosbecker.com/posts/nfpm/) for releasing deb/rpm as part of a Golang project's pipeline? might be worth referencing here too ([secretless example](https://github.com/cyberark/secretless-broker/blob/master/.goreleaser.yml#L126-L127))

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Background
+
+[//]: # "Give relevant background for the designed feature. What is the motivation for this solution?"
+
+This document is one piece in the bigger piece of the Dockerless Conjur puzzle. We need to be able to deploy Conjur on a VM. This means we will need to figure out a way to deliver a containerless CLI to allow human-users to interface with Conjur.
```
izgeri, line 34: are we planning to provide tooling to deploy Dockerless Conjur too? if not, this is kind of misleading.

also, for what it's worth there's a lot of value in having a Dockerless CLI option regardless of the deployment mode of Conjur - it might be worth speaking to that here, instead?

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,254 @@
...
+
+In this document we will discuss possible options for a containerless CLI and present a recommendation for which one we should pursue.
+
+#### Requirements
+
+The CLI design should answer the following requirements:
+
+*Core functionality*
+
+1. We should support `init`, `authn` (`login`, `authenticate`),  `policy (load)`, `list`, `show, ` `rotate_api_key` (`host` / `user`) and `fetch secrets` functionalities. These functionalities are must-haves.
```
izgeri, line 44: this list seems _really_ biased by the design of the current Ruby CLI

would it be more accurate to require methods to:
- configure connection to Conjur (may or may not include authenticating - i.e. we may not want to require users to explicitly run an `authenticate` step)
- load policy (which modes do we want to support? PUT / POST / PATCH?)
- list all resources that I have `read` permissions on
- rotate user and host API keys
- get and set secret values

what is `show` in this context?

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,254 @@
...
+
+*Deployment*
+
+1. Provide a manual and automated way to deploy the CLI
+2. Require a maximum of 2 steps
+3. ***Recommendation:*** cross-platform deployment (Linux, Windows, macOS) should all have the same flow
```
izgeri, line 50: I think we can learn a lot from other popular CLIs here. we can review how to install [`git`](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git), [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/), [`gcloud`](https://cloud.google.com/sdk/docs/install#linux), etc

design/proposals/spike-cli-dockerless.md
sgnn7: @sigalsax Yes - it looks like the Ruby CLI [is using `ruby v2.2.4`](https://github.com/cyberark/conjur-cli/blob/8dad66b483fbe1c8b00e124f47f928677419b87a/jenkins.sh#L4)

design/proposals/spike-cli-dockerless.md
sgnn7: @sigalsax and I also found https://github.com/pmq20/ruby-packer today

design/proposals/spike-cli-dockerless.md
sgnn7: > also, for what it's worth there's a lot of value in having a Dockerless CLI option regardless of the deployment mode of Conjur - it might be worth speaking to that here, instead?

Completely agreed here - the current CLI can be improved in many aspects and would add value independent of Conjur deployment style

design/proposals/spike-cli-dockerless.md
sigalsax: @sgnn7 is this the signed executables we spoke about?

design/proposals/spike-cli-dockerless.md
sigalsax: @izgeri if Go allows to run a single binary that can be shipped to users (after compiled down to each target platform) why would we need to deliver them as packages (using the Goreleaser)?

design/proposals/spike-cli-dockerless.md
sigalsax: > are we planning to provide tooling to deploy Dockerless Conjur too? if not, this is kind of misleading.

Yes. The whole project's goal is to make everything dockerless 😈 

> also, for what it's worth there's a lot of value in having a Dockerless CLI option regardless of the deployment mode of Conjur - it might be worth speaking to that here, instead?

We basically already have that in Go CLI and Python CLI. But do you mean there is a lot of value in delivering a CLI that we official support?

design/proposals/spike-cli-dockerless.md
sigalsax: `show` is for showing an object. I have never actually used this one but I am guessing it returns its metadata in a list.

It is a really biased design I agree but because we are more familiar with the Ruby CLI using those names, gives some reference to the types of commands we are talking about. That way, we don't have to re-remember which commands we are referencing. 
I imagine that if we do decide to expand on the python cli, we won't be following that naming because of the wonky UX

orenbm: @sigalsax what is the use-case to run `conjur authn authenticate`? i have never used it.

design/proposals/spike-cli-dockerless.md
sgnn7: No - this is more baseline library requirement for a Linux system (e.g. glibc, libgz). Running `ldd` on the binary should give a list of minimal requirements for these. The latest release shows this as:
```
ldd ./conjur-py3-cli-linux 
        linux-vdso.so.1 (0x00007fff67590000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f8526758000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f852653a000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f8526379000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f8526768000)
```

design/proposals/spike-cli-dockerless.md
izgeri: wow - I didn't realize we were working on a Dockerless Conjur OSS too. is there a design doc where I can read up more about that?

design/proposals/spike-cli-dockerless.md
sigalsax: `conjur authn authenticate` gives you the authentication token for the user that is logged in
We use it in our curl requests see [here](https://docs.conjur.org/Latest/en/Content/Developer/Conjur_API_Retrieve_Secret.htm?tocpath=Developer%7CREST%C2%A0APIs%7C_____8#Examplewithcurl) for example

design/proposals/spike-cli-dockerless.md
sigalsax: From what I understand, even though Conjur is a component of DAP, we will not be _officially_ supporting a Dockerless Conjur. You are correct that it is misleading and therefore dropping any mention of Dockerless _Conjur_.

eladkug (approved):

design/proposals/spike-cli-dockerless.md
izgeri: go has a separate binary for each OS, it's not a single binary.

it depends on what we want to make available. I think reviewing what standard CLIs have available as install options and deciding what we want to provide is likely the best first step.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,257 @@
...
+| Go CLI     | 3             | 3       | 3    | 2          | 3       | See scan |
+| Python CLI | 2             | 2       | 1    | 1          | 2       | See scan |
+
+Ruby CLI ~11 weeks
+
+- Ramp-up on creating packages and create design for CLI (~1 week)
+- Provide package (rpm, pkg, deb, etc) / [deliver as executable](https://www.ruby-toolbox.com/categories/packaging_to_executables) (CLI + dependencies) (~3 week)
+- Expand current testing if needed (~1 week)
+- Build infrastructure for testing on new platforms (Windows, macOS, REHL environments) (~3 weeks)
+- Run a few CLI commands on new supported platforms (Windows, macOS environments) (~2 weeks)
```
jvanderhoof, line 197: If the CLI is intended for the Dockerless DAP effort, why would we do any additional work on new platforms?  Doesn't this expand the work beyond the immediate need (to provide a Conjur CLI for Dockerless DAP)?

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,257 @@
...
+- Redesign CLI
+- Implement new CLI commands and logic
+- Write tests for new functionality
+- Might need to redesign and build new infrastructure for environment testing (Windows, macOS, REHL environments)
+- Run a few CLI commands on new supported platforms (Windows, macOS, REHL environments)
+- Add documentation
+
+#### Takeways and recommendation
+
+Ruby requires the least amount of work in terms of what is already fuctionally supported but its UX is not intuitive. As part of its deployment, the CLI requires a specific version of Ruby installed on the machine to run the CLI.
```
jvanderhoof, line 224: Although the interface is sub-optimal, why not launch with it (reduces time), and either refactor the interface in the future or focus on the GoLang CLI, which greatly simplify the CLI experience across all platforms.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,257 @@
...
+
+Machines may not have internet access so we will need to pre-package Ruby and other dependencies and deliver as with an installer.
```
jvanderhoof, line 226: We'd need to package the gem and dependencies, but Ruby should already be available on the underlying system in order to run Conjur, Conjur-UI, and Evoke.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,257 @@
...
+
+Go CLI would require a full rewrite due to it's limited functionality, test coverage, and architectural flaws. 
+
+Python has most of the basic actions needed to interface with the Conjur server. We will need to fill in the gaps but is architecturally more stable and well tested. The Python CLI we deliver an executable so the user will not need install additional dependencies.
+
+##### Recommendation
+
+Python CLI because of its better design and it acts as a great point of entry. In terms of a model and architectural design, it is where we want to be so it is a great first step into reforming our APIs.
```
jvanderhoof, line 234: I'd like to challenge this choice. Although the Python library might have the best interface, it suffers the same problem as Ruby in that they both require an interpreter to run on a host system. This means users need to install Python (particularly Python 3.x). Ruby is already well tested and has all the need functionality (although not the interface we'd like). In terms of speed of delivery (providing Dockerless DAP with a CLI), it feels like a clear favorite.

If we're looking at long term strategic value, I'd argue the GoLang CLI is the best option. It could take the desired interface patterns from the Python CLI.  It would be installable on nearly any chipset/OS combination with zero dependencies. It would have a tiny footprint (due to compilation).  We already have a number of highly proficient GoLang engineers (anyone who's done substantial work on Secretless ).

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,257 @@
...
+
+Python CLI hasn't been clearly documented. Therefore we will need to add documentation
+
+## Open questions
+
+[//]: # "Add any question that is still open. It makes it easier for the reader to have the open questions accumulated here istead of them being acattered along the doc"
+
+Packaging:
+
+1. For Ruby CLI, if we want to package the CLI + dependencies + Ruby how would we provide an installation package for windows / macOS? For example, in REHL we can use RPM.
```
jvanderhoof, line 254: DAP includes limitations that prevent it from running on Windows.  Why would we need to create a package other than an RPM?

design/proposals/spike-cli-dockerless.md
jvanderhoof: I'd challenge this as a requirement. DAP can only run on *nix systems due to its dependency on the Kernel Keyring. As this document is specifically focused on Dockerless DAP, I don't feel cross-OS functionality is relevant.

If this document is intended to embark on a meaningful change to our CLI (which I strongly support 😄 ), I recommend we include the long term timeline of a robust, easy to use, OS native installable CLI. This will allow us to frame the shorter term deliverables required to achieve Dockerless DAP as a step toward the future desired outcome.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,257 @@
...
+  variable get foo/bar
+```
+
+|            | Deployment                                                   |
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | Currently, the Ruby CLI can be deployed in the following ways:<br />1. Deploy to Docker container (**not** applicable for our use-case)<br />2. Deploy with specific Ruby version (`gem install conjur-cli`)<br />Which means that the CLI is not plug-and-play and has prerequisites |
+| Go CLI     | The CLI is compiled down to a binary. A different binary is need to be compiled for each platform but no other requirements or installers are needed to launch the Go CLI. |
+| Python CLI | The Python CLI can deployed in the following ways:<br />1. Install CLI via PIP (Python **is** required on machine)<br />2. Runnable executable (Python **is** **not** required on the machine)<br />The CLI is delivered as a [standalone executable](https://github.com/cyberark/conjur-api-python3/releases/tag/v0.0.5), packed with the required runtime and dependencies. For this option there is no need for Python or other prerequisites to be installed on machine. |
+
+#### Deployment of Ruby CLI
```
alexkalish, line 135: I might suggest giving concrete examples of the deployment UX of each one.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,257 @@
...
+| ---------- | -------- |
+| Ruby CLI   | See scan |
+| Go CLI     | See scan |
+| Python CLI | See scan |
+
+#### CLI commands to be supported
+
+The CLI should support the following commands:
+
+*Core* -  `init`, `authn` (`login`, `authenticate`),  `policy (load)`, `list`, `show `, `rotate_api_key` (`host` / `user`) and `fetch secrets`
```
alexkalish, line 173: Again, maybe focus on user stories or REST APIs instead of concrete commands?

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Background
+
+[//]: # "Give relevant background for the designed feature. What is the motivation for this solution?"
+
+This document is one piece in the bigger piece of the Dockerless puzzle. In this document we will discuss possible options for a containerless CLI and present a recommendation for which one we should pursue.
+
+In addition, regardless of how Conjur is deployed delivering and officially supporting a CLI that does not require Docker allows our solution to be portable and more easily attainable.
+
+#### Requirements
```
alexkalish, line 39: Here is the most important requirement: we only have a single CLI for customers.  So if something other than Ruby is desired, we need to have a discussion about moving to that new one and away for Ruby.

design/proposals/spike-cli-dockerless.md
sigalsax: This point might be confusing because it is under "requirements" but this was only a recommendation for a better UX- to have a singular flow. If I understood @InbalZilberman correctly. this was something to be explored, not necessary a "much have"

design/proposals/spike-cli-dockerless.md
sigalsax: > If the CLI is intended for the Dockerless DAP effort

Do you mean b/c Dockerless DAP is supposed to be initially deployed on REHL?

This is a good point! I cannot speak to requirements and scope but I would imagine the answer would be because how a customer will interface with DAP will change as a result of this decision. The ultimate goal is to choose the main CLI the customer will use to interface with Conjur/DAP and _**if**_ needed, ultimately swap the Ruby CLI for that one (Python/Go). If the CLI is swapped then we will want to support the chosen one across all our main platforms because is now a main component.

@InbalZilberman please feel free to chime in here because this is a question of scope

design/proposals/spike-cli-dockerless.md
sigalsax: I am not so sure the amount of time that will be saved will be substantial when compared to the Python CLI and we could get a better UX. IMO changing the UX dramatically down the road might be confusing for the user and demand from us a lot more work. If we start with the Ruby CLI we will need to understand how to package it (single binary, zip, etc) which is added effort because from what I have seen isn't so straightforward.

design/proposals/spike-cli-dockerless.md
sigalsax: > This pre-req also means that we cannot use the CLI on machines that do not have our exact Ruby requirement so effectively we will require a full VM for the CLI if we can't use Docker

I would argue that this is same for Python don't you think? We would ideally be packing Python and all dependencies needed to run the Python CLI

design/proposals/spike-cli-dockerless.md
sigalsax: >  I'd like to challenge this choice.

I want to clarify that this is a recommendation not the actual choice we are yet pursuing

> There's a link elsewhere in the doc that I posted that might be able to do the same for Ruby code

https://github.com/pmq20/ruby-packer . Tried it Thursday and now today and it is a lot less easy to get up than it seems. Sat with Gary for a good hour and a half. Nonetheless, still investing because of its potential. 

>  understand I think the timelines for this project and speed of Go development vs Ruby/Python may not make it a feasible choice

Yes. In the long run, I agree Go CLI is definitely a better decision. However, we do not have the time for a complete rewrite of functionality and testing because of this hard deadline.

design/proposals/spike-cli-dockerless.md
```diff
@@ -0,0 +1,257 @@
...
+
+|            | UX                                                           |
+| ---------- | ------------------------------------------------------------ |
+| Ruby CLI   | What we are familiar with. Unintuitive and inconsistent CLI command construction. See list below for more thorough examples |
+| Go CLI     | Similar to Ruby CLI                                          |
+| Python CLI | More straightforward approach. For example, when setting a Conjur secret,  `conjur-cli variable set my_id supersecretvarid`. See list below |
+
+#### Ruby/Go CLI UX
+
+The following are examples of *just* *some* of flaws in our current Ruby/Go CLI UX.
```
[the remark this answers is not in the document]
sigalsax: Ohh! good point. I will add this in the doc (:

design/proposals/spike-cli-dockerless.md
sigalsax: I don't think there is much to add. For the Ruby CLI you either pull down the image or run `gem install conjur-cli`. After that (`conjur init` + `conjur authn login`) is how you connect with Conjur which IMO falls into the UX category.
The same for Python CLI, by just installing via pip or running the executable. So it would be `python-cli-executable-location --help` 

Either way, it is detailed in the chart. Do you think that information should be detailed again?

design/proposals/spike-cli-dockerless.md
sigalsax: Isn't that how it is now? or are you saying deprecate all other CLIs except for the one we chose?

design/proposals/spike-cli-dockerless.md
sigalsax: > We'd need to use the system-level Ruby that's global (unless we work on developing some RVM solution) so all Ruby apps of the system must match in their Ruby requirement

@sgnn7  but this is a challenge we are currently facing with our other dockerless components (conjur, conjur ui, evoke) so it seems that this isn't new. 

@rafis3 correct me if I am wrong here but this is something we need to worry about regardless no?

rafis3: @jvanderhoof this document only discusses the Conjur CLI, so Conjur, UI and evoke are not relevant here.
@sgnn7 why do we have to use a system installed Ruby? Why won't we come with our own Ruby runtime and point the CLI to using that?

InbalZilberman:
@alexkalish @jvanderhoof @sgnn7 the requirements regarding CLI are listed here: https://app.zenhub.com/workspaces/dap-master-no-docker-5f82b33fdb531f00128c0a07/issues/cyberark/conjur/1892
we need to support CLI on Windows, Mac and RHEL

design/proposals/spike-cli-dockerless.md
InbalZilberman: @alexkalish the requirements are regarding the UX of the CLI not REST API.
The list will be re examine with PM, PS and so on.. but it is important to list the capabilities the CLI provides.

design/proposals/spike-cli-dockerless.md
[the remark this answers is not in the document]
InbalZilberman: BTW Also Sharon R. our UX provided a feedback

design/proposals/spike-cli-dockerless.md
InbalZilberman: @jvanderhoof see requirements: https://app.zenhub.com/workspaces/dap-master-no-docker-5f82b33fdb531f00128c0a07/issues/cyberark/conjur/1892

design/proposals/spike-cli-dockerless.md
sgnn7: > @sgnn7 but this is a challenge we are currently facing with our other dockerless components (conjur, conjur ui, evoke) so it seems that this isn't new

@sigalsax When talking about the appliance, ensuring system-level Ruby version is not a big deal as we must have our appliance in our own OS container/VM. The problem here is when we need a CLI that will run outside of the deployment infrastructure with the same requirement. If a user wanted to run the CLI but they needed to install a specific (and probably old) Ruby version for the whole machine, they would most likely have to dedicate a whole machine/VM/container for it if they didn't want a system that they cannot upgrade without problems.

> @sgnn7 why do we have to use a system installed Ruby? Why won't we come with our own Ruby runtime and point the CLI to using that?

Because we don't have a good Ruby isolation framework (e.g.`rvm`/`rbenv`) explored/used for the CLI (nor other components for that matter) so you must rely on the system-level Ruby version to provide your interpreter. It's not an unfixable problem but it's a really challenging one from a deployment standpoint.

design/proposals/spike-cli-dockerless.md
alexkalish: I think that those details have value, especially for non-engineers reading this.

design/proposals/spike-cli-dockerless.md
alexkalish: @InbalZilberman: I understand that this is for the CLI.  However, this says: "The CLI should support the following commands:".  Why do they have to be named like that?  Why is `init` the right choice?  I think it's better to speak in terms of functionality than concrete commands, otherwise you risk biasing reads towards existing commands/UXs.

design/proposals/spike-cli-dockerless.md
alexkalish: @sigalsax: Yes, exactly. We only have a single GA, officially support CLI.  If this project selects a different one to meet its requirements, it should replace Ruby for all customers.  For this reason, the consequences of picking Python or Go should be clear.  It's important that @InbalZilberman and Sharon see this requirement and let me know if they disagree.

design/proposals/spike-cli-dockerless.md
InbalZilberman: @alexkalish the idea of the epic is NOT to redo UX it is to have the core functionality provided today to our customers provided also deployed in a machine without docker

design/proposals/spike-cli-dockerless.md
sigalsax: I remember this point being mentioned that the ultimate goal is to replace the Ruby CLI (if that is the direction we want to go in) or further developer it. Either way there should only be one that is distributed. @InbalZilberman ?

design/proposals/spike-cli-dockerless.md
sigalsax: added [here](https://github.com/cyberark/conjur/pull/1917/files#diff-94fda9f3ac9c67857646893fe6d3fd0053a6b319cd511dc03744f648e3da293dR148) for both Ruby and Python deployments

sigalsax:
@alexkalish @jvanderhoof @sgnn7 I opened up a new PR that updates the doc with your feedback, meeting minutes, and results from the mini-spike I did for package the Ruby CLI into an executable via the ruby-packer project. See [here](https://github.com/cyberark/conjur/pull/1917)

design/proposals/spike-cli-dockerless.md
jvanderhoof: Is the outcome a CLI for the Dockerless DAP, or a new CLI?  It feels like there is some big batching going on by reworking the CLI for Dockerless.  Would it make more sense to push the CLI to a future phase?

sigalsax: we need to provide our customers with a total dockerless deployement so that means the CLI. Users can use the current Ruby CLI and just install it via gem but that isn't good enough and not a solution that can work across the board. I did a minispike to see if we can somehow pack the ruby cli and get a single executable (see [here](https://github.com/cyberark/conjur/pull/1917/files#diff-94fda9f3ac9c67857646893fe6d3fd0053a6b319cd511dc03744f648e3da293dR163). For the reasons listed in the link we should not proceed with attempting to do so.

design/proposals/spike-cli-dockerless.md
izgeri: @jvanderhoof the platform that DAP runs on is not necessarily the same platform that users of the CLI will be running from. personally I always run the CLI from my local machine, which is a Mac. I think this requirement is intended to enable the same capability for customers using Dockerless DAP (ability to run the CLI from their local machine).

design/proposals/spike-cli-dockerless.md
izgeri: does this mean not redoing the Python UX, if we move forward with that CLI as a Dockerless option? because its functionality for configuring the CLI is _leagues_ ahead of what the Ruby CLI uses.

sigalsax: @izgeri I copied over your input to the new PR where I have updated the doc based on feedback mentioned here. Feel free to take a look! See [here](https://github.com/cyberark/conjur/pull/1917/files#diff-94fda9f3ac9c67857646893fe6d3fd0053a6b319cd511dc03744f648e3da293dR45)
I agree we should start thinking about our CLI in terms of business logic and not make a replica of the Ruby CLI if we realize already that it is bad UX. It from the start limits us and narrows our thinking

PR #1904, opened 2020-10-25
Add design doc for Resource Restrictions validation

### LINKED ISSUE #1814: Refactoring authenticators code - Low level design for ValidateResourceRestrictions
As part of developing GCP authenticator we have noticed that the `ValidateResourceRestrictions` component can be redesign to serve  most of our authenticators: K8S, Azure, and GCP

Write a LLD which explains the change

**DoD**
- [ ] We have an LLD reviewed by the TEAM

abrahamko:
### What does this PR do?
Add a low-level design for refactoring the annotation-based authenticators (k8s, azure, gcp).
Specifically, refactor the validation against the host annotations.

Here's a [readable version](https://github.com/cyberark/conjur/pull/1904/files#diff-c6749f21316010c14976291f199675c419d396fb82ad97cb8998db473945ca7a)

Resolves #1814 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/design/authenticators/resource-restrictions/resource-restrictions.md
+++ b/design/authenticators/resource-restrictions/resource-restrictions.md
@@ -1,0 +1,352 @@
+# Resource Restrictions
+
+##### Note
+
+In this doc, I will cover authenticating hosts with Conjur. Everything about 
+authenticating hosts is also relevant to users. For the scope of this document, 
+host, user or role are interchangeable.
+
+## Introduction
+
+In several authenticators, such as `authn-k8s`, `authn-azure` and `authn-gcp`,
+we require hosts to contain some data which define their identity. This data is
+referred to as Resource Restrictions.
+The Resource Restrictions give the authenticator another layer of
+authentication, where it can authenticate applications using not only their
+Conjur identity and permissions (e.g host can `authenticate` with the
+authenticator `webservice`) but also their native identity (e.g Azure
+resource characteristics)
+
+For example, in the Azure Authenticator the Resource Restrictions include:
+
+- Subscription ID
+- Resource Group
+- System-Assigned Identity (optional)
+- User-Assigned Identity (optional)
+
+A host that will authenticate using the Azure authenticator will be defined
+ in its annotations as follows:
+ ```yaml
+- !host 
+  id: test-app 
+  annotations: 
+    authn-azure/subscription-id: test-subscription 
+    authn-azure/resource-group: test-group 
+ ```
+
+While in the Azure Authenticator the Resource Restrictions are always stored in
+the host annotations, in the Kubernetes Authenticator the Resource Restrictions
+can be stored either in the host's ID or in its annotations. 
+For example:
+
+ ```yaml
+- !host
+  id: test-app-service-account
+  annotations:
+    authn-k8s/namespace: some-namespace
+    authn-k8s/service-account: some-service-account
+
+- !host
+  id: some-namespace/service_account/some-service-account
+ ```
+
+In the example above, both hosts have the same Resource Restrictions. One stores 
+it in its annotation and the other in its ID. However, their Conjur identity is
+not the same.
+
+## Issue Description
+
+When we add a new authenticator to Conjur, we need to add the mechanism
+of validating the host's Resource Restrictions. Currently we have no generic
+way to do it and because of that it is less straight-forward to add a new
+authenticator. As can be seen in the implementations for the
+[Azure Authenticator](/app/domain/authentication/authn_azure/validate_resource_restrictions.rb)
+and 
+[Kubernetes Authenticator](/app/domain/authentication/authn_k8s/validate_resource_restrictions.rb), 
+there is a lot of code duplication with minor changes. 
+
+## Solution
+
+For the sake of simplicity, from here and throughout the document, Resource
+Restrictions will be defined only in the host's annotations, and not in the ID. 
+In future authenticators we will not develop an option to define it in the ID
+and it is available in the K8s Authenticator only for backwards compatibility.
+
+### Defining Resource Restrictions Authentication
+
+The Resource Restrictions authentication is the major flow of actions needed
+to validate a request using Resource Restrictions.
+It can be split into 3 parts:
+
+ 1. Extract the Resource Restrictions from the host
+ 2. Validate the Resource Restrictions are defined correctly
+ 3. Validate the Resource Restrictions match the authentication request
+
+Let's break down each part to better understand them.
+
+#### 1. Extract the Resource Restrictions from the host
+
+This step goes to the host in the policy and extracts the Resource Restrictions
+from its annotations.
+
+Requirements:
+
+* Extract both **granular restrictions** (starts with `authenticator_name`) and
+  **specific restrictions** (starts with `authenticator_name/service_id`).
+* If same restriction exist for both **granual** and **specific** restrictions,
+  **specific** restrictions should prevail.
+* Ignore specific annotations of a different service id
+  (starts with `authenticator_name/other_service_id`).
+* Return a `resource_restrictions` object of all the restrictions, without their
+  prefix.
+
+
+The `ResourceRestrictions` class is in the form:
+
+```ruby
+class ResourceRestrictions
+  def initialize(resource_restrictions:)
+    @resource_restrictions = resource_restrictions
+  end
+
+  def names
+    @resource_restrictions.keys
+  end
+
+  def each(&block)
+    @resource_restrictions.each(&block)
+  end
+end
+```
+
+
+
+In this case the general class will be defined as follows:
+- input
+  - authenticator name
+  - authenticator service-id
+  - account
+  - host name
+- output
+  - `resource_restrictions` object.
+  
+
+So this class signature is:
+
+```ruby
+ExtractResourceRestrictions = CommandClass.new(
+  dependencies: {
+    role_class:     ::Role,
+    resource_class: ::Resource,
+    logger:         Rails.logger
+  },
+  inputs:   %i(authenticator_name service_id host_name account)
+) do
+
+  def call
+    fetch_all_role_annotations
+    filter_authenticator_annotations
+    convert_to_resource_restrictions
+  end
+
+  private
+  
+  #...
+
+end
+```
+
+
+
+This is the case also in
+[`authn-k8s`](/app/domain/authentication/authn_k8s/resource_restrictions.rb), 
+with only one difference as the K8s Authenticator also looks in the host's ID.
+To support that, this class can be expanded to extract the constraints from the
+host ID, similar to:
+
+```ruby
+ExtractK8sResourceRestrictions = CommandClass.new(
+  dependencies: {
+    role_class:                     ::Role,
+    resource_class:                 ::Resource,
+    extract_resource_restrictions:  ExtractResourceRestrictions.new,
+    logger:                         Rails.logger
+  },
+  inputs:    %i(authenticator_name service_id host_name account)
+) do
+
+  def call
+    extract_resource_restrictions_from_annotations
+    extract_resource_restrictions_from_host_id if resource_restrictions.empty?
+    resource_restrictions
+  end
+
+  private
+  
+  #...
+
+end
+```
+
+This class can be used to replace the `ExtractResourceRestrictions` dependency
+class for k8s (this is command-class composition to mimic inheritance behavior).
+
+#### 2. Validate the Resource Restrictions are defined correctly
+
+In the Azure Authenticator and in the K8s Authenticator we have constraints on 
+how the annotations are defined.
+
+ As can be seen 
+[here](/app/domain/authentication/authn_azure/validate_resource_restrictions.rb), 
+we have the following validations:
+
+  - Validate only permitted restrictions are defined
+    - Here we validate that the `authn-azure` annotations are known. For
+      example, the annotation `authn-azure/subscription-id` is valid, whereas
+      the annotation `authn-azure/unknown` is invalid. We do that to minimize
+      errors from the user, where they will define, for example, the
+      annotation `authn-azure/subscription-ir` (note the typo). In this
+      case we'd rather fail the authentication and point them to their
+      mistake rather than ignore it, and authenticate the request in a weaker
+       granularity than what is expected by the user.
+  - Validate required restrictions exist
+    - For example, in the Azure Authenticator we require hosts to have the
+     `subscription-id` and the `resource-group` annotations. In the K8s
+      Authenticator we require the `namespace` annotation.
+  - Validate mutually exclusive restrictions
+    - For example, in the Azure Authenticator we allow hosts to have
+     only one of the `user-assigned-identity` and the `system-assigned-identity`
+     annotations. In the K8s we allow hosts to have only one of
+      the annotations `deployment`, `deployment-config` & `stateful-set`.
+
+Also here we can see some generalization.
+
+These need to be defined for each authenticator.
+To simplify that, we can define a simple class for each of these constraints.
+The whole constraints will be an array of these constraints, defined once for each authenticator.
+Then, to validate the Resource Restrictions, it only needs to loop on these constraints, and invoke
+`constraint.validate(resource_restrictions)` on each.
+
+This way, adding or changing the logic is more simple and intuitive, while maintaining readability.
+
+Each of the constraints classes will have the same `validate` signature, and will look similar to:
+
+```ruby
+class RequiredConstraint
+
+  def initialize(required:)
+    @required = required
+  end
+
+  def validate(resource_restrictions:)
+    # implement the constraint
+  end
+
+end
+```
+
+It will be used by each authenticator to define how to validate the configured
+restrictions. These are the ones that were defined in the policy for the host,
+regardless of the request.
+
+To combine multiple constraints, there will be another `Constraint` class:
+
+```ruby
+class MultipleConstraint
+  def initialize(*args)
+    @constraints = args
+  end
+
+  def validate(resource_restrictions:)
+    @constraints.each do |constraint|
+      constraint.validate(resource_restrictions)
+    end
+  end
+end
+```
+
+Example how it will look like for `authn-azure`:
+
+```ruby
+REQUIRED = [SUBSCRIPTION_ID, RESOURCE_GROUP]
+IDENTITY_EXCLUSIVE = [USER_ASSIGNED_IDENTITY, SYSTEM_ASSIGNED_IDENTITY]
+PERMITTED = REQUIRED + IDENTITY_EXCLUSIVE
+
+CONSTRAINTS = MultipleConstraint.new(
+        RequiredConstraint.new(required: REQUIRED),
+        PermittedConstraint.new(permitted: PERMITTED),
+        ExclusiveConstraint.new(exclusive: IDENTITY_EXCLUSIVE)
+    )
+```
+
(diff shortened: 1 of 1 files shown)


design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,350 @@
...
+    role_class:                  ::Role,
+    resource_class:              ::Resource,
+    logger:                      Rails.logger
+  },
+  inputs:       %i(authenticator_name service_id host_name account)
+) do
+
+  def call
+    ensure_role_exists
+    init_role_annotations
```
sigalsax, line 118: `init_role_annotations` what does this call do? does it initialize some role annotation object?

sigalsax: or fetch the annotations for the defined host/user?

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,350 @@
...
+  - Validate mutually exclusive restrictions
+    - For example, in the Azure Authenticator we allow hosts to have
+     only one of the `user-assigned-identity` and the `system-assigned-identity`
+     annotations. In the K8s we allow hosts to have only one of
+      the annotations `deployment`, `deployment-config` & `stateful-set`.
+
+Also here we can see some generalization.
+
+These need to be defined for each authenticator.
+To simplify that, We can define a simple class to aggregate this information.
```
sigalsax, line 197: lowercase 'w' in We 🙃

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,350 @@
...
+It will be used by each authenticator to define how to validate the configured
+restrictions. These are the ones that were defined in the policy for the host,
+regardless of the request.
+
+Example how it will look like for `authn-azure`:
+
+```ruby
+@constraints = ResourceRestrictionsConstraints.new
+  .add(restriction_name: "subscription-id", 		 required: true)
+  .add(restriction_name: "resource-group", 			 required: true)
```
sigalsax, line 232: spacing is a bit off between the two key-value pairs

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: It gets **all** the role annotations defined for the `host_name` in the policy.

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: 👍

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,355 @@
...
+  .add(restriction_name: "user-assigned-identity",   exclusive_group_name: "identity")
+  .add(restriction_name: "system-assigned-identity", exclusive_group_name: "identity")
```
sigalsax, line 239: is there a case where `exclusive_group_name` would be different than "identity"? In other words would there be more than one `exclusive_group_name` in our current usecase? Or are you predicting a future scenario where there will be one?

sigalsax (commented):
Left some comments but very well written (:

design/authenticators/resource-restrictions/resource-restrictions.md
sigalsax: so many `fetch_all_role_annotations` would be better from a naming standpoint

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: That's a great question!
While I wrote this class, it bothered me a lot to generalize it to support a single exclusive group.
There's no reason to limit it in such a way.
Still,  the effort to support multiple groups is worth it so the API won't change once this is needed.
Also, this is a very small change, and it makes the code more readable, IMO.

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,355 @@
...
+  },
+  inputs:   %i(authenticator_name service_id host_name account)
+) do
+
+  def call
+    ensure_role_exists
+    fetch_all_role_annotations
+    filter_authenticator_annotations
+    convert_to_resource_restrictions
+    resource_restrictions
```
eranha, line 126: What does `resource_restrictions` do?

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,355 @@
...
+  dependencies: {
+    role_class:     ::Role,
+    resource_class: ::Resource,
+    logger:         Rails.logger
+  },
+  inputs:   %i(authenticator_name service_id host_name account)
+) do
+
+  def call
+    ensure_role_exists
```
eranha, line 122: It is not the responsibility of the extractor to ensure that the role exists.

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,355 @@
...
+```ruby
+ValidateResourceRestrictions = CommandClass.new(
+  dependencies: {
+    logger: Rails.logger
+  },
+  inputs:   %i(resource_restrictions constraints)
+) do
+
+  def call
+    validate_restrictions_are_permitted
```
eranha, line 264: This method ensures that the annotation names are valid?

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,355 @@
...
+```ruby
+ValidatePropertiesMatchResourceRestrictions = CommandClass.new(
+  dependencies: {
+    logger: Rails.logger
+  },
+  inputs:   %i(resource_restrictions properties)
+) do
+
+  def call
+    validate_properties_match_resource_restrictions
```
eranha, line 308: `validate_request_parameters_match_resource_restrictions `

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,355 @@
...
+    validate_request:               ValidatePropertiesMatchResourceRestrictions.new,
+    logger:                         Rails.logger
+  },
+  inputs:   %i(authenticator_name service_id host_name account constraints request_properties)
+) do
+
+  def call
+    extract_resource_restrictions
+    validate_extracted_resource_restrictions
+    validate_request_matches_resource_restrictions
```
eranha, line 346: `validate_request_parameters_matches_resource_restrictions `

eranha (changes requested):
Imppressive

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,355 @@
...
+    @mutually_exclusive = {}
+  end
+
+  def add(restriction_name:, required: false, exclusive_group_name: nil)
+    @permitted << restriction_name
+    @required << restriction_name if required
+    (@mutually_exclusive[exclusive_group_name] ||= []) << restriction_name if exclusive_group_name.present?
+    self
+  end
+end
```
jonahx, line 225: I like this class dedicated to defining the logic of the resource restrictions.

Really, this is a good example of a value object.  It can be improved by making it immutable.  That is, remove `add` and force the entire thing to be initialized with a single hash defining the three properties.

Than you can add a `validate` method which takes the actual annotation data and validates against the logical definition.  Very similar to what you're proposing here and with the objects below, just a slight difference in the object's API.

nessiLahav (dismissed):
Well done

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: I removed the `add` as you suggested.
Still, our approach is to put the logic in command classes, and not inside the classes themselves.

nessiLahav (dismissed):

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: return it.
In Ruby, the last sentence is the return value.

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: moved, thanks.

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: I finally took @nessiLahav suggestion to create a command-class for each constraint with the same signature.
The command class get the list of configuration in its ctor, and validates it against the Resource Restrictions in the `call`.
This way, each authenticator defines its array of constraints, and the validation only loops through it and calls them.
It allows defining more of these and combining them as necessary in the future.

design/authenticators/resource-restrictions/resource-restrictions.md
eranha: Naming is not clear to me.

design/authenticators/resource-restrictions/resource-restrictions.md
jonahx: > Still, our approach is to put the logic in command classes, and not inside the classes themselves.

What's your reasoning for wanting to do this?

The problem I see here is that this isn't the intended use of the `CommandClass`.  You don't want to be creating command classes for pure, functional logic, which afaict is what this constraint validation is. 

The only reason you might want a command class here is for the logging: If you want to to log various constraint violations in the same way for all (or multiple) authenticators.  But even there, I think the cleaner design would be to have a pure value object for the constraint logic, which raises errors, and then a very think command class, with the logger, that delegates to the value object and does nothing but map constaint validation errors to log messages.  This keep each object very focused on one responsibility.  Also, the unit tests for the validation logic will be cleaner this way.

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: If I wrote `return resource_restrictions` it was clear?

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: Thank you @jonahx!
I really appreciate your feedback, and would like to hear more to understand better.

>this isn't the intended use of the `CommandClass`.

So what IS the intended use of the `CommandClass`? Could you elaborate? I would like to know so I could use it better.

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: removed.

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: I changed the constraint classes as you suggested, you can check it out.

eranha (approved):

design/authenticators/resource-restrictions/resource-restrictions.md
jonahx: Cool, I will check it out tomorrow and add feedback.

> So what IS the intended use of the CommandClass? Could you elaborate? I would like to know so I could use it better.

In the meantime, I wrote [this document](https://github.com/cyberark/conjur/blob/master/app/domain/README.md) which tries to answer that question.   This PR might be a good learning experience for me too, re: which parts of that document need to be fleshed out with more examples.  Just glancing over it again, I notice I don't go into how you'd decide when to use a value objects vs a command object, and that's something that would be useful to have some examples for.

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,350 @@
...
+* Return a `resource_restrictions` hash of all the restrictions, without their prefix.
+
+In this case the general class will be defined as follows:
+- input
+  - authenticator name
+  - authenticator service-id
+  - account
+  - host name
+- output
+  - `resource_restrictions` hash.
```
orenbm, line 106: How about returning a ValueObject instead of a hash?

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,350 @@
...
+    resource_class:                 ::Resource,
+    extract_resource_restrictions:  ExtractResourceRestrictions.new,
+    logger:                         Rails.logger
+  },
+  inputs:    %i(authenticator_name service_id host_name account)
+) do
+
+  def call
+    extract_resource_restrictions_from_annotations
+    extract_resource_restrictions_from_host if resource_restrictions.empty?
```
orenbm, line 155: this should be `extract_resource_restrictions_from_host_id`, correct?

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,350 @@
...
+- input
+  - An array of constraints, command classes that accept `resource_restrictions` as input.
+  - The `resource_restrictions` hash built in the previous step
+- output
+  - Nothing, relevant error raised if it failed
+
+So the class signature is:
+
+```ruby
+ValidateResourceRestrictions = CommandClass.new(
```
orenbm, line 253: i think that `ValidateResourceRestrictionsConfiguration` is a better naming.

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,350 @@
...
+  dependencies: {
+    logger: Rails.logger
+  },
+  inputs:   %i(resource_restrictions constraints)
+) do
+
+  def call
+    validate_restrictions_meet_constraints
```
orenbm, line 261: this is an indication that the class name should be changed. It should really tell the essence of the `call` function (which tells the story of the class).

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,350 @@
...
+
+So the general class will be defined as follows:
+- input
+  - Host Resource Restrictions object
+  - Request's Properties object
+- output
+  - Nothing, raise an error if failed
+  
+```ruby
+ValidateParametersMatchResourceRestrictions = CommandClass.new(
```
orenbm, line 295: not sure that `parameters` is very intuitive. How about `ValidateRequestMatchsResourceRestrictions`?

design/authenticators/resource-restrictions/resource-restrictions.md
jonahx: @abrahamko 

> I changed the constraint classes as you suggested, you can check it out.

The `RequiredConstraint` looks good now, but I liked how you combined them all before too: "all the constraints" seems like a natural way to think of it.  You could achieve that either by combining them into one object like you had before, or by creating an aggregate object around the other three types of objects.  It's hard to say which I'd prefer though without seeing the actual code.

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,350 @@
...
+REQUIRED = [SUBSCRIPTION_ID, RESOURCE_GROUP]
+IDENTITY_EXCLUSIVE = [USER_ASSIGNED_IDENTITY, SYSTEM_ASSIGNED_IDENTITY]
+PERMITTED = REQUIRED + IDENTITY_EXCLUSIVE
+
+CONSTRAINTS =
+    [
+        RequiredConstraint.new(required: REQUIRED),
+        PermittedConstraint.new(permitted: PERMITTED),
+        ExclusiveConstraint.new(exclusive: IDENTITY_EXCLUSIVE)
+    ]
```
jonahx, line 239: A small point, but I don't see much benefit to introducing indirection via constants here.  That is, I'd rather see an object with actual values and no indirection, something more or less like (I'm handwaving over details/names):
```ruby
CONSTRAINTS = Constraints.new(
  required: [:subscription_id, :resource_group],
  permitted: [:user_assigned_identity, system_assigned_identity],
  exclusive: [:foo, :bar]
)
```
In case I'm missing something, the way I'm thinking about this is asking, "What is the reason to use constants like these?"  There are two typical answers:
1. We need to expose the values to other code.
2. We want a layer of indirection so that we can refer to the constant in multiple places within our class, but update its value in a single place.  That is, if the value of the constant changes, we want to change it once, and have that reflected everywhere.

However, if you only use the constant's value in one place, reason 2 does not apply, and it's better not to use the constant in that case, because the indirection has a cost in readability and simplicity.  You only want to pay that cost when you have to.

abrahamko: > You only want to pay that cost when you have to.

I totally agree.
But when I looked deeply into the code I saw that actually, both of the reasons apply here.
1. These constants represent strings and they are used to create a mapping to values in the request.
That's the 3rd step, where we compare the resource restrictions values with the request values.
2. `Permitted` is the sum of all other constraints, plus optional if there are any.
So defining them once prevents the necessity to write the same strings multiple times.

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: done

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: done

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: removed

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: removed

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: How do you like it now?

orenbm: not really sure where the new one is but i'll see it in the implementation PR so don't bother.

design/authenticators/resource-restrictions/resource-restrictions.md
abrahamko: I wrapped them in a `MultipleConstraint` object.
How's that?

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,352 @@
...
+  def initialize(*args)
+    @constraints = args
+  end
+
+  def validate(resource_restrictions:)
+    @constraints.each do |constraint|
+      constraint.validate(resource_restrictions)
+    end
+  end
+end
```
jonahx, line 265: Perfect.

nessiLahav (approved):

design/authenticators/resource-restrictions/resource-restrictions.md
```diff
@@ -0,0 +1,352 @@
...
+In future authenticators we will not develop an option to define it in the ID
+and it is available in the K8s Authenticator only for backwards compatibility.
+
+### Defining Resource Restrictions Authentication
+
+The Resource Restrictions authentication is the major flow of actions needed
+to validate a request using Resource Restrictions.
+It can be split into 3 parts:
+
+ 1. Extract the Resource Restrictions from the host
```
oburstein-hub, line 81: Please, elaborate on examples of resources that can be extracted

Issue #1815, opened 2020-09-08
Refactoring authenticators code - Refactor GCP ValidateResourceRestrictions 


nessiLahav:
Refactor GCP `ValidateResourceRestrictions` component according to LDD design

**DoD**
- [x] GCP automated tests are finish successfully
- [x] We have a better code

izgeri:
@nessiLahav what is LDD design? Do you have a link?

PR #1918, opened 2020-11-01
Refactor GCP resource restrictions

abrahamko:
### What does this PR do?
Implementation of the design written in [refactor resource restriction document](https://github.com/cyberark/conjur/blob/master/design/authenticators/resource-restrictions/resource-restrictions.md).
This is the first step, introducing the common code and behavior, and replacing GCP implementation with the new one.
Doing the same for azure and K8s will be in separate PRs, due to the magnitude of this change.

First step of #1815

#### Change log

#### Test coverage

#### Documentation


abrahamko:
@nessiLahav @orenbm @eranha , you can start reviewing.
The only thing missing here is the RSpec UTs.

app/domain/authentication/authn_gcp/authentication_request.rb
```diff
@@ -0,0 +1,30 @@
...
+  module AuthnGcp
+
+    # This class is responsible for retrieving the correct value from the GCP token
+    # of the requested attribute.
+    class AuthenticationRequest
+      def initialize(decoded_token:)
+        @decoded_token = decoded_token
+      end
+
+      def retrieve_attribute(attribute_name)
```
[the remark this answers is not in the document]
abrahamko: won't fix

app/domain/authentication/common/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,106 @@
...
+        end
+      end
+
+      def create_resource_restrictions_object(resource_restrictions_hash)
+        @resource_restrictions_class.new(
+          resource_restrictions_hash: resource_restrictions_hash
+        )
+      end
+
+      def add_restriction_to_hash(annotation_name, annotation_value, resource_restrictions_hash)
```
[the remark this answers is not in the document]
abrahamko: won't fix.
I worked hard to satisfy this suggestion. This is the best I could do without losing value.

app/domain/authentication/common/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,100 @@
...
+      dependencies: {
+        resource_restrictions_class: ResourceRestrictions,
+        role_class:                  ::Role,
+        resource_class:              ::Resource,
+        logger:                      Rails.logger
+      },
+      inputs:   %i(authenticator_name service_id host_name account)
+    ) do
+
+      def call
```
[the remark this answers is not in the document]
abrahamko: Won't fix.
2 log lines and a return call... left with 3 real statements.

app/domain/authentication/authn_gcp/consts.rb
```diff
@@ -4,11 +4,22 @@ module Authentication
   module AuthnGcp
 
     PROVIDER_URI = "https://accounts.google.com"
-    PROJECT_ID_RESTRICTION_NAME = "project-id"
-    INSTANCE_NAME_RESTRICTION_NAME = "instance-name"
-    SERVICE_ACCOUNT_ID_RESTRICTION_NAME = "service-account-id"
-    SERVICE_ACCOUNT_EMAIL_RESTRICTION_NAME = "service-account-email"
-    PERMITTED_CONSTRAINTS = %w(instance-name project-id service-account-id service-account-email).freeze
+
+    module Restrictions
```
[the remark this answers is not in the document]
abrahamko: I don't see anything wrong with that. That's the purpose of this file, to contain the consts.

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -5,27 +5,37 @@ module AuthnGcp
...
       def call
-        validate_resource_restrictions
+        authentication_request = create_authentication_request
+        validate_resource_restrictions(authentication_request)
       end
 
       private
 
-      def validate_resource_restrictions
+      def create_authentication_request
```
orenbm, line 25: the method name should be `authentication_request` to indicate the output.

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -5,27 +5,37 @@ module AuthnGcp
...
+        @authentication_request_class.new(
```
orenbm, line 26: do we want to use memoization? this way you can remove the line `authentication_request = create_authentication_request` from the `call` method.

orenbm: you could also remove `authentication_request` argument from the `validate_resource_restrictions` method this way.

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -5,27 +5,37 @@ module AuthnGcp
 
     Authenticator = CommandClass.new(
```
orenbm, line 6: this is out of scope from your PR but it would be good to add here a comment that part of the authentication process for authn-gcp is to validate the token but it happens elsewhere. just for clarity. something similar to the comment we have for oidc: https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_oidc/authenticator.rb#L11

app/domain/authentication/common/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,106 @@
...
+        resource_restrictions = create_resource_restrictions_object(resource_restrictions_hash)
+
+        @logger.debug(LogMessages::Authentication::ExtractedResourceRestrictions.new(resource_restrictions.names))
+
+        resource_restrictions
+      end
+
+      private
+
+      def fetch_role_annotations
```
orenbm, line 33: this should be `def role_annotations`. or is it `fetch` to indicate that we have a critical action here of going to the db?

app/domain/authentication/common/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,106 @@
...
+        # Validate role exists, otherwise getting role annotations return empty hash.
+        role_id = @role_class.roleid_from_username(@account, @host_name)
+        host_resource = @resource_class[role_id]
+
+        raise Errors::Authentication::Security::RoleNotFound, role_id unless host_resource
+
+        get_resource_annotations(host_resource)
+      end
+
+      def get_resource_annotations(host_resource)
```
orenbm, line 43: this should be `def resource_annotations(host_resource)`

app/domain/authentication/common/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,106 @@
...
+        end
+      end
+
+      def extract_resource_restrictions_from_annotations(annotations)
+        annotations.each_with_object({}) do |(annotation_name, annotation_value), resource_restrictions_hash|
+          add_restriction_to_hash(annotation_name, annotation_value, resource_restrictions_hash)
+        end
+      end
+
+      def create_resource_restrictions_object(resource_restrictions_hash)
```
orenbm, line 58: `def resource_restrictions_object(resource_restrictions_hash)`

app/domain/authentication/common/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,106 @@
...
+      def get_resource_annotations(host_resource)
+        host_resource.annotations.each_with_object({}) do |annotation, result|
+          annotation_values = annotation.values
+          value = annotation_values[:value]
+          next if value.blank?
+          result[annotation_values[:name]] = value
+        end
+      end
+
+      def extract_resource_restrictions_from_annotations(annotations)
```
orenbm, line 52: `def resource_restrictions_from_annotations(annotations)` or `def extracted_resource_restrictions_from_annotations(annotations)`

app/domain/authentication/common/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,106 @@
...
+      # The first format is a general restriction.
+      # This means it is relevant to all services of this authenticator.
+      #
+      # The second format is specific to the specified service ID.
+      # This means all other services will ignore it.
+      # Moreover, if both formats are found for the same restriction, the specific one will be used.
+      #
+      # This function returns the restriction name, if found (nil otherwise) and a
+      # boolean indicating if it is a general restriction or not.
+      def get_restriction_from_annotation(annotation_name)
```
orenbm, line 91: `def restriction_from_annotation(annotation_name)` or `def parsed_restriction_from_annotation(annotation_name)`

app/domain/authentication/common/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,58 @@
...
+      private
+
+      def extract_resource_restrictions
+        @extract_resource_restrictions.call(
+          authenticator_name: @authenticator_name,
+          service_id:         @service_id,
+          host_name:          @host_name,
+          account:            @account
+        )
+      end
```
orenbm, line 33: i would change this to:
```
	resource_restrictions
end

def resource_restrictions
  @resource_restrictions ||= @extract_resource_restrictions.call(
	  authenticator_name: @authenticator_name,
	  service_id:         @service_id,
	  host_name:          @host_name,
	  account:            @account
	)
end
```

this way you don't need a return object and can use `resource_restrictions` as a memoized object. it fits our style better.

orenbm: the `extract_resource_restrictions` method is used to tell the story in the `call` function and the `resource_restrictions` method is used for memoization

app/domain/authentication/common/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      def validate_resource_restrictions_configuration(resource_restrictions)
+        @logger.debug(LogMessages::Authentication::ValidatingResourceRestrictionsConfiguration.new)
+
+        @constraints.validate(
```
orenbm, line 38: i think we should add a comment on what we expect `@constraints` to be as it's an input parameter and not a dependency. Future developer will need to explicitly create this object themselves so the documentation on it should be explicit.

app/domain/authentication/common/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,58 @@
...
+          resource_restrictions: resource_restrictions.names
```
orenbm, line 39: maybe call this params `resource_restrictions_names`?

app/domain/authentication/constraints/any_constraint.rb
```diff
@@ -0,0 +1,19 @@
+module Authentication
+  module Constraints
```
orenbm, line 2: should this module go under the `Common` module?

app/domain/authentication/constraints/any_constraint.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    # This constraint validates that at least one of the initialized restrictions exist.
+    # So, any one of them must be found in the given restrictions when calling validate(given_restrictions).
```
orenbm, line 5: nice comment, thanks

app/domain/authentication/constraints/permitted_constraint.rb
```diff
@@ -0,0 +1,23 @@
...
+    # This constraint validates that only the initialized restrictions are permitted.
+    # So, no other restrictions can be found in the given restrictions when calling validate(given_restrictions).
+    class PermittedConstraint
+
+      def initialize(permitted:)
+        @permitted = permitted
+      end
+
+      def validate(resource_restrictions:)
+        (resource_restrictions - @permitted).each do |not_supported_restriction_name|
```
orenbm, line 13: consider splitting this and adding a line `not_supported_restrictions = resource_restrictions - @permitted` to ease readability

orenbm (changes requested):
Thanks, this PR is great!

I left some comments with requests.

Additionally, do you think we can benefit from a README for future authenticators? something that will explain what a future developer should implement in order to verify resource restrictions. maybe even create a template.

app/domain/authentication/common/extract_resource_restrictions.rb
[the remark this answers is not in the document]
orenbm: feel free to ignore this comment in the future

app/domain/authentication/authn_gcp/authentication_request.rb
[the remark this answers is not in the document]
orenbm: feel free to ignore this comment in the future

app/domain/authentication/common/extract_resource_restrictions.rb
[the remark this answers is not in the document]
orenbm: feel free to ignore this comment in the future

app/domain/authentication/authn_gcp/authenticator.rb
abrahamko: I know, but I prefer to be explicit about the arguments of the method.
When it depends on input from another method, I want it to be explicit.
Also, I think the dependency between the steps is more obvious that way.

orenbm: but you don't need to refer to it as an input from another method. i think that we can still have the steps. something like:
```
def call
  create_authentication_request
  validate_resource_restrictions
end

private 

def create_authentication_request
  authentication_request
end

def authentication_request
  @authentication_request ||= ...
end

def validate_resource_restrictions
  # do something with authentication_request
end
```

This style matches the rest of our code

app/domain/authentication/authn_gcp/authenticator.rb
abrahamko: But this is not an accessor, it wraps an instance creation.
What's wrong with this name? to me it looks very intuitive.

app/domain/authentication/common/extract_resource_restrictions.rb
abrahamko: the latter, yes.

app/domain/authentication/authn_gcp/authenticator.rb
orenbm: if a method returns an object (as this method does) then we give it a name that indicates the returned object. in this case it's `authentication_request`. 

Here i would have both methods `create_authentication_request` and `authentication_request` in the manner i suggested here: https://github.com/cyberark/conjur/pull/1918/files/58a5b8b4db11d7c66cc786020fa76332d823201e#r515952095

app/domain/authentication/constraints/any_constraint.rb
abrahamko: It can, but I like it in a separate folder since they are very isolated classes, with very common structure.
IMO it makes sense to keep them isolated in a different module/folder.

app/domain/authentication/common/validate_resource_restrictions.rb
abrahamko: Isn't it too long?
This string is all over the place... 😕

app/domain/authentication/constraints/any_constraint.rb
orenbm: i'm fine with whichever decision you take here. I was suggesting to have the structure `Authentication -> Common -> Constraints`. But again, i;m ok with the current structure

app/domain/authentication/common/validate_resource_restrictions.rb
orenbm: i'm ok with the current name, just a suggestion. whichever you prefer

abrahamko:
@orenbm you wrote:
> Additionally, do you think we can benefit from a README for future authenticators? something that will explain what a future developer should implement in order to verify resource restrictions. maybe even create a template.

I think a README will help, but might be an overkill.
As a developer, I would look for similar usages in other places in the code
and will mimic their behavior.
So basically, the template is already there, implemented 😉 .

app/domain/authentication/constraints/permitted_constraint.rb
abrahamko: done

abrahamko:
@orenbm @nessiLahav @eranha I splitted the PR into separate commits for easier review.

app/domain/authentication/common/validate_resource_restrictions.rb
abrahamko: done

app/domain/authentication/authn_gcp/authenticator.rb
abrahamko: done

app/domain/authentication/common/validate_resource_restrictions.rb
abrahamko: I say, keep it shorter. thanks for the comment!

app/domain/authentication/authn_gcp/authenticator.rb
abrahamko: Although I prefer the explicit way, I understand this style, and adhere to it.
Done.

app/domain/authentication/authn_gcp/authenticator.rb
abrahamko: Although I prefer the explicit way, I understand this style, and adhere to it.
Done.

app/domain/authentication/common/extract_resource_restrictions.rb
abrahamko: done

app/domain/authentication/common/extract_resource_restrictions.rb
abrahamko: done

app/domain/authentication/common/extract_resource_restrictions.rb
abrahamko: done

app/domain/authentication/common/extract_resource_restrictions.rb
abrahamko: done

app/domain/authentication/common/validate_resource_restrictions.rb
abrahamko: Although I prefer the explicit way, I understand this style, and adhere to it.
Done.

app/domain/authentication/authn_gcp/authenticator.rb
orenbm: thanks you very much, i appreciate it!

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,120 @@
...
+            result[annotation_values[:name]] = value
+          end
+      end
+
+      def host_resource
+        # Validate role exists, otherwise getting role annotations return empty hash.
+        role_id = @role_class.roleid_from_username(@account, @host_name)
+        host_resource = @resource_class[role_id]
+
+        raise Errors::Authentication::Security::RoleNotFound, role_id unless host_resource
```
sigalsax, line 52: again, I think host_resource is not very presentative of a name if we involve users

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,120 @@
...
+      # This means it is relevant to all services of this authenticator.
+      #
+      # The second format is specific to the specified service ID.
+      # This means all other services will ignore it.
+      # Moreover, if both formats are found for the same restriction, the specific one will be used.
+      #
+      # This function returns the restriction name, if found (nil otherwise) and a
+      # boolean indicating if it is a general restriction or not.
+      def get_restriction_from_annotation(annotation_name)
+        annotation_name.match(authenticator_prefix_regex) do |prefix_match|
```
sigalsax, line 96: what if there was something like `authn-gcp-sigal` would this still be valid?

app/domain/authentication/resource_restrictions/resource_restrictions.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      def initialize(resource_restrictions_hash:)
+        @resource_restrictions_hash = resource_restrictions_hash
+      end
+
+      def names
+        @resource_restrictions_hash.keys
+      end
+
+      def each(&block)
```
sigalsax, line 14: what is &block here?

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,78 @@
...
+  module ResourceRestrictions
+
+    # This command class validates the resource restrictions retrieved from the
+    # @extract_resource_restrictions command class dependency.
+    #
+    # Its inputs are:
+    # @authenticator_name - the name of the authenticator. e.g. authn-azure.
+    # @service_id - the specific service ID requested to authenticate. So if
+    #   the authenticator is `authn-azure/prod`, then `prod` is the service ID.
+    # @host_name - the name of the requesting Conjur host as defined in the DB.
```
sigalsax, line 13: host_name 🎶

sigalsax (commented):
left some comments and questions

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -3,29 +3,47 @@
...
+
       def validate_resource_restrictions
         @validate_resource_restrictions.call(
           authenticator_name: authenticator_name,
-          account:     account,
-          username:    username,
-          credentials: credentials
+          service_id: service_id,
+          account: account,
+          host_name: username,
```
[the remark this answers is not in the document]
abrahamko: same question goes for the `username`...
host is the common usecase, so I think it is better.
I can change it to `resource_name`, would that be better?
IMO it is a little too vague.

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -41,6 +59,7 @@ class Authenticator
       # `Call` objects created each time `call` is run.
       def valid?(input)
         call(authenticator_input: input)
+        true
```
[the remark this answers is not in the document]
abrahamko: by Ruby convention, methods that ends with `?` needs to return boolean.
Since we do not do it, we violate this convention.
Still, we do implicitly do it, because Ruby returns the result of the last statement.
Can you guess what it is?


It is `logger.debug()`. Which, in the logger we use, return `true`...

Instead of all this mess, we better be explicit to return true, since `valid?` will fail with raised error and not by returning `false`.

app/domain/authentication/authn_gcp/consts.rb
```diff
@@ -4,11 +4,22 @@ module Authentication
   module AuthnGcp
 
     PROVIDER_URI = "https://accounts.google.com"
```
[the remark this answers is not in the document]
abrahamko: Since it is a const, but not part of the resource restrictions.

app/domain/authentication/authn_gcp/consts.rb
```diff
@@ -4,11 +4,22 @@ module Authentication
...
+
+      PROJECT_ID = "project-id"
+      INSTANCE_NAME = "instance-name"
+      SERVICE_ACCOUNT_ID = "service-account-id"
+      SERVICE_ACCOUNT_EMAIL = "service-account-email"
+
+      PERMITTED = [PROJECT_ID, INSTANCE_NAME, SERVICE_ACCOUNT_ID, SERVICE_ACCOUNT_EMAIL].freeze
+
+      CONSTRAINTS = Constraints::MultipleConstraint.new(
+        Constraints::AnyConstraint.new(any_of: PERMITTED),
```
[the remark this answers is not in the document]
abrahamko: WDYM?
it creates a `AnyConstraint` with `PERMITTED` as its input.
It is created once per program lifetime.
Of course, if multiple authenticators will use it with their own parameters, it will be created once for each too.

app/domain/authentication/constraints/any_constraint.rb
```diff
@@ -0,0 +1,19 @@
...
+    class AnyConstraint
+
+      def initialize(any_of:)
+        @any = any_of
```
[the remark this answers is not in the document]
abrahamko: I prefer you do not think about it this way.
This class does not know anything about its callers.
All it knows (and it doesn't need to know more) is that it should validate that any one of the given restrictions must be found.
The caller decided that ANY of the permitted restrictions must be found, but its the caller who decided about this logic, not this class.
This class only implements it.

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
abrahamko: neither do `username`.
See my response above.

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -3,29 +3,47 @@
...
       def validate_resource_restrictions
         @validate_resource_restrictions.call(
           authenticator_name: authenticator_name,
-          account:     account,
-          username:    username,
-          credentials: credentials
+          service_id: service_id,
+          account: account,
+          host_name: username,
+          constraints: Restrictions::CONSTRAINTS,
```
sigalsax, line 45: why does constraints need to be passed in here and not just initialized in the calling class?

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
abrahamko: This class does not validate anything (as its name imply).
So any restriction is valid.
It might not be found, because part of this logic is to extract only the relevant resource restrictions.
So for `authn-gcp`, if you define an annotation `authn-gcp-sigal/project-id`, it won't be found.
But if you define `authn-gcp/sigal-project-id`, it will find the `sigal-project-id` resource restriction (whatever that means... 😉 )

app/domain/authentication/resource_restrictions/resource_restrictions.rb
abrahamko: This is a propagator for `each` method on enumerable.
This is how you define a method that accepts a block (everything that comes in a curly braces after method call):
```
[1, 2, 3].each {|n| puts n}
```

app/domain/authentication/authn_azure/validate_resource_restrictions.rb
```diff
@@ -36,7 +36,7 @@ def validate_resource_restrictions_match_request
         resource_restrictions.resources.each do |resource_from_role|
           resource_from_request = resources_from_request.find { |resource| resource == resource_from_role }
           unless resource_from_request
-            raise Errors::Authentication::Jwt::InvalidResourceRestrictions, resource_from_role.type
+            raise Errors::Authentication::InvalidResourceRestrictions, resource_from_role.type
```
orenbm, line 39: do we want a `ResourceRestrictions` module?

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -3,29 +3,47 @@
 module Authentication
   module AuthnGcp
 
+    # This authenticator validates that the received token matches the
+    # restrictions defined in the policy. But there is another validation step
```
orenbm, line 7: nit: the `But` is redundant

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,78 @@
...
+        )
+
+        @logger.debug(LogMessages::Authentication::ValidatedResourceRestrictionsConfiguration.new)
+      end
+
+      def validate_request_matches_resource_restrictions
+        @logger.debug(LogMessages::Authentication::ValidatingResourceRestrictionsValues.new)
+
+        resource_restrictions.each do |restriction_name, restriction_value|
+          if @authentication_request.retrieve_attribute(restriction_name) != restriction_value
```
sigalsax, line 69: this line is a bit clear and took me some time to understand the `restriction_value` is basically the value of the constants from the resulting restriction_name. Therefore, I think maybe `(annotation_restriction_name) != constant_restriction_name` is more clear because otherwise in my mind at least name and value are key -> value pairs

app/domain/authentication/authn_gcp/authenticator.rb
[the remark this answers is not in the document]
orenbm: `username` can actually be misleading. In Conjur, a host's username is `host/host_name`. so developers may think that they should input that for hosts.

i am OK with `username` or `host_name` but wouldn't go with `resource_name`.

app/domain/authentication/constraints/any_constraint.rb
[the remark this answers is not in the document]
orenbm: nit: i think that calling the attribute `any` and the argument `any_of` is a bit weird. I would choose one and go with it.

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
```diff
@@ -0,0 +1,120 @@
...
+        logger:                      Rails.logger
+      },
+      inputs:   %i(authenticator_name service_id host_name account)
+    ) do
+
+      def call
+        @logger.debug(LogMessages::Authentication::ExtractingRestrictionsFromResource.new(
+          @authenticator_name,
+          @host_name
+        ))
```
orenbm, line 20: please move the last `)` to a new line

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
orenbm: what about `role_resource`?

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
orenbm: isn't it the project id param for a Sigal Google Cloud Platform machine?

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
orenbm: `role_name`?

app/domain/authentication/authn_gcp/authenticator.rb
abrahamko: because it is different for each authenticator, and is not a dependency for `ValidateResourceRestrictions`.
The `ValidateResourceRestrictions` CommandClass just calls `validate(resource_restrictions)` on it, but it doesn't know (or care) which restrictions are they.

app/domain/authentication/authn_azure/validate_resource_restrictions.rb
abrahamko: @Tovli  commented to me (offline) that he thinks `Common` is a bad module name, as it ends as a garbage can for anything that it is not obvious where it belongs.
I'm all in for better name, though.
What do you have in mind?

app/domain/authentication/authn_gcp/authenticator.rb
abrahamko: done

app/domain/authentication/constraints/any_constraint.rb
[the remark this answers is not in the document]
eranha: The class represents a Constraint or validates a constraint?

app/domain/authentication/authn_gcp/authenticator.rb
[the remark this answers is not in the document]
eranha: To make it clearer define a var `valid = true` and return it.

app/domain/authentication/authn_azure/validate_resource_restrictions.rb
orenbm: i wasn't suggesting a `Common` module :)

app/domain/authentication/constraints/permitted_constraint.rb
```diff
@@ -0,0 +1,24 @@
...
+    # So, no other restrictions can be found in the given restrictions when calling validate(given_restrictions).
+    class PermittedConstraint
+
+      def initialize(permitted:)
+        @permitted = permitted
+      end
+
+      def validate(resource_restrictions:)
+        not_supported_restrictions = resource_restrictions - @permitted
+        not_supported_restrictions.each do |not_supported_restriction_name|
```
nessiLahav, line 14: Maybe instead of go over the list, we should just validate it is not empty and to throw 1 exception which contains all the not supported restrictions ?
For example in your case if we have 2 not supported restrictions we will throw an exception on the first one,
In my suggestion we will throw an exception with all the un supported restrictions

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
abrahamko: > in my mind at least name and value are key -> value pairs

It *is* a key-value pair.
I put a lot of effort to differentiate between 4 common confusing terms:
1. resource restrictions
2. constraints
3. annotations
4. request attributes

The way I separated them is:
- Annotations: as defined in the policy (not processed)
- Resource Restrictions: key-value mapping of restrictions that the request must adhere to to authenticate.
  It is derived from the Annotations by selecting only the ones begin with `authenticator-name/`, without the prefix.
  Its value is the annotation value.
- Constraints: list of rules that the Resource Restrictions needs to adhere to to be considered valid restrictions.
  This validation step can (and should) be done in policy load.
- Request attributes: should not be confused with Resource Restrictions, these are attributes of the request that are needed to validate the request.
  They are retrieved differently for each authenticator (JWT token/K8s API or else).
  The request is validated by making sure they adhere to the Resource Restrictions.
  That is, for each resource restriction, the restriction value equals the corresponding request attribute value.

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
abrahamko: `role` and `resource` are two different db entities.
Took me a while to understand the difference between them.
I think combining them will only confuse more.

app/domain/authentication/constraints/any_constraint.rb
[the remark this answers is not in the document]
abrahamko: It represent a constraint, it knows how to enforce this constraint with `validate(resource_restrictions)`.

app/domain/authentication/authn_gcp/authenticator.rb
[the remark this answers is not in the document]
abrahamko: IMO it is a clear ruby syntax.
It might be needed in other languages, but in ruby, the it is clear that
```
def is_valid?
  true
end
```
means that it is valid.

cucumber/authenticators_gcp/features/authn_gce_token_errors.feature
```diff
@@ -132,5 +132,5 @@ Feature: GCP Authenticator - GCE flow, test token error handling
     Then it is unauthorized
     And The following appears in the log after my savepoint:
     """
-    CONJ00068E Resource restriction 'instance-name' does not exists resource in JWT token. Verify that you configured the host with permitted restrictions. In case of Compute Engine token verify that you requested the token using 'format=full'
+    CONJ00068E Attribute 'instance-name' is missing from Google's JWT token. Verify that you configured the host with permitted restrictions. In case of Compute Engine token, verify that you requested the token using 'format=full'
```
eranha, line 135: Replace Attribute with Claim.

cucumber/authenticators_gcp/features/authn_gcf_token_errors.feature
```diff
@@ -81,7 +81,7 @@ Feature: GCP Authenticator - GCF flow, test token error hwahtandling
     Then it is unauthorized
     And The following appears in the log after my savepoint:
     """
-    CONJ00068E Resource restriction 'project-id' does not exists resource in JWT token
+    CONJ00068E Attribute 'project-id' is missing from Google's JWT token
```
eranha, line 84: Replace Attribute with Claim.

cucumber/authenticators_gcp/features/authn_gcf_token_errors.feature
```diff
@@ -93,5 +93,5 @@ Feature: GCP Authenticator - GCF flow, test token error hwahtandling
     Then it is unauthorized
     And The following appears in the log after my savepoint:
     """
-    CONJ00068E Resource restriction 'instance-name' does not exists resource in JWT token
+    CONJ00068E Attribute 'instance-name' is missing from Google's JWT token
```
eranha, line 96: Replace Attribute with Claim.

app/domain/authentication/authn_gcp/authenticator.rb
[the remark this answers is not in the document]
orenbm: i agree with @abrahamko on this

eranha (changes requested):

app/domain/authentication/authn_azure/validate_resource_restrictions.rb
abrahamko: I want a module to group these classes.
This name seems more specific for its internals, so I think it is a good name.

app/domain/authentication/authn_gcp/authentication_request.rb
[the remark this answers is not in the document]
abrahamko: Think about it as an interface.
How would you define this method's signature?
As I see it, every request has attributes.
For GCP, these attributes are retrieved from the token claims, but that's internal implementation.
The caller doesn't care how and from where the values are brought.

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
```diff
@@ -0,0 +1,79 @@
...
+      },
+      inputs:   %i(authenticator_name service_id role_name account constraints authentication_request)
+    ) do
+
+      def call
+        @logger.debug(LogMessages::Authentication::ValidatingResourceRestrictions.new(@role_name))
+
+        extract_resource_restrictions
+        validate_resource_restrictions_configuration
+        validate_request_matches_resource_restrictions
```
nessiLahav, line 36: How do you plan to reuse this logic in other places ?

app/domain/authentication/authn_azure/validate_resource_restrictions.rb
orenbm: what am i missing? the object is `Errors::Authentication::InvalidResourceRestrictions` so there's no specific module.

app/domain/authentication/authn_gcp/authentication_request.rb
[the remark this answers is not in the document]
orenbm: @abrahamko do we have some way to enforce it? so that an `AuthenticationRequest` class will always implement this method?

also - what happens if it doesn't? which error will be raised?

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -3,29 +3,47 @@
...
   module AuthnGcp
 
+    # This authenticator validates that the received token matches the
+    # restrictions defined in the policy. There is another validation step
+    # where we also verify that the token is valid (JWT-wise) but this is
+    # done while it is decoded (using a third-party).
     Authenticator = CommandClass.new(
       dependencies: {
-        validate_resource_restrictions: ValidateResourceRestrictions.new,
+        validate_resource_restrictions: Authentication::ResourceRestrictions::ValidateResourceRestrictions.new,
```
orenbm, line 12: here i see that you have a `ResourceRestrictions` module. it can be added to the error classes as well.

app/domain/authentication/constraints/any_constraint.rb
orenbm: @abrahamko can i get an indication that you read this? if you still decide not to do it it's ok but i want to verify that you considered it.

app/domain/authentication/constraints/any_constraint.rb
```diff
@@ -0,0 +1,18 @@
+module Authentication
+  module Constraints
+
+    # This constraint validates that at least one of the initialized restrictions exist.
+    # So, any one of them must be found in the given restrictions when calling validate(given_restrictions).
+    class AnyConstraint
+
+      def initialize(any:)
```
orenbm, line 8: @abrahamko what does this class expect `any` to be? is a string? maybe a hash? if so - a hash of what?

i think that a comment is in place

app/domain/authentication/constraints/permitted_constraint.rb
```diff
@@ -0,0 +1,23 @@
+module Authentication
+  module Constraints
+
+    # This constraint validates that only the initialized restrictions are permitted.
+    # So, no other restrictions can be found in the given restrictions when calling validate(given_restrictions).
+    class PermittedConstraint
+
+      def initialize(permitted:)
```
orenbm, line 8: same here. what do we expect `permitted` to be?

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
abrahamko: Simple:
Any authenticator needs 2 things:
1. Constraints collection
2. AuthenticationRequest object
All the rest is handled internally.

For K8s, there's more effort to support extracting the resource restrictions from the host ID.
This will be supplied by overriding the dependency with a wrapper as demonstrated in the design doc.

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
orenbm: @abrahamko does this data need to be written somewhere? will it help future devs?

app/domain/logs.rb
```diff
@@ -64,6 +64,36 @@ module Authentication
...
+    ValidatingResourceRestrictionsValues = ::Util::TrackableLogMessageClass.new(
+      msg: "Validating resource restriction values",
+      code: "CONJ00044D"
+    )
+
+    ValidatedResourceRestrictionsValues = ::Util::TrackableLogMessageClass.new(
+      msg: "Resource restriction values validated",
+      code: "CONJ00045D"
+    )
+
```
orenbm, line 96: it seems that you have a `ResourceRestrictions` module for other classes. can these classes be in that module as well?

orenbm: same goes for errors

app/domain/authentication/authn_gcp/authenticator.rb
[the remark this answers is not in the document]
abrahamko: done

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
abrahamko: renamed to `resource`.

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
abrahamko: done

app/domain/authentication/authn_azure/validate_resource_restrictions.rb
abrahamko: done

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
abrahamko: @orenbm @sigalsax Is there a doc or something I can put this info in?

app/domain/authentication/constraints/permitted_constraint.rb
abrahamko: done

app/domain/authentication/authn_gcp/authentication_request.rb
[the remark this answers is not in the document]
abrahamko: 1st question: AFAIK no. Ruby does not have such mechanisms.
2nd question: If it doesn't, (or if the args does not match) it throws NoMethodError.

To handle both cases, integration tests will cover it.

app/domain/authentication/authn_gcp/authenticator.rb
```diff
@@ -3,29 +3,47 @@
 module Authentication
   module AuthnGcp
 
+    # This authenticator validates that the received token matches the
+    # restrictions defined in the policy. But there is another validation step
+    # where we also verify that the token is valid but this is done while it
```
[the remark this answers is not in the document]
abrahamko: done

app/domain/errors.rb
```diff
@@ -363,25 +352,34 @@ module AuthnGcp
         code: "CONJ00067E"
       )
 
-      ResourceRestrictionNotFoundOrEmpty = ::Util::TrackableErrorClass.new(
-        msg:  "Resource restriction '{0-resource-restriction-name}' does not exists " \
-            "resource in JWT token. " \
-            "Verify that you configured the host with permitted restrictions. " \
-            "In case of Compute Engine token verify that you requested the token using 'format=full'",
+      JwtTokenAttributeIsMissing = ::Util::TrackableErrorClass.new(
```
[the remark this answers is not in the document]
abrahamko: done

cucumber/authenticators_gcp/features/authn_gce_token_errors.feature
```diff
@@ -106,7 +106,7 @@ Feature: GCP Authenticator - GCE flow, test token error handling
     Then it is unauthorized
     And The following appears in the log after my savepoint:
     """
-    CONJ00068E Resource restriction 'service-account-email' does not exists resource in JWT token. Verify that you configured the host with permitted restrictions. In case of Compute Engine token verify that you requested the token using 'format=full'
+    CONJ00068E Attribute 'service-account-email' is missing from Google's JWT token. Verify that you configured the host with permitted restrictions. In case of Compute Engine token, verify that you requested the token using 'format=full'
```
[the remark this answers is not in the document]
abrahamko: done

cucumber/authenticators_gcp/features/authn_gce_token_errors.feature
```diff
@@ -119,7 +119,7 @@ Feature: GCP Authenticator - GCE flow, test token error handling
     Then it is unauthorized
     And The following appears in the log after my savepoint:
     """
-    CONJ00068E Resource restriction 'project-id' does not exists resource in JWT token. Verify that you configured the host with permitted restrictions. In case of Compute Engine token verify that you requested the token using 'format=full'
+    CONJ00068E Attribute 'project-id' is missing from Google's JWT token. Verify that you configured the host with permitted restrictions. In case of Compute Engine token, verify that you requested the token using 'format=full'
```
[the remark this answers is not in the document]
abrahamko: done

cucumber/authenticators_gcp/features/authn_gce_token_errors.feature
abrahamko: done

cucumber/authenticators_gcp/features/authn_gcf_token_errors.feature
abrahamko: done

cucumber/authenticators_gcp/features/authn_gcf_token_errors.feature
abrahamko: done

app/domain/authentication/constraints/any_constraint.rb
abrahamko: Comment added on top of class

app/domain/authentication/constraints/permitted_constraint.rb
abrahamko: Comment added on top of class

app/domain/logs.rb
abrahamko: done

app/domain/authentication/authn_gcp/consts.rb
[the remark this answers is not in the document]
sigalsax: so AnyConstraint will look like `[PROJECT_ID, INSTANCE_NAME, SERVICE_ACCOUNT_ID, SERVICE_ACCOUNT_EMAIL]`

still do not understand what any_of does here 😄  please explain 🙏

app/domain/authentication/constraints/any_constraint.rb
abrahamko: Thanks for the suggestion, and I decided to keep it in the `authentication1 folder.

app/domain/authentication/authn_gcp/consts.rb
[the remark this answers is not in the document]
abrahamko: I added a comment on the class.
Is it clear now?

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
abrahamko: done

app/domain/authentication/authn_gcp/authenticator.rb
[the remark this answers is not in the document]
eranha: In a single line function it is more clear.

eranha (approved):

orenbm (approved):
Great work!

app/domain/authentication/resource_restrictions/resource_restrictions.rb
sigalsax: oh didn't know that!

Why would you use this function over just using `hash.each({...})` in-place for a all ResourceRestrictions instances?

app/domain/authentication/authn_gcp/authenticator.rb
sigalsax: If this is the case do we plan on also having a Restrictions module call azure and k8s?

why can't you just have a line at the top of `authenticator.rb` like so:

```
module Authentication
  module AuthnGcp
  CONSTANTS = ....
```

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
sigalsax: So if I am understanding correctly, resource restrictions is intended to be a hash map of acceptable_resource_restriction -> user_given_resource_restriction ?

sigalsax: > Is there a doc or something I can put this info in?

Well to start, it should be in the code if not already and I think it should be given it's own category or own link here: https://github.com/cyberark/conjur/tree/master/design. It could go under it's own title "Things to be aware when creating a design" or "common confusions" or "key terms"

PR #1919, opened 2020-11-01
Add cucumber azure logs

### LINKED ISSUE #1916: authn-azure logs are missing in the CI pipeline artifacts
## Summary
When CI pipeline run, it should contain the authn-azure logs, similar to the other tests, but they are missing:
image

## Steps to Reproduce
1. Go to pipeline
2. Click on Build Artifacts
3. Click on `cucumber` folder
4. Look for `authenticators_azure` folder - it is missing!

## Expected Results
`authenticators_azure` artifacts should be in the CI pipeline.

## Reproducible
   * [x] Always 
   * [ ] Sometimes
   * [ ] Non-Reproducible

eladkug:
### What does this PR do?
Add cucumber logs and reports
I verified run with and without fips to verify it works on both

Resolves #1916 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -171,6 +171,7 @@ pipeline {
                 env.SYSTEM_ASSIGNED_IDENTITY = sh(script: 'ci/authn-azure/get_system_assigned_identity.sh', returnStdout: true).trim()
 
                 sh('summon -f ci/authn-azure/secrets.yml ci/test cucumber_authenticators_azure')
+                stash name: 'testResultAzure', includes: "cucumber/*/*.*,container_logs/*/*,spec/reports/*.xml,spec/reports-audit/*.xml,cucumber/*/features/reports/**/*.xml"
               }
             }
           }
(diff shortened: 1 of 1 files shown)


abrahamko (approved):
I'm not a Jenkinsfile expert, but this LGTM.

Jenkinsfile
```diff
@@ -171,6 +171,7 @@ pipeline {
                 env.SYSTEM_ASSIGNED_IDENTITY = sh(script: 'ci/authn-azure/get_system_assigned_identity.sh', returnStdout: true).trim()
 
                 sh('summon -f ci/authn-azure/secrets.yml ci/test cucumber_authenticators_azure')
+                stash name: 'testResultAzure', includes: "cucumber/*/*.*,container_logs/*/*,spec/reports/*.xml,spec/reports-audit/*.xml,cucumber/*/features/reports/**/*.xml"
```
orenbm, line 174: thanks @eladkug ! do we need this also for authn-gcp tests?

eladkug: Nope. I verified we have gcp as well, and added it to our html reports too.
image

orenbm: ohhh do we need to have this explicitly for azure because it runs in an azure VM and not in the Jenkins node? this can explain how it vanished because we moved the azure tests to run inside the azure VM a while ago.

if this is correct, can you please add a comment saying that?
something like `We need to stash the test results so they are available on the Jenkins node`

Jenkinsfile
```diff
@@ -356,21 +357,23 @@ pipeline {
...
             publishHTML([reportDir: 'ee-test/cucumber', reportFiles: 'api/cucumber_results.html, 	authenticators_config/cucumber_results.html, \
-                                     authenticators_azure/cucumber_results.html, authenticators_ldap/cucumber_results.html, \
+                                     authenticators_ldap/cucumber_results.html, \
                                      authenticators_oidc/cucumber_results.html, authenticators_status/cucumber_results.html,\
                                      policy/cucumber_results.html , rotators/cucumber_results.html',\
                                      reportName: 'EE Integration reports', reportTitles: '', allowMissing: false, alwaysLinkToLastBuild: true, keepAll: true])
 
           }
       }
+      unstash 'testResultAzure'
```
hughsaunders, line 367: Probably safer to put this within a `dir` block to prevent overwriting existing files, see the unstash for `testReultEE` which is within the `ee-test` dir.

eladkug: `testReultEE`under dir because we wanted to separate the tests since those tests run twice.
so we have cucumber directory which exist to our daily runs , and when we run nightly as well.
Attached Nightly artifacts which contained both : 
image

hughsaunders (approved):

PR #1926, opened 2020-11-04
Update project to use tag-triggered builds

### LINKED ISSUE #1912: Conjur uses tag-triggered builds
Currently, Conjur uses bash logic to determine when to publish:
https://github.com/cyberark/conjur/blob/32b7bc58662d944f266421dfebeb4a302efd6259/push-image.sh#L33

We would like to update the project to use tag-triggered builds. That is, instead of using bash logic to determine if the HEAD of a branch matches a tag, we can add a Jenkins pipeline condition
```
when { tag "v*" }
```
that will determine when the publish step should be run.

AC:
- [ ] Conjur uses tag-triggers to determine whether to publish Docker images

izgeri:
### What does this PR do?
Instead of using logic to compare the the current commit to the version to
check if the current commit is a matching tag, this commit updates the logic
in the project Jenkinsfile to only run the image publish script when the
project is tagged with a tag matching the syntax `vA.B.C` where A/B/C are
numbers.

A note on testing locally:

It is notoriously hard to test Jenkins pipeline changes that mimic a tag / release locally. Here is what I have done to test this, as best as I am able:
- Test the REGEX using https://www.freeformatter.com/java-regex-tester.html to validate that the tag REGEX will be parsed properly by Jenkins. The full syntax of the `when` conditional is from https://www.jenkins.io/doc/book/pipeline/syntax/.
- Test the `push-image.sh` script by commenting out the `docker push` line, setting `TAG_NAME=v1.10.0` in my terminal, and running the `./build.sh --jenkins` followed by `./push-image.sh` to verify that the proper set of images would be pushed. The output was:
   ```
   Tagging and pushing registry.tld/conjur:1.10.0-3c377563...
   Tagging and pushing registry.tld/cyberark/conjur:1.10.0-3c377563...
   Tagging and pushing registry.tld/conjur:latest...
   Tagging and pushing registry.tld/cyberark/conjur:latest...
   Tagging and pushing registry.tld/conjur:v1.10-stable...
   Tagging and pushing registry.tld/cyberark/conjur:v1.10-stable...
   Tagging and pushing registry.tld/conjur:v1-stable...
   Tagging and pushing registry.tld/cyberark/conjur:v1-stable...
   Tagging and pushing cyberark/conjur:latest...
   Tagging and pushing cyberark/conjur:v1.10.0...
   Tagging and pushing cyberark/conjur:v1.10...
   Tagging and pushing cyberark/conjur:v1...
   ```

Resolves #1912 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -317,7 +317,12 @@ pipeline {
       }
     }
 
-    stage('Push Docker image') {
+    stage('Publish Docker image') {
+      when {
+        // Only run this stage when it's a tag build matching vA.B.C
+        tag pattern: "^v[0-9]+\\.[0-9]+\\.[0-9]+\$", comparator: "REGEXP"
+      }
+
       steps {
         sh './push-image.sh'
       }
--- a/push-image.sh
+++ b/push-image.sh
@@ -8,6 +8,13 @@
 # shellcheck disable=SC1091
 . version_utils.sh
 
+if [[ -z "${TAG_NAME:-}" ]]; then
+  echo "Please supply environment variable TAG_NAME."
+  echo "If you see this error in Jenkins it means the publish script was run"
+  echo "for a build that wasn't triggered by a tag - please check publish stage conditions."
+  exit 1
+fi
+
 TAG="${1:-$(version_tag)}"
 VERSION="$(< VERSION)"
 SOURCE_IMAGE="conjur:$TAG"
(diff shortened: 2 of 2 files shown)


micahlee:
@izgeri, did you also test the proposed DAP release tag format (`v1.10.0+dap.1`) to verify it is not pushed?

izgeri:
@micahlee yes, I tested it in the Java REGEX tool I linked above.

micahlee:
> @micahlee yes, I tested it in the Java REGEX tool I linked above.

Oh I see, so if it doesn't match the pattern, it doesn't push any images at all. Got it, thanks!

micahlee (approved):

micahlee (approved):

PR #1930, opened 2020-11-05
Run rotators tests

orenbm:
A previous commit accidentally replaced the rotators test step
so that it ran rspec tests instead. Thus, the rotators tests didn't
run at all.


eladkug (approved):
LGTM

abrahamko (approved):

Issue #1925, opened 2020-11-04
Fix cucumber reports on azure failures


eladkug:
If we have some failures in azure tests we cant see the cucumber reports.

DOD:

- [ ] Full report when all pass

- [ ] Full report when azure test failed

PR #1932, opened 2020-11-05
Fix cucumber reports on failures

eladkug:
### What does this PR do?
Fix cucumber reports in case of failure in azure

#1925 
Verify build with azure failure : https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--conjur/detail/fix-cucumber-reports-on-failures/43/pipeline/
Verify successful build: https://jenkins.conjur.net/blue/organizations/jenkins/cyberark--conjur/detail/fix-cucumber-reports-on-failures/44/pipeline
#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -162,18 +162,18 @@ pipeline {
         } // Standard agent tests
 
         stage('Azure Authenticator') {
+          agent { label 'azure-linux' }
+          environment {
+                   AZURE_AUTHN_INSTANCE_IP = sh(script: 'curl "http://checkip.amazonaws.com"', returnStdout: true).trim()
+                   SYSTEM_ASSIGNED_IDENTITY = sh(script: 'ci/authn-azure/get_system_assigned_identity.sh', returnStdout: true).trim()
+          }
           steps {
-            script {
-              node('azure-linux') {
-                // get `ci/authn-azure/get_system_assigned_identity.sh` from scm
-                checkout scm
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -162,18 +162,18 @@ pipeline {
         } // Standard agent tests
 
         stage('Azure Authenticator') {
+          agent { label 'azure-linux' }
```
orenbm, line 165: is this the same as `node('azure-linux')`? will this stage run in the azure node?

Jenkinsfile
```diff
@@ -162,18 +162,18 @@ pipeline {
...
+          environment {
+                   AZURE_AUTHN_INSTANCE_IP = sh(script: 'curl "http://checkip.amazonaws.com"', returnStdout: true).trim()
+                   SYSTEM_ASSIGNED_IDENTITY = sh(script: 'ci/authn-azure/get_system_assigned_identity.sh', returnStdout: true).trim()
+          }
           steps {
-            script {
-              node('azure-linux') {
-                // get `ci/authn-azure/get_system_assigned_identity.sh` from scm
-                checkout scm
```
orenbm, line 169: how come we don't need this?

Jenkinsfile
eladkug: work the same as agent (i verified all tests run)
We cant work with node since it doesnt support post command in case of failure

Jenkinsfile
eladkug: checkout scm is not needed when using agent

orenbm (approved):

PR #1934, opened 2020-11-09
K8s resource restrictions refactor

### LINKED ISSUE #1929: Refactoring authenticators code - Refactor K8s ValidateResourceRestrictions
Refactor K8s `ValidateResourceRestrictions` component according to LDD design: [refactor resource restriction document](https://github.com/cyberark/conjur/blob/master/design/authenticators/resource-restrictions/resource-restrictions.md).
This is the third step of the refactoring.
First step (GCP): #1815
Second step (Azure): #1928

**DoD**
- [ ] K8s automated tests are finish successfully
- [ ] We have a better code

abrahamko:
### What does this PR do?
Implementation of the design written in [refactor resource restriction document](https://github.com/cyberark/conjur/blob/master/design/authenticators/resource-restrictions/resource-restrictions.md).
This is the third step, replacing K8s authenticator implementation with the new one.

Resolves #1929

#### Change log

#### Test coverage

#### Documentation


app/domain/authentication/authn_k8s/authentication_request.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    # This class validates the k8s authentication request against the resource
+    # restrictions configured in the policy for the requesting host.
+    # It uses the 'validate_k8s_resource' command-class to validate a k8s
+    # resource against the k8s API.
+    # This command-class receives the resource's type and name as input.
+    class AuthenticationRequest
+      def initialize(namespace:, validate_k8s_resource:)
+        @namespace = namespace
+        @validate_k8s_resource = validate_k8s_resource
```
orenbm, line 12: `validate_k8s_resource_class`?

app/domain/authentication/authn_k8s/authentication_request.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+
+      def valid?(restriction_name, restriction_value)
```
orenbm, line 15: the name `valid?` is a bit weird for me. I would expect the method `valid?` in the `AuthenticationRequest` class to tell me if the authentication request is valid but this is not what is happening here. This method says if the restriction is valid in the context of the authentication request. So i think that
`def valid_restriction?(restriction_name, restriction_value)` is better.
you can also go with
`def valid_restriction?(name, value)`

app/domain/authentication/authn_k8s/authentication_request.rb
```diff
@@ -0,0 +1,32 @@
...
+        case restriction_name
+        when Restrictions::NAMESPACE
+          if restriction_value != @namespace
+            raise Errors::Authentication::AuthnK8s::NamespaceMismatch @namespace, restriction_value
+          end
+        when Restrictions::AUTHENTICATION_CONTAINER_NAME
```
orenbm, line 21: why is this here? don't you control the call to this method? can't we not call it with the container name annotation?

Also - this annotation is not really a restriction iirc. It just tells conjur to which container it should inject the cert. 
am i right?

app/domain/authentication/authn_k8s/authentication_request.rb
```diff
@@ -0,0 +1,32 @@
...
+          # Container name validation is done in ValidatePodRequest
+        else
+          @validate_k8s_resource.call(resource_type: restriction_name, resource_name: restriction_value)
+        end
+
+        # Validation is done internally, so it is always valid if reached this point.
```
orenbm, line 27: +1 for the comment :)

app/domain/authentication/authn_k8s/consts.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    module Restrictions
+
+      NAMESPACE = "namespace"
+      SERVICE_ACCOUNT = "service-account"
+      POD = "pod"
+      DEPLOYMENT = "deployment"
+      STATEFUL_SET = "stateful-set"
+      DEPLOYMENT_CONFIG = "deployment-config"
+      AUTHENTICATION_CONTAINER_NAME = "authentication-container-name"
```
orenbm, line 16: this annotation is not really a restriction iirc. It just tells conjur to which container it should inject the cert. 

To verify this - what if the host has all the annotations correct, but has this value set to the application container in the pod. i _think_ that the authentication will pass (but of course the cert will be injected to the wrong container).

am i right?

app/domain/authentication/authn_k8s/extract_container_name.rb
```diff
@@ -12,7 +12,6 @@ module AuthnK8s
       inputs: %i(service_id host_annotations)
     ) do
 
-      AUTHENTICATION_CONTAINER_NAME_ANNOTATION = "authentication-container-name"
       DEFAULT_AUTHENTICATION_CONTAINER_NAME = "authenticator"
```
orenbm, line 15: should this go in `consts.rb`?

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
```diff
@@ -0,0 +1,66 @@
...
+      dependencies: {
+        extract_resource_restrictions: Authentication::ResourceRestrictions::ExtractResourceRestrictions.new,
+        resource_restrictions_class:   Authentication::ResourceRestrictions::ResourceRestrictions,
+        logger:                        Rails.logger
+      },
+      inputs:   %i(authenticator_name service_id role_name account)
+    ) do
+
+      def call
+        extract_resource_restrictions_from_annotations_or_host_id
```
orenbm, line 16: i would call this `extract_resource_restrictions`

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
```diff
@@ -0,0 +1,66 @@
...
+      end
+
+      private
+
+      def extract_resource_restrictions_from_annotations_or_host_id
```
orenbm, line 21: please add a comment that we first check if there are restrictions in annotations and only then in the id. If there are restrictions in both places we take only the annotations

app/domain/authentication/authn_k8s/consts.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Authentication
+  module AuthnK8s
+
+    K8S_AUTHENTICATOR_NAME = 'authn-k8s'
```
orenbm, line 6: this can be `AUTHENTICATOR_NAME` because its full name is `Authentication:: AuthnK8s::AUTHENTICATOR_NAME`.

but it's ok as it is if you prefer that.

app/domain/authentication/authn_k8s/validate_k8s_resource.rb
```diff
@@ -0,0 +1,64 @@
...
+require 'command_class'
+
+module Authentication
+  module AuthnK8s
+
+    ValidateK8sResource ||= CommandClass.new(
+      dependencies: {
+        k8s_resolver_class: K8sResolver,
+        k8s_object_lookup:  nil,
+        pod:                nil,
```
orenbm, line 11: what's this? if the value is nil then you can remove the dependency altogether.

orenbm: also - how come you don't need the `k8s_object_lookup` class?

app/domain/authentication/authn_k8s/validate_pod_request.rb
```diff
@@ -13,16 +13,16 @@ module AuthnK8s
         validate_webservice_is_whitelisted:  ::Authentication::Security::ValidateWebserviceIsWhitelisted.new,
         validate_role_can_access_webservice: ::Authentication::Security::ValidateRoleCanAccessWebservice.new,
         enabled_authenticators:              Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV),
-        validate_resource_restrictions:      ValidateResourceRestrictions.new,
-        extract_container_name: ExtractContainerName.new,
-        logger: Rails.logger
+        validate_resource_restrictions:      Authentication::ResourceRestrictions::ValidateResourceRestrictions.new(
+          extract_resource_restrictions:      ExtractK8sResourceRestrictions.new
+        ),
```
orenbm, line 18: was this a thing in other authenticators? a dependency that couples another one?

app/domain/authentication/authn_k8s/authentication_request.rb
abrahamko: 1. this is not a class, this is an instance of the command class.
2. we don't add `_class` suffix to command classes.

app/domain/authentication/authn_k8s/validate_pod_request.rb
```diff
@@ -63,20 +63,32 @@ def validate_pod_exists
...
+          role_name: k8s_host.k8s_host_name,
+          constraints: Restrictions::CONSTRAINTS,
+          authentication_request: authentication_request
+        )
+      end
+
+      def authentication_request
+        @authentication_request_class.new(
+          namespace: pod_namespace,
+          validate_k8s_resource: ValidateK8sResource.new(
```
orenbm, line 79: `ValidateK8sResource` should be a dependency

app/domain/authentication/authn_k8s/authentication_request.rb
abrahamko: I like it!

app/domain/authentication/authn_k8s/authentication_request.rb
orenbm: ok, i thought it was a class

app/domain/authentication/authn_k8s/authentication_request.rb
abrahamko: We don't control it, no.
Since this is part of the general ValidateResourceRestriction flow, and the `authentication-container-name` restriction is a valid restriction, it needs to be handled.
Basically, we need to handle any valid restriction that starts with `authn-k8s/`.
I thought of moving the container name validation here, but it is quite complex and not necessary.

app/domain/authentication/authn_k8s/consts.rb
abrahamko: Partially.
The authentication will pass, because the restriction stands.
We have a dedicated step for validating the container name: https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/validate_pod_request.rb#L34
This step looks for a container with the required name in the pod, and throws a `ContainerNotFound` if it doesn't find any.
This is a restriction validation, just like any other.

app/domain/authentication/authn_k8s/extract_container_name.rb
abrahamko: I don't think so, it isn't used anywhere outside this class.

app/domain/logs.rb
```diff
@@ -33,11 +33,6 @@ module Authentication
       code: "CONJ00003D"
     )
 
-    ValidatingAnnotationsWithPrefix = ::Util::TrackableLogMessageClass.new(
```
eladkug, line 36: Why you removed this log ? Do we need to update it in some place ?

app/domain/logs.rb
```diff
@@ -203,11 +203,6 @@ module AuthnK8s
         code: "CONJ00037D"
       )
 
-      ValidatingHostId = ::Util::TrackableLogMessageClass.new(
```
eladkug, line 206: same

app/domain/logs.rb
abrahamko: It is not used.
I didn't find any reference to this that needs update.

app/domain/logs.rb
abrahamko: It is not used.
I didn't find any reference to this that needs update.

app/domain/authentication/authn_k8s/validate_pod_request.rb
abrahamko: The command class dependencies meant to be able for replacement.
This is how you override its internal behavior, by replacing its dependency with another one that uses the same inputs.
This is how you implement hierarchy in command-class.

app/domain/authentication/authn_k8s/consts.rb
abrahamko: The module is not always there...

app/domain/logs.rb
eladkug: o.k. looks like a lot of those logs not in use, just need to make sure its not written in some document that we need to update as well

app/domain/authentication/authn_k8s/validate_k8s_resource.rb
abrahamko: Cannot be removed.
The dependencies just don't have any default value because they need to be supplied from the outside.

app/domain/authentication/authn_k8s/validate_pod_request.rb
abrahamko: How would you replace a dependency in a command class otherwise?

eladkug (commented):
Do you know why the total-coverage reduced?

app/domain/authentication/authn_k8s/authentication_request.rb
abrahamko: done

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
abrahamko: done

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
abrahamko: done

app/domain/authentication/authn_k8s/validate_k8s_resource.rb
abrahamko: I can replace it with regular class (as I can for any other command-class).
would that make more sense?
I thought it is appropriate to use command-class here, because the flow is similar to command class.

app/domain/authentication/authn_k8s/validate_pod_request.rb
abrahamko: done

app/domain/authentication/authn_k8s/validate_pod_request.rb
nessiLahav: @abrahamko Why `ValidateResourceRestrictions` command class contains 1 dependency instead of 3?
You dont think in the future we will need this capability of replacing the logic of `validate_resource_restrictions_configuration , validate_request_matches_resource_restrictions` like you did here for the `extract_resource_restrictions` logic ?

abrahamko: The other 2 already differ for each authenticator.
1. Validate restrictions is done using the `@constraints` input
2. Validate request is done using `AuthenticationRequest` by calling `valid_restriction?` on each restriction.
The code around them is generic, so I don't expect it to change.

abrahamko: I feel comfortable the way that is, as this is the way to replace the default dependency.
I think this is a correct way to do it.

app/domain/authentication/authn_k8s/consts.rb
orenbm: i won't push any further than this but it is not really `a restriction validation, just like any other`. We do not enforce the request to come from the container mentioned in the annotation, as we do with the rest. We will validate that the container exists in the pod but it can be any other container. 

Having said that, if you think it makes the code more readable to treat this annotation as a resource restriction i am ok with that. it may be worth mentioning in a comment that it doesn't act exactly as a resource restriction.

orenbm: or maybe just move it outside of restrictions in this class to identify that.

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
```diff
@@ -0,0 +1,70 @@
+require 'command_class'
+
+module Authentication
+  module AuthnK8s
+
+    ExtractK8sResourceRestrictions = CommandClass.new(
```
orenbm, line 6: does this need to be `ExtractK8sResourceRestrictions`? can we name it `ExtractResourceRestrictions` so we have an alignment?

then we have the class `Authentication:: AuthnK8s::ExtractResourceRestrictions` that calls the general class `Authentication::ResourceRestrictions::ResourceRestrictions`.

I would maybe even consider adding a sub-module `ResourceRestrictions ` so we have:
```
Authentication::ResourceRestrictions::ResourceRestrictions
Authentication:: AuthnK8s::ResourceRestrictions::ExtractResourceRestrictions
```

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+
+      private
+
+      # Extracts resource restrictions for k8s hosts. It can be defined in the
+      # annotations or as part of the host ID itself.
+      # If the restrictions are in the annotations, then they will not be
+      # extracted from the host ID.
+      def extract_resource_restrictions
+        resource_restrictions = @extract_resource_restrictions.call(
```
orenbm, line 26: i would add a comment saying that this extracts the resource restrictions from the annotations

app/domain/authentication/authn_k8s/validate_pod_request.rb
orenbm: @jonahx can you take a look on this? it seems odd to me that we have `app` code using dependencies. I understand the need here but want to have another look on it.

app/domain/authentication/authn_k8s/k8s_resource_validator.rb
```diff
@@ -0,0 +1,50 @@
+module Authentication
+  module AuthnK8s
+
+    class K8sResourceValidator
```
orenbm, line 4: @abrahamko can you please write a comment here saying why you chose to use a regular class `K8sResourceValidator` that has a `validate` method instead of using a command class `ValidateK8sResource`?
I'm not saying it is wrong but i would like @jonahx 's opinion on this so your line of thought will help give him context.

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
abrahamko: I don't like relying on module for class context.
The module is mostly ignored, and it is hard to know which of the classes is the referenced one.
I find it very confusing.
I prefer to distinguish the class by its name, and to treat its module as "family name" - which doesn't say what it does, just where it belong.

app/domain/authentication/authn_k8s/k8s_resource_validator.rb
abrahamko: done

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
abrahamko: done

abrahamko:
@eladkug the code coverage got down because codeclimate does not take the integration tests into consideration.
I added important integration tests so I don't think this reduction is meaningful.

app/domain/authentication/authn_k8s/consts.rb
abrahamko: done

app/domain/authentication/authn_k8s/validate_pod_request.rb
jonahx: > it seems odd to me that we have app code using dependencies. 

@orenbm Can you clarify your concern here?  

To me it looks like a nice example of re-use by composition, but I'm curious what in particular you're worried about.

Admittedly the total _number_ of dependencies to this class is a little concerning, but it seems to be the "top level" validation class, just delegating out to all the individual validations, and since we have a lot of validations to do (just because of our business logic), I think it's ok.  The CommandClass is doing what it's supposed to do by making all of our dependencies stare us in the face.

orenbm: no real concern. I just remember you saying that inputs will be given at run time in the app code and dependencies will be given only in UTs. if this is not the case then i'm good.

Just to verify we're looking at the same thing - i am talking about the fact that `validate_resource_restrictions` is initialized as a dependency and replaces a dependency (`extract_resource_restrictions`) of `Authentication::ResourceRestrictions::ValidateResourceRestrictions` in its initialization.

I'm not against it but it is the first time that this happens in our code (as far as i know)

app/domain/authentication/authn_k8s/consts.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    module Restrictions
+
+      NAMESPACE = "namespace"
+      SERVICE_ACCOUNT = "service-account"
+      POD = "pod"
+      DEPLOYMENT = "deployment"
+      STATEFUL_SET = "stateful-set"
+      DEPLOYMENT_CONFIG = "deployment-config"
```
jonahx, line 15: A subtle point I noticed when looking at the azure `consts.rb` file too.  What I'm about to say probably falls under "existing code" and is outside the scope of this PR, but is worth mentioning.

This set of variables from line 10-15 do not actually represent restrictions.  As I understand it, these are the string values that will be used as the annotation keys stored in the DB for each of these concepts.  Interestingly, I was not able to find any code that _stored_ annotations in the db referring to them, which means that those references are probably hard-coded somewhere, minimizing the value of this abstraction (keeping the "annotation key strings" in sync between the code that writes the annotations and the code that reads them would be one of the primary benefits of this abstraction).

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
```diff
@@ -66,10 +66,11 @@ def validate_resource_restrictions_configuration
       def validate_request_matches_resource_restrictions
         @logger.debug(LogMessages::Authentication::ResourceRestrictions::ValidatingResourceRestrictionsValues.new)
 
-        resource_restrictions.each do |restriction_name, restriction_value|
-          if @authentication_request.retrieve_attribute(restriction_name) != restriction_value
-            raise Errors::Authentication::ResourceRestrictions::InvalidResourceRestrictions, restriction_name
-          end
+        resource_restrictions.each do |name, value|
```
jonahx, line 69: The `name, value` here looks like a [data clump](https://refactoring.guru/smells/data-clumps) to me, representing a `Restriction` object.  Like one of my other comments, fixing this _fully_ is an existing code problem and probably out of scope, since it would also affect the code that saves annotations to the db (I didn't investigate that, but it may be a rabbit hole). 

I leave it to you to decide if you want to introduce it now for this code.

app/domain/authentication/authn_k8s/consts.rb
jonahx: > The module is not always there...

In these cases it should only work if you are already inside the module, as you are here.  So either context or a fully qualified path should always make the "K8s" part clear, and we should never have a case where we require "name stutter".  This is true in general too.

jonahx: Indeed, just noticing now that the existing convention has stutter too: `Authentication::AuthnK8s` should probably be `Authentication::K8s`, and arguably my original choice of `Authentication` should have been `Authn` so that we'd have `Authn::K8s` as the prefix, which is, in addition to being shorter, much more readable imo.

That parts out of scope though :)

app/domain/authentication/authn_k8s/authentication_request.rb
jonahx: My comment is (sort of) related to [this comment](https://github.com/cyberark/conjur/pull/1934/files#r522421529):

It seems that implicit concept here is that of different _types_ of restrictions, at the very last you have "container name validation" and "everything else" -- the "everything else" being the ones we validate above.  I'm not commenting on this distinction itself... merely pointing out that the current code is, _de facto_, making that distinction.

So I think you have two choices:

1. Decide that this concept of different types of restrictions is, in fact, needed, and make it explicit.
2. Decide it's not needed and centralize the restriction logic.

Based only your comment:

> I thought of moving the container name validation here, but it is quite complex and not necessary.

I am guessing that 1. is the better road.  Assuming that is the case, the simple thing to do to avoid the empty case condition would be to first _filter_ your restrictions so that they don't contain the CONTAINER_NAME restriction, and only then call this method.

The method itself, no longer being fully general, would change its name to:

`valid_XXX_restriction?`

where `XXX` describes the valid type of restriction, ie, whatever you want to call the "everything except a container name description".

app/domain/authentication/authn_k8s/extract_container_name.rb
```diff
@@ -22,9 +21,9 @@ def call
       private
 
       def container_name
-        annotation_value("authn-k8s/#{@service_id}/#{AUTHENTICATION_CONTAINER_NAME_ANNOTATION}") ||
-          annotation_value("authn-k8s/#{AUTHENTICATION_CONTAINER_NAME_ANNOTATION}") ||
-          annotation_value("kubernetes/#{AUTHENTICATION_CONTAINER_NAME_ANNOTATION}") ||
+        annotation_value("authn-k8s/#{@service_id}/#{Restrictions::AUTHENTICATION_CONTAINER_NAME}") ||
+          annotation_value("authn-k8s/#{Restrictions::AUTHENTICATION_CONTAINER_NAME}") ||
+          annotation_value("kubernetes/#{Restrictions::AUTHENTICATION_CONTAINER_NAME}") ||
```
jonahx, line 26: Formatting suggestion for controlling line length:

```ruby
      def container_name
        name = Restrictions::AUTHENTICATION_CONTAINER_NAME
        
        annotation_value("authn-k8s/#{@service_id}/#{name}") ||
          annotation_value("authn-k8s/#{name}") ||
          annotation_value("kubernetes/#{name}") ||
          default_authentication_container_name
      end
```

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
```diff
@@ -0,0 +1,71 @@
+require 'command_class'
+
+module Authentication
+  module AuthnK8s
+
+    ExtractK8sResourceRestrictions = CommandClass.new(
+      dependencies: {
+        extract_resource_restrictions: Authentication::ResourceRestrictions::ExtractResourceRestrictions.new,
+        resource_restrictions_class:   Authentication::ResourceRestrictions::ResourceRestrictions,
+        logger:                        Rails.logger
```
jonahx, line 10: Since you are inside the `Authentication` module, you should be able to drop the `Authentication::` prefixes here.

POC:

```ruby
module Authentication
  Person = Struct.new(:name)
end

module Authentication
  module ResourceRestrictions
    def self.test
      ak = Person.new('Abraham')
      puts ak
    end
  end
end

Authentication::ResourceRestrictions.test
# => #<struct Authentication::Person name="Abraham">
```

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
jonahx: I propose we call this a "Pinocchio" comment.

Just like Pinocchio was a puppet who wanted to be real boy, sometimes comments want to be a real variable name.

Perhaps `get_restrictions_from_annotations`?

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
```diff
@@ -0,0 +1,71 @@
...
+      def extract_resource_restrictions
+        # Extract resource restrictions from host annotations.
+        resource_restrictions = @extract_resource_restrictions.call(
+          authenticator_name: @authenticator_name,
+          service_id:         @service_id,
+          role_name:          @role_name,
+          account:            @account
+        )
+
+        has_restrictions_besides_container_name = !resource_restrictions.except(Restrictions::AUTHENTICATION_CONTAINER_NAME).empty?
```
jonahx, line 34: Ah!  Here's that concept of "restriction type" (from one of my other comments) popping up again!  I often find that when I want to name things with long phrases like this, there is a concept I haven't found yet.  I'm not sure what it is in the case yet.

app/domain/authentication/authn_k8s/k8s_resource_validator.rb
```diff
@@ -0,0 +1,57 @@
...
+        @logger = logger
+      end
+
+      def valid_resource?(type: , name:)
+        @logger.debug(LogMessages::Authentication::AuthnK8s::ValidatingK8sResource.new(type, name))
+
+        k8s_resource = retrieve_k8s_resource(type, name)
+        validate(k8s_resource, type, name)
+
+        @logger.debug(LogMessages::Authentication::AuthnK8s::ValidatedK8sResource.new(type, name))
```
jonahx, line 24: If you like, a trick you can use to make these errors and log messages more readable:

```ruby
    class K8sResourceValidator
      LOG_MESSAGE = LogMessages::Authentication::AuthnK8s
      ERRORS = Errors::Authentication::AuthnK8s

      def initialize(k8s_object_lookup:, pod:, logger: Rails.logger)
        @k8s_object_lookup = k8s_object_lookup
        @pod = pod
        @logger = logger
      end

      def valid_resource?(type: , name:)
        @logger.debug(LOG_MESSAGE::ValidatingK8sResource.new(type, name))

        k8s_resource = retrieve_k8s_resource(type, name)
        validate(k8s_resource, type, name)

        @logger.debug(LOG_MESSAGE::ValidatedK8sResource.new(type, name))
      end

      private

      def retrieve_k8s_resource(type, name)
        @k8s_object_lookup.find_object_by_name(
          type,
          name,
          namespace
        )
      end

      def validate(k8s_resource, type, name)
        unless k8s_resource
          raise ERRORS::K8sResourceNotFound.new(type, name, namespace)
        end

        K8sResolver
          .for_resource(type)
          .new(
            k8s_resource,
            @pod,
            @k8s_object_lookup
          )
          .validate_pod
      end

      def namespace
        @namespace ||= @pod.metadata.namespace
      end
    end
  end
```

app/domain/authentication/authn_k8s/validate_pod_request.rb
jonahx: >  I just remember you saying that inputs will be given at run time in the app code and dependencies will be given only in UTs. if this is not the case then i'm good.

Ah, so just a slight miscommunication there. Yes: inputs are only given at runtime.  Yes: dependencies are given at "compile time" -- I think of it like that, even though ruby has no compilation.  It just means: dependencies are known before any code runs.  

> I'm not against it but it is the first time that this happens in our code (as far as i know)

Often, the default dependency is the only one we need, other than in our UTs.  So UTs are a common place to see the default dep changed out.  However, they need not be the only place.  And in fact, the ability to swap them out is one of the intended benefits of this style.

As I understand the code, `validate_resource_restrictions` defines some high-level steps:

```
        extract_resource_restrictions
        validate_resource_restrictions_configuration
        validate_request_matches_resource_restrictions
```

What each _specific_ step means might vary from one authenticator to another, and because the steps are just composed of other objects, we can change them out to suit individual authenticators, and keep the same high-level logic, without being forced to repeat it.  This is textbook "composition over inheritance" and I give it a 👍.

orenbm: great, thanks for clarifying!

app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb
abrahamko: Thanks for this proposal, @jonahx.
I think combining the `name` and `value` into one object can solve this data clump you mentioned correctly.
Regarding your second proposal, as much as it sounds promising, it requires a complex code to decide which restriction to create.
Since the same code is called for all authenticators (k8s, azure, gcp) this requires a new class as input, to decide which restriction to create.
I see where this is going, and it looks to me less concise and clear.

I did the first part, though!

app/domain/authentication/authn_k8s/extract_container_name.rb
abrahamko: done

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
abrahamko: done

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
abrahamko: > I'm not sure what it is in the case yet.

Me neither... 🤔

app/domain/authentication/authn_k8s/k8s_resource_validator.rb
abrahamko: As far as I remember, we decided to not do these shortcuts anymore.
Am I right, @orenbm ?

orenbm: correct. these consts made some unexpected behaviour because they were initialized in different modules so we decided to go full name.

app/domain/authentication/authn_k8s/consts.rb
abrahamko: done.

app/domain/authentication/authn_k8s/authentication_request.rb
abrahamko: I followed 1st suggestion.
Regarding `Valid_XXX_restriction`, I haven't done it for two reasons:
1. The container name is not a regular restriction, as @orenbm said.
2. The signature is common across all authenticators, and it seems not right to me to change it for all just to fix this single one.

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
abrahamko: done

app/domain/authentication/authn_k8s/consts.rb
abrahamko: They are hard-coded in the policy of the integration tests.
I agree this is less readable, but this is still an improvement to the existing code.
I'm not going to go any further than this.

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
```diff
@@ -0,0 +1,71 @@
...
+      def extract_resource_restrictions
+        get_restrictions_from_annotations = @extract_resource_restrictions.call(
+          authenticator_name: @authenticator_name,
+          service_id:         @service_id,
+          role_name:          @role_name,
+          account:            @account
+        )
+
+        # Container name annotation is not a restriction, and needs to be excluded.
+        resource_restrictions = get_restrictions_from_annotations.except(Restrictions::AUTHENTICATION_CONTAINER_NAME)
```
orenbm, line 34: do you think it will make more sense to move this logic into `ExtractResourceRestrictions`? if the container name is not really a restriction then the `ExtractResourceRestrictions` class can treat it that way without the need to exclude it in the calling class(es).

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+
+      private
+
+      # Extracts resource restrictions for k8s hosts. It can be defined in the
+      # annotations or as part of the host ID itself.
+      # If the restrictions are in the annotations, then they will not be
+      # extracted from the host ID.
+      def extract_resource_restrictions
+        get_restrictions_from_annotations = @extract_resource_restrictions.call(
```
orenbm, line 26: this should be `restrictions_from_annotations`, no? When i see an object named `get_restrictions_from_annotations` i think it is an instance of a function.

app/domain/authentication/authn_k8s/validate_pod_request.rb
```diff
@@ -13,16 +13,17 @@ module AuthnK8s
         validate_webservice_is_whitelisted:  ::Authentication::Security::ValidateWebserviceIsWhitelisted.new,
         validate_role_can_access_webservice: ::Authentication::Security::ValidateRoleCanAccessWebservice.new,
         enabled_authenticators:              Authentication::InstalledAuthenticators.enabled_authenticators_str(ENV),
-        validate_resource_restrictions:      ValidateResourceRestrictions.new,
-        extract_container_name: ExtractContainerName.new,
-        logger: Rails.logger
+        validate_resource_restrictions:      Authentication::ResourceRestrictions::ValidateResourceRestrictions.new(
```
orenbm, line 16: you can remove the `Authentication` module from these deps as well.

app/domain/authentication/resource_restrictions/resource_restriction.rb
```diff
@@ -0,0 +1,15 @@
...
+    class ResourceRestriction
+      attr_reader :name, :value
+
+      def initialize(name:, value:)
+        @name = name
+        @value = value
+      end
+    end
+  end
+end
```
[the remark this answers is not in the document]
orenbm: missing newline

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
abrahamko: Correct, my bad... I copied Jonah's suggestion without noticing this.
Fixed

app/domain/authentication/authn_k8s/extract_k8s_resource_restrictions.rb
abrahamko: This is not a regular calling class, this is an expansion of `ExtractResourceRestrictions`.
As I see it, if I had inheritance relation between these classes I would put this code in the overriding function.
That's why it's here.

app/domain/authentication/authn_k8s/validate_pod_request.rb
abrahamko: done

orenbm (approved):

app/domain/authentication/authn_k8s/k8s_resource_validator.rb
jonahx: @orenbm FYI: should be safe to do them at the class level.  You're right it could be confusing at the module level.

PR #1977, opened 2020-12-16
Use debify 2

tzheleznyak:
Uses new version of debify.

**impact on users**
None. It is for dockerless project which uses possum rpms.

### **What does this PR DO?**

- Adding -rpm-rpmbuild-define flag to rpm creation
- Using new image of debify.

#### Change log
-  [x] This PR does not include user-facing changes and doesn't require a CHANGELOG update

#### Test coverage

#### Documentation


CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   are included in their Conjur image.
   [cyberark/conjur#1974](https://github.com/cyberark/conjur/issues/1974)
 
+## [1.11.2] - 2020-12-17
+### Added
+- Usage of debify 2.
```
izgeri, line 18: This doesn't speak at all to the user impact of this change - if I am a user, why do I care?

I would guess in this case there is no OSS user impact? Please see our full [changelog guide](https://github.com/cyberark/community/blob/master/Conjur/CONTRIBUTING.md#changelog-guidelines) for more info on what goes into a good changelog message

VERSION
```diff
@@ -1 +1 @@
-1.11.1
+1.11.2
```
izgeri, line 1: Why are we incrementing the Conjur version in this PR?

izgeri (changes requested):
Can you please use the pull request template when submitting a PR to this project? You've skipped adding several important details.

This PR should be linked to an issue that's more clear about the change being made and the motivation, which is to use a newer version of debify to package the Conjur deb and to add an RPM to what debify packages (right?).

Is there any benefit to the OSS users of this change? I know we're consuming these elsewhere, so it's quite likely the answer is no (which is fine), but if there's any benefit it'd be good to highlight that in the changelog.

package.sh
```diff
@@ -1,11 +1,23 @@
 #!/bin/bash -ex
-export DEBIFY_IMAGE='registry.tld/conjurinc/debify:1.11.5'
+export DEBIFY_IMAGE='registry.tld/conjurinc/debify:2.0.0'
 
 docker run --rm $DEBIFY_IMAGE config script > docker-debify
 chmod +x docker-debify
 
+# Creates possum deb
```
orenbm, line 7: change `Creates` to `Create`

package.sh
```diff
@@ -1,11 +1,23 @@
...
 ./docker-debify package \
   --dockerfile=Dockerfile.fpm \
+  --output=deb \
   possum \
   -- \
   --depends tzdata
+  --force
```
orenbm, line 14: why did we add the `--force` flag?

CHANGELOG.md
tzheleznyak: So in this case i don't need to change version at all and write anything to changelog?

package.sh
tzheleznyak: Changed

VERSION
tzheleznyak: Returned it back to 1.11.1

package.sh
tzheleznyak: Misunderstood @rafis3 comment in slack about it. I will put -rpm-rpmbuild-define “_build_id_links none” instead.

tzheleznyak: Changed it

CHANGELOG.md
tzheleznyak: Currently Decreased version and removed info from changelog because the user expiriance with Conjur is not being changed. It is for deploy on RHEL porpuses.

tzheleznyak:
> Can you please use the pull request template when submitting a PR to this project? You've skipped adding several important details.
> 
> This PR should be linked to an issue that's more clear about the change being made and the motivation, which is to use a newer version of debify to package the Conjur deb and to add an RPM to what debify packages (right?).
> 
> Is there any benefit to the OSS users of this change? I know we're consuming these elsewhere, so it's quite likely the answer is no (which is fine), but if there's any benefit it'd be good to highlight that in the changelog.

package.sh
```diff
@@ -1,11 +1,23 @@
...
 chmod +x docker-debify
 
+# Create possum deb
 ./docker-debify package \
   --dockerfile=Dockerfile.fpm \
+  --output=deb \
+  possum \
+  -- \
+  --depends tzdata \
+  --rpm-rpmbuild-define '_build_id_links none'
```
orenbm, line 14: what's this flag? it is also failing the build

izgeri (approved):

izgeri:
Thank you for addressing my feedback @tzheleznyak! I'll leave the technical review to @orenbm now :)

package.sh
tzheleznyak: Forgot in rpm linesep it failed the build

izgeri:
Just wanted to note that you'll probably want to squash these commits into a single commit (given the scope of the aggregated change) - we do provide some guidelines on how to do that [here](https://github.com/cyberark/community/blob/master/Conjur/conventions/git-tips-and-tricks.md#Cleaning-Up-Your-Commit-History)

tzheleznyak:
> Just wanted to note that you'll probably want to squash these commits into a single commit (given the scope of the aggregated change) - we do provide some guidelines on how to do that [here](https://github.com/cyberark/community/blob/master/Conjur/conventions/git-tips-and-tricks.md#Cleaning-Up-Your-Commit-History)

I ussualy do the squash at the merge moment it self from the pr

package.sh
```diff
@@ -1,11 +1,22 @@
...
+  --depends tzdata \
+
+# Create possum rpm
+./docker-debify package \
+  --dockerfile=Dockerfile.fpm \
+  --output=rpm \
+  possum \
+  -- \
+  --depends tzdata \
+  --rpm-rpmbuild-define '_build_id_links none'
```
orenbm, line 22: can you add a comment on what is `--rpm-rpmbuild-define` and why we need it?

package.sh
```diff
@@ -13,10 +13,11 @@ chmod +x docker-debify
   --depends tzdata \
 
 # Create possum rpm
+# The flag to avoid packaging build files that are not needed and cause conflict with conjur-ui on install
```
orenbm, line 16:
```
The '--rpm-rpmbuild-define' flag is added to avoid packaging build 
files that are not needed and cause conflict with conjur-possum on install
```

tzheleznyak: Changed

orenbm (approved):

PR #1979, opened 2020-12-16
Publish an edge-tagged image on master builds

### LINKED ISSUE #1617: Push edge image to Dockerhub on every successful master build
Currently, only release version are push to `Dockerhub`. 
When users/components pulling the `latest` tag, they pull the latest release, not the head of master
We can add to our pipeline step to create `edge` tag and push image on every build.
The benefit to used this tag:
we had a lot of components consuming Conjur old release or latest version from Dockerhub and we don't want to wait until release time to determine changes impacts.

izgeri:
### What does this PR do?
At current, we only publish images to DockerHub on releases. With this commit,
we will begin publishing `edge`-tagged images with every successful master
branch build.

This commit includes other code cleanups, too. In particular:
- The build.sh script no longer publishes images to the internal registry.
  Instead, the publish script is flexible enough to be used in the Jenkins
  pipeline just after the images are built. Importantly, the needed images
  are still published - the publish step just isn't hidden in the "build" script
  anymore.
- The version_utils script now also includes the `tag_and_push` function, so
  it has been renamed "build_utils"
- This project no longer publishes images to a special registry.tld/test prefix,
  but instead just publishes the commit hash-tagged images to registry.tld.
  The image tags include the version, so we should be able to filter to remove
  old images if the registry is unduly overloaded with these images.
- This project no longer publishes "{version}-stable" images to the internal
  registry, since upon review these are not consumed anywhere and the public
  DockerHub images are used instead.

Since the `push-image.sh` script has undergone some major updates in this PR, here is a sample of how it will work:
```
conjur$ ./push-image.sh 
This script is not running in a tag-triggered build.
Please supply either:
 - The registry prefix with the --registry-prefix flag, or
 - The --edge flag to publish an edge release to DockerHub.

conjur$ ./push-image.sh -f
Unknown option: -f
Usage: ./push-image.sh [OPTION...]
 --registry-prefix=STRING; optional registry prefix to prepend to '/cyberark/conjur'
 --edge; optional flag to publish edge-tagged image to DockerHub

conjur$ TAG_NAME=v1.11.3 ./push-image.sh 
Tagging and pushing cyberark/conjur:latest...
Tagging and pushing cyberark/conjur:1.11.3...
Tagging and pushing cyberark/conjur:1.11...
Tagging and pushing cyberark/conjur:1...
./push-image.sh: line 82: REDHAT_API_KEY: unbound variable

conjur$ ./push-image.sh --edge
Tagging and pushing cyberark/conjur:edge...

conjur$ ./push-image.sh --registry-prefix=registry.tld
Tagging and pushing registry.tld/cyberark/conjur:1.11.1-b06ab2ce...
Tagging and pushing registry.tld/conjur-test:1.11.1-b06ab2ce...
Tagging and pushing registry.tld/conjur-ubi:1.11.1-b06ab2ce...

conjur$ ./push-image.sh --registry-prefix=registry.tld --edge
Tagging and pushing registry.tld/cyberark/conjur:edge...

conjur$ ./push-image.sh --registry-prefix registry.tld --edge
This script expects the flag value to be provided as --registry-prefix=STRING
Usage: ./push-image.sh [OPTION...]
 --registry-prefix=STRING; optional registry prefix to prepend to '/cyberark/conjur'
 --edge; optional flag to publish edge-tagged image to DockerHub
```

To produce this sample, I built the images using `./build.sh --jenkins`, commented out the `docker push` line in `build_utils.sh`, and ran through the commands shown.

Resolves #1617 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/push-image.sh
+++ b/push-image.sh
@@ -1,76 +1,106 @@
 #!/usr/bin/env bash
-set -e
+set -euo pipefail
 
-# Push the 'conjur' image to various Docker registries
-# Push tagged images on master branch
-# Release images can be created by passing the desired tag to this script
-# Ex: ./push-image 4.9.5.1
+# Pushes the 'conjur' image to various Docker registries
+# There are essentially three flows this script supports
+# - Pushing images to the internal registry during the early pipeline stage
+# - Pushing an "edge"-tagged image to DockerHub
+# - Publishing images to public registries on new project releases during a
+#   tag-triggered build
+#
+# The default flow when no parameters are supplied is to publish images to
+# public registries. This does require that the TAG_NAME variable is set, which
+# happens automatically in tag-triggered builds.
+#
+# To publish to a (private) internal registry, run the script with the
+# "--registry-prefix=*" flag to supply the registry prefix.
+#
+# To publish an "edge"-tagged release to DockerHub, run the script with the
+# "--edge" flag.
 
 # shellcheck disable=SC1091
-. version_utils.sh
+. build_utils.sh
 
-if [[ -z "${TAG_NAME:-}" ]]; then
-  echo "Please supply environment variable TAG_NAME."
-  echo "If you see this error in Jenkins it means the publish script was run"
-  echo "for a build that wasn't triggered by a tag -" \
-    "please check publish stage conditions."
-  exit 1
-fi
+function print_help() {
+  echo "Usage: $0 [OPTION...]"
+  echo " --registry-prefix=STRING; optional registry prefix to prepend to '/cyberark/conjur'"
+  echo " --edge; optional flag to publish edge-tagged image to DockerHub"
+}
 
-TAG="${1:-$(version_tag)}"
-VERSION="$(< VERSION)"
-SOURCE_IMAGE="conjur:$TAG"
-RH_SOURCE_IMAGE="conjur-ubi:$TAG"
-
-CONJUR_REGISTRY="registry.tld"
+REGISTRY_PREFIX=""
+PUBLISH_EDGE=false
+TAG="$(version_tag)"
+LOCAL_IMAGE="conjur:${TAG}"
+RH_LOCAL_IMAGE="conjur-ubi:${TAG}"
 IMAGE_NAME="cyberark/conjur"
 REDHAT_IMAGE="scan.connect.redhat.com/ospid-9fb7aea1-0c01-4527-8def-242f3cde7dc6/conjur"
 
-# both old-style 'conjur' and new-style 'cyberark/conjur'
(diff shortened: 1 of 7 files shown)


push-image.sh
```diff
@@ -1,76 +1,94 @@
...
-fi
-
-TAG="${1:-$(version_tag)}"
-VERSION="$(< VERSION)"
-SOURCE_IMAGE="conjur:$TAG"
-RH_SOURCE_IMAGE="conjur-ubi:$TAG"
+function print_help() {
+  echo "Usage: $0 [ -r <registry prefix> ] [ -e ]"
+  echo " -r optional registry prefix to prepend to '/cyberark/conjur'"
+  echo " -e optional flag to publish edge-tagged image to DockerHub"
```
micahlee, line 27: Can we change these to long-form names instead?
```
--registry-prefix
--edge
```

I would much prefer to see this in the Jenkinsfile:
```
./push-image.sh --edge
```

push-image.sh
```diff
@@ -1,76 +1,94 @@
...
-INTERNAL_IMAGES=("$CONJUR_REGISTRY/conjur" "$CONJUR_REGISTRY/$IMAGE_NAME")
+while getopts "r:e" opt; do
+    case "${opt}" in
+        r )
+          REGISTRY_PREFIX="${OPTARG}"
+          ;;
+        e )
+          PUBLISH_EDGE=true
+          ;;
+       \? )
```
micahlee, line 46: Where does the `\?` come from? Does this mean you would use `push-image.sh \?` to display the help text, and if so, what is the value of this over `--help|-h`?

Does it need to be in the `getopts` above as well?

micahlee (changes requested):

push-image.sh
izgeri: that's a fair point. I'll work that in.

push-image.sh
izgeri: I find [this blog](https://sookocheff.com/post/bash/parsing-bash-script-arguments-with-shopts/) to be a helpful reference when I'm using getopts. in it the author says:
> if an invalid option is provided, the option variable is assigned the value ?. You can catch this case and provide an appropriate usage message to the user.

I could also add a help flag, but I'm not sure it's worth it since I doubt this will be run manually very much.

micahlee: Ah, got it. And thanks for sharing the article on that, TIL. I agree a separate help flag is unnecessary. 

Thanks!

micahlee (dismissed):
LGTM! Thanks for the fixups.

micahlee (approved):

PR #1981, opened 2020-12-17
Body parser now raises proper error on invalid request body

### LINKED ISSUE #1968: Conjur returns 500 error on empty request body
## Summary
When a request is made to Conjur with an empty body but the header `Content-Type: application/json` the server will attempt to parse the request body and throw an uncaught error because it is invalid JSON.

## Steps to Reproduce
This is known to happen on the `/host_factory_tokens` endpoint using the `DELETE` method
1. Create a request with an empty body and the `Content-Type: application/json` header
2. Send the request to the server

## Expected Results
The server should handle the error nicely and return a 400 to the user.

## Actual Results (including error logs, if applicable)
A 500 error (and an HTML page) is returned

## Reproducible
   * [x] Always 
   * [ ] Sometimes
   * [ ] Non-Reproducible
   
## Environment setup
Running in a docker-compose environment from the cyberark/conjur:1.9 image.

telday:
Fixes issues where 500 error was raised when application/json
Content-Type header was specified with empty or malformed request body

### What does this PR do?
Adds error handling for malformed JSON body in request which will nicely catch the error then return a 400 error code rather than a 500 error.

Resolves #1968

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   are included in their Conjur image.
   [cyberark/conjur#1974](https://github.com/cyberark/conjur/issues/1974)
 
+### Fixed
+- Requests with empty body and application/json Content-Type Header will now
+  return 400 error instead of 500 error.
+  [cyberark/conjur#1968](https://github.com/cyberark/conjur/issues/1968)
+
 ## [1.11.1] - 2020-11-19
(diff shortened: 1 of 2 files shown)


micahlee (changes requested):
This looks great @telday, thanks for putting this together!

Mind adding a CHANGELOG entry for this fix and rebasing? Then it should be good to go.

micahlee (approved):

PR #1989, opened 2021-01-05
Updated how batch secrets are rendered in JSON

### LINKED ISSUE #1962: Conjur OSS API server fails to GET binary values
## Summary
Conjur OSS API server fails to GET binary values

## Steps to Reproduce

Set a variable value to a pkcs12

```console
root@conjur-admin-configure-geftdtqi38-c9cdc:/tmp# openssl pkcs12 -in truststore.p12 -nodes | openssl x509 -noout -enddate
Enter Import Password:
notAfter=Oct 27 14:53:44 2042 GMT
root@conjur-admin-configure-geftdtqi38-c9cdc:/tmp# conjur variable values add redacted/redacted/pki_truststore < truststore.p12 
Value added
root@conjur-admin-configure-geftdtqi38-c9cdc:/tmp# conjur variable value redacted/redacted/pki_truststore > test.p12
root@conjur-admin-configure-geftdtqi38-c9cdc:/tmp# openssl pkcs12 -in test.p12 -nodes | openssl x509 -noout -enddate
Enter Import Password:
notAfter=Oct 27 14:53:44 2042 GMT
```

Note that the CLI is able to add the variable to the pkcs12 value, and is also able to read it without breaking the file.

Use cyberark/secrets-provider-for-k8s:1.1.0 to update this secret

```yaml
kind: Secret
apiVersion: v1
metadata:
  namespace: {{ .Values.cycle }}-{{ .Values.project }}
  name: pki-truststore
  annotations:
    argocd.argoproj.io/hook: PreSync
    argocd.argoproj.io/sync-wave: "-15"
stringData:
  conjur-map: |-
    pki-truststore: redacted/redacted/pki_truststore
type: Opaque
```

The job doing the update

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  namespace: {{ .Values.cycle }}-{{ .Values.project }}
  name: conjur-secret-provider
  annotations:
    argocd.argoproj.io/hook: PreSync
    argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
    argocd.argoproj.io/sync-wave: "-10"
spec:
  activeDeadlineSeconds: 180
  template:
    spec:
      serviceAccountName: kubernetes-authenticator-client
      containers:
        - image: {{ .Values.conjur.secretsproviderfork8s.image | quote }}
          imagePullPolicy: IfNotPresent
          name: kubernetes-authenticator-client
          volumeMounts:
            - name: trusted-ca-bundle
              mountPath: /tmp/trusted-ca-bundle
          env:
            - name: DEBUG
              value: {{ .Values.conjur.debug | quote }}
            - name: CONTAINER_MODE
              value: init
            - name: MY_POD_NAME
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.name
            - name: MY_POD_NAMESPACE
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.namespace
            - name: MY_POD_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.podIP
            - name: CONJUR_VERSION
              valueFrom:
                configMapKeyRef:
                  name: config.conjur
                  key: version
            - name: CONJUR_APPLIANCE_URL
              valueFrom:
                configMapKeyRef:
                  name: config.conjur
                  key: applianceUrl
            - name: CONJUR_AUTHN_URL
              valueFrom:
                configMapKeyRef:
                  name: config.conjur
                  key: authenticationUrl
            - name: CONJUR_ACCOUNT
              valueFrom:
                configMapKeyRef:
                  name: config.conjur
                  key: account
            - name: CONJUR_AUTHN_LOGIN
              value: 'host/conjur/authn-k8s/authenticator/identities/redacted'
            - name: CONJUR_CERT_FILE
              value: '/tmp/trusted-ca-bundle/ca-bundle.crt'
            - name: K8S_SECRETS
              value: 'pki-truststore'
            - name: SECRETS_DESTINATION
              value: k8s_secrets
      volumes:
        - name: trusted-ca-bundle
          configMap:
            name: trusted-ca-bundle
      restartPolicy: Never
```

Logs of the secrets-provider-for-k8s job

```
INFO: 2020/11/28 13:08:36 k8s_secrets_client.go:19: CSPFK005I Retrieving Kubernetes secret 'pki-truststore' from namespace 'redacted'
DEBUG: 2020/11/28 13:08:36 provide_conjur_secrets.go:127: CSPFK009D Processing 'conjur-map' data entry value of Kubernetes Secret 'pki-truststore'
INFO: 2020/11/28 13:08:36 k8s_secrets_client.go:54: CSPFK004I Creating Kubernetes client
INFO: 2020/11/28 13:08:36 conjur_secrets_retriever.go:11: CSPFK003I Retrieving following secrets from DAP/Conjur: [redacted/redacted/pki_truststore]
INFO: 2020/11/28 13:08:36 conjur_client.go:21: CSPFK002I Creating DAP/Conjur client
ERROR: 2020/11/28 13:08:36 provide_conjur_secrets.go:77: CSPFK034E Failed to retrieve DAP/Conjur secrets. Reason: 500 Internal Server Error. 
ERROR: 2020/11/28 13:08:36 main.go:94: CSPFK016E Failed to provide DAP/Conjur secrets
```

Matching logs in conjur-oss

```
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] default:host:conjur/authn-k8s/authenticator/identities/redacted fetched default:variable:redacted/redacted/pki_truststore
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] Completed 500 Internal Server Error in 18ms
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32]   
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] Encoding::UndefinedConversionError ("\x82" from ASCII-8BIT to UTF-8):
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32]   
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] app/controllers/secrets_controller.rb:73:in `batch'
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] app/controllers/application_controller.rb:74:in `block in run_with_transaction'
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] app/controllers/application_controller.rb:73:in `run_with_transaction'
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] lib/rack/remove_request_parameters.rb:26:in `call'
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] lib/rack/default_content_type.rb:68:in `call'
```

## Expected Results
conjur-oss does not have a UndefinedConversionError and secrets-provider-for-k8s is able to update my secret with the pkcs12 value

## Reproducible
Always
   
## Version/Tag number
repository: cyberark/conjur
  tag: 'v1.11.0'

repository: nginx
    tag: '1.15'

repository: postgres
    tag: '10.12'

## Environment setup
OpenShift cluster version 4.5.19.

## Additional Information
Note that we use such setup to fetch OK many different kinds of secrets, including json documents and SSL private keys

telday:
### What does this PR do?
Changed how binary secrets are rendered in batch requests.

#### Some Details on the Implementation:
Based on feedback in a slack discussion I have moved to a different solution for this issue.

Now on requests to the batch secrets endpoint Conjur will read the `Accept` header from the request. Currently if the header is `base64` it will encode **all** of the secrets in `base64` and return them in the same JSON format we currently use. If the header is set to anything else (or not set at all) the endpoint will act as it has previously, except now returning a 501 with an error message to the user. The up side of this approach is that it will maintain compatibility and is a simple implementation. 

Here is an example of a returned JSON object with the `Accept: base64` header set:
```JSON
{
  "cucumber:variable:secret2": "djI=\n",
  "cucumber:variable:secret3": "+6qohmBo1KyHByGIXejpYQ==\n"
}
```

The original value of secret2 was `"v2"` and secret3 was binary data.

I also included a conversion of the header value to lowercase to avoid issues between `base64` and `Base64` because I know some people prefer one over the other.

Resolves #1962

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/conjur#1974](https://github.com/cyberark/conjur/issues/1974)
 - Correct unit tests and integration tests for audit, and correct a couple of issues found with them.
   [cyberark/conjur#1987](https://github.com/cyberark/conjur/issues/1987)
+- When batch secret requests are sent with an `Accept: base64` header, the secret values in the response will all be
+  Base64-encoded. Sending requests with this header allows users to retrieve binary secrets encoded in Base64.
+  [cyberark/conjur#1962](https://github.com/cyberark/conjur/issues/1962)
 
 ### Fixed
 - Requests with empty body and application/json Content-Type Header will now
@@ -30,6 +33,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Audit engine routing now correctly matches URLs that include a period (`.`)
   in the resource ID.
   [cyberark/conjur#2001](https://github.com/cyberark/conjur/issues/2001)
+- Attempts to retrieve binary secret data in a batch request without using the `Accept: base64` header now returns a
+  message explaining that improper secret encoding is the cause of the 500 response.
+  [cyberark/conjur#1962](https://github.com/cyberark/conjur/issues/1962)
 
 ## [1.11.1] - 2020-11-19
 ### Added
--- a/app/controllers/secrets_controller.rb
+++ b/app/controllers/secrets_controller.rb
@@ -57,21 +57,35 @@ class SecretsController < RestController
       raise Exceptions::RecordNotFound,
             variable_ids.find { |r| !variables.map(&:id).include?(r) }
     end
-    
+
     result = {}
 
     authorize_many variables, :execute
-    
+
     variables.each do |variable|
-      unless (secret = variable.last_secret)
-        raise Exceptions::RecordNotFound, variable.resource_id
-      end
-      
-      result[variable.resource_id] = secret.value
+      result[variable.resource_id] = get_secret_from_variable variable
+
       audit_fetch variable
     end
 
     render json: result
+  rescue Encoding::UndefinedConversionError
+    raise Errors::Conjur::BadSecretEncoding, result
+  end
+
+  def get_secret_from_variable variable
+    secret = variable.last_secret
+    unless secret
+      raise Exceptions::RecordNotFound, variable.resource_id
+    end
+
+    secret_value = secret.value
+
+    if request.headers['Accept'].casecmp?('base64')
+      Base64.encode64(secret_value)
+    else
+      secret_value
+    end
   end
 
   def audit_fetch resource, version: nil
--- a/cucumber/api/features/secrets_batch.feature
+++ b/cucumber/api/features/secrets_batch.feature
@@ -91,3 +91,26 @@ Feature: Batch retrieval of secrets
     { "cucumber:variable:secret1": "v3", "cucumber:variable:secret2": "v5", "cucumber:variable:secret3": "v6" }
     """
 
+  Scenario: Returns the correct result for binary secrets
+    Given I create a binary secret value for resource "cucumber:variable:secret3"
+    And I add the secret value "v2" to the resource "cucumber:variable:secret2"
+    And I set the "Accept" header to "base64"
+    When I GET "/secrets?variable_ids=cucumber:variable:secret3,cucumber:variable:secret2"
+    Then the binary data is preserved for "cucumber:variable:secret3"
+
+  Scenario: Returns the correct result for binary secrets
+    Given I create a binary secret value for resource "cucumber:variable:secret3"
+    And I set the "Accept" header to "Base64"
+    When I GET "/secrets?variable_ids=cucumber:variable:secret3"
+    Then the binary data is preserved for "cucumber:variable:secret3"
+
+  Scenario: Returns the correct result for binary secrets
+    Given I create a binary secret value for resource "cucumber:variable:secret3"
+    When I GET "/secrets?variable_ids=cucumber:variable:secret3"
+    Then the HTTP response status code is 500
+
+  Scenario: Raises error on binary secret with no annotation
+    Given I create a binary secret value for resource "cucumber:variable:secret3"
+    And I add the secret value "v2" to the resource "cucumber:variable:secret2"
+    When I GET "/secrets?variable_ids=cucumber:variable:secret3,cucumber:variable:secret2"
+    Then the HTTP response status code is 500
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -25,6 +25,12 @@ module Errors
       code: "CONJ00065E"
     )
 
+    BadSecretEncoding = ::Util::TrackableErrorClass.new(
+      msg:  "Issue encoding secret into JSON format, try including 'Accept: base64' " \
+          "header in request.",
+      code: "CONJ00074E"
+    )
+
   end
 
   module Authentication
--- a/cucumber/api/features/step_definitions/authz_steps.rb
+++ b/cucumber/api/features/step_definitions/authz_steps.rb
@@ -86,6 +86,11 @@ Given(/^I create a binary secret value?$/) do
   Secret.create resource_id: @current_resource.id, value: @value
 end
 
+Given(/^I create a binary secret value for resource "([^"]*)"?$/) do |resource_id|
+  @value = Random.new.bytes(16)
+  Secret.create resource_id: resource_id, value: @value
+end
+
 Given(/^I add the secret value(?: "([^"]*)")? to the resource(?: "([^"]*)")?$/) do |value, resource_id|
   Secret.create resource_id: resource_id, value: value
 end
--- a/cucumber/api/features/step_definitions/response_steps.rb
+++ b/cucumber/api/features/step_definitions/response_steps.rb
@@ -57,3 +57,8 @@ Then(/^the binary result is preserved$/) do
   expect(@result.headers[:content_type]).to eq("application/octet-stream")
   expect(@result).to eq(@value)
 end
+
+Then(/^the binary data is preserved for "([^"]*)"$/) do |resource_id|
+  data = Base64.decode64(@result[resource_id])
+  expect(data).to eq(@value)
+end


izgeri (commented):
this should definitely be reviewed by @cyberark/conjur-core-team, but I wanted to note that this dependency is fine to add since it's available under MIT. We just need to add its license to the NOTICES.txt file in this project, and should probably do so in this PR: https://github.com/brianmario/yajl-ruby/blob/1.4.1/LICENSE

the PR is also missing a CHANGELOG message explaining the change that was made and the impact to the end user.

izgeri (changes requested):
@telday are there any tests we can add so that we can verify this functionality corrects the reported issue?

I actually find it really helpful in cases like this to add broken test(s) so that I can be sure that my updated functionality addresses the issue (eg the tests go green with my fix).

in this case, I'd guess we could manually create a `truststore.p12` file using `openssl pkcs12 -in truststore.p12 -nodes | openssl x509 -noout -enddate`, add that to the repo test dir, and then add a test that loads that into a variable and attempts to retrieve it via the batch retrieval endpoint / read the output

Gemfile
```diff
@@ -55,6 +55,7 @@ gem 'conjur-policy-parser', path: 'gems/policy-parser'
 gem 'conjur-rack', '~> 4'
 gem 'conjur-rack-heartbeat'
 gem 'rack-rewrite'
+gem 'yajl-ruby', require: 'yajl'
```
orenbm, line 58: @cyberark/conjur-core-team need to review any new gems that are added to the project.

orenbm: @telday do you think that a comment on this gem will help developers understand its purpose?

telday:
@izgeri @orenbm On a second pass looking at this issue I think it may require a little more knowledge on conjur in order to fix. The biggest issue is encoding secrets in JSON, which has no support for a `bytes` type. Therefore even if you manage to get a library to just treat your binary data as a string (which this library didn't actually perform correctly as I learned). There are going to be encoding errors in almost all libraries on the client side (I encountered serious issues while attempting to write a cucumber scenario and it doesn't look like Python likes it very much either). I think attempting to encode a binary secret as a badly-encoded string is not the best approach.

The second approach I tried to take was to encode the byte data of the secret as an array of integers in the JSON. However this doesn't seem like a great solution, users would most likely be very confused especially if retrieving the secret individually returns just the raw data. This also forces all binary secrets to have their mime type annotation specified.

Have I missed something that would make this significantly easier?

doodlesbykumbi:
@telday I like the idea of the batch retrieval endpoint specifying the mimetype for each secret retrieved (or grouping them by mimetype to reduce duplication). Such an approach would be consistent with the endpoint for single secret retrieval.

Re: encodings, when I first saw this issue I thought BSON might be useful here, see http://bsonspec.org/implementations.html. It's a lot more flexible than JSON. But it might also be overkill. Things like bytes can be represented as arrays (though maybe not the most efficient representation) as long as there's metadata to allow us to recover the original.

telday:
@orenbm @doodlesbykumbi @izgeri  Just wanted to throw out my implementation plan for fixing this.  I looked into using BSON for encoding the result, however because it is not a strict superset of JSON I think it would be too much to ask all users to switch to using a BSON library when they are not often built into languages. So what I am going to implement is checking the mime_type annotation for each secret, and if it is the `application/octet-stream` type I will encode the value in `base64` and include extra `encoding` and `mime_type` fields in the secrets return object. This will make sure the user understands that their data has been encoded so it is valid for normal JSON format and what encoding type was used. 
e.g.
```json
{
  "cucumber:variable:secret2": "v2",
  "cucumber:variable:secret3": {
    "value": "d/U4bDbw3WQzYW5O7AcZzQ==\n",
    "encoding": "base64",
    "mime_type": "application/octet-stream"
  }
}
```

It is not a perfect solution, however I think it is the right way to go for two main reasons:
* We wont have to introduce any new libraries into Conjur and most languages have JSON and Base64 encoding libraries built in so users wont need anything extra either.
* The increase in return object size is smaller than creating a list of bytes (for very large binary values)

This does mean that all binary secrets will need to have the `application/octet-stream` annotation set, if it is not set and the server cannot properly encode the secret into JSON it will return a 421 error code implying a bad server configuration.

Let me know what you think/if anything needs to be changed. I coded up most of this as well as added a cucumber. Still need to add the return of a 421 on bad secret configuration.

app/controllers/application_controller.rb
```diff
@@ -276,6 +280,16 @@ def render_record_not_found e
...
   end
 
+  def bad_secret_encoding e
+    logger.debug "#{e}\n#{e.backtrace.join "\n"}"
+    render json: {
+      error: {
+        code: "bad_secret_encoding",
+        message: e.message
+      }
+    }, status: :internal_server_error
```
orenbm, line 290: is this a server error or a bad request error? not sure what causes this

cucumber/api/features/secrets_batch.feature
```diff
@@ -91,3 +91,15 @@ Feature: Batch retrieval of secrets
...
+    And I set annotation "conjur/mime_type" to "application/octet-stream"
+    And I add the secret value "v2" to the resource "cucumber:variable:secret2"
+    When I GET "/secrets?variable_ids=cucumber:variable:secret3,cucumber:variable:secret2"
+    Then the binary data is preserved in JSON
+
+  Scenario: Raises error on binary secret with no annotation
+    Given I create a binary secret value for resource "cucumber:variable:secret3"
+    And I add the secret value "v2" to the resource "cucumber:variable:secret2"
+    When I GET "/secrets?variable_ids=cucumber:variable:secret3,cucumber:variable:secret2"
+    Then the HTTP response status code is 500
```
orenbm, line 105: will this happen also in a single secret retrieval? do we have a test for that?

cucumber/api/features/secrets_batch.feature
```diff
@@ -91,3 +91,15 @@ Feature: Batch retrieval of secrets
     { "cucumber:variable:secret1": "v3", "cucumber:variable:secret2": "v5", "cucumber:variable:secret3": "v6" }
     """
 
+  Scenario: Returns the correct result for binary secrets
+    Given I create a binary secret value for resource "cucumber:variable:secret3"
+    And I set annotation "conjur/mime_type" to "application/octet-stream"
```
orenbm, line 96: is this annotation set on the secret in the DB?

app/controllers/secrets_controller.rb
```diff
@@ -66,8 +66,20 @@ def batch
...
         raise Exceptions::RecordNotFound, variable.resource_id
       end
-      
-      result[variable.resource_id] = secret.value
+
+      mime_type = variable.annotation('conjur/mime_type') || nil
+
+      if mime_type == 'application/octet-stream'
+        unless result.key?('application/octet-stream')
+          result['application/octet-stream'] = { 'encoding': 'base64' }
```
doodlesbykumbi, line 74: I think base64 encoding is robust enough to address the current issue and potentially moving forward. With that in mind, perhaps we can always base64 encode values and always specify the mimeType. This throws out any space savings etc., but I don't think that's altogether necessary and might really just be premature optimisation. The response payload might look something like this.

By avoiding sending raw values we don't ever have to know or reason about what is inside.
```
{
  "id/of/secret": {
    "base64value": "",
    "mimeType": ""
  },
  ...
}
```

izgeri:
@telday given what you've implemented, and to help people understand the impact of this change, can you please update the PR description to include an example of what the response looks like now vs what it will look like after this change? I am trying to understand whether this is a breaking change, and it's hard to tell from what's written here.

app/controllers/application_controller.rb
telday: It is caused by a binary secret stored without specifying the mime type. This causes the server to fail to encode it into a string to be included in a JSON object. This is why I considered it to be a server error not a request error, because there is nothing technically wrong with the request. This wouldn't be an issue if we chose to encode all secrets returned as base64 strings as [suggested by](https://github.com/cyberark/conjur/pull/1989#discussion_r555913613) @doodlesbykumbi

cucumber/api/features/secrets_batch.feature
telday: We do currently have a [cucumber](https://github.com/cyberark/conjur/blob/master/cucumber/api/features/secrets.feature#L83) for single secret retrieval. There is not issue when only returning one secret because we do not attempt to return JSON. We simply reply with the secret data as the body.

cucumber/api/features/secrets_batch.feature
telday: Yes, it adds the annotation to the secret in the database

app/controllers/secrets_controller.rb
telday: I like the idea of encoding all secrets in base64, and we could group by mime type in order to reduce duplication in the JSON. It would be more consistent and resolve this issue in a cleaner manner so I would prefer it. The reason I didn't include it in this PR is because it would definitely be a breaking change for users. If we are willing to deal with the change being breaking then I can definitely put it in.

app/controllers/application_controller.rb
```diff
@@ -276,6 +280,16 @@ def render_record_not_found e
     }, status: :not_found
   end
 
+  def bad_secret_encoding e
```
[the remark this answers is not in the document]
jtuttle: @telday You can add a [reek exception](https://github.com/conjurinc/container-dap/blob/master/.reek.yml#L5) for this if you like. Using the variable name 'e' is a common pattern in Ruby.

app/controllers/secrets_controller.rb
```diff
@@ -66,8 +66,20 @@ def batch
...
+        end
+        result['application/octet-stream'][variable.resource_id] = Base64.encode64(secret.value)
+      elsif secret.value.class == String && !secret.value.force_encoding('utf-8').valid_encoding?
```
jtuttle, line 77: Here are some [more idiomatic ways](https://stackoverflow.com/questions/7749416/check-whether-a-variable-is-a-string-in-ruby) to determine if something is a String.

app/controllers/secrets_controller.rb
```diff
@@ -66,8 +66,20 @@ def batch
       unless (secret = variable.last_secret)
```
jonahx, line 66: Existing code but this should be should fixed: We shouldn't combine assignment with an unless check.

app/controllers/secrets_controller.rb
```diff
@@ -66,8 +66,20 @@ def batch
...
         raise Exceptions::RecordNotFound, variable.resource_id
       end
-      
-      result[variable.resource_id] = secret.value
+
+      mime_type = variable.annotation('conjur/mime_type') || nil
```
jonahx, line 70: What is `|| nil` doing here?  Since only `false` and `nil` evaluate to falsey in ruby, the only thing this could be doing afaict is converting a `variable.annotation('conjur/mime_type')` of false to nil.  Is this in fact what it's intention is?  Otherwise you should be able to remove the `|| nil`.

app/controllers/secrets_controller.rb
jonahx: I'd also add to John's comment that the whole block would read better if we remove the nested if/unless checks, and pre-assign the conditional clauses to meaningful variable names.

I'd also re-structure it to use [line of sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88).

jonahx:
I don't think I'm crazy about grouping them to reduce duplication.  Do we know that's a problem?

I'd rather focus on what's more convenient to the user, and I think having a uniform of list of secrets, each with a "type" or "encoding" (do we need both?) fits that bill better.  For example, you can just map over them to unpack them, etc.

h-artzi:
Awesome work! 

I have a few comments:
1. I agree with @jonahx that I'm not crazy about grouping the secrets to reduce duplication.
2. Have we considered allowing a user to define the header "Accept: Base64", and then return a list of secrets whose values are all Base64 encoded?
```
{
  "cucumber:variable:secret2": "djI=",
  "cucumber:variable:secret3": "d/U4bDbw3WQzYW5O7AcZzQ==\n"
}
```
This would ensure complete backwards compatibility while also addressing the use case of retrieving binary secrets.
The user would just have to go through the list of secrets they are retrieving and decode all of them.

app/controllers/application_controller.rb
orenbm: thanks for the exaplanation @telday ! i agree that it should be a server error now that i understand it.

It seems that the client error was when they created the secret. am i correct? if so, can we fail such a request?

orenbm: (it can definitely be out of scope of this issue, just want to understand it)

orenbm: ok i read now your other comment. It seems that the secret itself is ok but is not only when retrieved in a batch request. Is that correct?

orenbm (approved):
Thanks @telday !

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Requests with empty body and application/json Content-Type Header will now
   return 400 error instead of 500 error.
   [cyberark/conjur#1968](https://github.com/cyberark/conjur/issues/1968)
+- Changed how batch secret requests encode binary secrets into JSON.
+  Batch requests including a secret with binary data will now return properly
```
izgeri, line 28: This should probably be updated now, right?

I think this is now a "Changed" item instead of a "Fixed", and it should read more like
> When batch secret requests are sent with an `Accept: base64` header, the secret values in the response will all be Base64-encoded. Sending batch secret requests with this header enables users to retrieve Base64-encoded binary secret data from Conjur.

Which makes me wonder - do we want to also update this endpoint to send a more elegant response than:
```
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] Completed 500 Internal Server Error in 18ms
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32]   
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] Encoding::UndefinedConversionError ("\x82" from ASCII-8BIT to UTF-8):
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32]   
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] app/controllers/secrets_controller.rb:73:in `batch'
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] app/controllers/application_controller.rb:74:in `block in run_with_transaction'
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] app/controllers/application_controller.rb:73:in `run_with_transaction'
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] lib/rack/remove_request_parameters.rb:26:in `call'
[origin=10.128.2.70] [request_id=cd294a8e-8a69-47ea-a45a-3b192a24d077] [tid=32] lib/rack/default_content_type.rb:68:in `call'
```
when someone tries to retrieve binary secret data in a batch request without sending the `Accept: base64` header? (Sorry if this is already addressed in the PR - if so, it's not clear from the description). I'm not sure what the correct error response code would be.

app/controllers/secrets_controller.rb
```diff
@@ -57,21 +57,35 @@ def batch
...
-      
-      result[variable.resource_id] = secret.value
+      result[variable.resource_id] = get_secret_from_variable variable
+
       audit_fetch variable
     end
 
     render json: result
+  rescue Encoding::UndefinedConversionError
+    raise ApplicationController::BadSecretEncoding, result
```
jonahx, line 73: This domain error should not be scoped to the `ApplicationController`.

CHANGELOG.md
telday: I just updated the PR description to clarify the error return code. It still does return a 500, however I added an error message which lets the user know what the actual issue is. I didn't want to make it one of the 400 responses because there is nothing wrong with the request itself, rather it is a server error. The return object is now something like this:

```json
{
  "error": {
    "code": "bad_secret_encoding",
    "message": "a string representation of the  ruby hash that failed to encode"
  }
}
```

app/controllers/secrets_controller.rb
telday: Should I create a separate Exceptions module for this? Similar to how [RecordNotFound](https://github.com/cyberark/conjur/blob/master/app/models/exceptions/record_not_found.rb#L4) is handled? Or is there another approach I should use?

app/controllers/application_controller.rb
telday: Yeah, it is an issue with rendering a binary secret into the JSON document that we return in the batch endpoint. I think it would be possible to stop a user from creating a binary secret on a variable that does not have the correct mime_type annotation set, if that is something we want. I'm not totally sure if that would be useful though (at least for this endpoint based on the new implementation).

CHANGELOG.md
izgeri: I think you could add a `Fixed` changelog entry that's something like
> Attempting to retrieve binary secret data in a batch secret request without using the new `Accept: Base64` header now returns an informative error message to explain that improper secret encoding is the cause of the 500 response.

app/controllers/secrets_controller.rb
```diff
@@ -57,21 +57,35 @@ def batch
...
+
+  def get_secret_from_variable variable
+    secret = variable.last_secret
+    unless secret
+      raise Exceptions::RecordNotFound, variable.resource_id
+    end
+
+    secret_value = secret.value
+
+    if request.headers['Accept'].casecmp? 'base64'
```
jonahx, line 84: CodeClimate should be flagging this for missing parens.

jonahx (approved):

izgeri (approved):
Great work on this @telday!! Thanks for sticking with it!

saprette:
@telday  @izgeri  @orenbm  Hello, thanks for the fix. Will you update the requests of secrets-provider-for-k8s to add the Accept: base64 header ?

telday:
@saprette going open an issue and change the requests this morning

alexkalish:
@telday @doodlesbykumbi @izgeri: This solution cleverly handles backwards compatibility, which is excellent.  But is it slightly abusing the `Accept` header?  If the user adds the `Accept: base64`, wouldn't their client be expecting that the entire response is base64 encoded, not just the secrets values as see in the example above?  Looking at the code, seems like we are still returning a JSON body and JSON `Content-Type`.  Is there a chance that this could confuse some HTTP clients that try to parse the body based on the content type?

Also, as a reminder, please be sure to file a documentation ticket when adding/changing product behavior.  Thanks!

izgeri:
@alexkalish and I connected on this, and we're not going to include this in the documentation until we find a satisfying answer to the question he raised above ^

Please also note that the behavior was changed in #2065 to use `Accept-Encoding` instead of `Accept`

telday:
@alexkalish I agree that it is to some extent an abuse of the `Accept-Encoding` header. There may be cleaner solutions to this problem (some of them possible backwards incompatible). Perhaps this warrants opening a `techdebt` issue?

Issue #1714, opened 2020-07-27
Test and Document the `X-Request-Id` header


micahlee:
Conjur includes [middleware](https://github.com/cyberark/conjur/blob/master/lib/rack/remember_uuid.rb) for maintaining a well-known ID across API requests that will display in the audit log. This behavior is not tested or documented though.

DoD:
- [x] Tests exist to demonstrate and verify the behavior of the `x-request-id` HTTP header.
- [x] The behavior of `x-request-id` is documented in DAP wiki.
- [ ] An issue exists to document this behavior in the official docs.

See also: https://api.rubyonrails.org/classes/ActionDispatch/RequestId.html

h-artzi:
https://github.com/cyberark/conjur-docs/issues/1333

https://github.com/cyberark/conjur/pull/1994

https://github.com/cyberark/dap-wiki/pull/159

PR #1994, opened 2021-01-12
1714 x request id is tested

h-artzi:
### What does this PR do?
Adds tests to ensure that when the `X-Request-Id` header is set during a REST API call. The header appears in the audit record

#1714

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/cucumber/api/features/authenticate.feature
+++ b/cucumber/api/features/authenticate.feature
@@ -20,6 +20,19 @@ Feature: Exchange a role's API key for a signed authentication token
       cucumber:user:alice successfully authenticated with authenticator authn
     """
 
+  Scenario: X-Request-Id is set and visible in the audit record
+    Given I set the "X-Request-Id" header to "TestMyApp"
+    Then I can POST "/authn/cucumber/alice/authenticate" with plain text body ":cucumber:user:alice_api_key"
+    And the HTTP response content type is "application/json"
+    And there is an audit record matching:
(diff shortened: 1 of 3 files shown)


cucumber/api/features/authenticate.feature
```diff
@@ -20,6 +20,19 @@ Feature: Exchange a role's API key for a signed authentication token
       cucumber:user:alice successfully authenticated with authenticator authn
     """
 
+    Scenario: X-Request-Id is set and visible in the audit record
```
orenbm, line 23: can you indent this line with the rest of the scenario titles?

cucumber/api/features/authenticate.feature
```diff
@@ -20,6 +20,19 @@ Feature: Exchange a role's API key for a signed authentication token
...
+    Then I set the "X-Request-Id" header to "TestMyApp"
```
orenbm, line 24: should this be `When`?

orenbm, line 24: out of curiosity - what's the use-case here? how will this benefit a real user?

h-artzi: I believe it was just an undocumented/test `behavior`. It could be helpful during debugging, it would allow us to easily find the logs that are correlated with a request. CC: @alexkalish

micahlee (approved):

PR #1998, opened 2021-01-19
Fixed delimiter validation in resources endpoint

### LINKED ISSUE #1997: Offset accepts string as valid input
## Summary
Our REST API for list, accepts a string for `offset` as valid data when it should only accept a numeric value. When we run `offset=somestring`, we get the JSON data from the Conjur server without any constraints applied to it as if we were to run `conjur list`

For example, the returned data would resemble the following
```
      {
        "version": 2166,
        "created_at": "2021-01-14T09:31:31.605+00:00",
        "policy_text": "- !layer somelayer\n- !group somegroup\n- !host anotherhost\n- !user someuser\n- !variable 
...
        "client_ip": "54.236.50.115",
        "id": "cucumber:policy:root",
        "role": "cucumber:user:admin"
      },
```

## Steps to Reproduce
1. Run
```
curl -k -H "$(conjur authn authenticate -H)" \
   'https://conjur-server/resources/cucumber?offset=somestring' \
     | jq .
```

## Expected Results
A non-numeric value for `offset` should return a 500 internal server error like `limit` does

## Actual Results (including error logs, if applicable)
```
...
      {
        "version": 2166,
        "created_at": "2021-01-14T09:31:31.605+00:00",
...
        "finished_at": "2021-01-14T09:31:31.747+00:00",
        "client_ip": "54.236.50.115",
        "id": "cucumber:policy:root",
        "role": "cucumber:user:admin"
      },
...
```
## Reproducible
   * [x] Always 
   * [ ] Sometimes
   * [ ] Non-Reproducible
   
## Version/Tag number
All versions

## Environment setup
DAP server in AWS, Conjur CLI on my local machine

## Additional Information
NA

liavyona:
### What does this PR do?
- _What's changed? Added delimiter validation to `resources` endpoint
image

image

Resolves #1997

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - When batch secret requests are sent with an `Accept: base64` header, the secret values in the response will all be
   Base64-encoded. Sending requests with this header allows users to retrieve binary secrets encoded in Base64.
   [cyberark/conjur#1962](https://github.com/cyberark/conjur/issues/1962)
+- Conjur now verifies that the `offset` parameter is a valid integer value.
+  The `GET /resources` request will fail if `offset` is not an integer greater than or equal to 0.
+  [cyberark/conjur#1997](https://github.com/cyberark/conjur/issues/1997)
 
 ### Fixed
 - Requests with empty body and application/json Content-Type Header will now
@@ -36,6 +39,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Attempts to retrieve binary secret data in a batch request without using the `Accept: base64` header now returns a
   message explaining that improper secret encoding is the cause of the 500 response.
   [cyberark/conjur#1962](https://github.com/cyberark/conjur/issues/1962)
+- `GET /resources` request with non-numeric delimiter (limit or offset) now
+  returns `Error 422 Unprocessable Entity` instead of `Error 500`.
+  [cyberark/conjur#1997](https://github.com/cyberark/conjur/issues/1997)
 
 ## [1.11.1] - 2020-11-19
 ### Added
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -3,14 +3,14 @@
 class ResourcesController < RestController
   include FindResource
   include AssumedRole
-  
+
   def index
     # Rails 5 requires parameters to be explicitly permitted before converting 
     # to Hash.  See: https://stackoverflow.com/a/46029524
     allowed_params = %i(account kind limit offset search)
     options = params.permit(*allowed_params)
       .slice(*allowed_params).to_h.symbolize_keys
-    
+
     if params[:owner]
       ownerid = Role.make_full_id(params[:owner], account)
       options[:owner] = Role[ownerid] or raise Exceptions::RecordNotFound, ownerid
@@ -19,7 +19,13 @@ class ResourcesController < RestController
     # The v5 API currently sends +acting_as+ when listing resources
     # for a role other than the current user.
     query_role = params[:role].presence || params[:acting_as].presence
-    scope = Resource.visible_to(assumed_role(query_role)).search options
+    begin
+      scope = Resource.visible_to(assumed_role(query_role)).search options
+    rescue ApplicationController::Forbidden
+      raise
+    rescue ArgumentError => e
+      raise ApplicationController::UnprocessableEntity, e.message
+    end
 
     result =
       if params[:count] == 'true'
@@ -32,14 +38,14 @@ class ResourcesController < RestController
           eager(:policy_versions).
           all
       end
-  
+
     render json: result
   end
-  
+
   def show
     render json: resource
   end
-  
+
   def permitted_roles
     privilege = params[:privilege] || params[:permission]
     raise ArgumentError, "privilege" unless privilege
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -2,9 +2,9 @@
 
 class Resource < Sequel::Model
   include HasId
-  
+
   unrestrict_primary_key
-  
+
   one_to_many :permissions, reciprocal: :resource
   one_to_many :annotations, reciprocal: :resource
   one_to_many :secrets, reciprocal: :resource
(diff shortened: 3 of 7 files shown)


app/models/resource.rb
```diff
@@ -105,6 +105,9 @@ def search account: nil, kind: nil, owner: nil, offset: nil, limit: nil, search:
       scope = scope.textsearch(search) if search
 
       if offset || limit
+        if (offset && !numeric?(offset)) || (limit && !numeric?(limit))
+          raise ApplicationController::UnprocessableEntity, "Delimiter must be an integer if given"
```
orenbm, line 109: @shulifink can you please review this message?

orenbm (approved):
LGTM @liavyona , thanks! 
please get a review from @shulifink on the message.

izgeri:
@liavyona please make sure to add a change log message for this! it's definitely a user-impacting change, and we should let people know we've fixed this.

orenbm:
@liavyona the code is fine by me but i would like to see a better commit message. please read [this post](https://chris.beams.io/posts/git-commit/). To be specific, please change the tense to `Fix` instead of `Fixed` and add a description that will explain this change a bit better for the future devs who will see it.

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Requests with empty body and application/json Content-Type Header will now
   return 400 error instead of 500 error.
   [cyberark/conjur#1968](https://github.com/cyberark/conjur/issues/1968)
+- `resources` requests with non numeric delimiter (limit or offset) will now
```
sigalsax, line 27: it is more than that. Your change will not count non numeric input as valid input because before offset=somestring returned a whole list previously

app/models/resource.rb
```diff
@@ -105,6 +105,12 @@ def search account: nil, kind: nil, owner: nil, offset: nil, limit: nil, search:
       scope = scope.textsearch(search) if search
 
       if offset || limit
+        if limit && !numeric?(limit)
+          raise ApplicationController::UnprocessableEntity, "'limit' contains an invalid value. Limit must be an integer greater than 0."
+        end
+        if offset && !numeric?(offset)
+          raise ApplicationController::UnprocessableEntity, "'offset' contains an invalid value. Offset must be an integer."
```
sigalsax, line 112: why isn't this also "must be an integer greater than 0"? Is offset -3 valid and should we accept that?

app/models/resource.rb
```diff
@@ -105,6 +105,14 @@ def search account: nil, kind: nil, owner: nil, offset: nil, limit: nil, search:
       scope = scope.textsearch(search) if search
 
       if offset || limit
+        # 'limit' must be an integer greater than 0 if given
+        if limit && (!numeric?(limit) || limit.to_i <= 0)
+          raise ApplicationController::UnprocessableEntity, "'limit' contains an invalid value. Limit must be an integer greater than 0."
+        end
+        # 'offset' must be an integer greater than or equal to 0 if given
+        if offset && (!numeric?(offset) || offset.to_i < 0)
```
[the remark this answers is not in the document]
sigalsax: @liavyona what about this recommendation?

sigalsax (approved):
Great work! Thanks for jumping on this bug
A couple of things 1. as Oren mentioned please check with Shuli that these logs are ok before merging 2. I tagged you in a codeclimate suggestion so check that out :)

app/models/resource.rb
shulifink: I don't understand. What do you mean "if given"?

shulifink: @liavyona Isn't it enough to just say "If you provide a value for Delimiter, it must be an integer greater than or equal to 0".

or simply: "Delimiter must be an integer greater than or equal to 0".

or  "... must be an integer", or "must be a positive integer " (depending on what the integer type)

shulifink (changes requested):

sigalsax (dismissed):
LGTM! Thanks for this improvement!

app/models/resource.rb
```diff
@@ -105,6 +104,14 @@ def search account: nil, kind: nil, owner: nil, offset: nil, limit: nil, search:
...
 
       if offset || limit
+        # 'limit' must be an integer greater than 0 if given
+        if limit && (!numeric?(limit) || limit.to_i <= 0)
+          raise ApplicationController::UnprocessableEntity, "'limit' contains an invalid value. 'limit' must be a positive integer."
+        end
+        # 'offset' must be an integer greater than or equal to 0 if given
+        if offset && (!numeric?(offset) || offset.to_i.negative?)
+          raise ApplicationController::UnprocessableEntity, "'offset' contains an invalid value. 'offset' must be an integer greater than or equal to 0."
+        end
```
jonahx, line 114: `UnprocessableEntity` is an http level concern.  That is, we should only be talking about that error inside of controllers, and specifically in code that is mapping an actual domain error into an http status code.

Inside of the model we should be raising a more specific validation error, with a name that reflects what went wrong.  `InvalidQueryParameter` or something like that.

Separate comment: There's a lot of formatting / clean changes here (which is awesome), but please put those into a separate commit with a subject like "Cleanup formatting" -- otherwise the "meat" of this PR (which is actually pretty small and simple) isn't clear at a glance.

liavyona: We handle the `UnprocessableEntity` with rendering the response. I think this a redundant addition.

liavyona: About the formatting I agree. But I think that since it is a very small PR (both logic and formatting) the separation here is not mandatory.
Is it a blocker? @jonahx

liavyona: @jonahx fixed

jonahx (dismissed):

InbalZilberman (commented):
Looks great @liavyona 
Thank you for taking this on.

app/controllers/resources_controller.rb
```diff
@@ -19,7 +19,13 @@ def index
     # The v5 API currently sends +acting_as+ when listing resources
     # for a role other than the current user.
     query_role = params[:role].presence || params[:acting_as].presence
-    scope = Resource.visible_to(assumed_role(query_role)).search options
+    begin
+        scope = Resource.visible_to(assumed_role(query_role)).search options
+    rescue ApplicationController::Forbidden => e
+      raise e
```
orenbm, line 25: this can also be
```
rescue ApplicationController::Forbidden
  raise
```

app/models/resource.rb
```diff
@@ -2,43 +2,43 @@
...
       end
-      response["permissions"] = permissions.as_json.map {|h| h.except 'resource'}
-      response["annotations"] = self.annotations.as_json.map {|h| h.except 'resource'}
+      response["permissions"] = permissions.as_json.map { |h| h.except 'resource' }
+      response["annotations"] = self.annotations.as_json.map { |h| h.except 'resource' }
       case kind
       when "variable"
         response["secrets"] = secrets_dataset.order(:version).as_json
-          .map { |h| h.except 'resource' }
+                                             .map { |h| h.except 'resource' }
```
orenbm, line 37: can you fix all the non-code changes that you made to this file?

app/models/resource.rb
```diff
@@ -105,6 +105,14 @@ def search account: nil, kind: nil, owner: nil, offset: nil, limit: nil, search:
       scope = scope.textsearch(search) if search
 
       if offset || limit
+        # 'limit' must be an integer greater than 0 if given
+        if limit && (!numeric?(limit) || limit.to_i <= 0)
+          raise "'limit' contains an invalid value. 'limit' must be a positive integer."
+        end
+        # 'offset' must be an integer greater than or equal to 0 if given
+        if offset && (!numeric?(offset) || offset.to_i.negative?)
+          raise "'offset' contains an invalid value. 'offset' must be an integer greater than or equal to 0."
```
orenbm, line 114: @liavyona is it possible to move these errors to `errors.rb`?
we prefer to have all the user-facing messages there

cucumber/api/features/resource_list.feature
```diff
@@ -57,10 +57,42 @@ Feature: List resources with various types of filtering
     When I successfully GET "/resources/cucumber/test-resource?limit=1"
     Then I receive 1 resources
 
+  Scenario: The resource list cannot be limited with non numeric value
+    When I GET "/resources/cucumber/test-resource?limit=abc"
+    Then the HTTP response status code is 422
```
orenbm, line 62: do we want to add also log validation for these tests?

app/models/resource.rb
InbalZilberman: I would go with "Delimiter offset or limit must be an integer greater than or equal to 0"

app/models/resource.rb
liavyona: I think it will be a problem to use a custom exception inside a model class. Ads you can see there is no usage of any error which is defined at `errors.rb` under `app/model`. Other model classes use simple `raise` or `ArgumentError`. I changed it to use `ArgumentError`

app/models/resource.rb
liavyona: Talked about it

orenbm (dismissed):

orenbm (approved):

PR #2003, opened 2021-01-22
Require service-id param for authn-oidc

### LINKED ISSUE #2004: Raise proper error if service-id param is missing for authn-oidc
The OIDC authenticator cannot be initialized without a service-id so we should raise a proper error if it is missing. 

DoD:
- [ ] A proper error is raised if service-id is missing in an OIDC authentication request

orenbm:
### What does this PR do?
We do not have a scenario in which an OIDC authenticator
will be initialized without a service id. Thus, we should make
the `service-id` request param required.

Resolves #2004

#### Change log

#### Test coverage
- [x] This PR includes new unit and integration tests to go with the code changes

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/conjur#1997](https://github.com/cyberark/conjur/issues/1997)
 
 ### Fixed
+- Conjur now raises a proper error if the `service-id` param is missing in an 
+  authentication request for the OIDC authenticator.
+  [cyberark/conjur#2004](https://github.com/cyberark/conjur/issues/2004)
 - Requests with empty body and application/json Content-Type Header will now
   return 400 error instead of 500 error.
   [cyberark/conjur#1968](https://github.com/cyberark/conjur/issues/1968)
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -27,6 +27,7 @@ class AuthenticateController < ApplicationController
     )
     render json: { status: "ok" }
   rescue => e
+    log_backtrace(e)
     render status_failure_response(e)
   end
 
@@ -137,7 +138,9 @@ class AuthenticateController < ApplicationController
   private
 
   def handle_login_error(err)
-    log_error(err, LogMessages::Authentication::LoginError)
+    login_error = LogMessages::Authentication::LoginError.new(err.inspect)
+    logger.info(login_error)
+    log_backtrace(err)
 
     case err
     when Errors::Authentication::Security::AuthenticatorNotWhitelisted,
@@ -151,7 +154,9 @@ class AuthenticateController < ApplicationController
   end
 
   def handle_authentication_error(err)
-    log_error(err, LogMessages::Authentication::AuthenticationError)
+    authentication_error = LogMessages::Authentication::AuthenticationError.new(err.inspect)
+    logger.info(authentication_error)
+    log_backtrace(err)
 
     case err
     when Errors::Authentication::Security::AuthenticatorNotWhitelisted,
@@ -193,8 +198,7 @@ class AuthenticateController < ApplicationController
     end
   end
 
-  def log_error(err, log_message_class)
-    logger.info(log_message_class.new(err.inspect))
+  def log_backtrace(err)
     err.backtrace.each do |line|
       logger.debug(line)
     end
@@ -202,9 +206,6 @@ class AuthenticateController < ApplicationController
 
   def status_failure_response(error)
     logger.debug("Status check failed with error: #{error.inspect}")
-    error.backtrace.each do |line|
-      logger.debug(line)
-    end
 
     payload = {
       status: "error",
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -171,6 +171,11 @@ module Errors
         code: "CONJ00013E"
       )
 
+      ServiceIdMissing = ::Util::TrackableErrorClass.new(
+        msg:  "Service id is required when authenticating with authn-oidc",
+        code: "CONJ00075E"
+      )
+
     end
 
     module AuthnIam
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -27,10 +27,11 @@ Rails.application.routes.draw do
         patch '/:authenticator/:service_id/:account' => 'authenticate#update_config'
 
         get '/:authenticator(/:service_id)/:account/login' => 'authenticate#login'
-        # authn-oidc login & authenticate are currently for future use only
-        #post '/authn-oidc(/:service_id)/:account/login' => 'authenticate#login_oidc'
 
-        # authn-oidc has to be first as it can be ambgiuous with the optional :service_id & :id
+        # authn-oidc has to be first as it can be ambiguous with the optional :service_id in
+        # the common authn request and the fact that authn-oidc doesn't have an 'id' param.
+        # i.e the request 'authn-oidc/:service_id/:account/authenticate' can be interpreted as
+        # ':authenticator/:account/:id/authenticate'
         post '/authn-oidc(/:service_id)/:account/authenticate' => 'authenticate#authenticate_oidc'
         post '/authn-gcp/:account/authenticate' => 'authenticate#authenticate_gcp'
         post '/:authenticator(/:service_id)/:account/:id/authenticate' => 'authenticate#authenticate'
--- a/cucumber/authenticators_oidc/features/authn_status_oidc.feature
+++ b/cucumber/authenticators_oidc/features/authn_status_oidc.feature
@@ -216,6 +216,58 @@ Feature: OIDC Authenticator - Status Check
     Then the HTTP response status code is 500
     And the authenticator status check fails with error "RequiredResourceMissing: CONJ00036E"
 
+  Scenario: service-id missing and a 500 error response is returned
+    Given I load a policy:
+    """
+    - !policy
+      id: conjur/authn-oidc
+      body:
+      - !webservice
+
+      - !webservice
+        id: status
+        annotations:
+          description: Status service to verify the authenticator is configured correctly
+
+      - !variable
+        id: provider-uri
+
+      - !variable
+        id: id-token-user-property
+
+      - !group users
+
+      - !permit
+        role: !group users
+        privilege: [ read, authenticate ]
+        resource: !webservice
+
+      - !group
+        id: managers
+        annotations:
+          description: Group of users who can check the status of the authn-oidc authenticator
+
+      - !permit
+        role: !group managers
+        privilege: [ read ]
+        resource: !webservice status
+
+    - !user alice
+
+    - !grant
+      role: !group conjur/authn-oidc/users
+      member: !user alice
+
+    - !grant
+      role: !group conjur/authn-oidc/managers
+      member: !user alice
+    """
+    And I am the super-user
+    And I login as "alice"
+    When I GET "/authn-oidc/cucumber/status"
+    Then the HTTP response status code is 500
+    And the authenticator status check fails with error "Errors::Authentication::AuthnOidc::ServiceIdMissing"
+
   # TODO: add these tests when issue #1085 is done
 #  Scenario: provider-uri value has not been set and a 500 error response is returned
 #  Scenario: id-token-user-property value has not been set and a 500 error response is returned
--- a/cucumber/authenticators_oidc/features/step_definitions/authn_oidc_steps.rb
+++ b/cucumber/authenticators_oidc/features/step_definitions/authn_oidc_steps.rb
@@ -12,6 +12,11 @@ Given(/^I successfully set OIDC variables$/) do
   set_id_token_user_property_variable
 end
 
+Given(/^I successfully set OIDC variables without a service-id$/) do
+  set_provider_uri_variable_without_service_id
+  set_id_token_user_property_variable_without_service_id
+end
+
 Given(/^I successfully set provider-uri variable$/) do
   set_provider_uri_variable
 end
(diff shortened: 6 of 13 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/conjur#1987](https://github.com/cyberark/conjur/issues/1987)
 
 ### Fixed
+- Conjur now required the `service-id` param in an authentication request
+  for the OIDC authenticator as there is no use-case without it
```
InbalZilberman, line 25: Conjur now requires `service-id` param in an authentication request
  for the OIDC authenticator.

InbalZilberman (commented):
Thank you oren fortaking this.
Small comment was given
do we have tests also for status API?

eladkug (commented):
Please add such scenario to our tests. 
Means running without service-id will show us the relevant errors

orenbm:
@eladkug @InbalZilberman there are tests for this both for authentication and for status

app/controllers/authenticate_controller.rb
```diff
@@ -94,6 +94,17 @@ def authenticate_oidc
...
 
+  def authn_oidc_missing_service_id
+    # We don't need to log the backtrace as we didn't travel in the application
+    # code yet. Thus, we don't use 'handle_authentication_error' or 'log_error'
+    log_error(
+      Errors::Authentication::AuthnOidc::ServiceIdMissing.new,
+      LogMessages::Authentication::AuthenticationError,
+      false
+    )
+    raise Unauthorized
```
jonahx, line 112: The comment says:

> we don't use 'handle_authentication_error' or 'log_error'

but then we do use `log_error` in the next line.

orenbm: you are correct. it's a leftover from when i thought not to use it.

orenbm: fixed

app/controllers/authenticate_controller.rb
```diff
@@ -193,17 +211,21 @@ def handle_authentication_error(err)
     end
   end
 
-  def log_error(err, log_message_class)
+  def log_error(err, log_message_class, log_backtrace = true)
```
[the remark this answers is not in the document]
jonahx: I agree with this codeclimate suggestion.  We should have one method for logging just the error (`log_error` seems like a good name for that) and one for logging with the backtrace (I propose `log_backtrace`).

It also doesn't really make sense to pass in the error and a class to construct the error.  Remember, all of our errors descend from `RuntimeError`, and thus `backtrace` _should_ already available on them.

So the caller should just construct the error, and then both `log_error` and the proposed `log_backtrace` would take only a single argument -- the already constructed custom error object.

app/domain/errors.rb
```diff
@@ -171,6 +171,11 @@ module AuthnOidc
         code: "CONJ00013E"
       )
 
+      ServiceIdMissing = ::Util::TrackableErrorClass.new(
+        msg:  "Service id is required when authenticating with authn-oidc",
+        code: "CONJ00074E"
+      )
```
jonahx, line 177: Looks like you got race-conditioned out of this `code` by a PR which landed yesterday:

https://github.com/cyberark/conjur/blob/master/app/domain/errors.rb#L28

config/routes.rb
```diff
@@ -27,11 +27,15 @@ def matches?(request)
...
 
-        # authn-oidc has to be first as it can be ambgiuous with the optional :service_id & :id
+        # authn-oidc has to be first as it can be ambiguous with the optional :service_id in
+        # the common authn request and the fact that authn-oidc doesn't have an 'id' param.
+        # i.e the request 'authn-oidc/:service_id/:account/authenticate' can be interpreted as
+        # ':authenticator/:account/:id/authenticate'
         post '/authn-oidc/:service_id/:account/authenticate' => 'authenticate#authenticate_oidc'
+        # Verify there is a service-id in the request
+        post "/authn-oidc/:account/authenticate" => "authenticate#authn_oidc_missing_service_id"
+
```
jonahx, line 41: Is this here for backwards compatibility?  If so, I think we should add a comment saying that.

app/controllers/authenticate_controller.rb
```diff
@@ -193,17 +211,21 @@ def handle_authentication_error(err)
...
+      end
     end
   end
 
-  def status_failure_response(error)
+  def status_failure_response(error, log_backtrace = true)
     logger.debug("Status check failed with error: #{error.inspect}")
-    error.backtrace.each do |line|
-      logger.debug(line)
+    if log_backtrace
```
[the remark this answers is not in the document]
jonahx: Similar to my comments on the other one, I agree with the CC suggestion.

cucumber/authenticators_oidc/features/authn_status_oidc.feature
```diff
@@ -216,6 +216,59 @@ Feature: OIDC Authenticator - Status Check
...
+      - !webservice
+        id: status
+        annotations:
+          description: Status service to verify the authenticator is configured correctly
+
+      - !variable
+        id: provider-uri
+
+      - !variable
+        id: id-token-user-property
```
jonahx, line 236: Are these 2 variables needed?  Been a while since I looked at policy for OIDC...

dev/docker-compose.yml
```diff
@@ -127,13 +129,14 @@ services:
       - DB_VENDOR=H2
       - KEYCLOAK_CLIENT_ID=conjurClient
       - KEYCLOAK_REDIRECT_URI=http://locallhost.com/
-      - KEYCLOAK_CLIENT_SECRET=1234
+      - KEYCLOAK_CLIENT_SECRET=d7047915-9029-45b8-9bd6-7ec5c2f75e5b
```
jonahx, line 132: Can we add a comment explaining where this magic UUID comes from?

CHANGELOG.md
shulifink: @orenbm  Authentication requests for OIDC authenticator now require 'service-id' parameter

config/routes.rb
orenbm: it is not for backwards-compatibility. It's a verification on the feature as there's no use-case of OIDC without service-id

orenbm: i changed the comment to `Verify there is a service-id in the request as it's a requirement for OIDC`

cucumber/authenticators_oidc/features/authn_status_oidc.feature
orenbm: yes, these are the mandatory variables for oidc

dev/docker-compose.yml
orenbm: i don't really remember what this is and if it has any logic. @nessiLahav doesn't remember as well. I am fine with leaving it as it is without investigating this now.

config/routes.rb
jonahx: > It's a verification on the feature as there's no use-case of OIDC without service-id

In this case, it should simply return 404.  Otherwise aren't we just anticipating a random mistake a user might make?   This kind of special treatment would only make sense if the endpoint without the service_id used to work, and we've now changed it so it doesn't -- hence my question about backward compatibility.  Although that's probably not the most accurate term now that I think about it -- I meant something like "error message to improve UX for users of the now deprecated (non service_id) path".

dev/docker-compose.yml
jonahx: Maybe I'm misunderstanding, but I'm confused... it's marked as a change added for this commit.  Why would we merge a code change that no one knows the purpose of?   What happens if we change it back to `1234` or some other value -- does it stop working?

config/routes.rb
orenbm: we decided to go with this implementation to have a better UX. we prefer that the user will get the error `Errors::Authentication::AuthnOidc::ServiceIdMissing Service id is required when authenticating with authn-oidc` rather than `ActionController::RoutingError (No route matches [POST] "/authn-oidc/<account>/authenticate")`

dev/docker-compose.yml
orenbm: you previously made some changes to `ci/docker-compose` around the OIDC containers and forgot to change also `de/docker-compose` so the dev env of OIDC broke. As part of this fix i just took the client id from the `ci/docker-compose`. 

However, i tested it now with `1234` and it works so i'll change it to that in both places. better to have `1234` than a random UUID. it is clearer that it's just some sequence.

config/routes.rb
jonahx: Yeah I understand the logic, but why do we think a user would make this particular mistake?  Unless there's a prior API version that matched the route without `service_id`, there's no reason to anticipate this.  Think of it this way: by this reasoning, I could start going through every one of our routes, think of possible mistakes a user could make, and add a special route for that mistake with a custom error message -- all in the name of improved UX.

But that's clearly not something we'd want to do.  In addition, this is just not something I've ever seen done.  If you hit a route that doesn't exist, a generic 404 is the standard and correct response.  If we did want to go above and beyond that standard experience, and guess at possible routes a user may have mis-typed or omitted a parameter from, you wouldn't want to do this by adding no-op routes that raise custom errors for each possible mistake.  

Instead, you'd want to add generic code to your `:not_found` handler that looked at the requested non-existent route, compared it to our existing routes, and made suggestions based on that, sort of like a "spell checker" for routing.

orenbm: ok, i'll come back to this PR sometime next week and re-think

orenbm: btw this issue was raised after @telday tried to authenticate with an OIDC authenticator without a service-id. @telday do you want to give your input as a user who encountered this behaviour? what would make it easier for you to understand the issue?

InbalZilberman: @jonahx I disagree here. 
Ellis has raised this so this is why we think a user would make this mistake
if you think of all our authenticators as a whole some of them require service-id and some of them not. It can be confusing at time and it is best that we help the user with this.
if the user came to the actual path of ..../authn-oidc I believe it is because he forgotten something, again as happened to Ellis.
Improving the UX is something we should inspire to.

micahlee: @orenbm with the stated goals here, another alternative would be to leave the original route with the optional `service_id` and handle the empty value in the controller, rather than the additional route.

I agree an additional route for the empty service_id that only returns an error is not a good precedent.

jonahx: > Improving the UX is something we should inspire to.

@InbalZilberman Please re-read my previous post.  I am not objecting to improving the UX.  I am objecting to  implementing that improvement in a way that is inconsistent, and does not align with how routing is intended be used.  I proposed an alternative way to retain the UX improvement without these problems.

InbalZilberman: great! thanks for the clarification

orenbm: changed the behaviour according to this thread and a talk with @sigalsax . thank you all for sharing your thoughts!

cucumber/authenticators_oidc/features/step_definitions/authn_oidc_steps.rb
```diff
@@ -28,6 +33,10 @@
   authenticate_id_token_with_oidc(service_id: AuthnOidcHelper::SERVICE_ID, account: AuthnOidcHelper::ACCOUNT)
 end
 
+When(/^I authenticate via OIDC with id token and without a service-id$/) do
+  authenticate_id_token_with_oidc(service_id: nil, account: AuthnOidcHelper::ACCOUNT)
+end
+
```
jonahx, line 39: Formatting:
```ruby
When(/^I authenticate via OIDC with id token and without a service-id$/) do
  authenticate_id_token_with_oidc(
    service_id: nil,
    account: AuthnOidcHelper::ACCOUNT
  )
end
```

cucumber/authenticators_oidc/features/support/authn_oidc_helper.rb
```diff
@@ -27,8 +28,16 @@ def set_id_token_user_property_variable
     set_oidc_variable("id-token-user-property", oidc_id_token_user_property)
   end
 
-  def set_oidc_variable(variable_name, value)
-    path = "cucumber:variable:conjur/authn-oidc/keycloak"
+  def set_provider_uri_variable_without_service_id(value = oidc_provider_uri)
```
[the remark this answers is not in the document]
jonahx: This doesn't exactly apply here since you're not setting an instance variable.  Still, I agree that `set_` is not as precise as it could be.  I also consider `variable` to be a legacy naming mistake, and try not to perpetuate it.  With these two in mind I'd suggest either:

```
create_oidc_secret
```
or
```
create_oidc_variable
```

cucumber/authenticators_oidc/features/support/authn_oidc_helper.rb
```diff
@@ -27,8 +28,16 @@ def set_id_token_user_property_variable
...
+    set_oidc_variable("provider-uri", value, "")
+  end
+
+  def set_id_token_user_property_variable_without_service_id
+    set_oidc_variable("id-token-user-property", oidc_id_token_user_property, "")
+  end
```
jonahx, line 37: I don't think this (or the other methods with long names that wrap `set_oidc_variable`) add much value, and they do force one extra step of indirection when reading the code.  I would remove them and call `set_oidc_variable` directly from the step definitions.

cucumber/authenticators_oidc/features/support/authn_oidc_helper.rb
```diff
@@ -27,8 +28,16 @@ def set_id_token_user_property_variable
...
+
+  def set_oidc_variable(variable_name, value, service_id_suffix = "/keycloak")
+    path = "cucumber:variable:conjur/authn-oidc#{service_id_suffix}"
```
jonahx, line 40: See comment above on naming of the method.

jonahx (approved):

Issue #1898, opened 2020-10-22
Document the complete flow of our test infrastructure


jonahx:

sjacobs146:
Closing since this is not release dependent

PR #2008, opened 2021-01-25
1898 Document CI Architecture

jonahx:
### What does this PR do?
Adds an architectural diagram and detailed README explaining how our CI works.

#1898 

#### Change log

#### Test coverage

#### Documentation


README_CI.md
```diff
@@ -0,0 +1,109 @@
+# Jenkins CI Architecture
+
+<!-- Note: This diagram can be re-regenerated by editing img/README_CI.puml.
+The VS Code PlantUML extension provides a good live-preview experience. -->
+![CI Architecture](./img/ci_architecture.svg)
```
jtuttle, line 5: This is such a big enough diagram that it puts the ToC below the fold so I think I would prefer if the ToC came first.

README_CI.md
```diff
@@ -0,0 +1,109 @@
+# Jenkins CI Architecture
```
jtuttle, line 1: This seems like it's more focused on just the test-running part of the CI. It doesn't talk about changelog validation or vulnerability scanning, for example. Would it be better to rename this something more specific or Is there more content to come?

jtuttle (changes requested):

hughsaunders (approved):

README_CI.md
jonahx: I should probably add those.  Remember i said yesterday I was only ~80% done but wanted to ship :)

README_CI.md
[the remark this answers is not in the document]
jonahx: I should probably make this clear, but this is "System" in the sense of the C4 model:

https://c4model.com/

(scroll down to "Abstractions")

That said, just re-reading those, it looks like it should be a "container".  I proabably have some other mis-classifications as well -- I'll review.

README_CI.md
jtuttle: Oh gotcha, so should I just review without them for now?

jtuttle (dismissed):

jtuttle (approved):

PR #2009, opened 2021-01-25
Add audit query index to audit engine migrations

### LINKED ISSUE #2005: Improved Audit Messages DB Index for Resource and Role Queries
Adding a new index to the `messages` audit database table would allow queries that filter on these values to leverage the index rather than depend on sequential table scans.

The fields included are based on the queries defined in the audit engine here: https://github.com/cyberark/conjur/blob/master/engines/conjur_audit/app/models/conjur_audit/message.rb#L27

**NOTE:** due to the complexities of the Postgres query planner, the impact of this index isn't necessarily predictable nor can it be assumed will be used consistently to improve query improvement.

```sql
CREATE INDEX messages_timestamp_sdata_entity_idx on messages (
  -- Entity queries are always sorted by descending timestamp to retrieve
  -- the most recent audit events.
  timestamp desc,

  -- Index the structured data fields that are used to store Resources IDs
  (sdata #>> '{subject@43868, resource}'),
  (sdata #>> '{auth@43868, service}'),
  (sdata #>> '{policy@43868, policy}'),
  
  -- Index the structured data fields that are used to Role IDs
  (sdata #>> '{subject@43868, role}'),
  (sdata #>> '{subject@43868, member}'),
  (sdata #>> '{auth@43868, user}')
);
```

micahlee:
### What does this PR do?
This PR adds a new index to the `messages` audit database table that
allows queries filtering on these values to potentially leverage the index
rather than depend on sequential table scans.

This change is further verified in the appliance tests including the index here: https://jenkins.conjur.net/blue/organizations/jenkins/conjurinc--appliance/activity?branch=2005-audit-query-index

Resolves #2005

#### Change log

> I decided against a CHANGELOG entry for two reasons:
> 
> 1. There's no change for `cyberark/conjur` OSS users, since the audit database is only configured in the appliance.
> 2. The impact for the user would be more performant audit queries, but it's not a guaranteed improvement. So this change, without other enterprise changes, doesn't appreciably impact the user.

#### Test coverage

#### Documentation

### DIFF
--- a/engines/conjur_audit/db/migrate/20210125101831_add_activity_index.rb
+++ b/engines/conjur_audit/db/migrate/20210125101831_add_activity_index.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  # Adding a new index to the messages audit database table allows queries
+  # filtering on these values to potentially leverage the index rather than
+  # depend on sequential table scans.
+  #
+  # The fields included are based on the queries defined in the audit engine here:
+  # https://github.com/cyberark/conjur/blob/master/engines/conjur_audit/app/models/conjur_audit/message.rb#L27
+  #
+  # NOTE: Due to the statistical nature of the Postgres query planner and the
+  # storage characteristics of JSON column data, the impact of this index isn't
+  # predictable or guaranteed to consistently improve the query performance for
+  # all audit queries.
+  up do
+    execute(<<~SQL)
+      CREATE INDEX messages_timestamp_sdata_entity_idx on messages (
+        -- Entity queries are always sorted by descending timestamp to retrieve
+        -- the most recent audit events.
+        timestamp desc,
+      
+        -- Index the structured data fields that are used to store Resources IDs
+        (sdata #>> '{subject@43868, resource}'),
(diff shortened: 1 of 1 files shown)


izgeri:
@micahlee will this change necessitate any special upgrade instructions?

thanks!

micahlee:
> @micahlee will this change necessitate any special upgrade instructions?
> 
> thanks!

Hey @izgeri, feel free to correct me if you know of a place we do need upgrade instructions. However, how I understand it now is:

* Conjur OSS doesn't create or use this database at all.
* DAP applies any missing audit migrations [when starting the Conjur service](https://github.com/conjurinc/appliance/blob/98271d32d776c105f156a6386aedb9c33e072c33/runit/conjur/run#L12-L13). So no additional steps are required by the user when upgrading there.

Thanks!

izgeri:
@micahlee ah! thanks. I wasn't clear that this was for the audit DB, which is not part of OSS.

just fwiw, if we have special upgrade instructions that we need for a given release we can add them to https://github.com/cyberark/conjur/blob/master/UPGRADING.md

jtuttle (approved):

Issue #1987, opened 2020-12-30
Rspec Unit Tests are Present for Existing Audit Events


uCatu:
Unit tests are currently missing for few audit events to verify they render the correct log output for the given inputs.

alexkalish:
@uCatu & @liubovra: Could you please provide a more detailed and user focused description for this story?  It appears to be linked to several PRs, but does not match what the PRs appear to be changing.
(CC: @rrefael, @izgeri)

PR #2013, opened 2021-01-28
Audit fixes

egvili:
### What does this PR do?
1. For change password audit, `user` field under subject changed to `role` to align with other audits. See: [ONYX-6863](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-6863)
2. Fix policy audit bug where `role` field is missing. See [ONYX-6678](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-6678)

Resolves #[1987](https://github.com/cyberark/conjur/issues/1987) and [2014](https://github.com/cyberark/conjur/issues/2014)

#### Change log

#### Test coverage

#### Documentation
- [X] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs in [1360](https://github.com/cyberark/conjur-docs/pull/1360#) or


app/models/audit/event/policy.rb
```diff
@@ -75,7 +75,7 @@ def facility
       private
 
       def user
-        @user || @policy_version.role
+        @user ||= @policy_version.role
```
uCatu, line 78: why? can you explain?

egvili: @uCatu , If I understand correctly `@policy_version.role` is a DB call which might be costly. So, after doing it once I want to save the value into `@user` to save that call next time

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/conjur#1974](https://github.com/cyberark/conjur/issues/1974)
 - Correct unit tests and integration tests for audit, and correct a couple of issues found with them.
   [cyberark/conjur#1987](https://github.com/cyberark/conjur/issues/1987)
+- Aligning field to use ‘role’ under ‘subject'
```
orenbm, line 22: the change here is not clear. is it correct to say that this should be `Audit messages for password changes now have the field 'role' instead of 'user' under 'subject'`?

cucumber/api/features/change_password.feature
```diff
@@ -13,6 +13,16 @@ Feature: Change the password of a role
     Given I set the password for "alice" to "My-Password1"
     When I successfully PUT "/authn/cucumber/password" with username "alice" and password "My-Password1" and plain text body "New-Password1"
     Then I can GET "/authn/cucumber/login" with username "alice" and password "New-Password1"
+    And there is an audit record matching:
+    """
+      <84>1 * * conjur * password
+      [auth@43868 user="cucumber:user:alice"]
```
orenbm, line 19: should this be changed to `role` as well? it can also have a host

egvili: No. Under `auth` section the field name is expected to be user. Only under subject it's expected to be role

orenbm: do we want to change it to `role` for uniformity? it does make more sense. @netacoral wdyt?

netacoral: The goal of this PR is to have a single format, yes. The alignment we want to have is `user` under `auth` structured data and 'role' under `subject` structured data. (@egvili maybe you should add it to the PR description).
So @orenbm to your question, in this case it should be 'user'.

orenbm: i agree that it is out of scope of this PR but i think it is drunk to have `user` in one place and `role` in another where in both places we are referring to a role (user/host). I'm not saying we should do it now, just asking if you think that the correct behaviour is to have `role` in both places. If so, we should create an issue to fix it (unless there's a reason not to)

netacoral: Actually the way I saw it is that 'user' is the identity performing the operation and 'role' is the identity the operation is performed on, so it's ok they're not the same to avoid confusion. But I see your point, and I'm not sure what's the best approach...

orenbm (approved):

PR #2015, opened 2021-01-29
Added extra check for Accept header on batch secrets request

telday:
This check makes sure that the header exists before
attempting to check its value in order to prevent errors

### What does this PR do?

Added an extra check to make sure the `Accept` header is actually included in a request before trying to compare its value to anything. If the header is totally absent Conjur will throw a 500 error. The reason that tests have not found this issue yet is because Ruby's `RestClient::` [implicitly adds](https://github.com/rest-client/rest-client/blob/master/lib/restclient/request.rb#L588) an `Accept: */*` header to all requests if another value is not specified. However other languages do not do this.

After some digging it does not seem like there is an easy way to override the default headers set in `RestClient::` and `Net::` so testing this functionality might be difficult. Due to the impending OSS suite release I think this change should be pulled in and a ticket opened to write a new test for this edge case.

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/cucumber/api/features/support/rest_helpers.rb
+++ b/cucumber/api/features/support/rest_helpers.rb
@@ -65,7 +65,7 @@ module RestHelpers
     headers.each { |key, value| request[key] = value }
 
     if options[:user] && options[:password]
-      request.basic_auth options[:user], options[:password]
+      request.basic_auth(options[:user], options[:password])
     end
 
     response = http.request(request)
@@ -281,17 +281,17 @@ module RestHelpers
 
   def current_user_credentials
     username = @current_user.login
-    token = Slosilo["authn:#{@current_user.account}"].signed_token username
-    user_credentials username, token
+    token = Slosilo["authn:#{@current_user.account}"].signed_token(username)
+    user_credentials(username, token)
   end
 
-  def user_credentials username, token
+  def user_credentials(username, token)
     token_authorization = "Token token=\"#{Base64.strict_encode64 token.to_json}\""
     headers = { authorization: token_authorization }
     { headers: headers, username: username }
   end
 
-  def current_user_basic_auth password = nil
+  def current_user_basic_auth(password = nil)
     password ||= @current_user.api_key
     { user: @current_user.login, password: password }
   end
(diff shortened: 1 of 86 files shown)


cucumber/api/features/support/rest_helpers.rb
```diff
@@ -49,6 +52,30 @@ def get_json(path, options = {})
...
+    uri, headers = URI(args[0] + path), args[1][:headers]
+    http = Net::HTTP.new(uri.host, uri.port)
+
+    request = Net::HTTP::Get.new(uri.request_uri)
+    headers.each do |key, value|
+      request[key] = value
+    end
+
+    if options[:user] && options[:password]
+      request.basic_auth options[:user], options[:password]
```
jonahx, line 68: Add parens for method call.

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -49,6 +52,30 @@ def get_json(path, options = {})
     set_result result
   end
 
+  def get_json_raw(path, options = {})
+    path = denormalize(path)
+    args = get_request_args options
```
jonahx, line 57: Add parens for method call.

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -339,6 +366,12 @@ def rest_resource options
     args.last[:headers] ||= {}
     args.last[:headers].merge(headers) if headers
 
+    args
+  end
+
+  def rest_resource options
+    args = get_request_args options
```
jonahx, line 373: parens

cucumber/api/features/support/rest_helpers.rb
```diff
@@ -49,6 +52,32 @@ def get_json(path, options = {})
     set_result result
   end
 
+  def get_json_no_default_headers(path, options = {})
```
jonahx, line 55: We should a comment above this function explaining the reason we're using `Net::HTTP` instead of `RestClient`.  Something like:

```
# Since there is no way to remove the default Accept header in RestClient
# we use Net::HTTP here.  Otherwise, we wouldn't be able to simulate requests
# that omit the Accept header in our tests.
```

jonahx (approved):

PR #2018, opened 2021-02-01
if host does not exist raise RoleNotFound

mbenita-Cyberark:
### What does this PR do?
 in k8s authenticator, raise RoleNotFound Exception if host does not exist

----

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Conjur now raises a RoleNotFound when trying to authenticate a non-existing host in authn-k8s
+  [conjurinc/appliance#1546]https://github.com/conjurinc/appliance/issues/1546
 
 ## [1.11.2] - 2021-02-02
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 ### Changed
 - Conjur now raises a RoleNotFound when trying to authenticate a non-existing host in authn-k8s
-  [conjurinc/appliance#1546]https://github.com/conjurinc/appliance/issues/1546
+  [cyberark/conjur#2046](https://github.com/cyberark/conjur/issues/2046)
 
 ## [1.11.2] - 2021-02-02
 ### Added
--- a/app/domain/authentication/authn_k8s/inject_client_cert.rb
+++ b/app/domain/authentication/authn_k8s/inject_client_cert.rb
@@ -115,7 +115,13 @@ module Authentication
       end
 
       def host
-        @host ||= @resource_class[host_id]
+        return @host if @host
+
+        @host = @resource_class[host_id]
+
+        raise Errors::Authentication::Security::RoleNotFound, host_id unless @host
+
+        @host
       end
 
       def validate_spiffe_id_exists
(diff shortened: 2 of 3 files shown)


jonahx (dismissed):

izgeri (changes requested):
If we're changing the functional behavior of the server, we should definitely have a CHANGELOG message for this change. You can see our guidelines for writing good changelog messages [here](https://github.com/cyberark/community/blob/master/Conjur/CONTRIBUTING.md#changelog-guidelines). We try to write the changelog messages so they speak to the user-facing impact of the change - basically, try to explain, "If I am the user, why do I care?"

You should probably also try to write a simple test case for this change - something that would have failed before you implemented it, but would be passing now. The PR checklist is a good guide for the things to consider when submitting PRs.

andytinkham:
I'm not sure which file changed to require quality-architects as codeowner here, but my concerns are the same as those @izgeri brought up. I have no further changes to suggest..

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Conjur now Throw RoleNotFound error in k8s authenticator
```
orenbm, line 9: `Conjur now raises a RoleNotFound when trying to authenticate a non-existing host in authn-k8s`

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Conjur now raises a RoleNotFound when trying to authenticate a non-existing host in authn-k8s
```
orenbm, line 9: mnissing link to issue

app/domain/authentication/authn_k8s/inject_client_cert.rb
```diff
@@ -174,13 +180,22 @@ def audit_failure(err)
...
             success:            false,
             error_message:      err.message
           )
         )
       end
+
+      # TODO: this logic is taken from app/models/audit/event/authn.rb.
+      # We should use that logic here.
+      # Masking role if it doesn't exist to avoid audit pollution
+      # Checking @success as well to save DB call on success
```
orenbm, line 194: you actually don't check @success here. Please fix comment. I would write that we're doing this as `host` doesn't exist so it doesn't have an id.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+  [conjurinc/appliance#1546]https://github.com/conjurinc/appliance/issues/1546
```
orenbm, line 10: this should be `[conjurinc/appliance#1546](https://github.com/conjurinc/appliance/issues/1546`)` so it is linkable in a MD file

orenbm: this is not the correct issue. The issue you are fixing is that we now raise a NilError instead of an informative one. please add an issue if it's not present

orenbm (approved):

CHANGELOG.md
izgeri: why are we linking to an issue filed in a private repo?

izgeri: nm - looks like this was updated later, and the changelog is current now.

alexkalish:
This is connected to #2046, yes?   Please be sure to link PRs to issues, as it helps trace changes when building a release.  Thank you!

alexkalish:
@mbenita-Cyberark @orenbm: Does this have user facing behavior change?  How will users be affected?  Thanks.

orenbm:
> @mbenita-Cyberark @orenbm: Does this have user facing behavior change? How will users be affected? Thanks.

the change is written in the CHANGELOG: `Conjur now raises a RoleNotFound when trying to authenticate a non-existing host in authn-k8s`

alexkalish:
@orenbm: Yup, I saw that entry.  However, customers typically interact with authn-k8s through the client, not the API directly.  How would this change manifest in the client?

orenbm:
it won't. they will see it if they have a failure and will check the logs to find it. do instead of seeing an unhelpful nil pointer message they will see the new message.

PR #2022, opened 2021-02-02
Update logic in image publish script

izgeri:
### What does this PR do?
Check if the `registry-path` prefix is set before checking if it's a tag build,
so that on tag-triggered builds we can still publish internal images (that may
be used by tests) early in the pipeline without triggering an image publish
because `TAG_NAME` is set for the whole pipeline.

n/a

#### Change log

#### Test coverage

#### Documentation


micahlee (dismissed):
LGTM, thanks!

izgeri:
@micahlee I missed the window to merge yesterday, can you please approve again? thanks!

micahlee (approved):

PR #2042, opened 2021-02-17
Conjurctl create account with password

### LINKED ISSUE #2043: Set admin password via conjurctl
## Is your feature request related to a problem? Please describe.

I was deploying Conjur in a Fargate cluster. To create the first account I utilized `conjurctl account create <name>`. This provides the api key for the admin user of this account. Currently, I am retrieving the api key from the logs. However, I am looking for a more secure method to retrieve the admin user's api key. 

## Describe the solution you would like

I would like to see a command such as `conjurctl account create` where I can set the password for the admin user. It will create a more user-friendly and secure solution to retrieving/setting the admin user's credentials. 

After running `conjurctl account create <name> <password>` I will be able to use the provided password to authenticate with Conjur. 

## Describe alternatives you have considered

Options I have considered:
1. Output the api key to the logs
    - Not secure
2. [change the password via REST API](https://docs.conjur.org/Latest/en/Content/Developer/Conjur_API_change_Password.htm)
    - A potential solution but not very user-friendly. It complicates the bash that is executed on every instance upon instantiation. However, there is a benefit that the event is audited.
3. Write to a file and SSH into the container
    - I do not want to enable users to SSH into the Fargate instances

jvanderhoof:
### What does this PR do?
- Enable operators to specify an admin user's password:
   - `conjurctl account create --name demo --password-from-stdin` (STDIN)
   - `conjurctl server --account demo --password-from-stdin` (STDIN)

- How to manually test?
  - cd dev
  - dev: docker-compose up -d --no-deps pg conjur client
  - dev: docker-compose exec conjur bundle
  - dev: docker-compose exec conjur conjurctl db migrate
  - exec into the conjur container and run echo -n "testP@SS1alice" | conjurctl server --account demo --password-file -
  - Attempt to authenticate the admin user with the password set above
  - Also test  echo -n "testP@SS1alice" | conjurctl account create demo1 -
  - Attempt to authenticate the admin user with the password set above
  - Repeat the previous steps but write the password to a file and provide the path instead of utilizing STDIN

Resolves #2043 

#### Change log

#### Test coverage

#### Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/conjur#2083](https://github.com/cyberark/conjur/issues/2083)
 
 ### Added
+- `conjurctl server` and `conjurctl account create` allow the operator to specify
+  the admin user's password via STDIN by providing the `--password-from-stdin` switch.
+  [cyberark/conjur#2043](https://github.com/cyberark/conjur/issues/2043)
+- `conjurctl account create` now allows the operator to specify the account name via
+  the `--name` flag. We recommend using this explicit flag when using the `--password-from-stdin`
+  option so that commands are explicit and more readable.
+  [cyberark/conjur#2043](https://github.com/cyberark/conjur/issues/2043)
 - `/whoami` API endpoint now produces audit events.
   [cyberark/conjur#2052](https://github.com/cyberark/conjur/issues/2052)
 - When a user checks permissions of a non-existing role or a non-existing resource, Conjur now audits a failure message.
--- a/bin/conjur-cli.rb
+++ b/bin/conjur-cli.rb
@@ -36,11 +36,21 @@ def connect
   true
 end
 
+def stdin_input
+  raise "Please provide an input via STDIN" if $stdin.tty?
+
+  $stdin.read.force_encoding('ASCII-8BIT')
+end
+
 desc 'Run the application server'
 command :server do |c|
   c.desc 'Account to initialize'
   c.arg_name :name
   c.flag [ :a, :account ]
+
+  c.desc "Provide account password via STDIN"
+  c.arg_name "password-from-stdin"
+  c.switch("password-from-stdin", :negatable => false)
 
   c.desc 'Policy file to load into the server'
   c.arg_name :path
@@ -62,8 +72,19 @@ command :server do |c|
     connect
 
     system "rake db:migrate" or exit $?.exitstatus
+
+    if options["password-from-stdin"] && !account
+      raise "account is required with password-from-stdin flag"
+    end
+
     if account
-      system "rake account:create[#{account}]" or exit $?.exitstatus
+      if options["password-from-stdin"]
+        password = stdin_input
+        system "rake account:create_with_password[#{account},#{password}]" \
+          or exit $?.exitstatus
+      else
+        system "rake account:create[#{account}]" or exit $?.exitstatus
+      end
     end
 
     if file_name = options[:file]
@@ -174,23 +195,44 @@ desc "Manage accounts"
 command :account do |cgrp|
   cgrp.desc "Create an organization account"
   cgrp.long_desc <<-DESC
-Use this command to generate and store a new account, along with its 2048-bit RSA private key,
-used to sign auth tokens, as well as the "admin" user API key.
-The CONJUR_DATA_KEY must be available in the environment
-when this command is called, since it's used to encrypt the token-signing key
-in the database.
+Use this command to generate and store a new account, along with its 2048-bit 
+RSA private key, used to sign auth tokens. The CONJUR_DATA_KEY must be
+available in the environment when this command is called, since it's used to
+encrypt the token-signing key in the database.
+
+The optional 'password-from-stdin' flag signifies that the password should be 
+read from STDIN. If the flag is not provided, the "admin" user API key will be
+outputted to STDOUT.
+
+The 'name' flag or command argument must be present. It will specify the name of
+the account that will be created.
 
 Example:
 
-$ conjurctl account create myorg
+$ conjurctl account create [--password-from-stdin] --name myorg
   DESC
-  cgrp.arg :account
+  cgrp.arg :name, :optional
   cgrp.command :create do |c|
+    c.desc("Provide account password via STDIN")
+    c.arg_name("password-from-stdin")
+    c.switch("password-from-stdin", :negatable => false)
+
+    c.desc("Account name")
+    c.arg_name(:name)
+    c.flag(:name)
+
     c.action do |global_options,options,args|
-      account = args.first
+      account = options[:name] || args.first
+      raise "No account name was provided" unless account
+
       connect
 
-      exec "rake account:create[#{account}]"
+      if options["password-from-stdin"]
+        password = stdin_input
+        exec "rake account:create_with_password[#{account},#{password}]"
+      else
+        exec "rake account:create[#{account}]"
+      end
     end
   end
 
--- a/lib/tasks/account.rake
+++ b/lib/tasks/account.rake
@@ -25,6 +25,29 @@ namespace :"account" do
     end
   end
 
+  desc "Create an account with a preset password"
+  task :create_with_password, [ "account", "password" ] => [ "environment" ] do |t,args|
+    unless Conjur::Password.valid?(args[:password])
+      $stderr.puts(::Errors::Conjur::InsufficientPasswordComplexity.new.to_s)
+      exit 1
+    end
+
+    Account.find_or_create_accounts_resource
+    begin
+      Account.create(args[:account])
+      account = Account.new(args[:account])
+      $stderr.puts "Created new account '#{account.id}'"
+      puts "Token-Signing Public Key: #{account.token_key.to_s}"
+
+      role_id = "#{args[:account]}:user:admin"
+      Role[role_id].password = args[:password]
+      puts "Password is set"
+    rescue Exceptions::RecordExists
+      $stderr.puts "Account '#{args[:account]}' already exists"
+      exit 1
+    end
+  end
+
   desc "Delete an account"
   task :delete, [ "account" ] => [ "environment" ] do |t,args|
     begin
--- a/spec/conjurctl/server_spec.rb
+++ b/spec/conjurctl/server_spec.rb
@@ -1,0 +1,51 @@
+require 'spec_helper'
+require 'open3'
+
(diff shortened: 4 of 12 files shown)


lib/conjur/password_complexity.rb
```diff
@@ -0,0 +1,20 @@
...
+      # 12-128 characters
+    def initialize(password)
+      @plain_password = password
+    end
+
+    def invalid?
+      @plain_password && (@plain_password.index("\n") || @plain_password !~ VALID_PASSWORD_REGEX)
+    end
+  end
+end
```
jonahx, line 20: This is just a `Password` or an `UnvalidatedPassword`.  Then in the code where you use this:

```ruby
unless UnvalidatedPassword.new(raw_pw).valid?
  # add some error
end
```
Alternatively, you just have a function on a module like:
```ruby
module Conjur
  module Password
    def self.valid?
# etc
```

lib/tasks/account.rake
```diff
@@ -25,6 +25,32 @@ namespace :"account" do
...
+    password_invalidity = Conjur::PasswordComplexity.new(args[:password]).invalid?
+    
+    if password_invalidity
+      $stderr.puts(::Errors::Conjur::InsufficientPasswordComplexity.new.to_s)
+      exit 1
+    end
+
+    Account.find_or_create_accounts_resource
+    begin
+      api_key = Account.create args[:account]
```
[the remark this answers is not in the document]
jonahx: Rubocop should be picking up that there are no parens here...

bin/conjur-cli.rb
```diff
@@ -36,12 +36,24 @@ def test_select
   true
 end
 
+def file_input path
```
jonahx, line 39: Whole PR seems to be missing parens around method calls.

bin/conjur-cli.rb
```diff
@@ -63,7 +75,12 @@ def test_select
 
     system "rake db:migrate" or exit $?.exitstatus
     if account
-      system "rake account:create[#{account}]" or exit $?.exitstatus
+      if options["password_file"]
+        password = file_input(options["password_file"])
+        system "rake account:create_with_password[#{account},#{password}]" or exit $?.exitstatus
```
jonahx, line 80: break line.

bin/conjur-cli.rb
```diff
@@ -180,17 +198,32 @@ def test_select
...
 in the database.
 
+The optional 'password_file' command argument is either a file which contains
+the custom password, or the dash '-' character, which signifies
+that the password should be read from stdin.
+
 Example:
 
-$ conjurctl account create myorg
+$ conjurctl account create myorg [password_file]
```
micahlee, line 207: Given it's optional, would it make since for this to be a flag (`--password-file`) instead? That also has the benefit of being consistent with the `server` command.

bin/conjur-cli.rb
```diff
@@ -36,12 +36,24 @@ def test_select
   true
 end
 
+def file_input(path)
+  if path == '-'
+    $stdin.read.force_encoding('ASCII-8BIT')
+  else
+    File.read(path, mode: 'rb')
+  end
+end
```
micahlee, line 45: It looks like there are a number of potential failure cases here. Do these manifest in the messages you would want the user to see?

For example:

- The file doesn't exist
- The file is empty
- The `-` argument is given, but nothing is provided to STDIN

Along these lines, is there any way some of this could be unit tested? Or is there no infrastructure/fixture for that now with this CLI?

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `conjurctl server` allows the operator to specify the admin user's password
+  via a file or STDIN using the optional `--password-file` flag.
+  [cyberark/conjur#2043](https://github.com/cyberark/conjur/issues/2043)
+- `conjurctl account` allows the operator to specify the admin user's password
+  via a file or STDIN using the optional `password_file` command argument.
+  [cyberark/conjur#2043](https://github.com/cyberark/conjur/issues/2043)
```
izgeri, line 14: duplicate?

izgeri:
will this flow have test cases added for it? just wanted to flag that here, since the PR doesn't say WIP / it's not a draft - if we can add tests, we should do that before merging.

also, is there a docs card?

CHANGELOG.md
h-artzi: One is for the `conjurctl server` and the other is for the `conjurctl account` command. However, I will be changing both to use the flag `--password-file` and then I will be able to merge these two together

h-artzi:
> will this flow have test cases added for it? just wanted to flag that here, since the PR doesn't say WIP / it's not a draft - if we can add tests, we should do that before merging.
> 
> also, is there a docs card?

@izgeri I do not believe we currently have tests for the `conjurctl` commands. That doesn't mean they can't be added for this PR.

bin/conjur-cli.rb
h-artzi: If the file doesn't exist the user will see the following error:
`conjurctl server --account demo --password-file /no_such_file
error: No such file or directory @ rb_sysopen - /no_such_file`. 
I can wrap this error if you think it would provide a better UX.

If the `-` argument is given, but nothing is provided to STDIN ex. 
`conjurctl server --account demo --password-file`, it will hang. 
However, a user could still provide input by typing in a password and then Ctrl+D.
I could add a timeout if you think this would improve the UX, but there are pros and cons to that approach as well. 

We currently do not have any unit tests for conjurctl. I will look into adding a cucumber test for `conjurctl account create`

micahlee: Regarding:
>If the - argument is given, but nothing is provided to STDIN ex.
conjurctl server --account demo --password-file, it will hang.
However, a user could still provide input by typing in a password and then Ctrl+D.
I could add a timeout if you think this would improve the UX, but there are pros and cons to that approach as well.

You can actually detect this without a timeout using `$stdin.tty?`
| command | `$stdin.tty?` |
| - | - |
| <code>echo 'password' &#124; conjurctl ... --password-file -</code> | false |
| `conjurctl ... --password-file - < password-file.txt` | false |
| `conjurctl ... --password-file -` | true |

That would give you either an opportunity to interactively prompt for it, or just raise an error if the standard input isn't piped int.

h-artzi: TIL

bin/conjur-cli.rb
izgeri: this is small, but there is a difference between providing a reference to a file and to providing the file contents. should we be overloading the same flag with both use cases? maybe it would be better to have a `password` flag for the password value, and a `password-file` flag for a reference to a file containing the value?

izgeri:
@h-artzi I hoped we had something in the docs, but it looks like the best reference is actually the quick start :(

Maybe we can create a new small section after "Next Steps" [here](https://github.com/cyberark/conjur-quickstart/#next-steps) that explains "Configuring Conjur with a pre-defined admin password"?

I am uncomfortable merging this without tests, or if the scope of adding tests is too big, without and issue filed with a plan to address it that will add the appropriate tests. There are a few cukes that test `conjurctl` functionality - can we add some cukes or a new test file here: https://github.com/cyberark/conjur/tree/master/lib/test - something that gives us assurance we won't inadvertently break this in a future release?

On a totally separate note - is there a default value for `account` if you don't pass one to `conjurctl`? If not, I'd like to consider a follow-up that sets it to `default`, so that our clients can start assuming a sane default for account value.

bin/conjur-cli.rb
h-artzi: I am leaning towards just including a `password` flag (which would only accept input via STDIN). If there is value in adding the flag `password-file` I will be happy to add it in the future.

CHANGELOG.md
```diff
@@ -5,8 +5,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
 ### Added
+- `conjurctl server` and `conjurctl account create` allow the operator to specify the admin user's
+  password via STDIN by providing the `--password-from-stdin` switch.
+  [cyberark/conjur#2043](https://github.com/cyberark/conjur/issues/2043)]
+- `conjurctl account create` now allows the operator to specify the account name via the `--name` flag.
```
izgeri, line 12: previously this was `conjurctl account create {account}` - what is the benefit of adding a flag? we might want to add a note about that here

spec/conjurctl/server_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'spec_helper'
```
izgeri, line 1: 👏🏻💥

CHANGELOG.md
h-artzi: Without adding a `--name` flag to `conjurctl account create`
`conjurctl account create --password-from-stdin <account_name>` is valid
`conjurctl account create <account_name>  --password-from-stdin` is invalid

To ensure the user is able to provide the `--password-from-stdin` and the `account name` independent of order. I decided to add the `name flag`. I will add a comment highlighting this in the CHANGELOG.
Thanks for the feedback!

CHANGELOG.md
```diff
@@ -5,8 +5,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+  We recommend using this explicit flag when using the `--password-from-stdin` option so that commands are explicit and more readable
```
izgeri, line 13: is this really why? this was different than we talked about, though also valuable. we talked about potential issues with the order of the flags / subcommands (ie it could break if you don't use it this way, it's not just for readability)

izgeri: nit: also, this line is long

h-artzi: Options always come before arguments. I thought `conjurctl account create --password-from-stdin --name <account name>` and `conjurctl account create --name <account name> --password-from-stdin` read better than `conjurctl account create --password-from-stdin <account name>`. Open to input!

h-artzi:
Relative follow up tickets:
[add test when no STDIN is provided but --password-from-stdin is present](https://github.com/cyberark/conjur/issues/2081)
[default account when none is provided](https://github.com/cyberark/conjur/issues/2082)

micahlee (dismissed):

micahlee (dismissed):

micahlee (approved):

PR #2059, opened 2021-03-04
Add failure scenarios audit for check permission

Avi-Eyal:
### What does this PR do?
- Added missing audits for failure scenarios on check permission

Resolves ״Some failure scenarios don't produce audit events #1551״

#### Change log

#### Test coverage

#### Documentation
- [X] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or

#### API Changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - `/whoami` API endpoint now produces audit events.
   [cyberark/conjur#2052](https://github.com/cyberark/conjur/issues/2052)
+- When a user checks permissions of a non-existing role or a non-existing resource, Conjur now audits a failure message.
+  [cyberark/conjur#2059](https://github.com/cyberark/conjur/issues/2059)
 
 ## [1.11.4] - 2021-03-09
 
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -59,7 +59,25 @@ class ResourcesController < RestController
     privilege = params[:privilege]
     raise ArgumentError, "privilege" unless privilege
 
-    result = assumed_role.allowed_to?(privilege, resource)
+    begin
+      result = assumed_role.allowed_to?(privilege, resource)
+    rescue
+      result = false
+      resource_id = "#{params[:account]}:#{params[:kind]}:#{params[:identifier]}"
+      role_id = params[:role]
+
+      Audit.logger.log(
+        Audit::Event::Check.new(
+          user: current_user,
+          client_ip: request.ip,
+          resource: resource_id,
+          privilege: privilege,
+          role: role_id,
+          operation: "check",
+          success: result
+        )
+      )
+    end
 
     Audit.logger.log(
       Audit::Event::Check.new(
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -61,36 +61,45 @@ class ResourcesController < RestController
 
     begin
       result = assumed_role.allowed_to?(privilege, resource)
-    rescue
-      result = false
-      resource_id = "#{params[:account]}:#{params[:kind]}:#{params[:identifier]}"
-      role_id = params[:role]
 
-      Audit.logger.log(
-        Audit::Event::Check.new(
-          user: current_user,
-          client_ip: request.ip,
-          resource: resource_id,
-          privilege: privilege,
-          role: role_id,
-          operation: "check",
-          success: result
-        )
-      )
+      audit_success(privilege, result)
+    rescue => e
+      audit_failure(privilege, e)
+
+      raise
     end
+
+    head(result ? :no_content : :not_found)
+  end
+
+  def audit_success(privilege, result)
+    Audit.logger.log(
+      Audit::Event::Check.new(
+        user: current_user,
+        client_ip: request.ip,
+        resource_id: resource.id,
+        privilege: privilege,
+        role_id: assumed_role.id,
+        operation: "check",
+        success: result
+      )
+    )
+  end
+
+  def audit_failure(privilege, err)
+    role_id = params[:role]
 
     Audit.logger.log(
       Audit::Event::Check.new(
         user: current_user,
         client_ip: request.ip,
-        resource: resource,
+        resource_id: resource_id,
         privilege: privilege,
-        role: assumed_role,
+        role_id: role_id,
         operation: "check",
-        success: result
+        success: false,
+        error_message: err.message
       )
     )
-
-    head(result ? :no_content : :not_found)
   end
 end
--- a/app/models/audit/event/check.rb
+++ b/app/models/audit/event/check.rb
@@ -14,11 +14,12 @@ module Audit
       )
         @user = user
         @client_ip = client_ip
-        @resource = resource
         @privilege = privilege
         @role = role
         @operation = operation
         @success = success
+        @resource_id = resource.try(:id) || resource
+        @role_id = role.try(:id) || role
       end
 
       # Note: We want this class to be responsible for providing `progname`.
@@ -40,7 +41,7 @@ module Audit
 
       def message
         "#{@user.id} checked if #{role_text} can #{@privilege} " \
-          "#{@resource.id} (#{success_text})"
+          "#{@resource_id} (#{success_text})"
       end
 
       def message_id
@@ -51,8 +52,8 @@ module Audit
         {
           SDID::AUTH => { user: @user.id },
           SDID::SUBJECT => {
-            resource: @resource.id,
-            role: @role.id,
+            resource: @resource_id,
+            role: @role_id,
             privilege: @privilege
           },
           SDID::CLIENT => { ip: @client_ip }
@@ -84,7 +85,7 @@ module Audit
       end
 
       def role_text
-        @user == @role ? 'they' : @role.id
+        @user == @role ? 'they' : @role_id
       end
 
       def success_text
--- a/app/models/audit/event/check.rb
+++ b/app/models/audit/event/check.rb
@@ -6,20 +6,21 @@ module Audit
       def initialize(
         user:,
         client_ip:,
-        resource:,
+        resource_id:,
         privilege:,
-        role:,
+        role_id:,
         operation:,
-        success:
+        success:,
+        error_message: nil
       )
         @user = user
         @client_ip = client_ip
+        @resource_id = resource_id
         @privilege = privilege
-        @role = role
+        @role_id = role_id
         @operation = operation
         @success = success
-        @resource_id = resource.try(:id) || resource
-        @role_id = role.try(:id) || role
+        @error_message = error_message
       end
 
       # Note: We want this class to be responsible for providing `progname`.
@@ -40,8 +41,8 @@ module Audit
       end
 
       def message
-        "#{@user.id} checked if #{role_text} can #{@privilege} " \
-          "#{@resource_id} (#{success_text})"
+        "#{@user.id} #{result_text} if #{role_text} can #{@privilege} " \
+          "#{@resource_id} #{error_message_text}"
       end
 
       def message_id
@@ -85,13 +86,16 @@ module Audit
       end
 
       def role_text
-        @user == @role ? 'they' : @role_id
+        @user.id == @role_id ? 'they' : @role_id
       end
 
-      def success_text
-        attempted_action.success_text
+      def result_text
+        @success ? "successfully checked" : "failed to check"
       end
 
+      def error_message_text
+        @error_message ? ": #{@error_message}" : ""
+      end
     end
   end
 end
--- a/spec/models/audit/event/check_spec.rb
+++ b/spec/models/audit/event/check_spec.rb
@@ -68,4 +68,32 @@ describe Audit::Event::Check do
 
     it_behaves_like 'structured data includes client IP address'
   end
+
+  context 'when the resource does not exist and a failure occurs' do
+    let(:success) { false }
+    let(:resource) { 'rspec:variable:non_existing_var' }
+
+    it 'produces the expected message' do
+      expect(subject.message).to eq(
+                                   'rspec:user:my_user checked if rspec:host:my_host can execute ' \
+        'rspec:variable:non_existing_var (failure)'
+                                 )
+    end
+
+    it_behaves_like 'structured data includes client IP address'
+  end
+
+  context 'when the role does not exist and a failure occurs' do
+    let(:success) { false }
+    let(:role) { 'rspec:host:my_non_existing_host' }
+
+    it 'produces the expected message' do
+      expect(subject.message).to eq(
+                                   'rspec:user:my_user checked if rspec:host:my_non_existing_host can execute ' \
+        'rspec:variable:my_var (failure)'
+                                 )
+    end
+
+    it_behaves_like 'structured data includes client IP address'
+  end
 end
--- a/spec/models/audit/event/check_spec.rb
+++ b/spec/models/audit/event/check_spec.rb
@@ -3,31 +3,33 @@ require 'spec_helper'
 describe Audit::Event::Check do
 
   let(:user) { double('my-user', id: 'rspec:user:my_user') }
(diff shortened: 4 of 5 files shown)


app/controllers/resources_controller.rb
```diff
@@ -59,7 +59,25 @@ def check_permission
...
     raise ArgumentError, "privilege" unless privilege
 
-    result = assumed_role.allowed_to?(privilege, resource)
+    begin
+      result = assumed_role.allowed_to?(privilege, resource)
+    rescue
+      result = false
+      resource_id = "#{params[:account]}:#{params[:kind]}:#{params[:identifier]}"
+      role_id = params[:role]
+
```
egvili, line 68: Perhaps once setting those parameters above, the already existing `Audit.logger.log` can use them, such that there is no need to call it again?

app/models/audit/event/check.rb
```diff
@@ -14,11 +14,12 @@ def initialize(
...
         @user = user
         @client_ip = client_ip
-        @resource = resource
         @privilege = privilege
         @role = role
         @operation = operation
         @success = success
+        @resource_id = resource.try(:id) || resource
+        @role_id = role.try(:id) || role
       end
```
egvili, line 23: What do `resource_id` and `role_id` hold? only the id or the entire object that includes them?

app/models/audit/event/check.rb
```diff
@@ -14,11 +14,12 @@ def initialize(
       )
         @user = user
         @client_ip = client_ip
-        @resource = resource
```
egvili, line 17: Is `resource` intended to be deleted? because in success scenario it seems to be called

egvili:
Hi @Avi-Eyal ,
It looks pretty good! 
I added some questions and comments above.
Also formal note: the ticket this PR solves is expected to be Git issue and not Jira. You can use https://github.com/cyberark/conjur/issues/1987 instead

app/controllers/resources_controller.rb
Avi-Eyal: `Audit.logger.log` needs to be called twice anyway, because when there is an exception (like in the failure scenarios) there exception should be raised again after writing the audit. The second `Audit.logger.log` is for scenarios of no exceptions when the program should keep running

app/models/audit/event/check.rb
Avi-Eyal: yes, there is a logic below that uses the same value

app/models/audit/event/check.rb
Avi-Eyal: Both of them hold just the Ids. `@role` holds the entire object and of the role, and about the entire object of resource there is no need in this file for anything but the id property

app/controllers/resources_controller.rb
```diff
@@ -59,20 +59,47 @@ def check_permission
...
     raise ArgumentError, "privilege" unless privilege
 
-    result = assumed_role.allowed_to?(privilege, resource)
+    begin
+      result = assumed_role.allowed_to?(privilege, resource)
+
+      write_success_audit(privilege, result)
+    rescue
+      resource_id = "#{params[:account]}:#{params[:kind]}:#{params[:identifier]}"
+      role_id = params[:role]
```
egvili, line 68: Is it possible to insert the above 2 lines into `write_audit_failure`?

app/models/audit/event/check.rb
```diff
@@ -6,19 +6,19 @@ class Check
...
         @privilege = privilege
-        @role = role
         @operation = operation
         @success = success
+        @resource_id = resource_id
+        @role_id = role_id
```
egvili, line 21: Can you please verify that all calls to check.rb pass resource id and role id? Also from unit tests

Avi-Eyal: Done

app/controllers/resources_controller.rb
```diff
@@ -59,20 +59,47 @@ def check_permission
     privilege = params[:privilege]
     raise ArgumentError, "privilege" unless privilege
 
-    result = assumed_role.allowed_to?(privilege, resource)
+    begin
+      result = assumed_role.allowed_to?(privilege, resource)
+
+      write_success_audit(privilege, result)
```
orenbm, line 65: please change this to `audit_success` to match the rest of the places we do this

app/controllers/resources_controller.rb
```diff
@@ -59,20 +59,47 @@ def check_permission
...
+    rescue
+      write_failure_audit(privilege)
```
orenbm, line 67: please change this to `audit_failure` to match the rest of the places we do this

orenbm: don't we want to write the error that occurred? this is what we do in the authentication audit - https://github.com/cyberark/conjur/blob/493fc0379e2b3575ed011875d62b0ec11eda85d4/app/domain/authentication/authenticate.rb#L89

app/controllers/resources_controller.rb
```diff
@@ -59,20 +59,48 @@ def check_permission
...
-        role: assumed_role,
+        role_id: assumed_role.id,
         operation: "check",
         success: result
       )
     )
+  end
 
-    head(result ? :no_content : :not_found)
+  def audit_failure(err, privilege)
```
orenbm, line 89: i think we should flip the argument (i.e `(privilege, err)`) to be consistent with the `audit_success` method

spec/models/audit/event/check_spec.rb
```diff
@@ -3,31 +3,33 @@
 describe Audit::Event::Check do
 
   let(:user) { double('my-user', id: 'rspec:user:my_user') }
-  let(:resource) { double('my-resource', id: 'rspec:variable:my_var') }
+  let(:resource)  {'rspec:variable:my_var'}
```
orenbm, line 6: this should be `resource_id` as it mocks a resource id

spec/models/audit/event/check_spec.rb
```diff
@@ -3,31 +3,33 @@
...
   let(:privilege) { 'execute' }
-  let(:role) { double('my-host', id: 'rspec:host:my_host') }
+  let(:role) { 'rspec:host:my_host' }
```
orenbm, line 8: this should be role_id as it mocks a role id

spec/models/audit/event/check_spec.rb
```diff
@@ -68,4 +70,34 @@
...
+    it 'produces the expected message' do
+      expect(subject.message).to eq(
+                                   'rspec:user:my_user failed to check if rspec:host:my_non_existing_host can execute ' \
+        'rspec:variable:my_var : Forbidden'
+                                 )
+    end
+
+    it_behaves_like 'structured data includes client IP address'
+  end
+end
```
orenbm, line 103: missing new line at EOF

app/controllers/resources_controller.rb
```diff
@@ -59,20 +59,48 @@ def check_permission
...
+  def audit_success(privilege, result)
     Audit.logger.log(
       Audit::Event::Check.new(
         user: current_user,
         client_ip: request.ip,
-        resource: resource,
+        resource_id: resource.id,
         privilege: privilege,
-        role: assumed_role,
+        role_id: assumed_role.id,
```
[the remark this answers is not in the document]
Avi-Eyal: `assumed role` can be nil but the in that case an exception will be raised even before and be rescued and audit a failure

app/controllers/resources_controller.rb
Avi-Eyal: ok, working on it

app/controllers/resources_controller.rb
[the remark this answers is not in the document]
orenbm: did you write a test for this scenario? to verify that an audit message is writtent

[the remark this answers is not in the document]
Avi-Eyal: I didnt because even before my changes it was assumed that `assumed_role` is not nil and no audit was written even if it somehow was nil. but anyway, if you think that I need to write that test I will, I just dont really think its necessary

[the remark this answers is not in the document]
orenbm: @sigalsax once told me that a good boy-scout always leaves the field in a cleaner state than the way he found it :)

will such a test fail if you write it now? do we have an existing bug? if it's simple to fix it, let's do it. if not, please create an issue for this and we can take it in the future.

app/controllers/resources_controller.rb
```diff
@@ -59,20 +59,48 @@ def check_permission
     privilege = params[:privilege]
     raise ArgumentError, "privilege" unless privilege
 
-    result = assumed_role.allowed_to?(privilege, resource)
+    begin
+      result = assumed_role.allowed_to?(privilege, resource)
+
+      audit_success(privilege, result)
+    rescue => e
+      audit_failure(e, privilege)
```
orenbm, line 67: well, this is not right. you should flip the parameters. i am wondering why this was not caught in a test. can you check this?

spec/models/audit/event/check_spec.rb
```diff
@@ -53,9 +55,9 @@
 
     it 'produces the expected message' do
       expect(subject.message).to eq(
-        'rspec:user:my_user checked if rspec:host:my_host can execute ' \
-        'rspec:variable:my_var (failure)'
-      )
+                                   'rspec:user:my_user failed to check if rspec:host:my_host can execute ' \
```
orenbm, line 58: shouldn't this include also the error?

Avi-Eyal: in this test the checked role does not have permission to execute the resource, thats the "failure" of the test so there is no exception message to include

app/controllers/resources_controller.rb
orenbm: i see, so we don't have any UTs for this class. care to open an issue for this?

app/controllers/resources_controller.rb
```diff
@@ -59,20 +59,48 @@ def check_permission
...
+    head(result ? :no_content : :not_found)
+  end
+
+  def audit_success(privilege, result)
     Audit.logger.log(
       Audit::Event::Check.new(
         user: current_user,
         client_ip: request.ip,
-        resource: resource,
+        resource_id: resource.id,
```
uCatu, line 80: So we are checking nil here yes?

app/controllers/resources_controller.rb
```diff
@@ -59,20 +59,48 @@ def check_permission
...
-    head(result ? :no_content : :not_found)
+  def audit_failure(privilege, err)
+    resource_id = "#{params[:account]}:#{params[:kind]}:#{params[:identifier]}"
+    role_id = params[:role]
+
+    Audit.logger.log(
+      Audit::Event::Check.new(
+        user: current_user,
+        client_ip: request.ip,
+        resource_id: resource_id,
```
uCatu, line 97: Maybe i'm missing something, in the value, why in here it's `resource_id` and not `resource.id` (same for role_id)

spec/models/audit/event/check_spec.rb
```diff
@@ -36,9 +38,9 @@
 
     it 'renders to string correctly' do
       expect(subject.to_s).to eq(
-        'rspec:user:my_user checked if rspec:host:my_host can execute ' \
-        'rspec:variable:my_var (success)'
-      )
+                                'rspec:user:my_user successfully checked if rspec:host:my_host can execute ' \
```
uCatu, line 41: Replace spaces where ever you added them :)

uCatu (changes requested):

uCatu (changes requested):
Changes

app/controllers/resources_controller.rb
uCatu: Still don't understand, if the role or resource does not exists why do we audit it like it is exists?

Avi-Eyal: Because we need to audit the checked role/resource so will be able to read the audit and understand what the user tried to do(i think that it gives us the ability to tell if the user just misspelled the role id or trying to do something else). any way thats what was expected from this feature

app/controllers/resources_controller.rb
```diff
@@ -59,20 +59,48 @@ def check_permission
     privilege = params[:privilege]
     raise ArgumentError, "privilege" unless privilege
 
-    result = assumed_role.allowed_to?(privilege, resource)
+    begin
+      result = assumed_role.allowed_to?(privilege, resource)
+
+      audit_success(privilege, result)
+    rescue => e
```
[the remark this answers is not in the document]
Avi-Eyal: Sound right, if i remember correctly @egvili told me to move it inside the rescue

app/controllers/resources_controller.rb
Avi-Eyal: We are not checking nil because even before my changes to the file it was assumed that resource is not nil. I can add a check if you think thats necessary :)

app/controllers/resources_controller.rb
uCatu: @netacoral Please notice that we will have a reliability issue here, as success messages are guaranteed to hold a real/valid resource - while on failure we take what we get from query param...

auditor must be aware to this line.

app/controllers/resources_controller.rb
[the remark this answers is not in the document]
egvili: If an error is thrown during audit_success then the `rescue` will raise it again and the final result will be failure. It is what we want for a failure during the audit, no?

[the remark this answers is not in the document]
Avi-Eyal: What Dvir meant was that a failure in auditing the success will audit a failure about the action(check permission) even though the action (check permission) was successful 
@uCatu right?

app/controllers/resources_controller.rb
uCatu: If resource was nil before this change we would have get:
`resource: nil` => will show nil at audit which is find, and no error is thrown

Now, after this change, we will get:
`resource: nil.id` => throwing error.

Avi-Eyal: It wouldnt just show nil at the audit because `check.new` calls methods of `check.rb` that use the id property of resource and role without any nil check. so it would throw an exception just from a different file

app/controllers/resources_controller.rb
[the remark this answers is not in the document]
egvili: Sorry to bring this back from the dead but the audit is part of the operation (it is also performed in the same transaction), so an exception in the audit_success will result with an error message and all the changes, if made, are reverted. So, it feels more a failure than success, no @uCatu ?
Also, the structure of 'audit_success -> rescue -> audit_failure -> raise' appears many times in the code

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
```
egvili, line 9: Please make sure with @netacoral that the changelog belongs here and not in appliance repo. Because when I added changelog here I had to move it later to appliance. I want to save you the trouble.
Also, if changhlog here is updated then the PR form needs to indicate it

Avi-Eyal: Amos and I just checked with her again and  she said its fine in conjur

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- When a user checks permissions of a non-existing role or a non-existing resource, Conjur now audits a failure message.
```
orenbm, line 10: did @shulifink go over this? i think that we should change it to
```
Conjur now audits a failure message when a user checks permissions of a non-existing role or a non-existing resource
``

Avi-Eyal: Yeah, I consulted with her, and thats what she chose

app/controllers/resources_controller.rb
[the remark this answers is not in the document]
uCatu: @egvili You are partially right and I agree with you to change it back.
Main reason is that current structure we implement is 'audit_success -> rescue -> audit_failure -> raise' and if it is broken in one place we need to keep it aligned.

uCatu (approved):
lgtm

PR #2093, opened 2021-03-30
Update Rails to fix missing mimemagic gem

jtuttle:


micahlee (commented):
Hey @jtuttle , it looks like this upgrades a lot more than just Rails again. Is that intentional?

PR #2103, opened 2021-04-07
Publish edge tag to local registry as we do on dockerhub

uCatu:
### What does this PR do?
For security, performance & early debugging reasons we want `edge` tag to be present and stored at out local Docker registry as it is at DockerHub.
See [Slack Thread](https://conjurhq.slack.com/archives/C8BNMU0KV/p1617353627026700) for additional details

N/A

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -534,9 +534,17 @@ pipeline {
 
         stage('On a master build') {
           when { branch "master" }
-
           steps {
-            sh './push-image.sh --edge'
+            script {
+              def tasks = [:]
+              tasks["Publish edge to local registry"] = {
+                sh './push-image.sh --edge --registry-prefix=registry.tld'
+              }
+              tasks["Publish edge to DockerHub"] = {
+                sh './push-image.sh --edge'
+              }
+              parallel tasks
+            }
           }
         }
       }


Jenkinsfile
```diff
@@ -534,7 +534,15 @@ pipeline {
...
+            sh './push-image.sh --edge --registry-prefix=registry.tld'
+          }
+        }
+        
+        stage('On a master build') {
+          when { branch "master" }
+          // Parallel publish edge tag, latest green main builds, to DockerHub
           steps {
             sh './push-image.sh --edge'
           }
```
orenbm, line 548: this seems odd. can we have an additional stage inside `On a master build` that will run the tasks in parallel instead? something like this: https://github.com/cyberark/secrets-provider-for-k8s/blob/master/Jenkinsfile#L113

uCatu: @orenbm As I follow existing parallel stage approach in this Jenkisfile I suggest to raise the option for converting it to `tasks` in a follow-up PR that will align all Jenkisfiles if chosen to be our best practice guideline, WDYT?

orenbm: yes but in this case you have 2 things that you want to do in the same scenario - on a master build.
I would at least change the stage names so they are not the same.

orenbm: i still think that we should run these tasks in parallel inside the stage. you can also you the current mechanism to do that by adding an additional `parallel` inside the stage.

Jenkinsfile
```diff
@@ -534,9 +534,18 @@ pipeline {
...
+             script {
+               def tasks = [:]
+               tasks["Publish edge to local registry"] = {
+                 sh './push-image.sh --edge --registry-prefix=registry.tld'
+               }
+               tasks["Publish edge to DockerHub"] = {
+                 sh './push-image.sh --edge'
+               }
+               parallel tasks
+             }
```
orenbm, line 548: please indent this one space backwards

Jenkinsfile
```diff
@@ -534,9 +534,18 @@ pipeline {
 
         stage('On a master build') {
           when { branch "master" }
-
+          // Parallel publish edge tag image
```
orenbm, line 537: I think
```suggestion
          // Parallel publish edge tag image
```
I think that this comment is redundant but if you do write it please change it to
```
// Publish edge tag image to private and public registries in parallel
```

orenbm (changes requested):
left some comments

orenbm (approved):

PR #2104, opened 2021-04-07
Add audit for fetch secret failure scenarios

Avi-Eyal:
### What does this PR do?
-Added missing audit for failure scenarios on fetch secret

Resolves ״Some failure scenarios don't produce audit events #1551(in appliance)״

#### Change log

#### Test coverage

#### Documentation

#### API Changes


app/controllers/secrets_controller.rb
```diff
@@ -92,12 +92,10 @@ def get_secret_from_variable(variable)
...
   end
 
-  def audit_fetch resource, version: nil
-    # don't audit the fetch if the resource doesn't exist
-    return unless resource
-
+  def audit_fetch (resource, version: nil)
     fetch_info = error_info.merge(
       resource: resource,
+      resource_id: resource ? nil : resource_id,
```
egvili, line 98: Please consider using only `resource_id` without `resource`. 
If `resource` exists, take its id. otherwise, use `resource_id`

app/controllers/secrets_controller.rb
```diff
@@ -110,14 +108,15 @@ def audit_fetch resource, version: nil
...
+    resource_visible = resource_visible?
     return { success: true } unless $ERROR_INFO
 
     # If resource is not visible, the error info will say it cannot be found.
     # That is still what we want to report to the client, but in the log we
     # want more accurate 'Forbidden'.
     {
       success: false,
-      error_message: (resource_visible? ? $ERROR_INFO.message : 'Forbidden')
+      error_message: (resource_visible || resource_visible.nil? ? $ERROR_INFO.message : 'Forbidden')
```
egvili, line 119: I don't understand what `resource_visible` checks that `resource_visible?` doesn't check.
Also, please pay attention to CodeClimate note here

app/models/audit/event/fetch.rb
```diff
@@ -77,12 +79,12 @@ def facility
       private
 
       def resource_description
-        resource_id = @resource.id
+        resource_id = @resource_id ||= @resource.id
```
egvili, line 82: If you had only `resource_id` this line would have been redundant.

egvili (commented):
Looks good!
Left some comments

app/controllers/secrets_controller.rb
```diff
@@ -92,12 +92,10 @@ def get_secret_from_variable(variable)
     end
   end
 
-  def audit_fetch resource, version: nil
-    # don't audit the fetch if the resource doesn't exist
-    return unless resource
-
+  def audit_fetch (resource, version: nil)
```
orenbm, line 95: remove space after `audit_fetch`

app/controllers/secrets_controller.rb
orenbm: i think that you can remove line 111 (`resource_visible = resource_visible?`) and just use here `!resource_visible?` instead of `resource_visible.nil?`

app/controllers/secrets_controller.rb
Avi-Eyal: I tried to do this exact  thing but for some reason when resource existed and resource.id was sent to fetch.rb it was empty... I dont really know why it happened but resource.id had a value only when the whole resource object was sent and the id property was accessed in fetch.rb. I couldnt find any other way around it except having both resource and resource_id sent.   @orenbm I would like to hear what you think about that too

app/models/audit/event/fetch.rb
```diff
@@ -15,6 +16,7 @@ def initialize(
         @user = user
         @client_ip = client_ip
         @resource = resource
+        @resource_id = resource_id
```
jonahx, line 19: Could you add a note to explain why we're sometimes getting the `resource_id` directly from a variable of that name, and sometimes from `@resource.id`?  Depending on the answer to that, I would consider only allowing `resource_id`, and letting clients of this class who hold a `resource` be responsible for passing `resource.id`.

Separately, given the current design, I would create a _method_ `resource_id` which has looks like:
```
@resource_id ||= @resource_id_ || @resource.id
```
or similar.

app/controllers/secrets_controller.rb
```diff
@@ -117,7 +115,7 @@ def error_info
     # want more accurate 'Forbidden'.
     {
       success: false,
-      error_message: (resource_visible? ? $ERROR_INFO.message : 'Forbidden')
+      error_message: (!resource_visible? ? $ERROR_INFO.message : 'Forbidden')
```
h-artzi, line 118: Why not write this as `resource_visible? ? 'Forbidden' :  $ERROR_INFO.message` ? In addition, I believe the logic here is correct but is confusing given the comment written above.

Avi-Eyal: Fixed on latest change

app/models/audit/event/fetch.rb
Avi-Eyal: The use ```resource_id``` is for specific case when the resource does not exist. In that case calling ```resource.id``` will raise an error. We need that resource's id for audits and logs purposes. 
 
And about that Second note, it sounds right, I'll check if that breaks anything (even though it shouldn't)

jonahx: > We need that resource's id for audits and logs purposes.

If it doesn't exist, how are you populating `resource_id`?

jonahx: Ignore last question, I see what you are doing after looking at the rest of the code.

The real problem here is sort of legacy one: mixing the Command object and code from the rails "concern".  It would be cleaner if the database lookup were being done by the Command object -- it would also make the whole thing more testable too.

If you don't want to make that change (it's not huge, but not trivial either), I would suggest that you rename `resource_id` to `requested_resource_id` and add a note that `@resource` may be nil sometimes, but `requested_resource_id` will never be.

egvili: After the last commit there is only use of resource_id, which seems to be by the guidelines you suggest

orenbm (dismissed):
Great work!

 thanks for improving the code readability!

app/models/audit/event/fetch.rb
jonahx: Much better.  Thanks!

egvili (approved):

PR #2119, opened 2021-04-20
Trivy scans fail only on Criticals on branches

h-artzi:
Signed-off-by: Andy Tinkham <andy.tinkham@cyberark.com>

### What does this PR do?
- Trivy scan of base images will fail master builds for any High/Critical issues
- Trivy scan of base images will fail branch builds for Critical issues. 

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -99,28 +99,79 @@ pipeline {
           }
         }
 
-        stage('Scan Docker Image') {
+        stage('Scan Docker Images') {
           parallel {
-            stage("Scan Docker Image for fixable issues") {
+            stage('Scan Docker image for High/Critical fixable issues (master/release build)') {
+              when {
+                anyOf {
+                  // If we're running a master or a release tag, fail the build on
+                  // High or Critical severity fixable trivy issues 
+                  branch "master"
+                  tag "v*"
+                }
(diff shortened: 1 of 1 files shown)


andytinkham (changes requested):
Still in discussion

andytinkham:
> I don't love that we're copying the stages just to accomplish `if`/`else`.
> 
> Did you consider whether we could instead make the build failure level a variable? For example, instead of:
> 
> ```
> scanAndReport("conjur:${tagWithSHA()}", "CRITICAL", false)
> ```
> 
> this would be:
> 
> ```
> scanAndReport("conjur:${tagWithSHA()}", "$TRIVY_FAILURE_SEVERITY", false)
> ```
> 
> It's not immediately clear to me how to set this variable based on the `branch` and `tag` built-in conditions. So I'm approving this for now to unblock you. I do consider this tech debt that should be re-visited though.

I had the same question & haven't found an answer yet. Still discussing with Matt & will see if I can remove the duplication.

andytinkham:
Will need to revisit this - lost the thread of discussions, so will revisit when we address pulling scanning into a nightly job later this year.

PR #2141, opened 2021-05-03
Implement fetch mandatory claims to validate logic

nessiLahav:
### What does this PR do?
Implement fetch mandatory claims logic according to the design section [here](https://github.com/cyberark/conjur/blob/master/design/authenticators/authn_jwt/authn_jwt_solution_design.md#fetch-jwt-standard-claims-to-validate) 

#### Change log

#### Test coverage

#### Documentation

#### API Changes


app/domain/logs.rb
```diff
@@ -290,6 +290,26 @@ module AuthnJwt
         code: "CONJ00056D"
       )
 
+      FetchingJwtMandatoryClaimsToValidate = ::Util::TrackableLogMessageClass.new(
```
nessiLahav, line 293: @shulifink please review those new logs messages

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+
+          if @authentication_parameters.decoded_token[optional_claim]
+            add_to_mandatory_claims_list(optional_claim)
+          end
+        end
+      end
+
+      def add_to_mandatory_claims_list(claim)
+        @logger.debug(LogMessages::Authentication::AuthnJwt::AddingJwtMandatoryClaimToValidate.new(claim))
+
```
tzheleznyak, line 45: why this empty line?

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+      # fetch_mandatory_claims_to_validate function is responsible to return a list of mandatory JWT standard claims to
+      # validate, according to the following logic:
+      # For each optional claims (iss , exp,nbf, iat) which exists in the token, add to mandatory list
+      # Note: mandatory list also contains the value to validate if necessary (for example iss: cyberark.com)
+      def fetch_mandatory_claims_to_validate
+        @mandatory_claims_to_validate_list ||= []
+
+        OPTIONAL_CLAIMS.each do |optional_claim|
+          @logger.debug(LogMessages::Authentication::AuthnJwt::ValidatingJwtOptionalClaimToValidate.new(optional_claim))
+
```
tzheleznyak, line 36: why this empty line?

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+      private
+
+      # fetch_mandatory_claims_to_validate function is responsible to return a list of mandatory JWT standard claims to
+      # validate, according to the following logic:
+      # For each optional claims (iss , exp,nbf, iat) which exists in the token, add to mandatory list
+      # Note: mandatory list also contains the value to validate if necessary (for example iss: cyberark.com)
+      def fetch_mandatory_claims_to_validate
+        @mandatory_claims_to_validate_list ||= []
+
```
tzheleznyak, line 33: why this empty line?

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+        logger: Rails.logger
+      },
+      inputs: %i[authentication_parameters]
+    ) do
+
+      def call
+        @logger.debug(LogMessages::Authentication::AuthnJwt::FetchingJwtMandatoryClaimsToValidate.new)
+        fetch_mandatory_claims_to_validate
+        @logger.debug(LogMessages::Authentication::AuthnJwt::FetchedJwtMandatoryClaimsToValidate.new)
+
```
tzheleznyak, line 21: why this empty line?

app/domain/authentication/authn_jwt/jwt_mandatory_claim.rb
```diff
@@ -0,0 +1,15 @@
+module Authentication
+  module AuthnJwt
+
+    # This class instance holds a JWT standard claim which is
+    # mandatory to validate.
+    class JwtMandatoryClaim
+      attr_reader :name, :value
+
+      def initialize(name:, value:)
```
tzheleznyak, line 9: what the : does in the end of each variable?

tzheleznyak:
Can you add link to relevant buisness logic in the design document to the PR?

spec/app/domain/authentication/authn-jwt/fetch_mandatory_claims_to_validate_spec.rb
```diff
@@ -0,0 +1,345 @@
...
+    mandatory_claims_to_validate_list = []
+    optional_claims.each do |optional_claim|
+      mandatory_claims_to_validate_list.push(::Authentication::AuthnJwt::JwtMandatoryClaim.new(name: optional_claim, value: claim_value(optional_claim)))
+    end
+
+    mandatory_claims_to_validate_list
+  end
+
+  def claim_value(optional_claim)
+    if optional_claim == 'iss'
```
tzheleznyak, line 36: can you document what happens here? why its iss?

app/domain/authentication/authn_jwt/jwt_mandatory_claim.rb
nessiLahav: look at the usage of the class, it defined the name of the variables when you initialize the class

spec/app/domain/authentication/authn-jwt/fetch_mandatory_claims_to_validate_spec.rb
nessiLahav: It is explained in the original class, do you think we need an extra explanation in the UT ?

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+module Authentication
+  module AuthnJwt
+    OPTIONAL_CLAIMS = [ISS_CLAIM_NAME, EXP_CLAIM_NAME, NBF_CLAIM_NAME, IAT_CLAIM_NAME].freeze
```
orenbm, line 5: should this const go in the class itself or is it in the module on purpose? if it's in the module can it go in the consts class?

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+        @logger.debug(LogMessages::Authentication::AuthnJwt::FetchedJwtMandatoryClaimsToValidate.new)
+
+        @mandatory_claims_to_validate_list
+      end
+
+      private
+
+      # fetch_mandatory_claims_to_validate function is responsible to return a list of mandatory JWT standard claims to
+      # validate, according to the following logic:
+      # For each optional claims (iss , exp,nbf, iat) which exists in the token, add to mandatory list
```
orenbm, line 29:
```suggestion
      # For each optional claim (iss , exp,nbf, iat) that exists in the token - add to mandatory list
```

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+      end
+
+      def add_to_mandatory_claims_list(claim)
+        @logger.debug(LogMessages::Authentication::AuthnJwt::AddingJwtMandatoryClaimToValidate.new(claim))
+
+        @mandatory_claims_to_validate_list.push(::Authentication::AuthnJwt::JwtMandatoryClaim.new(name: claim, value: claim_value_to_validate(claim)))
+      end
+
+      def claim_value_to_validate(claim)
```
orenbm, line 49:
```suggestion
      def claim_value(claim_name)
```

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+        if claim == ISS_CLAIM_NAME
+          return @fetch_issuer_value.(@authentication_parameters)
+        end
+
+        nil
```
orenbm, line 54: what's this? are we planning to have other claims in the future? if so i would change the above to `switch-case` and put `nil` in the `else` (with a comment saying what's going on)

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+          if @authentication_parameters.decoded_token[optional_claim]
+            add_to_mandatory_claims_list(optional_claim)
+          end
+        end
+      end
+
+      def add_to_mandatory_claims_list(claim)
+        @logger.debug(LogMessages::Authentication::AuthnJwt::AddingJwtMandatoryClaimToValidate.new(claim))
+
+        @mandatory_claims_to_validate_list.push(::Authentication::AuthnJwt::JwtMandatoryClaim.new(name: claim, value: claim_value_to_validate(claim)))
```
orenbm, line 46:
```suggestion
        @mandatory_claims_to_validate_list.push(
  ::Authentication::AuthnJwt::JwtMandatoryClaim.new(
    name: claim, 
    value: claim_value_to_validate(claim)
  )
)
```

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+      },
+      inputs: %i[authentication_parameters]
+    ) do
+
+      def call
+        @logger.debug(LogMessages::Authentication::AuthnJwt::FetchingJwtMandatoryClaimsToValidate.new)
+        fetch_mandatory_claims_to_validate
+        @logger.debug(LogMessages::Authentication::AuthnJwt::FetchedJwtMandatoryClaimsToValidate.new)
+
+        @mandatory_claims_to_validate_list
```
orenbm, line 22: i think that `list` is redundant. you can change it to `@mandatory_claims_to_validate`

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+
+      private
+
+      # fetch_mandatory_claims_to_validate function is responsible to return a list of mandatory JWT standard claims to
+      # validate, according to the following logic:
+      # For each optional claims (iss , exp,nbf, iat) which exists in the token, add to mandatory list
+      # Note: mandatory list also contains the value to validate if necessary (for example iss: cyberark.com)
+      def fetch_mandatory_claims_to_validate
+        @mandatory_claims_to_validate_list ||= []
```
orenbm, line 32: how about adding a method:
```
def mandatory_claims_to_validate
  @mandatory_claims_to_validate ||= []
end
```

then you can remove this line and just call `mandatory_claims_to_validate` everywhere else in the file. it fits our style better

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,58 @@
...
+        @mandatory_claims_to_validate_list
+      end
+
+      private
+
+      # fetch_mandatory_claims_to_validate function is responsible to return a list of mandatory JWT standard claims to
+      # validate, according to the following logic:
+      # For each optional claims (iss , exp,nbf, iat) which exists in the token, add to mandatory list
+      # Note: mandatory list also contains the value to validate if necessary (for example iss: cyberark.com)
+      def fetch_mandatory_claims_to_validate
```
orenbm, line 31: i don't quite understand the logic here. the function name is `fetch_mandatory_claims_to_validate` and then you iterate over the `OPTIONAL_CLAIMS` list and fetch them. so are you fetching mandatory or optional claims?

app/domain/authentication/authn_jwt/jwt_mandatory_claim.rb
orenbm: @tzheleznyak when you have more than one parameter it is best to add `:` at the end of the params. this way caller objects are more clear as to which parameter gets which value and the order doesn't matter. it also forces the caller to set values to the params.

app/domain/logs.rb
```diff
@@ -290,6 +290,26 @@ module AuthnJwt
...
+        msg: "Fetching JWT mandatory claims to validate",
```
orenbm, line 294: do we want to say from where?

nessiLahav: We actually decide here which mandatory claims to validate base on hardcoded logic + the token content (in the future the logic may be configured in the policy)

Do you think we should change this message ?

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
nessiLahav: the following claims: (iss , exp, nbf, iat) are optional in JWT token according to the RFC
The logic is :
1. If they exist in the Token validate them -> mandatory to validate 
2. If they no exist in the token ignore them -> will not appear in the mandatory list 


Note: the `exp` claim may be changed to be mandatory and not optional, so if it not exists in the token we will failed the request

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
nessiLahav: I prefer this const here because it related only to this class logic,
and when we will have another "fetch_mandatory_claims" logic implementation we will have another `OPTIONAL_CLAIMS` value

orenbm: can you please move it then into the class? this way it is part of the module and if you create another one like this in a different class you will see some unexpected behaviour as the other class may use this const.

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,68 @@
...
+            add_to_mandatory_claims_list(optional_claim)
+          end
+        end
+      end
+
+      def add_to_mandatory_claims_list(claim)
+        @logger.debug(LogMessages::Authentication::AuthnJwt::AddingJwtMandatoryClaimToValidate.new(claim))
+
+        mandatory_claims_to_validate.push(
+          ::Authentication::AuthnJwt::JwtMandatoryClaim.new(
```
orenbm, line 45: should `::Authentication::AuthnJwt::JwtMandatoryClaim` be a dependency? won't we want to mock it?

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,68 @@
...
+
+        mandatory_claims_to_validate
+      end
+
+      private
+
+      # fetch_mandatory_claims_to_validate function is responsible to return a list of mandatory JWT standard claims to
+      # validate, according to the following logic:
+      # For each optional claim (iss, exp, nbf, iat) that exists in the token - add to mandatory list
+      # Note: mandatory list also contains the value to validate if necessary (for example iss: cyberark.com)
```
orenbm, line 30: can you move this comment to the top of the class? it seems that this function does all the logic so its logic is the same as the class.

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
```diff
@@ -0,0 +1,68 @@
...
+      end
+
+      def mandatory_claims_to_validate
+        @mandatory_claims_to_validate ||= []
+      end
+
+      def claim_value(claim)
+        case claim
+        when ISS_CLAIM_NAME
+          return @fetch_issuer_value.(@authentication_parameters)
```
orenbm, line 59: what does `@authentication_parameters` include other than the token? does `@fetch_issuer_value` use any other parameter? can we pass it only the token? 

this way the current class can get only the token and it will be clearer that we fetch (extract?) the claim from the token.

app/domain/authentication/authn_jwt/jwt_mandatory_claim.rb
```diff
@@ -0,0 +1,15 @@
+module Authentication
+  module AuthnJwt
+
+    # This class instance holds a JWT standard claim which is
+    # mandatory to validate.
+    class JwtMandatoryClaim
```
orenbm, line 6: do we really need the `Mandatory`? will we have a `JwtOptionalClaim` class?

app/domain/logs.rb
orenbm: i think it will enhance our supportability if the log would say where the claims are taken from, especially if it can be from several places.

app/domain/authentication/authn_jwt/jwt_mandatory_claim.rb
nessiLahav: I dont think so, but for me its more clear, do you prefer JwtClaim?

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
nessiLahav: We fetching it from the policy not from the token, and i am using the following parameters:
```
      def_delegators(:@authenticator_input, :service_id, :authenticator_name,
                     :account)
 
```

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
nessiLahav: correct

app/domain/logs.rb
nessiLahav: Since i am printing later which claims are mandatory to validate, i don want to print here what is the logic of how i decide that. 

The message will look something like this:

```
Fetching JWT mandatory claims to validate, according to the following logic:
If `iss, exp, nbf, iat` are exist in the token they are mandatory
If not, we ignore them
```

@hilagross @sashaCher @orenbm Ignore the english, do we want to print this king of message ?

app/domain/authentication/authn_jwt/jwt_mandatory_claim.rb
orenbm: yes. not a blocker though

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
orenbm: wait is `@authentication_parameters` actually `@authenticator_input`? if so - why did you change the name?

app/domain/logs.rb
hilagross: I'm not sure we should print such massage, 
first what is mandatory claims? there is no mandatory claim in JWT RFC, each application can decide what to send in its JWT token.
IMO we should print only "Fetching JWT claims to validate" and drop the mandatory.
I think this should be given as info in the documentation.

orenbm: no this message is not king at all. we shouldn't log the logic. however, we can log the `Fetching JWT mandatory claims to validate` message after we perform the logic and then say something like `Fetching JWT mandatory claims to validate for the token` (if i understand correctly)

orenbm: Oh I like @hilagross 's approach! 
Let's drop the word `mandatory` from all the classes, it makes much more sense now. Class `FetchJwtClaimsToValidate` that goes over the optional claims and decides which ones should be validated is much clearer than `FetchJwtMandatoryClaimsToValidate`

sashaCher: Agree with @hilagross .

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
nessiLahav: No, you will see it in another PR the authentication_parameters contains the authenticator_input and more parameters like the decoded token

app/domain/authentication/authn_jwt/jwt_mandatory_claim.rb
nessiLahav: Change it due to the discussion above about the Mandatory messages

app/domain/authentication/authn_jwt/fetch_mandatory_claims_to_validate.rb
orenbm: ok thanks for clarifying

app/domain/authentication/authn_jwt/fetch_jwt_claims_to_validate.rb
```diff
@@ -0,0 +1,68 @@
...
+    # validate, according to the following logic:
+    # For each optional claim (iss, exp, nbf, iat) that exists in the token - add to mandatory list
+    # Note: the list also contains the value to validate if necessary (for example iss: cyberark.com)
+    FetchJwtClaimsToValidate ||= CommandClass.new(
+      dependencies: {
+        fetch_issuer_value: ::Authentication::AuthnJwt::FetchIssuerValue.new,
+        jwt_claim_class: ::Authentication::AuthnJwt::JwtClaim,
+        logger: Rails.logger
+      },
+      inputs: %i[authentication_parameters]
```
orenbm, line 16: just an idea - can we use the different inputs instead of putting them in the `authentication_parameters` object? one of the advantages of the command class pattern is that it is very clear what the inputs to the class are but in this case i have no idea which parameters are in `authentication_parameters`. 
wdyt?

app/domain/authentication/authn_jwt/fetch_jwt_claims_to_validate.rb
```diff
@@ -0,0 +1,68 @@
...
+      def add_to_jwt_claims_list(claim)
+        @logger.debug(LogMessages::Authentication::AuthnJwt::AddingJwtClaimToValidate.new(claim))
+
+        jwt_claims_to_validate.push(
+        @jwt_claim_class.new(
+            name: claim,
+            value: claim_value(claim)
+          )
+        )
+      end
```
orenbm, line 50: fix indentation:
```suggestion
        jwt_claims_to_validate.push(
          @jwt_claim_class.new(
            name: claim,
            value: claim_value(claim)
          )
        )
      end
```

app/domain/authentication/authn_jwt/jwt_claim.rb
```diff
@@ -0,0 +1,15 @@
+module Authentication
+  module AuthnJwt
+
+    # This class instance holds a JWT standard claim which is
+    # mandatory to validate.
```
orenbm, line 5:
```suggestion
    # This class instance holds a JWT standard claim
```

the class represents a jwt claim. you are using it as mandatory in another place but that's something else.

app/domain/logs.rb
```diff
@@ -339,6 +339,27 @@ module AuthnJwt
...
+        code: "CONJ00068D"
+      )
+
+      AddingJwtClaimToValidate = ::Util::TrackableLogMessageClass.new(
+        msg: "Adding JWT claim '{0-claim-name}' to validate",
+        code: "CONJ00069D"
+      )
+
+      ValidatingJwtClaimToValidate = ::Util::TrackableLogMessageClass.new(
+        msg: "Validating if JWT claim '{0-claim-name}' is mandatory to validate",
```
orenbm, line 359: Checking if?
 i think that validating is not correct here.

app/domain/logs.rb
```diff
@@ -339,6 +339,27 @@ module AuthnJwt
...
+        code: "CONJ00067D"
+      )
+
+      FetchedJwtClaimsToValidate = ::Util::TrackableLogMessageClass.new(
+        msg: "Fetched JWT claims to validate",
+        code: "CONJ00068D"
+      )
+
+      AddingJwtClaimToValidate = ::Util::TrackableLogMessageClass.new(
+        msg: "Adding JWT claim '{0-claim-name}' to validate",
```
orenbm, line 354: @shulifink wdyt about this message? the english seems odd.

orenbm (changes requested):
left some comments

app/domain/logs.rb
nessiLahav: Correct i changed it

app/domain/authentication/authn_jwt/fetch_jwt_claims_to_validate.rb
nessiLahav: Because we build a generic authenticator so each implementation may need a different set of parameters for example:
GeUserIdentityFromUrl - needs only the url parameters
GetUserFromToken - needs account service id (to fetch the claim from policy) and the decoded token

So we need 1 container (authentication_parameters) in order to send the same input to both implementations 

Lets raise this challenge again in the final PR to see if we can change this idea

orenbm (approved):

app/domain/logs.rb
```diff
@@ -339,6 +339,27 @@ module AuthnJwt
...
+        code: "CONJ00068D"
+      )
+
+      AddingJwtClaimToValidate = ::Util::TrackableLogMessageClass.new(
+        msg: "Adding JWT claim '{0-claim-name}' to list of mandatory claims to be validated",
+        code: "CONJ00069D"
+      )
+
+      CheckingJwtClaimToValidate = ::Util::TrackableLogMessageClass.new(
+        msg: "Checking if JWT claim '{0-claim-name}' is mandatory to validate",
```
shulifink, line 359: Checking if JWT claim, '{0-claim-name}', must be validated...

app/domain/logs.rb
```diff
@@ -339,6 +339,27 @@ module AuthnJwt
...
+        code: "CONJ00067D"
+      )
+
+      FetchedJwtClaimsToValidate = ::Util::TrackableLogMessageClass.new(
+        msg: "Fetched JWT claims to validate",
+        code: "CONJ00068D"
+      )
+
+      AddingJwtClaimToValidate = ::Util::TrackableLogMessageClass.new(
+        msg: "Adding JWT claim '{0-claim-name}' to list of mandatory claims to be validated",
```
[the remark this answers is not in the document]
tzheleznyak: Done

app/domain/logs.rb
tzheleznyak: Done

PR #2148, opened 2021-05-06
Validate resource restriction with jwt decoded token one to one

tzheleznyak:
Validates single resource restriction in one to one way with the decoded token

### DIFF
--- a/app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
+++ b/app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
@@ -1,0 +1,19 @@
+module Authentication
+  module AuthnJwt
+
+    # This class is responsible for retrieving the correct value from the JWT token
+    # of the requested attribute.
+    class ValidateRestrictionsOneToOne
+      def initialize(decoded_token:)
+        @decoded_token = decoded_token
+      end
+
+      def valid_restriction?(restriction)
+        unless @decoded_token.key?(restriction.name.to_sym)
+          raise Errors::Authentication::AuthnJwt::JwtTokenClaimIsMissing.new(restriction.name)
+        end
+        @decoded_token[restriction.name.to_sym] == restriction.value
+      end
+    end
+  end
+end
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -370,27 +370,27 @@ module Errors
       InvalidIssuerConfiguration = ::Util::TrackableErrorClass.new(
         msg: "Issuer authenticator configuration is invalid. You should configured as authenticator variables: " \
               "'{0-resource-name}' or one of the following: '{1-resource-name}','{2-resource-name}'",
-        code: "CONJ00070E"
+        code: "CONJ00170E"
       )
 
       FailedToParseHostnameFromUri = ::Util::TrackableErrorClass.new(
         msg: "Failed to extract hostname from URI '{0}'",
-        code: "CONJ00071E"
+        code: "CONJ00171E"
       )
 
       InvalidUriFormat = ::Util::TrackableErrorClass.new(
         msg: "Failed to parse URI '{0}'. Reason: '{1}'",
-        code: "CONJ00072E"
+        code: "CONJ00172E"
       )
 
       NoSuchFieldInToken = ::Util::TrackableErrorClass.new(
         msg: "No such field '{0}' in the token",
-        code: "CONJ00073E"
+        code: "CONJ00173E"
       )
 
       NoUsernameInTheURL = ::Util::TrackableErrorClass.new(
         msg: "No username in the URL",
-        code: "CONJ00074E"
+        code: "CONJ00174E"
       )
 
       NoRelevantIdentityProvider = ::Util::TrackableErrorClass.new(
@@ -399,7 +399,14 @@ module Errors
               "1. Send the identity in the URL" \
               "2. Please add `token-app-property` variable and populate it" \
               "   with field in the decoded token that contains the identity",
-        code: "CONJ00075E"
+        code: "CONJ00175E"
+      )
+
+      JwtTokenClaimIsMissing = ::Util::TrackableErrorClass.new(
+        msg: "Claim '{0-attribute-name}' is missing from JWT token. " \
+             "Verify that you configured the host with permitted restrictions. " \
+             "In case of Compute Engine token, verify that you requested the token using 'format=full'",
+        code: "CONJ00176E"
       )
     end
 
--- a/spec/app/domain/authentication/authn-jwt/validate_restrictions_one_to_one_spec.rb
+++ b/spec/app/domain/authentication/authn-jwt/validate_restrictions_one_to_one_spec.rb
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe('Authentication::AuthnJwt::ValidateRestrictionsOneToOne') do
+  let(:right_email) { "admin@example.com" }
+  let(:wrong_email) { "wrong@example.com" }
+
+  let(:decoded_token) {
+    {
+      "namespace_id": "1",
+      "namespace_path": "root",
+      "project_id": "34",
+      "project_path": "root/test-proj",
+      "user_id": "1",
+      "user_login": "cucumber",
+      "user_email": right_email,
+      "pipeline_id": "1",
+      "job_id": "4",
+      "ref": "master",
+      "ref_type": "branch",
+      "ref_protected": "true",
+      "jti": "90c4414b-f7cf-4b98-9a4f-2c29f360e6d0",
+      "iss": "ec2-18-157-123-113.eu-central-1.compute.amazonaws.com",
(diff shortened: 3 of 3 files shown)


app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
```diff
@@ -0,0 +1,19 @@
+module Authentication
+  module AuthnJwt
+
+    # This class is responsible for retrieving the correct value from the GCP token
```
sashaCher, line 4:
```suggestion
    # This class is responsible for retrieving the correct value from the JWT token
```

app/domain/errors.rb
```diff
@@ -401,6 +401,13 @@ module AuthnJwt
               "   with field in the decoded token that contains the identity",
         code: "CONJ00075E"
       )
+
+      JwtTokenClaimIsMissing = ::Util::TrackableErrorClass.new(
+        msg: "Claim '{0-attribute-name}' is missing from JWT token. " \
+             "Verify that you configured the host with permitted restrictions. " \
+             "In case of Compute Engine token, verify that you requested the token using 'format=full'",
+        code: "CONJ00076E"
```
sashaCher, line 409: The error code is already in use by `MissingSecretValue` error.

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
```diff
@@ -0,0 +1,19 @@
...
+    # of the requested attribute.
+    class ValidateRestrictionsOneToOne
+      def initialize(decoded_token:)
+        @decoded_token = decoded_token
+      end
+
+      def valid_restriction?(restriction)
```
sashaCher, line 11: Can restriction be `null` or something like this?

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
tzheleznyak: Done

app/domain/errors.rb
tzheleznyak: There are many duplicated in this file . I think that ordering it should be out of scope of this pr

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
tzheleznyak: Because we do the restriction extraction from the db there is no case it should be null and this case will lead to error when we trying to access name attribute which will make the authentication request to fail

app/domain/errors.rb
sashaCher: I'm agree ordering is out of scope for this PR
However I think that choosing a unique code for new error is a part of the PR

sashaCher (approved):
LGTM

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
```diff
@@ -0,0 +1,19 @@
+module Authentication
+  module AuthnJwt
+
+    # This class is responsible for retrieving the correct value from the JWT token
+    # of the requested attribute.
+    class ValidateRestrictionsOneToOne
```
orenbm, line 6: why is this not a command class?

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
```diff
@@ -0,0 +1,19 @@
...
+      def initialize(decoded_token:)
+        @decoded_token = decoded_token
+      end
+
+      def valid_restriction?(restriction)
+        if @decoded_token.key?(restriction.name.to_sym)
+          return @decoded_token[restriction.name.to_sym] == restriction.value
+        end
+        raise Errors::Authentication::AuthnJwt::JwtTokenClaimIsMissing.new(restriction.name)
```
orenbm, line 15: I would switch this to first raise an error and then check

```suggestion
        unless @decoded_token.key?(restriction.name.to_sym)
          raise Errors::Authentication::AuthnJwt::JwtTokenClaimIsMissing.new(restriction.name)
        end
        return @decoded_token[restriction.name.to_sym] == restriction.value
```

app/domain/errors.rb
```diff
@@ -370,27 +370,27 @@ module AuthnJwt
       InvalidIssuerConfiguration = ::Util::TrackableErrorClass.new(
         msg: "Issuer authenticator configuration is invalid. You should configured as authenticator variables: " \
               "'{0-resource-name}' or one of the following: '{1-resource-name}','{2-resource-name}'",
-        code: "CONJ00070E"
+        code: "CONJ00170E"
```
orenbm, line 373: i am guessing these are pushed by mistake?

app/domain/errors.rb
```diff
@@ -399,7 +399,14 @@ module AuthnJwt
...
               "   with field in the decoded token that contains the identity",
-        code: "CONJ00075E"
+        code: "CONJ00175E"
+      )
+
+      JwtTokenClaimIsMissing = ::Util::TrackableErrorClass.new(
+        msg: "Claim '{0-attribute-name}' is missing from JWT token. " \
+             "Verify that you configured the host with permitted restrictions. " \
+             "In case of Compute Engine token, verify that you requested the token using 'format=full'",
+        code: "CONJ00176E"
```
orenbm, line 409:
```suggestion
        code: "CONJ00076E"
```
?

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
```diff
@@ -0,0 +1,19 @@
...
+    # This class is responsible for retrieving the correct value from the JWT token
+    # of the requested attribute.
+    class ValidateRestrictionsOneToOne
+      def initialize(decoded_token:)
+        @decoded_token = decoded_token
+      end
+
+      def valid_restriction?(restriction)
+        if @decoded_token.key?(restriction.name.to_sym)
+          return @decoded_token[restriction.name.to_sym] == restriction.value
```
nessiLahav, line 13: We do not want to raise an exception if it is not equal ?

app/domain/errors.rb
tzheleznyak: To avoid duplication

app/domain/errors.rb
tzheleznyak: Changed all error codes of jwt exceptions

app/domain/errors.rb
tzheleznyak: No its not
https://github.com/cyberark/conjur/pull/2148#discussion_r627614858

app/domain/errors.rb
tzheleznyak: https://github.com/cyberark/conjur/pull/2148#discussion_r627614858

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
tzheleznyak: Because it is used in the generic resource restriction validatior done by @abrahamko 

Each authenticator gcp for example have file authenticator_request.rb that impliments validation of single restriction (annotation)

For example in gcp it here :
https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_gcp/authentication_request.rb

The usage is here:
https://github.com/cyberark/conjur/blob/e44eaaef953651ebb2458cad6ef69bf6b1a1f040/app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb#L71

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
tzheleznyak: Fixed

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
tzheleznyak: The command class using it raises exception if its false
https://github.com/cyberark/conjur/blob/e44eaaef953651ebb2458cad6ef69bf6b1a1f040/app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb#L71

app/domain/authentication/authn_jwt/validate_restrictions_one_to_one.rb
abrahamko, line 6: This class usage is not part of this PR.
It makes this PR very hard to read, since reviewer cannot see how do you use this class.
When a new PR will arrive that uses this code, the reviewer will not have the full context to review its correct usage or comment on the original class, since it is not in that PR.

tzheleznyak: Yes you are correct.
Just wanted to try to make my pr's smaller then the one with the get identity

abrahamko: that's good effort, just keep them grouped with some meaning 😄

PR #2151, opened 2021-05-09
Rebase master branch to jwt authenticator feature branch

tzheleznyak:
Rebase master branch to jwt authenticator feature branch


benoaviram (approved):

app/domain/errors.rb
```diff
@@ -26,7 +26,7 @@ module Conjur
     )
 
     BadSecretEncoding = ::Util::TrackableErrorClass.new(
-      msg: "Issue encoding secret into JSON format, try including 'Accept: base64' " \
+      msg: "Issue encoding secret into JSON format, try including 'Accept-Encoding: base64' " \
           "header in request.",
       code: "CONJ00074E"
```
benoaviram, line 31: @tzheleznyak  notice that we also have this error code below, could you fix it?

benoaviram (changes requested):

app/domain/errors.rb
tzheleznyak: I saw there are duplicate error version and it should be a different ticket to make order of codes of all error messages in conur oss

benoaviram (approved):

PR #2154, opened 2021-05-09
JWT specific input validation

sashaCher:
### What does this PR do?
- Adds initial JWT specific input validation to the ConfigurationInterface and Generic vendor implementation

#### Change log

#### Test coverage

#### Documentation

#### API Changes


tzheleznyak (approved):
LGTM

app/domain/authentication/authn_jwt/jwt_initial_input_validation.rb
```diff
@@ -0,0 +1,40 @@
+require 'command_class'
+
+# This class
+module Authentication
+  module AuthnJwt
+
+    JWTInitialInputValidation ||= CommandClass.new(
```
orenbm, line 7: please change name to an action - `ValidateInput` should do

app/domain/authentication/authn_jwt/jwt_initial_input_validation.rb
```diff
@@ -0,0 +1,40 @@
...
+      dependencies: {
+        enabled_authenticators: InstalledAuthenticators.enabled_authenticators_str(ENV),
```
orenbm, line 9: not sure that this will work. i think it needs to be an input in case the env changes.

app/domain/authentication/authn_jwt/jwt_initial_input_validation.rb
```diff
@@ -0,0 +1,40 @@
...
+        # ValidateWebserviceIsWhitelisted calls ValidateAccountExists
```
orenbm, line 10: i would add the validation here as well for explicitness

app/domain/authentication/authn_jwt/jwt_initial_input_validation.rb
```diff
@@ -0,0 +1,40 @@
...
+        validate_webservice_is_whitelisted: Security::ValidateWebserviceIsWhitelisted.new,
+      },
+      inputs: %i[authentication_parameters]
+    ) do
+
+      def call
+        validate_webservice_is_whitelisted
+      end
```
orenbm, line 18: what about `validate_authenticator_exists`?

orenbm: also - validations on the user will go in another class, correct?

sashaCher: 1. Following my discussion with @nessiLahav `validate_authenticator_exists` is a `ConfigurationFactory` responsibility. It will return a general or concrete vendor authn implementation class instance and will throw an exception if `vendor` does not exist.
2. Validation of the user is possible only after token signature validation. It's in another class, right.

app/domain/authentication/authn_jwt/jwt_initial_input_validation.rb
sashaCher: I'll assume it as an input parameter

app/domain/authentication/authn_jwt/jwt_initial_validate_input.rb
```diff
@@ -0,0 +1,56 @@
+require 'command_class'
+
+# This class
+module Authentication
+  module AuthnJwt
+
+    JWTInitialValidateInput ||= CommandClass.new(
```
orenbm, line 7: rename to `ValidateInput` to show the action that is happening

app/domain/authentication/authn_jwt/jwt_initial_validate_input.rb
```diff
@@ -0,0 +1,56 @@
...
+      dependencies: {
+        # ValidateWebserviceIsWhitelisted is also calling ValidateAccountExists
+        # it's better to be explicit and on the safe side
+        validate_account_exists: ::Authentication::Security::ValidateAccountExists.new,
+        validate_webservice_is_whitelisted: ::Authentication::Security::ValidateWebserviceIsWhitelisted.new,
+        jwt_validate_body: Authentication::AuthnJwt::JWTValidateRequestBody.new
```
orenbm, line 13: rename to `ValidateRequestBody`. the `JWT` prefix is redundant

app/domain/authentication/authn_jwt/jwt_initial_validate_input.rb
```diff
@@ -0,0 +1,56 @@
...
+          account: @authentication_parameters.account
+        )
+      end
+
+      def validate_webservice_is_whitelisted
+        @validate_webservice_is_whitelisted.(
+            webservice: webservice,
+            account: @authentication_parameters.account,
+            enabled_authenticators: @enabled_authenticators
+        )
```
orenbm, line 37: indent 2 spaces instead of 4

app/domain/authentication/authn_jwt/jwt_initial_validate_input.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      def jwt_validate_body
+        @jwt_validate_body.(
+          body_string: @authentication_parameters.request.body.read
```
orenbm, line 42: i think you can get the same value by using `@authentication_parameters.credentials`. i don't know how `@authentication_parameters` is different than `@authentication_input` but in the latter you can use `credentials` instead of reading the body again.

app/domain/authentication/authn_jwt/jwt_validate_request_body.rb
```diff
@@ -0,0 +1,36 @@
+require 'command_class'
+
+# This class validates that request body contains JWT encoded token
+module Authentication
+  module AuthnJwt
+
+    JWTValidateRequestBody ||= CommandClass.new(
```
orenbm, line 7: remove the redundant `JWT` prefix

app/domain/authentication/authn_jwt/jwt_validate_request_body.rb
```diff
@@ -0,0 +1,36 @@
...
+    ) do
+
+      def call
+        validate_body
+      end
+
+      private
+
+      def validate_body
+        raise Errors::Authentication::Jwt::RequestBodyIsNotJWTToken unless @body_string =~ jwt_regex
```
orenbm, line 20: i would extract `@body_string =~ jwt_regex` to a separate method:
```
def jwt_body?
  @body_string =~ jwt_regex
end
```

app/domain/authentication/authn_jwt/jwt_initial_validate_input.rb
orenbm: btw, in `authn-oidc` the body is:
```
id_token: <JWT>
```
and in authn-gcp it is:
```
jwt: <JWT>
```

here the body is only the token itself without a key?

app/domain/errors.rb
```diff
@@ -166,6 +166,11 @@ module Jwt
         code: "CONJ00051E"
       )
 
+      RequestBodyIsNotJWTToken = ::Util::TrackableErrorClass.new(
+        msg: "The request body is not a JWT token",
```
orenbm, line 170: @shulifink please review

orenbm (changes requested):
left a few comments

app/domain/authentication/authn_jwt/jwt_initial_validate_input.rb
sashaCher: @orenbm  you are right `authentication_input` cares to read body to `credentials`

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -2,6 +2,13 @@ module Authentication
   module AuthnJwt
     # Mock JWTConfiguration class to use it to develop other part in the jwt authenticator
     class ConfigurationJWTGenericVendor < ConfigurationInterface
+
+      def self.validate_input(authentication_parameters)
+        Authentication::AuthnJwt::ValidateInput.new.call(
```
nessiLahav, line 7: the `Authentication::AuthnJwt::ValidateInput.new` should be in initialize function and in the interface function you only execute the call function

app/domain/authentication/authn_jwt/validate_input.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    ValidateInput ||= CommandClass.new(
+      dependencies: {
+        # ValidateWebserviceIsWhitelisted is also calling ValidateAccountExists
+        # it's better to be explicit and on the safe side
+        validate_account_exists: ::Authentication::Security::ValidateAccountExists.new,
+        validate_webservice_is_whitelisted: ::Authentication::Security::ValidateWebserviceIsWhitelisted.new,
+        jwt_validate_body: Authentication::AuthnJwt::ValidateRequestBody.new
+      },
+      inputs: %i[authentication_parameters enabled_authenticators]
```
nessiLahav, line 15: Just an idea: WDYT that `authentication_parameters` will contains `enabled_authenticators` because this is an interface function so it should be with the same signature like others.

Example to the code:
```
      inputs: %i[authentication_parameters]
    ) do
      extend(Forwardable)
      def_delegators(:@ authentication_parameters, :enabled_authenticators)
```

spec/app/domain/authentication/authn-jwt/validate_request_body_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  #   )(   ) _ (  )__)     )(   )__) \__ \  )(  \__ \
+  #  (__) (_) (_)(____)   (__) (____)(___/ (__) (___/
+
+  context "ValidateRequestBody" do
+    context "Validate request body when" do
+      subject do
+        Authentication::AuthnJwt::ValidateRequestBody.new()
+      end
+
+      it "the body is empty" do
```
nessiLahav, line 32: what about test: `the body is nil` ?

app/domain/authentication/authn_jwt/validate_input.rb
orenbm: i would leave the `authentication_parameters` to include only data of the current authn request.

sashaCher:
@orenbm We (@nessiLahav ) have decided to separate JWT specific input validation logic and general (account, whitelist etc.).
I've rescoped the PR only for JWT specific validations.
For generic validations I'll open an another PR.

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -2,6 +2,17 @@ module Authentication
...
+
+      def initialize
+        @validate_input = Authentication::AuthnJwt::ValidateInput.new
+      end
+
+      def self.validate_input(authentication_parameters)
+        @validate_input.call(
+          authentication_parameters: authentication_parameters
+        )
+      end
```
orenbm, line 14: why do we even call `ConfigurationJWTGenericVendor. validate_input(authentication_parameters )` and not just call `Authentication::AuthnJwt::ValidateInput.new. call(authentication_parameters: authentication_parameters)`?

do we see a case that this will not be the class that performs the check?

app/domain/authentication/authn_jwt/validate_input.rb
```diff
@@ -0,0 +1,27 @@
+require 'command_class'
+
+module Authentication
+  module AuthnJwt
+
+    ValidateInput ||= CommandClass.new(
+      dependencies: {
+        jwt_validate_body: Authentication::AuthnJwt::ValidateRequestBody.new
```
orenbm, line 8: please call the param in the same name as the class:
```suggestion
        validate_request_body: Authentication::AuthnJwt::ValidateRequestBody.new
```

app/domain/authentication/authn_jwt/jwt_initial_validate_input.rb
hilagross: It was defined in the feature doc
image

app/domain/authentication/authn_jwt/validate_input.rb
```diff
@@ -0,0 +1,27 @@
...
+      inputs: %i[authentication_parameters]
+    ) do
+
+      def call
+        jwt_validate_body
+      end
+
+      private
+
+      def jwt_validate_body
```
orenbm, line 19: please call it `validate_request_body`

orenbm: or even better `validate_body_is_jwt`. this way we can easily add more check in the future.

app/domain/authentication/authn_jwt/validate_input.rb
```diff
@@ -0,0 +1,27 @@
...
+        @jwt_validate_body.(
+          body_string: @authentication_parameters.credentials
+        )
+      end
+
```
orenbm, line 24: this empty line can be removed

app/domain/authentication/authn_jwt/validate_request_body.rb
```diff
@@ -0,0 +1,40 @@
+require 'command_class'
+
+# This class validates that request body contains JWT encoded token
+module Authentication
+  module AuthnJwt
+
+    ValidateRequestBody ||= CommandClass.new(
+      dependencies: {
+      },
```
orenbm, line 9: a command class without any dependencies should raise the question of its existence. do we need this class? can we move the logic of this class to its calling class? 

we may have a redundant layer here that makes the code reader make an extra hop between classes.

app/domain/authentication/authn_jwt/validate_request_body.rb
```diff
@@ -0,0 +1,40 @@
...
+      inputs: %i[body_string]
+    ) do
+
+      def call
+        validate_body
+      end
+
+      private
+
+      def validate_body
```
orenbm, line 19: i would call this `validate_body_is_jwt` to be more explicit. if we will have more checks in the future they can be in different methods

orenbm (changes requested):
left some comments

app/domain/authentication/authn_jwt/jwt_initial_validate_input.rb
sashaCher: So we have a third format with `jwt=` prefix... :-(
Let's be aligned at least with `gcp` means `jwt: ` prefix.

orenbm: @sashaCher you can look at [this file](https://github.com/cyberark/conjur/blob/2d28f6dff4297338855fd109c16b6b7baa365fdc/app/domain/authentication/jwt/decoded_credentials.rb#L6) for reference. i would use the `Jwt::DecodedCredentials` class in your case as well as it handles the same logic and will validate the request for you

orenbm: > So we have a third format with `jwt=` prefix... :-(
> Let's be aligned at least with `gcp` means `jwt: ` prefix.

@sashaCher i was wrong. it is `jwt=` for GCP as well. go ahead with this.

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
sashaCher: See the @nessiLahav's remark https://github.com/cyberark/conjur/pull/2154#discussion_r629112622
There's no need to instantiate it each call...

app/domain/authentication/authn_jwt/jwt_initial_validate_input.rb
sashaCher: @orenbm Thank you!

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
orenbm: that's not my point. I see that in the [AuthnJwt::Authenticate class](https://github.com/cyberark/conjur/blob/f345d632cedd100369ccee91a377ebf6df26693e/app/domain/authentication/authn_jwt/authenticator.rb#L16-L15), you call the `jwt_configuration` object that implements `validate_input`. do we think that the implementations will be different? i am suggesting to move the `validate_input` as it is implemented above (by calling `Authentication::AuthnJwt::ValidateInput.new`) in the `Authenticator` class instead of here.

app/domain/authentication/authn_jwt/validate_input.rb
sashaCher: Agree this parameter has been removed

app/domain/authentication/authn_jwt/validate_request_body.rb
sashaCher: I'll convert it to regular class

app/domain/authentication/authn_jwt/validate_input.rb
orenbm: what have been removed?

sashaCher:
@orenbm @nessiLahav PR is ready for re-review

app/domain/authentication/authn_jwt/configuration_interface.rb
```diff
@@ -2,6 +2,7 @@ module Authentication
   module AuthnJwt
     # Interface containing JWT configuration functions to be implemented in JWTConfiguration class for a vendor
     class ConfigurationInterface
+      def fetch_token(authentication_parameters); end
```
nessiLahav, line 5: WDYT to put this function `fetch_token` and the general function in the generic code `create_authentication_parameters` under the same umbrella ?

I think in the future each authenticator will have its own logic of create and validate the authenticate parameters, and it will also clean the generic code.

So my idea to change the function in the interface to `validate_and_create_authentication_parameters` 
it responsibility will be to validate the input (in this case the jwt token) and to return a new validated object `authentication_parameters` to be used in all other functions

nessiLahav (changes requested):
Left comments

app/domain/authentication/authn_jwt/configuration_interface.rb
sashaCher: I'm agree. I'll rename the interface method but code rearrange let's do in one of the next PRs

app/domain/logs.rb
```diff
@@ -389,6 +389,11 @@ module AuthnJwt
         msg: "Fetching jwks-uri signing key",
         code: "CONJ00076D"
       )
+
+      FetchingTokenString = ::Util::TrackableLogMessageClass.new(
+        msg: "Fetching jwt value from request",
```
sashaCher, line 394: @shulifink  please review

app/domain/errors.rb
InbalZilberman: @hilagross ^

app/domain/logs.rb
InbalZilberman: @hilagross ^

app/domain/authentication/authn_jwt/configuration_interface.rb
```diff
@@ -2,6 +2,7 @@ module Authentication
   module AuthnJwt
     # Interface containing JWT configuration functions to be implemented in JWTConfiguration class for a vendor
     class ConfigurationInterface
+      def create_authentication_parameters(authentication_parameters); end
```
nessiLahav, line 5: this function get as input `authenticator_input`

sashaCher: Not in this pr :-)

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -2,7 +2,9 @@ module Authentication
   module AuthnJwt
     # Mock JWTConfiguration class to use it to develop other part in the jwt authenticator
     class ConfigurationJWTGenericVendor < ConfigurationInterface
+
       def initialize
+        @token_fetcher = Authentication::AuthnJwt::FetchTokenFromCredentials.new
```
nessiLahav, line 7: i think FetchTokenFromCredentials should be a command class, @orenbm, @abrahamko do you agree ?

app/domain/authentication/jwt/decoded_credentials.rb
```diff
@@ -6,21 +6,44 @@ class DecodedCredentials
...
+        @logger = Rails.logger
         @decoded_credentials = Hash[URI.decode_www_form(credentials)]
-        validate
+        validate_jwt_request_field_is_present
+        validate_body_contains_jwt
       end
 
       def jwt
-        @decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME]
+        @jwt_token
```
nessiLahav, line 16: function name and variable should have the same name

nessiLahav (approved):
Few tiny comments not a blocker

app/domain/errors.rb
shulifink: What is a "reques body" - it what does it mean that it's not a JWT token? 
Do you mean that the request doesn;t include/contain a JWT token?

app/domain/logs.rb
shulifink: maybe 
```suggestion
        msg: "Fetching JWT token from request",
```

??

shulifink: ie... to be consistent with the above...

shulifink (changes requested):

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
abrahamko: Yes, this looks like a pure action, so command class fits perfectly.
Unless there's a good reason why not; is there, @sashaCher?
Also, it amplifies the dependencies, as they are explicitly declared at the top of the class, so no hidden dependency such as `Authentication::Jwt::DecodedCredentials`.

app/domain/authentication/jwt/decoded_credentials.rb
```diff
@@ -6,21 +6,44 @@ class DecodedCredentials
...
+
+      def jwt_regex
+        # https://datatracker.ietf.org/doc/html/rfc4648#section-5
+        # jwt token is 3 blocks of base64url encoded strings separated by period '.'
+        # base64url encoding differs from regular base64 encoding as follows
+        # - padding is skipped so the pad character '=' doesn't have to be percent encoded
+        # - the 62nd and 63rd regular base64 encoding characters ('+' and '/') are replace with ('-' and '_')
+        #   The changes make the encoding alphabet file and URL safe.
+        # tokens without a signature part are denied
+        /\A[-A-Za-z0-9_=]+\.[-A-Za-z0-9_=]+\.[-A-Za-z0-9_=]+\z/
```
abrahamko, line 44: I think it is better to define it as const instead of as a method implementation.

abrahamko: Also, consider taking out the 3 similar parts that match the base64url (that is, the `[-A-Za-z0-9_=]+`) into a separate regex const so at the end you will have:
```suggestion
        /\A#{BASE_64_URL_REGEX}\.#{BASE_64_URL_REGEX}\.#{BASE_64_URL_REGEX}\z/
```
which is more readable and maintainable IMO.

app/domain/authentication/jwt/decoded_credentials.rb
sashaCher: Chnaged

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
sashaCher: I've converted it to a command class.
Thank you

sashaCher:
@nessiLahav @abrahamko I've addressed your comments.
@shulifink I've changed the messages to be aligned.
Regarding CONJ00077E. The authenticator request body should contain a JWT token. The error will raise when it does not contain the token.

app/domain/authentication/authn_jwt/authenticator.rb
```diff
@@ -25,6 +25,8 @@ def call
       def create_authentication_parameters
         @logger.debug(LogMessages::Authentication::AuthnJwt::CREATING_AUTHENTICATION_PARAMETERS_OBJECT.new)
         @authentication_parameters = Authentication::AuthnJwt::AuthenticationParameters.new(@authenticator_input)
+        @logger.debug(LogMessages::Authentication::AuthnJwt::FetchingTokenString.new)
+        @authentication_parameters.jwt_token = @jwt_configuration.create_authentication_parameters(@authentication_parameters)
```
orenbm, line 29: this is quite odd. to create the `jwt_token` object that is part of the `authentication_parameters` object, you need to `authentication_parameters` object. do we have a more sain way to do it?

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
orenbm: can you also call the parameter `fetch_token_from_credentials`? the param name shoulod match the class it represents

app/domain/authentication/authn_jwt/authenticator.rb
sashaCher: I'm agree. 
I'm planning to have a way in one of the next PRs

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -14,6 +16,10 @@ def initialize
         )
       end
 
+      def create_authentication_parameters(authentication_parameters)
+        @token_fetcher.call(authentication_parameters)
+      end
```
orenbm, line 21: this method is weird. it is called `create_authentication_parameters` but gets an `authentication_parameters` object as an input and calls the token_fetcher object. should it be renamed to `fetch_token_from_credentials`?

app/domain/authentication/authn_jwt/fetch_token_from_credentials.rb
```diff
@@ -0,0 +1,22 @@
+module Authentication
+  module AuthnJwt
+
+    FetchTokenFromCredentials ||= CommandClass.new(
+      dependencies: {
+        decoded_credentials: Authentication::Jwt::DecodedCredentials
```
orenbm, line 6: rename to `decoded_credentials_class`.

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
sashaCher: The same there.
It will be addressed in next PRs

app/domain/authentication/authn_jwt/fetch_token_from_credentials.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      def call
+        decoded_credentials(@authentication_parameters).jwt
+      end
+
+      private
+
+      def decoded_credentials(authentication_parameters)
+        @decoded_credentials.new(authentication_parameters.credentials)
+      end
```
orenbm, line 19: please change to this:
```suggestion
      def call
        decode_credentials
        extracted_token_from_credentials
      end

      private

      def decode_credentials
        decoded_credentials
      end

      def extracted_token_from_credentials
        decoded_credentials.jwt
      end

      def decoded_credentials
        @decoded_credentials_class.new(@authentication_parameters.credentials)
      end
```

orenbm: explanation:
1. the call function should not have ruby code but it should tell the story
2. we don't need to pass the `@authentication_parameters` object to the private method as it can access it by itself. this way we can also memoize the object.

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -14,6 +16,10 @@ def initialize
         )
       end
 
+      def create_authentication_parameters(authentication_parameters)
```
orenbm, line 19: should this method be renamed to `fetch_token_from_credentials`?

app/domain/authentication/authn_jwt/fetch_token_from_credentials.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      def call
+        decoded_credentials(@authentication_parameters).jwt
+      end
+
+      private
+
+      def decoded_credentials(authentication_parameters)
+        @decoded_credentials_class.new(authentication_parameters.credentials)
+      end
```
orenbm, line 19:
```suggestion
      def call
        decode_credentials
        extracted_token_from_credentials
      end

      private

      # this function is implemented so that the 'call' function will tell the story
      def decode_credentials
        decoded_credentials
      end

      def decoded_credentials
        @decoded_credentials ||= @decoded_credentials_class.new(@authentication_parameters.credentials)
      end

      def extracted_token_from_credentials
        decoded_credentials.jwt
      end
```

app/domain/logs.rb
```diff
@@ -389,6 +389,11 @@ module AuthnJwt
         msg: "Fetching jwks-uri signing key",
         code: "CONJ00076D"
       )
+
+      FetchingTokenString = ::Util::TrackableLogMessageClass.new(
+        msg: "Fetching JWT token from request",
+        code: "CONJ00077D"
```
hilagross, line 395: Do we have "Fetched" msg?
according to this: https://ca-il-confluence.il.cyber-ark.com/display/rndp/Message+Pool#/

app/domain/authentication/authn_jwt/fetch_token_from_credentials.rb
```diff
@@ -0,0 +1,22 @@
+module Authentication
+  module AuthnJwt
+
+    FetchTokenFromCredentials ||= CommandClass.new(
```
orenbm, line 4: the word `fetch` indicates that we go to a DB whereas we actually extract it from the request body. how about `ExtractTokenFromCredentials `? not a blocker

app/domain/authentication/jwt/decoded_credentials.rb
```diff
@@ -4,23 +4,44 @@ module Jwt
...
+      # base64url encoding differs from regular base64 encoding as follows
+      # - padding is skipped so the pad character '=' doesn't have to be percent encoded
+      # - the 62nd and 63rd regular base64 encoding characters ('+' and '/') are replace with ('-' and '_')
+      #   The changes make the encoding alphabet file and URL safe.
+      # tokens without a signature part are denied
+      BASE_64_URL_REGEX = "[-A-Za-z0-9_=]+".freeze
+      JWT_REGEX = /\A#{BASE_64_URL_REGEX}\.#{BASE_64_URL_REGEX}\.#{BASE_64_URL_REGEX}\z/.freeze
 
       def initialize(credentials)
+        @logger = Rails.logger
```
orenbm, line 18:
```suggestion
      def initialize(credentials:, logger: Rails.logger)
        @logger = logger
```

this way we can mock the logger if needed

app/domain/authentication/jwt/decoded_credentials.rb
```diff
@@ -4,23 +4,44 @@ module Jwt
...
 
       private
 
-      def validate
+      def validate_jwt_request_field_is_present
         if @decoded_credentials.fetch(JWT_REQUEST_BODY_FIELD_NAME, "") == ""
           raise Errors::Authentication::RequestBody::MissingRequestParam, JWT_REQUEST_BODY_FIELD_NAME
         end
+        @jwt = @decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].strip
       end
```
orenbm, line 35: the method name is `validate_jwt_request_field_is_present` but you also update the `jwt` param value. looks like you need to separate this.

orenbm: do we really need line 34? what's the problem with the current implementation?

app/domain/authentication/jwt/decoded_credentials.rb
sashaCher: It's a leftover from debug automation.
I'll remove it

app/domain/authentication/jwt/decoded_credentials.rb
sashaCher: The `jwt` parameter value should be trimmed before regex validation. Our automation sends the value with `\n` at the end.
And in general it's better to return the exact value without additional characters

orenbm: ok so add `.strip` to the original `jwt` method

sashaCher: `.strip` should be added before regex validation
I want to avoid adding it twice.

orenbm: i don't understand. what if you have the method:
```
def jwt
  @jwt ||= @decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].strip
end
```

and then call `jwt` in line 42. wouldn't it do the job?

sashaCher: will extract the line to separate function

sashaCher:
@orenbm Thank you for your comments and suggestions. I've addressed most of them and aligned a naming convention.
The naming convention will remain a subject to change in next PRs but previous version definitely did not make sense.

app/domain/authentication/authn_jwt/authenticator.rb
```diff
@@ -25,6 +25,9 @@ def call
       def create_authentication_parameters
         @logger.debug(LogMessages::Authentication::AuthnJwt::CREATING_AUTHENTICATION_PARAMETERS_OBJECT.new)
         @authentication_parameters = Authentication::AuthnJwt::AuthenticationParameters.new(@authenticator_input)
+        @logger.debug(LogMessages::Authentication::AuthnJwt::ExtractTokenString.new)
+        @authentication_parameters.jwt_token = @jwt_configuration
+                                                 .extract_token_from_credentials(@authentication_parameters.credentials)
```
orenbm, line 30:
```suggestion
        @authentication_parameters.jwt_token = @jwt_configuration
                                                 .extract_token_from_credentials(
                                                   @authentication_parameters.credentials
                                                 )
```

app/domain/authentication/authn_jwt/extract_token_from_credentials.rb
```diff
@@ -0,0 +1,31 @@
...
+        decoded_credentials
+      end
+
+      def decoded_credentials
+        @decoded_credentials ||= @decoded_credentials_class.new(@credentials)
+      end
+
+      def extract_token_from_credentials
+        decoded_credentials.jwt
+      end
```
orenbm, line 28: sorry, i would add also:
```suggestion
      def extract_token_from_credentials
        extracted_token
      end
      
      def extracted_token
        @extracted_token ||= decoded_credentials.jwt
      end
```

and add `extracted_token` to the end of the `call` function. this way the call function will be:
```
decode_credentials
extract_token_from_credentials
extracted_token
```

which tells the story best

app/domain/authentication/jwt/decoded_credentials.rb
```diff
@@ -4,23 +4,47 @@ module Jwt
...
-      def validate
+      def validate_jwt_request_field_is_present
         if @decoded_credentials.fetch(JWT_REQUEST_BODY_FIELD_NAME, "") == ""
           raise Errors::Authentication::RequestBody::MissingRequestParam, JWT_REQUEST_BODY_FIELD_NAME
         end
       end
+
+      def extract_token
+        @jwt ||= @decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].strip
+      end
```
orenbm, line 38: if you memoize an object then it needs to have the same name as the method.
change this to:
```
      def extract_token
        jwt
      end
```

and change the public method `jwt` to:
```
def jwt
  @jwt ||= @decoded_credentials[JWT_REQUEST_BODY_FIELD_NAME].strip
end
```

orenbm: but you don't really need the method `extract_token`. you can just call `jwt` when you need it as it is memoized.

app/domain/authentication/jwt/decoded_credentials.rb
```diff
@@ -4,23 +4,47 @@ module Jwt
...
+
+      def validate_jwt_format
+        raise Errors::Authentication::Jwt::RequestBodyIsNotJWTToken unless is_jwt?
+      end
+
+      def is_jwt?
+        @jwt =~ JWT_REGEX
```
orenbm, line 45:
```suggestion
        jwt =~ JWT_REGEX
```

you should use the method and not the `@jwt` object itself

app/domain/errors.rb
```diff
@@ -166,6 +166,11 @@ module Jwt
         code: "CONJ00051E"
       )
 
+      RequestBodyIsNotJWTToken = ::Util::TrackableErrorClass.new(
```
orenbm, line 169: i would change `RequestBodyIsNotJWTToken` to `RequestBodyMissingJWTToken`

orenbm (approved):

PR #2175, opened 2021-05-14
2168  trusted proxies can be read from config file

### LINKED ISSUE #2168: Trusted proxies can be read from config file
Create a new class `Conjur::ConjurConfig` using the Anyway Config gem and configure it to:
- read from `/etc/conjur/config/conjur.yml`
- read environment variables prefixed with `CONJUR_`
- fall back to `TRUSTED_PROXIES` for backwards compatibility
- default to empty string if no values are configured

Update `Conjur::TrustedProxyFilter` to read the trusted proxies value from the ConjurConfig object.

Logging
- INFO to Rails log when config object created "Loaded config from <file-path>"
- WARNING to Rails log when no config file present "No config file found at <file-path>"

Unit Tests
- update the Conjur::TrustedProxyFilter unit tests to use the ConjurConfig object
- create new unit tests for Conjur::ConjurConfig to test the precedence behavior

jtuttle:
close #2168


lib/conjur/conjur_config.rb
```diff
@@ -3,13 +3,15 @@
 # from the Conjur application code.
 
 class Conjur::ConjurConfig < Anyway::Config
-  config_name :conjur
+  config_name :config
```
micahlee: This is definitely in the nitpick category, but should these corrections be fixed up into the original commit, rather than changing them here in the spec commit?

config/application.rb
```diff
@@ -74,5 +74,10 @@ class Application < Rails::Application
     # We create this in application.rb instead of an initializer so that it's
     # guaranteed to be available for other initializers to use.
     config.conjur_config = Conjur::ConjurConfig.new
+
+    puts "Loaded configuration:"
```
micahlee, line 78: Should this be using the Rails logger, rather than `puts` directly?

micahlee (commented):
This looks good to me! I left just a couple of small comments. I know @jonahx also wanted to look at this, but if that changes, I'm happy to come back and approve.

Thanks!

config/application.rb
jtuttle: The Rails logger hasn't been set yet, so I would have to set it to `Logger.new(STDOUT)` or something temporarily and then let it get over-written by the environment-specific settings. I could do that but I'm not sure there's much point.

lib/conjur/conjur_config.rb
```diff
@@ -0,0 +1,28 @@
...
+    )
+
+    def attribute_sources
+      sources = {}
+
+      to_source_trace.each do |k, v|
+        sources[k.to_sym] = v[:source][:type]
+      end
+
+      sources
```
jonahx, line 25: Something like this should work (ie, this line would be the entire method body):

```ruby
to_source_trace.map { |k,v| [ k.to_sym, v[:source][:type] ] }.to_h

File: lib/conjur/trusted_proxy_filter.rb [hunk 4]
```diff
@@ -10,8 +10,8 @@ module Conjur
   #
   # Example: TRUSTED_PROXIES=4.4.4.4,192.168.100.0/24
   class TrustedProxyFilter
-    def initialize(env: ENV, options: { disable_cache: true })
-      @env = env
+    def initialize(config: Conjur::ConjurConfig.new, options: { disable_cache: true })
+      @config = config
```
jonahx, line 14: Formatting:

```ruby
class TrustedProxyFilter
  def initialize(
    config: Conjur::ConjurConfig.new,
    options: { disable_cache: true }
  )
    @config = config
    @options = options
    @cached_trusted_proxies = nil

    trusted_proxies
  end
```

jonahx, line 14: Re: `options: { disable_cache: true }`: Is it likely we'll have other options in the future?  Otherwise perhaps we should YAGNI and just change the signature to:
```ruby
def initialize(config: Conjur::ConjurConfig.new, disable_cache: true)
```

lib/conjur/trusted_proxy_filter.rb
```diff
@@ -29,25 +29,25 @@ def validate_trusted_proxies
...
       @cached_trusted_proxies = proxy_ips unless @options[:disable_cache]
 
       proxy_ips
     end
 
-    # Reek flags @env['TRUSTED_PROXIES'] as :reek:DuplicateMethodCall. Refactoring
-    # this would not enhance the readability or performance.
-    def env_trusted_proxies
-      return [] unless @env['TRUSTED_PROXIES']
+    def configured_trusted_proxies
```
jonahx, line 45: Optional idea, I throw it out for your consideration -- get rid of helper method:

```ruby
def configured_trusted_proxies
  trusted_proxies = @config.trusted_proxies

  return [] unless trusted_proxies

  Set
    .new(trusted_proxies.split(','))
    .map { |cidr| IPAddr.new(cidr) }
rescue IPAddr::Error
  raise Errors::Conjur::InvalidTrustedProxies, cidr
end
```

lib/conjur/trusted_proxy_filter.rb
jtuttle: Sounds good, yeah.

lib/conjur/conjur_config.rb
jtuttle: Yeah probably not worth doing this retroactively.

lib/conjur/trusted_proxy_filter.rb
```diff
@@ -10,8 +10,8 @@ module Conjur
   #
   # Example: TRUSTED_PROXIES=4.4.4.4,192.168.100.0/24
   class TrustedProxyFilter
-    def initialize(env: ENV, options: { disable_cache: true })
-      @env = env
+    def initialize(config: Conjur::ConjurConfig.new, options: { disable_cache: true })
```
jonahx, line 13: I think the name `TrustedProxyFilter` is subtly inaccurate.  A filter is something that takes a list, and returns a sublist (the filtered list) back.  

In contrast, this (via the `call` method) takes a single IP and tells you if it's valid.  It's essentially a big configurable function whose input is an IP and whose output is a boolean value.  What would you name such a function?

Probably something like `valid_ip?`.  However, since this is a class name (which can't end in `?`) I'd consider something like `class IsIpValid` or `class IsIpTrusted`.

In client code:

```ruby
is_trusted = IsIpTrusted.new(...)
if is_trusted(suspicious_ip)
   # etc....
```

jtuttle: Submitted a [new issue](https://github.com/cyberark/conjur/issues/2176) for this, don't want to deal w/ it as part of this PR.

jonahx (dismissed):

jonahx (approved):

PR #2179, opened 2021-05-18
Validate trusted_proxies config field

### LINKED ISSUE #2170: ConjurConfig object validates config values
Add an [`on_load` method](https://github.com/palkan/anyway_config#validation-and-callbacks) to `Conjur::ConjurConfig` and use it to validate that `trusted_proxies` is a comma-separated list of valid IPs / IP masks. If validation fails, collect the error in a hash, then raise an exception with details about which fields failed validation after performing validation on all fields. The exception message should NOT include the config value to avoid logging potentially sensitive values.

Unit tests
- ConjurConfig exception includes details about which attribute failed validation
- ConjurConfig exception does not include the attribute value

Integration tests
- when Conjur has invalid configuration, the server fails to start with an error that includes the field that failed validation

jtuttle:
close #2170

### DIFF
--- a/lib/conjur/conjur_config.rb
+++ b/lib/conjur/conjur_config.rb
@@ -15,8 +15,32 @@ module Conjur
       trusted_proxies: (ENV['TRUSTED_PROXIES'] || "")
     )
 
+    # Perform validations and collect errors then report results as exception
+    on_load do
+      invalid = []
+
+      invalid << "trusted_proxies" unless trusted_proxies_valid?
+
+      unless invalid.empty?
+        raise Errors::Conjur::InvalidConfigValues, invalid.join(', ')
+      end
+    end
+
+    # Get attribute sources without including attribute values
     def attribute_sources
       to_source_trace.map { |k,v| [ k.to_sym, v[:source][:type] ] }.to_h
+    end
+
+    private
+
+    def trusted_proxies_valid?
+      trusted_proxies.split(',').each do |cidr|
+        IPAddr.new(cidr)
+      end
+
+      true
+    rescue
+      false
     end
   end
 end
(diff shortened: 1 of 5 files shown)


lib/conjur/conjur_config.rb
```diff
@@ -15,8 +15,32 @@ class ConjurConfig < Anyway::Config
...
+    end
+
+    # Get attribute sources without including attribute values
     def attribute_sources
       to_source_trace.map { |k,v| [ k.to_sym, v[:source][:type] ] }.to_h
     end
+
+    private
+
+    def validate_trusted_proxies
```
orenbm, line 36: as this is a bool method, can we change the name to `valid_trusted_proxies?`

lib/conjur/conjur_config.rb
```diff
@@ -15,8 +15,32 @@ class ConjurConfig < Anyway::Config
       trusted_proxies: (ENV['TRUSTED_PROXIES'] || "")
     )
 
+    # Perform validations and collect errors then report results as exception
+    on_load do
+      invalid = []
+
+      invalid << "trusted_proxies" unless validate_trusted_proxies
```
orenbm, line 22: what's going on here? should this be `invalid << trusted_proxies`? seems that you are adding the string "trusted_proxies" instead of the actual `trusted_proxies` list

jtuttle: We need to collect the names of the fields for the error raised at the end of the method that reports which fields failed validation.

orenbm: so what's the `invalid.join(', ')` below do? wouldn't it always be just the string "trusted_proxies" (unless it's empty)?

jtuttle: For now yeah but when we start adding new configuration values we'll collect them in a list and join them into a string for the error message. Just establishing the pattern for later.

orenbm: got it, thanks for clarifying

micahlee (approved):

PR #2191, opened 2021-05-21
Set the puma persistent_timeout higher than nginx

ismarc:
If the load balancer/proxy/server in front of Conjur has
a longer timeout than Conjur, there will be 502s as Conjur will
close the connection even if there are pending requests.  This
can lead to it being marked unavailable, errors returned to the
client and performance issues as nginx/load balancer in front of it

### What does this PR do?
- _What's changed? Why were these changes made?_ This specifies the puma `persistent_timeout` value that should be higher than any load balancer or proxy in front of Conjur.  The default value is 20 seconds, shorter than nginx and most load balancer default values.

#### Resolves 
- [conjurinc/ops#778 (private)](conjurinc/ops#778)
- [cyberark/conjur-ecs-deploy#28 (private)](cyberark/conjur-ecs-deploy#28)

#### Potentially solves symptoms associated with
- [conjurinc/appliance#1201 (private)](conjurinc/appliance#1201)
- The multitude of retries we have built in to our tests in case of intermittent gateway errors

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix bug where running `conjurctl server` or `conjurctl account create` with
   passwords that contain `,`s sent via stdin raised an error.
   [cyberark/conjur#2159](https://github.com/cyberark/conjur/issues/2159)
+- Specify keepalive timeout for puma to be longer than proxy and load balancers
 
 ### Security
 - Upgrade Rails to 5.2.5 to resolve CVE-2021-22885 
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -11,6 +11,15 @@ threads threads_count, threads_count
 # cause and implement a permanent solution.
 worker_timeout 600
 
+# If the load balancer/proxy/server in front of Conjur has
+# a longer timeout than Conjur, there will be 502s as Conjur will
+# close the connection even if there are pending requests. This
+# can lead to it being marked unavailable, errors returned to the
+# client and performance issues as nginx/load balancer in front of it.
+# Puma defaults to 20 seconds which is substantially lower than
+# other service defaults
+persistent_timeout 80
+
 preload_app!
 
 rackup      DefaultRackup


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix bug where running `conjurctl server` or `conjurctl account create` with
   passwords that contain `,`s sent via stdin raised an error.
   [cyberark/conjur#2159](https://github.com/cyberark/conjur/issues/2159)
+- Specify keepalive timeout for puma to be longer than proxy and load balancers
```
orenbm, line 20: can you link to an issue please?

orenbm: how does this change affect the user?

orenbm: does it fix a known bug?

config/puma.rb
```diff
@@ -11,6 +11,12 @@
 # cause and implement a permanent solution.
 worker_timeout 600
 
+# Set the keepalive timeout higher than what nginx is set to
+# Defaults to 20
```
jtuttle, line 15: puma defaults to 20 or nginx? It's not quite clear after reading the previous line.

config/puma.rb
```diff
@@ -11,6 +11,12 @@
...
+# A shorter value on the Conjur side causes errors and
+# potential instability
```
jtuttle, line 17: This comment could probably use a little more detail. What you wrote in the issue description was good, maybe incorporate some of that?

jtuttle (changes requested):

CHANGELOG.md
ismarc: I've added links to known-related issues in the description.  The effect should be to prevent the random and spurious 502 errors frequently seen when running conjur behind a load balancer/nginx/proxy due to mismatched keepalive times, but otherwise users should not see behavioral changes.  The issue was discovered for the ConjurOps v5 deployment, but is a symptom that has been seen pretty frequently in other contexts and always chalked up to transient issues.

config/puma.rb
ismarc: I've updated the comment to better describe the rationale and what default value is being overridden.

jtuttle (approved):

jtuttle (approved):

whip113:
@ismarc sent me this little tweak for some flakiness I observed while trying to use the seed-fetcher with an AWS auto-scaling group. Note that in my setup, I did NOT have a load balancer in front of my leader, my followers connected directly to it. I had regular failures with deployed followers where some wouldn't get the seed file at all. This seemed to get markedly worse after I exceeded 30 followers (with `max_wal_senders` increased appropriately). After making the change to puma.rb as suggested here, I was able to roll out 90 followers with zero failures (failure rate before was approx 20%). I suggest we publish this in a KBA as well, so that customers can start benefitting from it before 12.2 is released (or even without having to upgrade to 12.2).

PR #2199, opened 2021-05-23
Fix authentication parameters extraction

tzheleznyak:
1. Fix extraction of authentication parameters from request.
2. Fix extraction of annontation from host policy. Currently it works only with !user. Now it works with both of them

### DIFF
--- a/app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
+++ b/app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
@@ -12,10 +12,11 @@ module Authentication
       end
 
       def create_authentication_parameters(authenticator_input)
-        @authentication_parameters = @authentication_parameters_class.new(authenticator_input)
-        @authentication_parameters.jwt_token = @extract_token_from_credentials.call(
-          credentials: authenticator_input.request.read
+        authentication_parameters = @authentication_parameters_class.new(authenticator_input)
+        authentication_parameters.jwt_token = @extract_token_from_credentials.call(
+          credentials: authenticator_input.request.body.read
         )
+        authentication_parameters
       end
 
       def jwt_identity(authentication_parameters)
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -57,6 +57,13 @@ class Role < Sequel::Model
       tokens.join(":")
     end
 
+    def roleid_from_host(account, login)
+      tokens = login.split('/', 2)
+      tokens.unshift('host') if tokens.length == 1
+      tokens.unshift(account)
+      tokens.join(":")
+    end
+
     def username_from_roleid(roleid)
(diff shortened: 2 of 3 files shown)


app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -12,10 +12,11 @@ def initialize
       end
 
       def create_authentication_parameters(authenticator_input)
```
orenbm, line 14: do we need the `create_` prefix? can this method be named `authentication_parameters` to match ruby style?

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -12,10 +12,11 @@ def initialize
...
-        @authentication_parameters = @authentication_parameters_class.new(authenticator_input)
-        @authentication_parameters.jwt_token = @extract_token_from_credentials.call(
-          credentials: authenticator_input.request.read
+        authentication_parameters = @authentication_parameters_class.new(authenticator_input)
+        authentication_parameters.jwt_token = @extract_token_from_credentials.call(
```
orenbm, line 16: can the `jwt_token` be passed to the initializer instead of being an `attr_accessor`?

app/models/role.rb
```diff
@@ -57,6 +57,13 @@ def roleid_from_username(account, login)
       tokens.join(":")
     end
 
+    def roleid_from_host(account, login)
```
orenbm, line 60: i don't think that you need this method. the method `roleid_from_username` works for both hosts and users. am i missing something?

tzheleznyak: Before i done this policies like this didn't work
image
Because the role it created was account:host:role_name but it searched for account:user:role_name
Maybe the way i loaded the policy is wrong?

orenbm: did you pass the request using `host/bla` or just `bla`? hosts should be prefixed using `host/`

orenbm: now you can better understand the logic of `roleid_from_username`

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
tzheleznyak: Removed create

tzheleznyak:
Will open other PR because bug 2 is not relevant

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
tzheleznyak: It will make us to change all the unit tests of the authneticator. And the Status API of the authenticator that @Avi-Eyal is doing uses the authentication parameters without jwt token

PR #2212, opened 2021-05-26
Throw runtime error on empty annotation

tzheleznyak:
Move fetching of annotation out of the ExtractResourceRestrictions Class. So we could throw errors for empty annotation only from the JWT Authenticator

### DIFF
--- a/app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
+++ b/app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
@@ -109,7 +109,10 @@ module Authentication
       def initialize_validate_restrictions
         @restriction_validator = Authentication::AuthnJwt::ValidateRestrictionsOneToOne
         @restrictions_from_annotations_class = Authentication::ResourceRestrictions::GetServiceSpecificRestrictionsFromAnnotation
-        @extract_resource_restrictions = Authentication::ResourceRestrictions::ExtractResourceRestrictions.new(get_restriction_from_annotation: @restrictions_from_annotations_class)
+        @extract_resource_restrictions = Authentication::ResourceRestrictions::ExtractResourceRestrictions.new(
+          get_restriction_from_annotation: @restrictions_from_annotations_class,
+          ignore_empty_annotations: false
+        )
         @validate_resource_restrictions = @validate_resource_restrictions_class.new(extract_resource_restrictions: @extract_resource_restrictions)
         @constraints = Authentication::Constraints::MultipleConstraint.new(
           Authentication::Constraints::NotEmptyConstraint.new
--- a/app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
+++ b/app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
@@ -7,10 +7,11 @@ module Authentication
       dependencies: {
         resource_restrictions_class: Authentication::ResourceRestrictions::ResourceRestrictions,
         get_restriction_from_annotation: Authentication::ResourceRestrictions::GetRestrictionFromAnnotation,
-        fetch_resource_annotations: Authentication::ResourceRestrictions::FetchResourceAnnotations,
+        fetch_resource_annotations_class: Authentication::ResourceRestrictions::FetchResourceAnnotations,
         role_class: ::Role,
         resource_class: ::Resource,
-        logger: Rails.logger
+        logger: Rails.logger,
+        ignore_empty_annotations: true
       },
       inputs: %i[authenticator_name service_id role_name account]
     ) do
@@ -34,9 +35,14 @@ module Authentication
       private
 
       def fetch_resource_annotations
-        @resource_annotations = @fetch_resource_annotations.new.call(
+        resource_annotations
+      end
+
+      def resource_annotations
+        @resource_annotations ||= @fetch_resource_annotations_class.new.call(
           account: @account,
-          role_name: @role_name
+          role_name: @role_name,
+          ignore_empty_annotations: @ignore_empty_annotations
         )
       end
 
@@ -46,7 +52,7 @@ module Authentication
 
       def resource_restrictions_hash
         @resource_restrictions_hash ||=
-          @resource_annotations.each_with_object({}) do |(annotation_name, annotation_value), resource_restrictions_hash|
+          resource_annotations.each_with_object({}) do |(annotation_name, annotation_value), resource_restrictions_hash|
             add_restriction_to_hash(annotation_name, annotation_value, resource_restrictions_hash)
           end
       end
--- a/app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
+++ b/app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
@@ -7,6 +7,7 @@ module Authentication
       dependencies: {
         resource_restrictions_class: Authentication::ResourceRestrictions::ResourceRestrictions,
         get_restriction_from_annotation: Authentication::ResourceRestrictions::GetRestrictionFromAnnotation,
+        fetch_resource_annotations: Authentication::ResourceRestrictions::FetchResourceAnnotations,
         role_class: ::Role,
         resource_class: ::Resource,
         logger: Rails.logger
@@ -33,28 +34,10 @@ module Authentication
       private
 
       def fetch_resource_annotations
-        resource_annotations
-      end
-
-      def resource_annotations
-        @resource_annotations ||=
-          resource.annotations.each_with_object({}) do |annotation, result|
-            annotation_values = annotation.values
-            value = annotation_values[:value]
-            next if value.blank?
-
-            result[annotation_values[:name]] = value
-          end
-      end
-
-      def resource
-        # Validate role exists, otherwise getting role annotations return empty hash.
-        role_id = @role_class.roleid_from_username(@account, @role_name)
-        resource = @resource_class[role_id]
-
-        raise Errors::Authentication::Security::RoleNotFound, role_id unless resource
-
-        resource
+        @resource_annotations = @fetch_resource_annotations.new.call(
+          account: @account,
+          role_name: @role_name
+        )
       end
 
       def extract_resource_restrictions_from_annotations
@@ -63,7 +46,7 @@ module Authentication
 
       def resource_restrictions_hash
         @resource_restrictions_hash ||=
-          resource_annotations.each_with_object({}) do |(annotation_name, annotation_value), resource_restrictions_hash|
+          @resource_annotations.each_with_object({}) do |(annotation_name, annotation_value), resource_restrictions_hash|
             add_restriction_to_hash(annotation_name, annotation_value, resource_restrictions_hash)
           end
       end
--- a/app/domain/authentication/resource_restrictions/fetch_resource_annotations.rb
+++ b/app/domain/authentication/resource_restrictions/fetch_resource_annotations.rb
@@ -6,9 +6,9 @@ module Authentication
     FetchResourceAnnotations = CommandClass.new(
       dependencies: {
         role_class: ::Role,
-        resource_class: ::Resource,
+        resource_class: ::Resource
       },
-      inputs: %i[account role_name]
+      inputs: %i[account role_name ignore_empty_annotations]
     ) do
 
       def call
@@ -25,20 +25,22 @@ module Authentication
         resource.annotations.each_with_object({}) do |annotation, result|
           annotation_values = annotation.values
           value = annotation_values[:value]
-          next if value.blank?
+          next if value.blank? and @ignore_empty_annotations
 
           result[annotation_values[:name]] = value
         end
       end
 
       def resource
-        # Validate role exists, otherwise getting role annotations return empty hash.
-        role_id = @role_class.roleid_from_username(@account, @role_name)
-        resource = @resource_class[role_id]
+        return @resource if @resource
 
-        raise Errors::Authentication::Security::RoleNotFound, role_id unless resource
+        @resource = @resource_class[role_id]
+        raise Errors::Authentication::Security::RoleNotFound, role_id unless @resource
+        @resource
+      end
 
-        resource
+      def role_id
+        @role_id ||= @role_class.roleid_from_username(@account, @role_name)
       end
 
     end
--- a/app/domain/authentication/resource_restrictions/fetch_resource_annotations.rb
+++ b/app/domain/authentication/resource_restrictions/fetch_resource_annotations.rb
@@ -1,0 +1,46 @@
+require 'command_class'
+
+module Authentication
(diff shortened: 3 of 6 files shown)


app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -109,7 +109,11 @@ def fetch_jwt_claims_to_validate
       def initialize_validate_restrictions
         @restriction_validator = Authentication::AuthnJwt::ValidateRestrictionsOneToOne
         @restrictions_from_annotations_class = Authentication::ResourceRestrictions::GetServiceSpecificRestrictionsFromAnnotation
-        @extract_resource_restrictions = Authentication::ResourceRestrictions::ExtractResourceRestrictions.new(get_restriction_from_annotation: @restrictions_from_annotations_class)
+        @fetch_resource_annotations = Authentication::ResourceRestrictions::FetchNonEmptyResourceAnnotations
```
orenbm, line 112: match param name with class name

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
```diff
@@ -7,6 +7,7 @@ module ResourceRestrictions
       dependencies: {
         resource_restrictions_class: Authentication::ResourceRestrictions::ResourceRestrictions,
         get_restriction_from_annotation: Authentication::ResourceRestrictions::GetRestrictionFromAnnotation,
+        fetch_resource_annotations: Authentication::ResourceRestrictions::FetchResourceAnnotations,
```
orenbm, line 10: the `.new` should be here

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
```diff
@@ -33,28 +34,10 @@ def call
...
-
-      def resource
-        # Validate role exists, otherwise getting role annotations return empty hash.
-        role_id = @role_class.roleid_from_username(@account, @role_name)
-        resource = @resource_class[role_id]
-
-        raise Errors::Authentication::Security::RoleNotFound, role_id unless resource
-
-        resource
+        @resource_annotations = @fetch_resource_annotations.new.call(
```
orenbm, line 37: change to
```
def fetch_resource_annotations
  resource_annotations
end

def resource_annotations
  @resource_annotations ||= @fetch_resource_annotations.new.call(
    account: @account,          
    role_name: @role_name        
  )
end
```

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
```diff
@@ -63,7 +46,7 @@ def extract_resource_restrictions_from_annotations
 
       def resource_restrictions_hash
         @resource_restrictions_hash ||=
-          resource_annotations.each_with_object({}) do |(annotation_name, annotation_value), resource_restrictions_hash|
+          @resource_annotations.each_with_object({}) do |(annotation_name, annotation_value), resource_restrictions_hash|
```
orenbm, line 49:
```suggestion
          resource_annotations.each_with_object({}) do |(annotation_name, annotation_value), resource_restrictions_hash|
```

orenbm: use memoized object

app/domain/authentication/resource_restrictions/fetch_non_empty_resource_annotations.rb
```diff
@@ -0,0 +1,46 @@
+require 'command_class'
+
+module Authentication
+  module ResourceRestrictions
+
+    FetchNonEmptyResourceAnnotations = CommandClass.new(
```
sashaCher, line 6: The name is confusing. Class does not fetch non empty annotations purely rather throws exception on empty annotation...

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
tzheleznyak: Done

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
tzheleznyak: Done

tzheleznyak: Done

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
tzheleznyak: Done

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
tzheleznyak: Done

app/domain/authentication/resource_restrictions/fetch_non_empty_resource_annotations.rb
tzheleznyak: You are right. But i cant think of better names. Any suggestions?

app/domain/errors.rb
```diff
@@ -496,6 +496,11 @@ module ResourceRestrictions
         code: "CONJ00049E"
       )
 
+      EmptyAnnotationGiven = ::Util::TrackableErrorClass.new(
+        msg: "Empty value for '{0-annotation-name}' has been given",
```
tzheleznyak, line 500: Hey @shulifink @hilagross 
Can i get review on this error message?

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -109,7 +109,11 @@ def fetch_jwt_claims_to_validate
       def initialize_validate_restrictions
         @restriction_validator = Authentication::AuthnJwt::ValidateRestrictionsOneToOne
         @restrictions_from_annotations_class = Authentication::ResourceRestrictions::GetServiceSpecificRestrictionsFromAnnotation
-        @extract_resource_restrictions = Authentication::ResourceRestrictions::ExtractResourceRestrictions.new(get_restriction_from_annotation: @restrictions_from_annotations_class)
+        @fetch_non_empty_resource_annotations = Authentication::ResourceRestrictions::FetchNonEmptyResourceAnnotations
```
orenbm, line 112: should this be `Authentication::ResourceRestrictions::FetchNonEmptyResourceAnnotations.new`?

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
orenbm: well this should be fixed everywhere. when a command class is a dependency then we put the `.new` in the dependency and not in the code itself so that it is initialized only once.

app/domain/authentication/resource_restrictions/fetch_non_empty_resource_annotations.rb
orenbm: i think it can still be `FetchResourceAnnotations` and raise an error in case an annotation is empty. For example, the `Authenticate` class does not allow wrong credentials and it is not called `AuthenticateCorrectCredentials`. The correct credentials is an implementation detail, the same as the non-empty annotations.

app/domain/authentication/resource_restrictions/fetch_non_empty_resource_annotations.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      def resource
+        # Validate role exists, otherwise getting role annotations return empty hash.
+        role_id = @role_class.roleid_from_username(@account, @role_name)
+        resource = @resource_class[role_id]
+
+        raise Errors::Authentication::Security::RoleNotFound, role_id unless resource
+
+        resource
+      end
```
orenbm, line 42:
```suggestion
      def resource
        return @resource if @resource
        
        @resource = @resource_class[role_id]
        raise Errors::Authentication::Security::RoleNotFound, role_id unless @resource
        @resource
      end
      
      def role_id
        @role_id ||= @role_class.roleid_from_username(@account, @role_name)
      end
```

app/domain/authentication/resource_restrictions/fetch_non_empty_resource_annotations.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      def fetch_resource_annotations
+        resource_annotations
+      end
+
+      def resource_annotations
+        resource.annotations.each_with_object({}) do |annotation, result|
+          annotation_values = annotation.values
+          value = annotation_values[:value]
+          raise Errors::Authentication::ResourceRestrictions::EmptyAnnotationGiven, annotation_values[:name] if value.blank?
```
orenbm, line 28: line too long guard imo

```suggestion
          if value.blank?
            raise Errors::Authentication::ResourceRestrictions::EmptyAnnotationGiven, annotation_values[:name]
          end
```

orenbm: also - i understand that this is the only change from the other class. I would leave it as one class and pass a param named `allow_empty_annotations` with default value `true`. this way you can call it with `false` in `authn-jwt`, leave the other calls as-is and it is very clear from reading the code that this is an option and that it allows by default.

orenbm (changes requested):
Added a suggestion on how to approach this.

app/domain/authentication/resource_restrictions/fetch_resource_annotations.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      def fetch_resource_annotations
+        resource_annotations
+      end
+
+      def resource_annotations
+        resource.annotations.each_with_object({}) do |annotation, result|
+          annotation_values = annotation.values
+          value = annotation_values[:value]
+          next if value.blank? and @ignore_empty_annotations
```
orenbm, line 28: i'm ok with this approach but why not raise an error? 
how about this:
```
        if value.blank?
          if @ignore_empty_annotations
            next
          else          
            raise Errors::Authentication::ResourceRestrictions::EmptyAnnotationGiven, restriction.name 
       end

```

orenbm: this way you don't need to handle it for jwt and we can utilize this in other places by just passing `ignore_empty_annotations: false`

app/domain/authentication/resource_restrictions/fetch_resource_annotations.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      def resource
+        # Validate role exists, otherwise getting role annotations return empty hash.
+        role_id = @role_class.roleid_from_username(@account, @role_name)
+        resource = @resource_class[role_id]
+
+        raise Errors::Authentication::Security::RoleNotFound, role_id unless resource
+
+        resource
+      end
```
orenbm, line 42:
```suggestion
      def resource
        return @resource if @resource

        @resource = @resource_class[role_id]
        raise Errors::Authentication::Security::RoleNotFound, role_id unless @resource
        @resource
      end
      
      def role_id
        @role_id ||= @role_class.roleid_from_username(@account, @role_name)
      end
```

app/domain/errors.rb
hilagross: I would write: Value '{0-annotation-name}' is empty
but would like to hear @shulifink

app/domain/authentication/resource_restrictions/fetch_resource_annotations.rb
tzheleznyak: Because in JWT we want to throw exception for annotation that is empty but relevant (according to its name) to the vendor (service id) trying to authenticate. The vendor check happens only after the extraction. So in JWT we will want to pass on the empty annotation in order to handle it later.

orenbm: ok thanks for clarifying

tzheleznyak: For the case of supporting third behaviour of throwing exception i think it better to inject another class doing it instead of featch_resource_annotations.

But i can also agree if you say for a const specifiying it that have three states (one for ignore, one for error, and one for passing it empty as is). In jwt i need the passing it as is to check later if exception need to be thrown (in the validator)

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
```diff
@@ -7,9 +7,11 @@ module ResourceRestrictions
       dependencies: {
         resource_restrictions_class: Authentication::ResourceRestrictions::ResourceRestrictions,
         get_restriction_from_annotation: Authentication::ResourceRestrictions::GetRestrictionFromAnnotation,
+        fetch_resource_annotations: Authentication::ResourceRestrictions::FetchResourceAnnotations,
```
sashaCher, line 10:
```suggestion
        fetch_resource_annotations_class: Authentication::ResourceRestrictions::FetchResourceAnnotations,
```

app/domain/authentication/resource_restrictions/fetch_non_empty_resource_annotations.rb
sashaCher: @orenbm What's the meaning of making `@resource` a class member?

sashaCher (approved):
Leaved one naming convention remark.
All other things are looking good for me.

app/domain/authentication/resource_restrictions/fetch_non_empty_resource_annotations.rb
orenbm: it's just for memoization. so you don't run the whole logic every time you call the `resource` method.

app/domain/errors.rb
shulifink, line 500:
```suggestion
        msg: "'{0-annotation-name}' has no value",
```

or 

```suggestion
        msg: "'{0-annotation-name}' is empty",
```

shulifink (changes requested):
small change

shulifink (changes requested):
see change

app/domain/authentication/authn_jwt/configuration_jwt_generic_vendor.rb
```diff
@@ -109,7 +109,10 @@ def fetch_jwt_claims_to_validate
       def initialize_validate_restrictions
         @restriction_validator = Authentication::AuthnJwt::ValidateRestrictionsOneToOne
         @restrictions_from_annotations_class = Authentication::ResourceRestrictions::GetServiceSpecificRestrictionsFromAnnotation
```
orenbm, line 111: this can be done in a separate effort but param names should match their class. And also command classes should be with `.new` in the dependencies

orenbm (approved):

PR #2217, opened 2021-05-26
Enabled authenticators can be configured using configuration file

### LINKED ISSUE #2173: Enabled authenticators can be configured using configuration file
Add an `:authenticators` attribute to the `Conjur::ConjurConfig` object. It should default to an empty string and should be validated to ensure it's a comma-separated list of strings starting w/ "authn-".

Update `Authentication::InstalledAuthenticators` to use read the authenticators attribute from `ConjurConfig` instead of the `CONJUR_AUTHENTICATORS` environment variable.

Tests
- update spec / integration tests for Authentication::InstalledAuthenticators

h-artzi:
### What does this PR do?
Add authenticators to Conjur Config

Resolves #2173

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Enabled authenticators can now be configured via a configuration file, or the
+  CONJUR_AUTHENTICATORS environment variable.
+  [cyberark/conjur##2173](https://github.com/cyberark/conjur/issues/#2173)
 - Trusted Proxies can now be configured with a configuration file or by setting
   the CONJUR_TRUSTED_PROXIES environment variable.
   [cyberark/conjur#2168](https://github.com/cyberark/conjur/issues/2168)
--- a/app/domain/authentication/installed_authenticators.rb
+++ b/app/domain/authentication/installed_authenticators.rb
@@ -33,21 +33,18 @@ module Authentication
           .push(::Authentication::Common.default_authenticator_name)
       end
 
-      def enabled_authenticators(env)
+      def enabled_authenticators
         # Enabling via environment overrides enabling via CLI
-        env_enabled_authenticators(env) || db_enabled_authenticators
+        authenticators = 
+          Rails.application.config.conjur_config.authenticators
+        authenticators.empty? ? db_enabled_authenticators : authenticators
       end
 
-      def enabled_authenticators_str(env)
-        enabled_authenticators(env).join(',')
+      def enabled_authenticators_str
+        enabled_authenticators.join(',')
       end
 
       private
-
-      def env_enabled_authenticators(env)
-        authenticators = env["CONJUR_AUTHENTICATORS"]
-        authenticators.present? ? authenticators.split(',') : nil
-      end
       
       def db_enabled_authenticators
         # Always include 'authn' when enabling authenticators via CLI so that it
--- a/lib/conjur/conjur_config.rb
+++ b/lib/conjur/conjur_config.rb
@@ -16,7 +16,8 @@ module Conjur
 
     attr_config(
       # Read TRUSTED_PROXIES before default to maintain backwards compatibility
-      trusted_proxies: (ENV['TRUSTED_PROXIES'] || [])
+      trusted_proxies: (ENV['TRUSTED_PROXIES'] || []),
+      authenticators: []
     )
 
     # Perform validations and collect errors then report results as exception
@@ -24,6 +25,7 @@ module Conjur
       invalid = []
 
       invalid << "trusted_proxies" unless trusted_proxies_valid?
+      invalid << "authenticators" unless authenticators_valid?
 
       unless invalid.empty?
         raise Errors::Conjur::InvalidConfigValues, invalid.join(', ')
@@ -32,7 +34,7 @@ module Conjur
 
     # Get attribute sources without including attribute values
     def attribute_sources
-      to_source_trace.map { |k,v| [ k.to_sym, v[:source][:type] ] }.to_h
+      to_source_trace.map { |k, v| [ k.to_sym, v[:source][:type] ] }.to_h
     end
 
     # The Anyway config gem automatically converts a comma-separated env var to
@@ -43,9 +45,15 @@ module Conjur
     #
     # To address this inconsistent behavior, we use custom setters to ensure
     # list attributes are always converted to a a list type.
+    # We filed an issue regarding this behavior:
+    #   https://github.com/palkan/anyway_config/issues/82
 
     def trusted_proxies=(val)
-      super(str_to_list(val))
+      super(str_to_list(val).uniq)
+    end
+
+    def authenticators=(val)
+      super(str_to_list(val).uniq)
     end
 
     private
@@ -60,6 +68,19 @@ module Conjur
       end
 
       true
+    rescue
+      false
+    end
+
+    def authenticators_valid?
+      # TODO: Ideally we would check against the enabled authenticators
+      # in the DB. However, we need to figure out how to use code from the
+      # application without introducing warnings.
+      authenticators_regex = 
+        %r{^(authn|authn-(k8s|oidc|iam|ldap|gcp|azure)(/.+)?)$}
+      authenticators.all? do |authenticator|
+        authenticators_regex.match?(authenticator.strip)
+      end
     rescue
       false
     end
--- a/lib/conjur/conjur_config.rb
+++ b/lib/conjur/conjur_config.rb
@@ -49,11 +49,11 @@ module Conjur
     #   https://github.com/palkan/anyway_config/issues/82
 
     def trusted_proxies=(val)
-      super(str_to_list(val).uniq)
+      super(str_to_list(val)&.uniq)
     end
 
     def authenticators=(val)
-      super(str_to_list(val).uniq)
+      super(str_to_list(val)&.uniq)
     end
 
     private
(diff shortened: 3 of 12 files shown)


lib/conjur/conjur_config.rb
```diff
@@ -63,5 +71,15 @@ def trusted_proxies_valid?
     rescue
       false
     end
+
+    def authenticators_valid?
+      authenticators.each do |authenticator|
+        return false unless /^(authn|authn-.+)$/.match?(authenticator.strip)
+      end
+      
+      true
```
micahlee, line 80: You should be able to replace this with `#all?`: https://apidock.com/ruby/Enumerable/all%3F

```suggestion
      authenticators.all? do |authenticator|
        /^(authn|authn-.+)$/.match?(authenticator.strip)
      end
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Enabled authenticators can now be configured via a configuration file, the
```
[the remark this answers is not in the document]
micahlee: We should clean this up.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Enabled authenticators can now be configured via a configuration file, the 
+  CONJUR_AUTHENTICATORS environment variable, or the authenticator allowlist
```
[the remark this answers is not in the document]
micahlee: We should clean this up.

micahlee (commented):
Looks good! I left a couple of small comments for you. Would you also mind updating the PR title to something descriptive?

Thanks!

app/domain/authentication/installed_authenticators.rb
```diff
@@ -33,20 +33,20 @@ def configured_authenticators
...
+      def enabled_authenticators_str
+        enabled_authenticators.join(',')
       end
 
       private
 
-      def env_enabled_authenticators(env)
-        authenticators = env["CONJUR_AUTHENTICATORS"]
-        authenticators.present? ? authenticators.split(',') : nil
+      def env_enabled_authenticators
```
jtuttle, line 47: I would call this `configured_authenticators` now.

app/domain/authentication/installed_authenticators.rb
```diff
@@ -33,20 +33,20 @@ def configured_authenticators
...
+        authenticators = Conjur::ConjurConfig.new.authenticators.uniq
```
jtuttle, line 48: Please use `Rails.application.config.conjur_config.authenticators` within the application code. It gets loaded once when the app starts instead of loading configuration fresh each time ConjurConfig is instantiated.

lib/conjur/conjur_config.rb
```diff
@@ -32,7 +34,7 @@ class ConjurConfig < Anyway::Config
 
     # Get attribute sources without including attribute values
     def attribute_sources
-      to_source_trace.map { |k,v| [ k.to_sym, v[:source][:type] ] }.to_h
+      to_source_trace.map { |k, v| [ k.to_sym, v[:source][:type] ] }.to_h
```
jtuttle, line 37: Weird, I had it like this originally but CodeClimate complained.

spec/lib/conjur/conjur_config_spec.rb
```diff
@@ -6,8 +6,8 @@
   end
 
   it "reports the attribute source as :defaults" do
-    expect(Conjur::ConjurConfig.new.attribute_sources[:trusted_proxies]).
-      to eq(:defaults)
+    expect(Conjur::ConjurConfig.new.attribute_sources[:trusted_proxies])
+      .to eq(:defaults)
```
jtuttle, line 10: I think we use trailing dots more often than leading so can we change these back?

app/domain/authentication/installed_authenticators.rb
h-artzi: configured_authenticators is already used

jtuttle: Hah, okay. In that case (and knowing that we're going to tear out the DB-configured authenticators sooni-ish) I would get rid of this helper method and just build the logic into the `enabled_authenticators` method.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
+- Enabled authenticators can now be configured via a configuration file, the
+  CONJUR_AUTHENTICATORS environment variable, or the authenticator allowlist
+  API.
+  [cyberark/conjur##2173](https://github.com/cyberark/conjur/issues/#2173)
```
orenbm, line 12: @h-artzi i'm probably missing something but i couldn't find in the code how authenticators can still be enabled by the CONJUR_AUTHENTICATORS environment variable. can you point me to it?

also - i am not sure that the authenticator allowlist  API is GA so not sure that it should be in the changelog. @alexkalish wdyt?

h-artzi: The [anyway_config](https://github.com/palkan/anyway_config#environment-variables) gem takes care of the usage of the environment variable CONJUR_AUTHENTICATORS.

app/domain/authentication/installed_authenticators.rb
```diff
@@ -33,21 +33,18 @@ def configured_authenticators
           .push(::Authentication::Common.default_authenticator_name)
       end
 
-      def enabled_authenticators(env)
+      def enabled_authenticators
         # Enabling via environment overrides enabling via CLI
-        env_enabled_authenticators(env) || db_enabled_authenticators
+        authenticators = 
+          Rails.application.config.conjur_config.authenticators.uniq
```
jvanderhoof, line 39: What do you think about moving the `uniq` call to the `authenticators` method in the config?  (We may need to create a new method that extends the parent class behaviour).  To me, it feels out of place here to "fix-up" config.

jvanderhoof: We might also want to strip empty/nil items from the array.

app/domain/authentication/installed_authenticators.rb
```diff
@@ -33,21 +33,18 @@ def configured_authenticators
...
+        authenticators.empty? ? db_enabled_authenticators : authenticators
```
jvanderhoof, line 40: I rarely feel like the ternary operator reads well.  What do you think about something like:
```ruby
def enabled_authenticators
  Rails.application.config.conjur_config.authenticators.select { |e| !e.to_s.empty? } ||
  db_enabled_authenticators
end
```

lib/conjur/conjur_config.rb
```diff
@@ -63,5 +71,13 @@ def trusted_proxies_valid?
     rescue
       false
     end
+
+    def authenticators_valid?
```
jvanderhoof, line 75: What do you think about checking for a valid authenticator (`authn-k8s`, `authn-ldap`, `authn-oidc`, etc.) rather than just the form of a possible authenticator?

app/domain/authentication/installed_authenticators.rb
jtuttle: I think I prefer the ternary in this case. Having a negated `select` that calls several methods on the elements it's selecting hurts my head more than the ternary does. 😄

jtuttle: Also note that we are hopefully going to tear out db_enabled_authenticators soon-ish so whatever we choose here is going to be a temporary inconvenience.

jonahx: I personally think the ternary is clear here, but here's a non-ternary alternative which does the same thing:

```rb
authenticators = 
  Rails.application.config.conjur_config.authenticators.uniq
return authenticators if authenticators

db_enabled_authenticators
```

spec/lib/conjur/conjur_config_spec.rb
jonahx: Dissenting vote :)

Consider:

```rb
        Resource
          .where(identifier.like("#{AUTHN_RESOURCE_PREFIX}%"))
          .where(kind => "webservice")
          .select_map(identifier)
          .map { |id| id[%r{^conjur/(authn(?:-[^/]+)?(?:/[^/]+)?)$}, 1] } # filter out nested status webservice
          .compact
          .push(::Authentication::Common.default_authenticator_name)
```
leading dot does two things:
1. Makes clear at a glance that we're dealing with a continuation (otherwise, if you don't notice the dot at the end of the previous line, it could be a new method call on `self`)
2. Makes cut and paste more ergonomic (otherwise the final item is a special case).

lib/conjur/conjur_config.rb
```diff
@@ -63,5 +71,18 @@ def trusted_proxies_valid?
...
+      # TODO: Ideally we would check against the enabled authenticators
+      # in the DB. However, we need to figure out how to use code from the
+      # application without introducing warnings.
+      authenticators_regex = 
+        %r{^(authn|authn-(k8s|oidc|iam|ldap|gcp|azure|config)(/.+)?)$}
```
jtuttle, line 80: You can remove `config` from here now, right?

jtuttle (changes requested):

lib/conjur/conjur_config.rb
h-artzi: great catch!

jtuttle (approved):

PR #2218, opened 2021-05-26
Remove the Conjur API dependency from the policy cucumber tests

### LINKED ISSUE #559: Tests should not depend on the Conjur API Ruby gem
The Policy cucumber tests currently use the Conjur API Ruby gem to implement the tests. This introduces a circular dependency where some changes to the API have to be made in both this project and the ruby client in order for Conjur's tests to pass.

Consider removing this dependency and using direct HTTP calls in the tests, for example.

mFelgate:
### What does this PR do?
The change removes the Conjur Ruby API dependency from the policy and rotator cucumber tests. Instead, the cucumber test access Conjur directly through its endpoints. 
To make sure the code is up to the standard set by the team.
No
Resolves #559

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/cucumber/policy/features/support/policy_helpers.rb
+++ b/cucumber/policy/features/support/policy_helpers.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module FullId
-  def make_full_id id, account: Conjur.configuration.account
+  def make_full_id(id, account: "cucumber")
     tokens  = id.split(":", 3)
     prepend = tokens.size == 2 ? [account] : []
     (prepend + tokens).join(':')
@@ -20,7 +20,7 @@ module PolicyHelpers
   def invoke status: nil, &block
     begin
       @result = yield
-      raise "Expected invocation to be denied" if status && status != 200
+      # raise "Expected invocation to be denied" if status && status != 200
 
       @result.tap do |result|
         puts(result) if @echo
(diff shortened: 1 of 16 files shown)


cucumber/rotators/features/password.feature
```diff
@@ -28,6 +28,6 @@ Feature: Postgres password rotation
   #       3 seconds of time.
   #
   Scenario: Values are rotated according to the policy
-    Given I moniter "db-reports/password" and db user "test" for 3 values in 20 seconds
+    Given I moniter "db-reports/password" and db user "test" for 3 values in 60 seconds
```
jonahx, line 31: This one should be part of the main commit.  But it's not super important.

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module FullId
-  def make_full_id id, account: Conjur.configuration.account
+  def make_full_id id, account: "cucumber"
```
jonahx, line 4: missing parens.

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -32,68 +32,127 @@ def invoke status: nil, &block
   end
 
   def load_root_policy policy
-    conjur_api.load_policy("root", policy, method: Conjur::API::POLICY_METHOD_PUT)
+    resource('root').put(policy, header)
```
jonahx, line 35: parens again.  looks like there's a number of these.

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -32,68 +32,127 @@ def invoke status: nil, &block
...
 
-  def extend_policy id, policy
-    conjur_api.load_policy(id, policy, method: Conjur::API::POLICY_METHOD_POST)
+  def extend_policy(id, policy)
+    resource(id).post(policy, header)
   end
 
-  def make_full_id *tokens
-    super(tokens.join(":"))
+  def create_api_key role
```
jonahx, line 58: parens again.

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -32,68 +32,127 @@ def invoke status: nil, &block
...
+    )
+  end
+
+  def admin_api_key
+    admin_resource.get
+  end
+
+  def get_login_token login, key
+    RestClient.post(
+      uri('authn', CGI.escape(login), 'authenticate'),
```
jonahx, line 70: I would move `CGI.escape` inside `uri`, so that it only exists in one place and client code doesn't have to worry about it.

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -32,68 +32,127 @@ def invoke status: nil, &block
...
+  end
+
+  def login_resource
+    RestClient::Resource.new(
+      uri('authn', 'api_key', ''), 'admin', admin_password
+    )
+  end
+
+  def make_full_id *tokens
+    super(tokens.join(":"))
```
jonahx, line 102: I think this is confusing.  We have two methods of the same name, doing similar things, but with different signatures, and one over-riding the other and calling `super` on it.  

There is a larger point here that we want to avoid this kind of mixin programming too, but we don't need to address that yet since you'll be rewriting much of this.  But just keep that in the back of your mind.

Also, a point on naming: the `make_` here is superfluous (as it is in the existing method name in `FullId`).  To avoid that and to fix the confusion problem I mentioned above, I'd suggest:

`full_id` and `full_id_from_tokens`

The first change from `make_full_id` to `full_id` you don't need to make now if you don't want, but lets at least give the new one a new name.

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -32,57 +32,100 @@ def invoke status: nil, &block
...
-      JSON.parse(@result.to_json)
     when Hash
       @result
     end
   end
 
-  def admin_api_key
-    @admin_api_key ||= Conjur::API.login('admin', admin_password)
+  def create_token_header token = nil
+    if token.nil?
```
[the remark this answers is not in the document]
jonahx: Parens again.

Also this whole thing can fit on one line, and you can also do:
```rb
token ||= admin_token
```

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -32,68 +32,127 @@ def invoke status: nil, &block
...
+      token = admin_token
+    end
+    %Q(Token token="#{token}")
+  end
+
+  def header token = nil
+    { Authorization: create_token_header(token) }
+  end
```
jonahx, line 123: Do we need both `header` and `create_token_header`?  It seems like `header` is the only one you'd use at the end of the day?

Also re: naming, similar to my comment above: `create_` is not needed.

cucumber/policy/features/support/resource_helpers.rb
```diff
@@ -0,0 +1,34 @@
...
+    client(uri('secrets', kind, id)).get(header(token))
+  end
+
+  def get_roles(kind, id)
+    client(uri('roles', kind, id)).get(header)
+  end
+
+  def client url
+    RestClient::Resource.new(url, 'Content-Type' => 'application/json')
+  end
```
jonahx, line 30: What's the reason we use `client` in some of the helpers, but in others we call `RestClient` directly?  This is another one we don't necessarily have to fix on this PR, but it struck me as inconsistent.

cucumber/policy/features/support/policy_helpers.rb
mFelgate: hmm yeah, the only time I don't use the header function but use "create_token_header"  is when I also use params in the request, which happens I think twice. in those cases I write out both in one hash. though there is probably a simpler way to handle both cases in one function

jonahx (approved):

PR #2222, opened 2021-05-27
Refactor authn jwt directories

tzheleznyak:
Reorder of the directories in the authn-jwt module and its UT

These is the explanation of each folder

identity_providers - retrieving the authenticating identity from the request. Currently from url provider and from decoded token provider
input_validation - responsible for first steps of validating the input of the authenticators
signing_key - all classes responsible for fetching signing key from the provider. Currently jwks and provider-uri
validate_and_decode - classes responsible for validating the jwt token and the mandatory claims there
vendor configurations - configurations of different jwt vendors. Currently there is just one for the generic vendor.

image


app/domain/authentication/authn_jwt/consts.rb
```diff
@@ -2,16 +2,16 @@
...
-    JWKS_URI_RESOURCE_NAME = "jwks-uri".freeze
-    ISSUER_RESOURCE_NAME = "issuer".freeze
-    IDENTITY_FIELD_VARIABLE = "token-app-property".freeze
-    IDENTITY_NOT_RETRIEVED_YET = "Identity not retrieved yet".freeze
-    PRIVILEGE_AUTHENTICATE="authenticate".freeze
-    ISS_CLAIM_NAME = "iss".freeze
-    EXP_CLAIM_NAME = "exp".freeze
-    NBF_CLAIM_NAME = "nbf".freeze
-    IAT_CLAIM_NAME = "iat".freeze
+    PROVIDER_URI_RESOURCE_NAME = "provider-uri"
```
nessiLahav, line 5: Without the freeze is not a real const, can we return it ?

nessiLahav (changes requested):
just move the `validate_restrictions_one_to_one` to new `validate_restrictions` directory

nessiLahav (approved):

PR #2230, opened 2021-05-31
Jwt dynamic jwks uri automation

sashaCher:
### What does this PR do?

Adds dynamic `jwks-uri` variable scenario.

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/cucumber/authenticators_jwt/features/authn_jwt.feature
+++ b/cucumber/authenticators_jwt/features/authn_jwt.feature
@@ -1,0 +1,62 @@
+Feature: JWT Authenticator - JWKs Basic sanity
+
+  In this feature we define a JWT authenticator in policy and perform authentication
+  with Conjur. In successful scenarios we will also define a variable and permit the host to
+  execute it.
+
+  Background:
+    Given I initialize JWKs endpoint with file "myJWKs.json"
+    And I load a policy:
+    """
+    - !policy
+      id: conjur/authn-jwt/raw
+      body:
+      - !webservice
+        annotations:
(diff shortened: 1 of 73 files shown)


cucumber/authenticators_jwt/features/authn_jwt.feature
```diff
@@ -38,25 +38,56 @@ Feature: JWT Authenticator - JWKs Basic sanity
...
+    When I authenticate via authn-jwt with the JWT token issued by "first.json" key
+    Then it is unauthorized
+    And The following appears in the audit log after my savepoint:
+    """
+    CONJ00035E
+    """
+
+    Given I successfully set authn-jwt jwks-uri variable with value of "first.json" endpoint
+    When I authenticate via authn-jwt with the JWT token issued by "first.json" key
+    Then host "myapp" has been authorized by Conjur
```
eladkug, line 93: Some comments here:
1. Looks like we have 2 different tests here and i think its better to separate between them. **Why?** If the first scenario success and the second one fail it doesnt mean the dynamic is not working.
2. Given is relevant only before we start the test, if you want to verify more issues it will be with `When` after the `Then`.  **Why?** If you have failure on `Given` it means something was failed before we start the test , with pre configure. If something was failed during the `When` it means during the test it self it failed. If something was failed during `Then` it means you didnt get what you want
3. I always prefer to verify the full audit log . **Why?** If we change the log i want the test will fail if we not fix it, and its more readable to the user while reading the test

tzheleznyak:
Will be solved in another PR including provider uri and straight to the master

PR #2228, opened 2021-05-31
Make authentication_parameters be a member of vendor configuration

tzheleznyak:
Make authentication_parameters be a member of vendor configuration

Makes order in absraction levels between authenticator.rb and vendor configuration. In addition it allows to remove jwt_identity accessor from the authentication parameters

PR #2232, opened 2021-06-01
PR fixes for jwt authenticator

tzheleznyak:
1. Small refactors according to here https://github.com/cyberark/conjur/pull/2228
2. Refactor in dependencies of command classes related to validate resource restrictions. Dependencies that are command classes them selves will be initialized in the dependency injection itself
3. Bugfix to audit failiure when there is still no identity. Caused because of moving to memoized get identity function.

### DIFF
--- a/app/domain/authentication/authn_jwt/authenticator.rb
+++ b/app/domain/authentication/authn_jwt/authenticator.rb
@@ -40,10 +40,10 @@ module Authentication
       end
 
       def get_jwt_identity
-        # Will be changed when real get identity implemented
         @logger.debug(LogMessages::Authentication::AuthnJwt::GET_JWT_IDENTITY.new)
         jwt_identity
         @logger.debug(LogMessages::Authentication::AuthnJwt::FOUND_JWT_IDENTITY.new(jwt_identity))
+        @jwt_identity_initialized = true
       end
 
       def jwt_identity
@@ -99,14 +99,8 @@ module Authentication
       end
 
       def role
-        if jwt_identity
-          return @role_class.by_login(
-            jwt_identity,
-            account: account
-          )
-        end
         @role_class.by_login(
-          "",
+          audited_username,
           account: account
         )
       end
@@ -115,8 +109,15 @@ module Authentication
         ::Audit::Event::Authn::RoleId.new(
           role: role,
           account: account,
-          username: jwt_identity
+          username: audited_username
         ).to_s
+      end
+
+      def audited_username
+        if @jwt_identity_initialized
+          return jwt_identity
+        end
+        NOT_INITIALIZED_IDENTITY
       end
 
       def webservice
--- a/app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
+++ b/app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
@@ -11,7 +11,7 @@ module Authentication
           @logger = Rails.logger
         end
 
-        def provide_jwt_identity
+        def jwt_identity
           token_field_name = fetch_token_field_name
           @logger.debug(LogMessages::Authentication::AuthnJwt::CHECKING_IDENTITY_FIELD_EXISTS.new(token_field_name))
           jwt_identity = @decoded_token[token_field_name]
--- a/app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
+++ b/app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
@@ -12,15 +12,17 @@ module Authentication
         end
 
         def jwt_identity
+          return @jwt_identity if @jwt_identity
+
           token_field_name = fetch_token_field_name
           @logger.debug(LogMessages::Authentication::AuthnJwt::CHECKING_IDENTITY_FIELD_EXISTS.new(token_field_name))
-          jwt_identity = @decoded_token[token_field_name]
-          if jwt_identity.blank?
+          @jwt_identity ||= @decoded_token[token_field_name]
+          if @jwt_identity.blank?
             raise Errors::Authentication::AuthnJwt::NoSuchFieldInToken, token_field_name
           end
 
           @logger.debug(LogMessages::Authentication::AuthnJwt::FOUND_JWT_FIELD_IN_TOKEN.new(token_field_name, jwt_identity))
-          jwt_identity
+          @jwt_identity
         end
 
         # Checks if variable that defined from which field in decoded token to get the id is configured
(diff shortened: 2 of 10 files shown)


app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
```diff
@@ -11,7 +11,7 @@ def initialize(authentication_parameters)
           @logger = Rails.logger
         end
 
-        def provide_jwt_identity
+        def jwt_identity
```
orenbm, line 14: you can change the method to:
```
        def jwt_identity
          return @jwt_identity if @jwt_identity

          token_field_name = fetch_token_field_name          @logger.debug(LogMessages::Authentication::AuthnJwt::CHECKING_IDENTITY_FIELD_EXISTS.new(token_field_name))          @jwt_identity = @decoded_token[token_field_name]          if @jwt_identity.blank?            raise Errors::Authentication::AuthnJwt::NoSuchFieldInToken, token_field_name          end
          @logger.debug(LogMessages::Authentication::AuthnJwt::FOUND_JWT_FIELD_IN_TOKEN.new(token_field_name, @jwt_identity))          @jwt_identity        end
```

to memoize it

app/domain/authentication/authn_jwt/vendor_configurations/configuration_jwt_generic_vendor.rb
```diff
@@ -5,29 +5,27 @@ module VendorConfigurations
...
             authenticator_name: @authentication_parameters.authenticator_name,
             service_id: @authentication_parameters.service_id,
             account: @authentication_parameters.account,
             role_name: jwt_identity,
-            constraints: @constraints,
-            authentication_request: @restriction_validator.new(
+            constraints: constraints,
+            authentication_request: restriction_validator_class.new(
               decoded_token: @authentication_parameters.decoded_token
             )
```
orenbm, line 30: is this correct? why does the `restriction_validator_class` return an `authentication_request` object?

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
```diff
@@ -6,8 +6,8 @@ module ResourceRestrictions
     ExtractResourceRestrictions = CommandClass.new(
       dependencies: {
         resource_restrictions_class: Authentication::ResourceRestrictions::ResourceRestrictions,
-        get_restriction_from_annotation: Authentication::ResourceRestrictions::GetRestrictionFromAnnotation,
-        fetch_resource_annotations_class: Authentication::ResourceRestrictions::FetchResourceAnnotations,
+        get_restriction_from_annotation: Authentication::ResourceRestrictions::GetRestrictionFromAnnotation.new,
+        fetch_resource_annotations_class: Authentication::ResourceRestrictions::FetchResourceAnnotations.new,
```
orenbm, line 10: you can remove the `_class` suffix

orenbm (changes requested):
left a few comments

app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
tzheleznyak: ValidateResourceRestrictions recivies the validator as authentication_request
https://github.com/cyberark/conjur/blob/ba3a03e841eebe427b340bda5cf5983be1029f6e/app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb#L28

I also don't agree with the naming but don't think its in scope for the jwt authenticator to fix this for all authenticators.

app/domain/authentication/authn_jwt/vendor_configurations/configuration_jwt_generic_vendor.rb
tzheleznyak: ValidateResourceRestrictions recivies the validator as authentication_request
https://github.com/cyberark/conjur/blob/ba3a03e841eebe427b340bda5cf5983be1029f6e/app/domain/authentication/resource_restrictions/validate_resource_restrictions.rb#L28

I also don't agree with the naming but don't think its in scope for the jwt authenticator to fix this for all authenticators.

app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
tzheleznyak: Done

app/domain/authentication/resource_restrictions/extract_resource_restrictions.rb
tzheleznyak: Done but renamed first function in call to call_fetch_resource_annotations

app/domain/authentication/authn_jwt/authenticator.rb
```diff
@@ -115,10 +116,17 @@ def audit_role_id
         ::Audit::Event::Authn::RoleId.new(
           role: role,
           account: account,
-          username: jwt_identity
+          username: audited_username
         ).to_s
       end
 
+      def audited_username
```
nessiLahav, line 123: Please validate with PO and in our Code (in other flows i think we had the same issue in OIDC) what we writing to the audit when the user is not exists, i am not sure empty string is the answer

nessiLahav (changes requested):
left comments

app/domain/authentication/authn_jwt/authenticator.rb
tzheleznyak: Already discussed this with @hilagross and its okay for our authenticator. The username is the identity. No identity no username its seems legit

tzheleznyak: There will be enough information in the message of the audit to understand this

app/domain/authentication/authn_jwt/authenticator.rb
```diff
@@ -40,10 +40,10 @@ def validate_and_decode_token
       end
 
       def get_jwt_identity
-        # Will be changed when real get identity implemented
         @logger.debug(LogMessages::Authentication::AuthnJwt::GET_JWT_IDENTITY.new)
         jwt_identity
         @logger.debug(LogMessages::Authentication::AuthnJwt::FOUND_JWT_IDENTITY.new(jwt_identity))
+        @jwt_identity_initialized = true
```
nessiLahav, line 46: WDYT instead of using this bool to initialize the `@jwt_identity` at the start of this class with decided const on an "empty user/ not identified/ not initilized "

and before writing the audit check if its value was updated to a real value

nessiLahav: @tzheleznyak also can you check if we have this flow in the test plan of failing before identified the user  and to check the audit message ?

nessiLahav (changes requested):

app/domain/authentication/authn_jwt/authenticator.rb
nessiLahav: + @sashaCher WDYtT we should write in audit when we do not have the user value ?

app/domain/authentication/authn_jwt/authenticator.rb
tzheleznyak: Talked about this. Moved to const so we will be able to configure the empty username later.
I will stay with the boolean

nessiLahav (approved):
LGTM

PR #2227, opened 2021-05-30
Add JWT authenticator

tzheleznyak:
### What does this PR do?
Adds JWT Authenticator to Conjur. Currently supports generic authenticator.
Read the design here https://github.com/cyberark/conjur/blob/master/design/authenticators/authn_jwt/authn_jwt_solution_design.md

### What ticket does this PR close?
Resolves https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-7022

#### Change log
- [ ] The CHANGELOG has been updated, or
- [x] This PR does not include user-facing changes and doesn't require a CHANGELOG update - Changelog need to be added

#### Test coverage
- [x] This PR includes new unit and integration tests to go with the code changes, or
- [ ] The changes in this PR do not require tests

#### Documentation
- [x] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or
- [ ] This PR does not require updating any documentation

#### API Changes
- [x] The [OpenAPI spec](https://github.com/cyberark/conjur-openapi-spec) has been updated to meet new API changes (or an issue has been opened), or
- [ ] The changes in this PR do not affect the Conjur API

PR #2233, opened 2021-06-01
Fix code climate style comments in JWT Authenticator

tzheleznyak:
In this PR we got 170 Code Climate Issues 
https://github.com/cyberark/conjur/pull/2227

Here in this PR i done refactore that  reduced errors from 170 to 105. Many of them just with rubocop -a command.

This issues need to be marked as won't fix 

Whats left are and need to be marked as won't fix. Maybe missed some in complexity but wrote the main ones.
**9 Style Issues**

- 4 Issues with Of them are because prefix get_  but this is our convention for function in command classes.
- 4 Issues with Of them is because we dont use super  in classes that inherit. But they inherit interface so there is nothing to to in constructor. Talked with @Avi-Eyal  and @nessiLahav  about this
- 1 Redundant else-clause : Its in switch case 

**96 Complexity Issues**

- 1 Issues with lines of code in logs.rb . This is logs file what can we do here
- 2 Issues : ConfigurationJWTGenericVendors and AuthenticationController exceed 20 methods. We cant do anything about it many one line functions needed for memoization and readbility. 
- 3 Issues : flog Complex Method. Don't even know what they want
- 7 Issues : Class doesn't depend on instance state. They are all in spec helpers and they need to be this way to work in the test step
- 2 Issues of attr_accessor in AuthenticationParameters needed for tests and deligators. Tried to remove but didn't found way.
- 9 has at least X instance variables. This classes are like AuthenticationParameters and they need to have this number of members. In some we store logger as member for readbility . We cant reduce number of members there
- 1 Issue of block has to many lines is authenticator.rb. Nothing we cant do about it.
- 25 Issues of not descriptive comments. It wants comments on Modules or Interfaces or classes that are self explanatory and it would be shame to add comment there 
- 2 Places with "manually dispatches method call"  ???
- 6 Issues of "assumes too much for instance variable" but @Nessi Lahav @Oren Ben Meir we could take a look.
- 10 Issues of "performs a nil-check". The nil? function helps readbility and its a must in most places it is used
- 7 Issues of "Has unused parameter in constructor" . This is in mocks and interfaces what can we do?
- 4 Issues of "controlled by argument". Not sure what it means or if we can do anything but we can check ???
- 1 Issues of has boolean parameter in Authentication::AuthnJwt::ValidateAndDecode#fetch_signing_key. It just passes it.

### DIFF
--- a/app/domain/authentication/authn_jwt/consts.rb
+++ b/app/domain/authentication/authn_jwt/consts.rb
@@ -2,22 +2,24 @@
 
 module Authentication
   module AuthnJwt
-    PROVIDER_URI_RESOURCE_NAME = "provider-uri".freeze
-    JWKS_URI_RESOURCE_NAME = "jwks-uri".freeze
-    ISSUER_RESOURCE_NAME = "issuer".freeze
-    IDENTITY_FIELD_VARIABLE = "token-app-property".freeze
-    IDENTITY_NOT_RETRIEVED_YET = "Identity not retrieved yet".freeze
-    PRIVILEGE_AUTHENTICATE="authenticate".freeze
-    NOT_INITIALIZED_IDENTITY="".freeze
-    ISS_CLAIM_NAME = "iss".freeze
-    EXP_CLAIM_NAME = "exp".freeze
-    NBF_CLAIM_NAME = "nbf".freeze
-    IAT_CLAIM_NAME = "iat".freeze
+    PROVIDER_URI_RESOURCE_NAME = "provider-uri"
+    JWKS_URI_RESOURCE_NAME = "jwks-uri"
+    ISSUER_RESOURCE_NAME = "issuer"
+    IDENTITY_FIELD_VARIABLE = "token-app-property"
+    IDENTITY_NOT_RETRIEVED_YET = "Identity not retrieved yet"
+    PRIVILEGE_AUTHENTICATE="authenticate"
+    NOT_INITIALIZED_IDENTITY="not initialized user"
+    ISS_CLAIM_NAME = "iss"
+    EXP_CLAIM_NAME = "exp"
+    NBF_CLAIM_NAME = "nbf"
+    IAT_CLAIM_NAME = "iat"
     RSA_ALGORITHMS = %w[RS256 RS384 RS512].freeze
     ECDSA_ALGORITHMS = %w[ES256 ES384 ES512].freeze
-    SUPPORTED_ALGORITHMS = (RSA_ALGORITHMS + ECDSA_ALGORITHMS).freeze
-    CACHE_REFRESHES_PER_INTERVAL = 10.freeze
-    CACHE_RATE_LIMIT_INTERVAL = 300.freeze # 300 seconds (every 5 mins)
-    CACHE_MAX_CONCURRENT_REQUESTS = 3.freeze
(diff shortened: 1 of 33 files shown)


app/domain/authentication/authn_jwt/restriction_validators/validate_restrictions_one_to_one.rb
```diff
@@ -9,14 +9,16 @@ def initialize(decoded_token:)
         end
 
         def valid_restriction?(restriction)
```
nessiLahav, line 11: there is any possibility that the restriction will be blank and we will get nil function exception ?

app/domain/authentication/authn_jwt/signing_key/fetch_jwks_uri_signing_key.rb
```diff
@@ -15,15 +15,16 @@ def initialize(authentication_parameters:,
...
           @resource_class = resource_class
+          @resource_id = authentication_parameters.authenticator_resource_id
           @http_lib = http_lib
           @create_jwks_from_http_response = create_jwks_from_http_response
         end
 
-        def has_valid_configuration?
-          @jwks_uri_resource_exists ||= jwks_uri_resource_exists?
+        def jwks_uri_resource_exists
+          @jwks_uri_resource_exists = jwks_uri_resource_exists?
```
nessiLahav, line 27: why we deleted the memoization ?

app/domain/authentication/authn_jwt/consts.rb
```diff
@@ -2,22 +2,24 @@
...
-    EXP_CLAIM_NAME = "exp".freeze
-    NBF_CLAIM_NAME = "nbf".freeze
-    IAT_CLAIM_NAME = "iat".freeze
+    PROVIDER_URI_RESOURCE_NAME = "provider-uri"
+    JWKS_URI_RESOURCE_NAME = "jwks-uri"
+    ISSUER_RESOURCE_NAME = "issuer"
+    IDENTITY_FIELD_VARIABLE = "token-app-property"
+    IDENTITY_NOT_RETRIEVED_YET = "Identity not retrieved yet"
+    PRIVILEGE_AUTHENTICATE="authenticate"
+    NOT_INITIALIZED_IDENTITY="not initialized user"
```
nessiLahav, line 11: Validate if we already have this functionality in code (look at the file `role_id.rb`)

nessiLahav (changes requested):
Few comments, great work 👍

app/domain/authentication/authn_jwt/restriction_validators/validate_restrictions_one_to_one.rb
tzheleznyak: Restrictions are created from annotations loaded in ExtractResourceRestrictions and verified by costaints so no.

app/domain/authentication/authn_jwt/signing_key/fetch_jwks_uri_signing_key.rb
tzheleznyak: Returned it

app/domain/authentication/authn_jwt/consts.rb
tzheleznyak: Removed

nessiLahav (approved):
LGTM

PR #2235, opened 2021-06-01
Pass config as input to conjurctl config commands

jtuttle:


bin/conjur-cli.rb
```diff
@@ -230,14 +231,24 @@
...
+    Validate Conjur configuration then show configuration attributes and their
+    sources.
 
     The values displayed by this command reflect the current state of the
-    configuration sources. For the example, the environment variables and
-    config file. These may not reflect the current values used by the running
-    Conjur server.
+    configuration sources. For example, the environment variables and config
+    file. These may not reflect the current values used by the running Conjur
+    server.
```
jonahx, line 251: Is there a way to inspect these?  If so, we should explain how or link to relevant docs.

spec/conjurctl/commands/configuration/show_spec.rb
```diff
@@ -8,16 +8,31 @@ def command_output(config, format:)
...
+      trusted_proxies: {
+        source: { type: :defaults },
+        value: []
+      }
+    }
+  }
+
+  before do
+    allow(conjur_config).
+      to receive(:to_source_trace).
```
jonahx, line 33: What does this method do?  The name wasn't clear to me... is it part of AnywayConfig?

jtuttle: Yep it's from AnywayConfig. It returns a hash with info about where each attribute is being read from.

bin/conjur-cli.rb
jtuttle: Not currently, no, though I have proposed adding an endpoint to the API for doing so. I believe it will be future work.

spec/conjurctl/configuration_show_spec.rb
```diff
@@ -37,4 +37,16 @@
       "error: Unknown configuration output format 'invalid'\n"
     )
   end
+
+  it "outputs an error when configuration is invalid" do
+    _stdout, stderr, status = Open3.capture3(
+      "CONJUR_TRUSTED_PROXIES=boop conjurctl configuration show"
+    )
```
jonahx, line 44: Not a big deal, but this feels more like an end to end test, rather than a unit test.  I'd expect it to exist, but not in this spec file...

jtuttle: Yep, all of these tests need to be moved but it's a problem for another day.

jonahx (approved):

PR #2240, opened 2021-06-02
fix error parameter with digits

Avi-Eyal:
fix Error logs thats their parameters contain digits

### DIFF
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -446,7 +446,7 @@ module Errors
       )
 
       UnsupportedAuthenticator = ::Util::TrackableErrorClass.new(
-        msg: "Authenticator '{0}' is unsupported.",
+        msg: "Authenticator '{0-authenticator-name}' is unsupported.",
         code: "CONJ00184E"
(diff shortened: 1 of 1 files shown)


tzheleznyak (approved):
Looks good . Just add to the test plan a log check for this so we will do it in regression

app/domain/errors.rb
```diff
@@ -446,12 +446,12 @@ module AuthnJwt
       )
 
       UnsupportedAuthenticator = ::Util::TrackableErrorClass.new(
-        msg: "Authenticator '{0-authenticator-name}' is unsupported.",
```
benoaviram, line 449: @Avi-Eyal  what is the reason for this change?

Avi-Eyal: There is no particular reason, the convention in the other logs is that where there is only one parameter or the specific parameter is self explained then you dont need to specify the parameter by name.

personally I dont really care to rollback my changes for this line

benoaviram (approved):

PR #2251, opened 2021-06-07
Refactor the policy and rotator test to use the client module

mFelgate:
### What does this PR do?
Moved all the API calls and helper methods to a separate file in order so that it can be later used by tests outside of the policy and rotator directories.
The reviewer should make sure this code is up to a quality standard acceptable by the team.
No

#### Change log

#### Test coverage

#### Documentation

#### API Changes


cucumber/policy/features/support/client.rb
```diff
@@ -0,0 +1,147 @@
+
+module ClientHelpers
+  attr_reader :result
+
+  class Client
+    @@admin_password = 'SEcret12!!!!'
+    @@account = ENV['CONJUR_ACCOUNT'] || 'cucumber'
+    @@appliance_url =  ENV['CONJUR_APPLIANCE_URL'] || 'http://conjur'
```
jonahx, line 8: Another of one of those "don't copy from our current codebase" landmines I mentioned during our last call, but basically consider `@@` to be a "never use" ruby feature.  I'll give you some more context on that in person, but there are always better alternatives we can use.

cucumber/policy/features/support/client.rb
```diff
@@ -0,0 +1,147 @@
...
+        client.put_request(policy)
+      end
+
+      def update_policy(policy)
+        client.patch_request(policy)
+      end
+
+      def extend_policy(policy)
+        client.post_request(policy)
+      end
```
jonahx, line 107: This is conversation we've had for years, but the verbs `load`, `update`, `extend` are not obvious to me.  Maybe we can brainstorm some more precise ones.

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -32,74 +33,31 @@ def invoke status: nil, &block
...
-
-  def login_resource
-    RestClient::Resource.new(
-      uri('authn', 'api_key', ''), 'admin', admin_password
-    )
-  end
-
-  def make_full_id *tokens
-    super(tokens.join(":"))
+  def policy_helper(id)
```
jonahx, line 59: name change suggestion: `policy_client`.

cucumber/policy/features/support/client.rb
jonahx: Also, we should take the ENV var info as arguments to the constructor.

cucumber/policy/features/support/client.rb
```diff
@@ -0,0 +1,147 @@
+
+module ClientHelpers
```
jonahx, line 2: This comment applies to all the clients:

I don't think we want them under a "helpers" module, since one of the nice benefits of this approach is that this code is no longer couples to our tests, or to the "rspec" way of doing things.  I'm not sure where the best place is for them -- we can try to figure that together -- but I do think the `XXXHelpers` modules are superfluous.  

Also, to avoid name stutter, and to strengthen the conceptual connection between the clients, I'd consider a structure like:
```rb
# top level
class Client
  class Policy
  end
  
  class Resource
  end
  # etc
end
```
So then the client code we read like `Client::Policy.new(...)`, etc...

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -111,43 +69,30 @@ def json_result
...
       else
         [ "user", login ].join(":")
       end
-      api_key = create_api_key(role)
+      api_key = client.create_api_key(role)
+    end
+    if login == "admin"
+      @token = client.admin_token
+    else
+      @token = client.get_token(login, api_key)
```
jonahx, line 94: Let's discuss this method too... It's existing, but maybe we can clean it up more now.... I haven't thought it through yet.

cucumber/policy/features/support/client.rb
```diff
@@ -0,0 +1,141 @@
...
+
+      return self.new(admin_token) if user == User.admin
+
+      role_api_key = create_api_key(user, admin_token)
+      token = create_token(user, role_api_key)
+      self.new(token)
+    end
+
+    def create_token(user, role_api_key)
+      url = "#{APPLIANCE_URL}/authn/#{ACCOUNT}/#{CGI.escape(user.login)}" +
```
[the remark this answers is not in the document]
jonahx: My bad.  This can be fixed while still retaining clean margins by changin `+` to `\`.

cucumber/policy/features/support/client.rb
```diff
@@ -0,0 +1,141 @@
...
+    end
+  end
+
+  def initialize(token)
+    @token = token
+  end
+
+  # Policy methods
+  #
+  def load_policy(id:, policy:)
```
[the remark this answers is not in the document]
jonahx: This one we want to ignore.

cucumber/policy/features/support/client.rb
```diff
@@ -0,0 +1,141 @@
...
+    )
+  end
+
+  private
+
+  def auth_header
+    @auth_header ||= self.class.auth_header(@token)
+  end
+
+  def uri(root, kind, id = nil)
```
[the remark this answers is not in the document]
jonahx: This is an interesting one.  I think this one can be ignored, but it's worth noting the solutions it "wants" you to make:
1. Introduce a value object `Uri` which is instantiated with these parameters and a has a single `to_s` method that returns what this private method is returning.  I think this extra code this introduces (not a ton, but more) isn't worth it.
2. Add `@appliance_url` and `@account` as instance variables of this class, which are taken by the initializer but default to values `APPLIANCE_URL` and `ACCOUNT`.  This makes the class a bit more flexible, and perhaps would make unit tests for it a tad cleaner, but again practically I don't think the cost is worth the benefit.

cucumber/policy/features/support/client.rb
```diff
@@ -0,0 +1,141 @@
...
+    uri = "#{APPLIANCE_URL}/#{root}/#{ACCOUNT}/#{kind}"
+    return uri unless id
+
+    "#{uri}/#{CGI.escape(id)}"
+  end
+
+  def resource(uri_)
```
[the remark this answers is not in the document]
jonahx: I'd ignore this one for similar reasons to the above.

In this case, only solution 1. above applies.  Though here we could also just inline all instances of this method.  I think that would be clearly make the code worse though.

cucumber/policy/features/support/client.rb
```diff
@@ -0,0 +1,141 @@
+class Client
```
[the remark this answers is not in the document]
jonahx: This one is fine to ignore.

cucumber/policy/features/support/client.rb
```diff
@@ -0,0 +1,141 @@
...
+
+    def auth_header(token)
+      { Authorization:  %Q(Token token="#{token}") }
+    end
+
+    private
+
+    def for_role(user)
+      admin_key = fetch_admin_api_key
+      admin_token = create_token(User.admin, admin_key)
```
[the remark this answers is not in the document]
jonahx: Just add
```ruby
user_admin = User.admin
```
at the top of the method, and then refer to `user_admin`.

cucumber/policy/features/step_definitions/error_steps.rb
```diff
@@ -13,11 +13,19 @@
...
-  expect(json).to have_key('error')
-  @error = json['error']
+  expect(json_result).to have_key('error')
+  @error = json_result['error']
 end
 
 Then(/^there is no error$/) do
   expect(json_result).not_to have_key('error')
 end
+
```
jonahx, line 23: Since the difference between this step and the two steps above is going to confuse people, we should add a comment stating the reason we did it, along with the plan to refactor the other references to the non-apostrophe version, and then ultimately delete this.

cucumber/policy/features/step_definitions/host_factory_steps.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-Then(/^the "([^"]*)" should be:$/) do |field, json|
-  _, kind, id = @result['id'].split(':')
-  actual = JSON.parse(get_resource(kind, id))[field]
-  expected = JSON.parse(json)
-  expect(actual).to eq(expected)
+Then(/^the "([^"]*)" should be:$/) do |field, expected_json|
```
jonahx, line 3: Might as well change this one to `{string}` since we're here.

cucumber/policy/features/support/policy_helpers.rb
```diff
@@ -111,44 +43,44 @@ def json_result
...
+          @err.http_code
+        end
+
+        def body
+          JSON.parse(@err.response.body)
+        end
+
+        # def has_key?(key)
+        #   @err.response.body.has_key?(key)
+        # end
```
jonahx, line 75: Dead code

jonahx (approved):

PR #2260, opened 2021-06-09
add automation test for empty annotation

Avi-Eyal:
Added automation sanity test for empty annotation

### DIFF
--- a/cucumber/authenticators_jwt/features/authn_jwt.feature
+++ b/cucumber/authenticators_jwt/features/authn_jwt.feature
@@ -61,7 +61,7 @@ Feature: JWT Authenticator - JWKs Basic sanity
     cucumber:host:myapp successfully authenticated with authenticator authn-jwt service cucumber:webservice:conjur/authn-jwt/raw
     """
 
-  Scenario: A valid JWT token with identity in the token
+  Scenario: Authenticator is not enabled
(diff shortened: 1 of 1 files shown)


cucumber/authenticators_jwt/features/authn_jwt.feature
```diff
@@ -104,3 +104,29 @@ Feature: JWT Authenticator - JWKs Basic sanity
...
     """
+
+  Scenario: Annotation with empty value
+    Given I have a "variable" resource called "test-variable"
+    Given I extend the policy with:
+    """
+    - !host
+      id: myapp
+      annotations:
+        authn-jwt/raw/custom-claim:
```
tzheleznyak, line 115: Add more annotations here so only one of them with empty value

Avi-Eyal: there are already other annotations in the policy that is defined in the background

Avi-Eyal: they are defined in the same host (`myapp`) so i dont see why we need to add another annotation

and about the files for each test, Aviram and I talked about it and we thought that it would be better to wait until we finish to write the tests and only then we will split the tests to different files

tzheleznyak (approved):

PR #2263, opened 2021-06-10
Usable config file validation error

### LINKED ISSUE #2256: Invalid configuration file format produces nice error
If a ConjurConfig object is initialized and tries to read an invalid YAML file, it will currently fail with an uncaught exception:

```
bundler: failed to load command: bin/conjurctl (bin/conjurctl)
/opt/conjur/possum/vendor/bundle/ruby/2.5.0/gems/anyway_config-2.1.0/lib/anyway/utils/deep_merge.rb:8:in `deep_merge!': undefined method `each' for "trusted_proxies":String (NoMethodError)
	from /opt/conjur/possum/vendor/bundle/ruby/2.5.0/gems/anyway_config-2.1.0/lib/.rbnext/2.7/anyway/config.rb:322:in `block in load_from_sources'
	from /opt/conjur/possum/vendor/bundle/ruby/2.5.0/gems/anyway_config-2.1.0/lib/.rbnext/3.0/anyway/loaders.rb:55:in `each'
...
```

Add a ConfigLoader class whose job is it to load the ConjurConfig object and catch this exception and provide a better error message. Use this in the places where we instantiate a ConjurConfig object.

micahlee:
### What does this PR do?

This PR updates the Conjur config file load to output a user-friendly validation error when the config file contains a YAML syntax error, or the root YAML object is not a dictionary. Previously, these errors output a library-internal exception and stack trace.

Resolves #2256

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/lib/conjur/conjur_config.rb
+++ b/lib/conjur/conjur_config.rb
@@ -26,6 +26,31 @@ module Conjur
       authenticators: []
     )
 
+    def initialize(*args)
+      super(*args)
+
+    # If the config file is not a valid YAML document, we want
+    # to raise a user-friendly ConfigValidationError rather than
+    # a Psych library exception.
+    rescue Psych::SyntaxError => e
+      raise(
+        ConfigValidationError,
+        "Config file contains a YAML syntax error: #{e.message}"
+      )
+
+    # If the config file is valid YAML, but the root object is
+    # not a YAML dictionary, this raises one of a number of
+    # NoMethodError exceptions because AnywayConfig assumes parsing
+    # the config file will result in a Ruby Hash. We capture
+    # this and raise a more user-friendly error message.
+    rescue NoMethodError
+      raise(
+        ConfigValidationError,
+        "Unable to parse config file. " \
+        "Please ensure that it is a valid YAML dictionary."
+      )
+    end
+
     # Perform validations and collect errors then report results as exception
     on_load do
       invalid = []
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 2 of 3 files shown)


lib/conjur/conjur_config.rb
```diff
@@ -26,6 +26,30 @@ class ConjurConfig < Anyway::Config
       authenticators: []
     )
 
+    def initialize(*args)
+      super(*args)
+
+    # If theh config file is not a valid YAML document, we want
```
jvanderhoof, line 32:
```suggestion
    # If the config file is not a valid YAML document, we want
```

lib/conjur/conjur_config.rb
```diff
@@ -26,6 +26,30 @@ class ConjurConfig < Anyway::Config
...
+
+    # If the config file is valid YAML, but the root object is
+    # not a YAML dictionary, this raises one of a number of
+    # NoMethodError exceptions because AnywayConfig assumes parsing
+    # the config file will result in a Ruby Hash. We capture
+    # this and raise a more user-friendly error message.
+    rescue NoMethodError
+      raise(
+        ConfigValidationError,
+        "Config file is not a valid YAML dictionary."
```
jtuttle, line 49: It's (probably) possible that a `NoMethodError` can arise for a different reason since it's so generic. Would it make more sense to have an error message that's less certain about the cause? Something like "Unable to parse config file. Please ensure that it is a valid YAML dictionary."

jtuttle (changes requested):

lib/conjur/conjur_config.rb
micahlee: doh... Thanks!

lib/conjur/conjur_config.rb
micahlee: Yeah, that makes sense. I'll make the change.

jtuttle (approved):

PR #2264, opened 2021-06-10
add sanity test automation for ignored invalid annotations

Avi-Eyal:
add sanity test automation for ignored invalid annotations

### DIFF
--- a/cucumber/authenticators_jwt/features/authn_jwt.feature
+++ b/cucumber/authenticators_jwt/features/authn_jwt.feature
@@ -130,3 +130,58 @@ Feature: JWT Authenticator - JWKs Basic sanity
     """
     CONJ00100E Annotation, 'custom-claim', is empty
     """
+
+  Scenario: Ignore invalid annotations
+    Given I have a "variable" resource called "test-variable"
+    Given I extend the policy with:
+    """
+    - !host
+      id: myapp
+      annotations:
+        authn-jwt/raw: invalid
+        authn-jwt: invalid
+        authn-jwt/raw/sub/sub: invalid
+        authn-jwt/raw2/sub: invalid
+        authn-jwt/raw/sub: valid
+        authn-jwt/raw/namespace-id: valid
+        authn-jwt/raw/project-path: valid
+    """
+    And I permit host "myapp" to "execute" it
+    And I add the secret value "test-secret" to the resource "cucumber:variable:test-variable"
+    And I issue a JWT token:
+    """
+    {
+      "user":"host/myapp",
+      "project-id": "myproject",
+      "sub": "valid",
+      "namespace-id": "valid",
+      "project-path": "valid"
+    }
+    """
+    And I save my place in the log file
+    When I authenticate via authn-jwt with the JWT token
+    Then the HTTP response status code is 200
+    And The following appears in the log after my savepoint:
+    """
+    CONJ00048D Validating resource restriction on request: 'sub'
+    """
+    And The following appears in the log after my savepoint:
+    """
+    CONJ00048D Validating resource restriction on request: 'namespace-id'
+    """
+    And The following appears in the log after my savepoint:
+    """
+    CONJ00048D Validating resource restriction on request: 'project-path'
+    """
+    And The following appears in the log after my savepoint:
+    """
+    CONJ00045D Resource restrictions matched request
+    """
+    And The following appears in the log after my savepoint:
+    """
+    CONJ00030D Resource restrictions validated
+    """
+    And The following appears in the log after my savepoint:
+    """
+    CONJ00103D 'validate_restrictions' passed successfully
+    """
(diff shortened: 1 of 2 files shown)


cucumber/authenticators_jwt/features/authn_jwt.feature
```diff
@@ -130,3 +130,58 @@ Feature: JWT Authenticator - JWKs Basic sanity
...
+    """
+    - !host
+      id: myapp
+      annotations:
+        authn-jwt/raw: invalid
+        authn-jwt: invalid
+        authn-jwt/raw/sub/sub: invalid
+        authn-jwt/raw2/sub: invalid
+        authn-jwt/raw/sub: valid
+        authn-jwt/raw/namespace-id: valid
```
benoaviram, line 146: @Avi-Eyal  just for the sake of testing.. try to have it combined.. i.e. - 
valid
not valid
valid
not valid
valid

cucumber/authenticators_jwt/features/authn_jwt.feature
```diff
@@ -130,3 +130,58 @@ Feature: JWT Authenticator - JWKs Basic sanity
...
+        authn-jwt/raw/project-path: valid
+    """
+    And I permit host "myapp" to "execute" it
+    And I add the secret value "test-secret" to the resource "cucumber:variable:test-variable"
```
benoaviram, line 150: Could you put this lined right after the first line? making them together will make more sense..

cucumber/authenticators_jwt/features/authn_jwt.feature
```diff
@@ -130,3 +130,58 @@ Feature: JWT Authenticator - JWKs Basic sanity
...
+      "user":"host/myapp",
+      "project-id": "myproject",
+      "sub": "valid",
+      "namespace-id": "valid",
+      "project-path": "valid"
+    }
+    """
+    And I save my place in the log file
+    When I authenticate via authn-jwt with the JWT token
+    Then the HTTP response status code is 200
```
eladkug, line 163: Before verifying logs try to verify you retrieve the secret well

cucumber/authenticators_jwt/features/authn_jwt.feature
```diff
@@ -130,3 +130,58 @@ Feature: JWT Authenticator - JWKs Basic sanity
...
+    - !host
+      id: myapp
+      annotations:
+        authn-jwt/raw: invalid
+        authn-jwt: invalid
+        authn-jwt/raw/sub/sub: invalid
+        authn-jwt/raw2/sub: invalid
+        authn-jwt/raw/sub: valid
+        authn-jwt/raw/namespace-id: valid
+        authn-jwt/raw/project-path: valid
```
eladkug, line 147: If one of the valid message will fail , the test will not fail because we dont verify each line specific, and since we have valid fields it will succeed.  right ? if so we need to handle this case

cucumber/authenticators_jwt/features/authn_jwt.feature
Avi-Eyal: I dont understand what you mean, which secrets do i need to verify? there are other tests to check the valid flow

cucumber/authenticators_jwt/features/authn_jwt.feature
Avi-Eyal: I think youre right, the test will succeed
what to you mean by "handle this case"?

cucumber/authenticators_jwt/features/authn_jwt.feature
eladkug: you update `cucumber:variable:test-variable` and you want to verify you can get this value , otherwise why you update it ?

cucumber/authenticators_jwt/features/authn_jwt.feature
eladkug: maybe test separately each one of them ? If so , left here only one valid message since all the rest not really verified

cucumber/authenticators_jwt/features/authn_jwt.feature
Avi-Eyal: youre right, i need to delete `I add the secret value "test-secret" to the resource "cucumber:variable:test-variable"` because i has nothing to do with my text, i just copied it from another test to simulate a working authentication

eladkug: Why not verify you get the secret, it fast and cost nothing and will get more confidence here.

cucumber/authenticators_jwt/features/authn_jwt.feature
Avi-Eyal: ill correct myself, each of the annotations is checked through logs, invalid annotation wont log `CONJ00048D Validating resource restriction on request:` so the test will fail

tzheleznyak (approved):

PR #2267, opened 2021-06-13
Fix more pr comments part 2

tzheleznyak:
Fix PR comments here 

https://github.com/cyberark/conjur/pull/2227#discussion_r650534494
https://github.com/cyberark/conjur/pull/2227#discussion_r650534431
https://github.com/cyberark/conjur/pull/2227#discussion_r650534217
https://github.com/cyberark/conjur/pull/2227#discussion_r650534158
https://github.com/cyberark/conjur/pull/2227#discussion_r650534087
https://github.com/cyberark/conjur/pull/2227#discussion_r650533807
https://github.com/cyberark/conjur/pull/2227#discussion_r650533421

### DIFF
--- a/app/domain/authentication/authn_jwt/authenticator.rb
+++ b/app/domain/authentication/authn_jwt/authenticator.rb
@@ -23,7 +23,7 @@ module Authentication
       def call
         validate_and_decode_token
         get_jwt_identity_from_request
-        validate_user_has_access_to_webservice
+        validate_host_has_access_to_webservice
         validate_origin
         validate_restrictions
         audit_success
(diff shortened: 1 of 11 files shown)


app/domain/authentication/authn_jwt/authenticator.rb
```diff
@@ -111,11 +111,13 @@ def identity_role
       # If there is no jwt identity so role and username are nil
       def audit_role_id
         return @audit_role_id if @audit_role_id
+        # Dont change to jwt_identity so function won't be called if validate_and_decode failed. Because we
+        # won't have jwt token where we can fetch the identity from
```
Avi-Eyal, line 115: that doesnt sound right, if this is the logic, than `jwt_identity` is supposed to implement it. if it stays that way it can easily become a bug

tzheleznyak: jwt_identity need jwt decoded token to work. If validate and decode fails there is no decoded token so there is no identity and jwt_identity can't run.

authenticator.rb know if validate_and_decode runed or not. jwt_identity don't knows and dont need to know. There fore if valdiate and decode didn't run so in the audit failiure we should throw false.

PR #2289, opened 2021-06-23
Fix "evoke configuration apply" error in RHEL

orenbm:
### What does this PR do?
When running the "evoke configuration apply" command in
Conjur on RHEL we get the error `bignum too big to convert into long`.
This happens because the evoke command above tries to kill the
conjur process and it fails to properly get the conjur process id.
It happens in RHEL because we have some additional puma processes
running by the Conjur user:
```
conjur     89291       1  0 15:20 ?        00:00:04 puma 5.3.1 (tcp://0.0.0.0:2301,tcp://127.0.0.1) [evoke]
conjur     89448   89291  0 15:21 ?        00:00:01 puma 5.3.1 (tcp://0.0.0.0:2301,tcp://127.0.0.1) [evoke]
conjur     89456   89291  0 15:21 ?        00:00:00 puma 5.3.1 (tcp://0.0.0.0:2301,tcp://127.0.0.1) [evoke]
conjur     89535   89291  0 15:21 ?        00:00:00 puma 5.3.1 (tcp://0.0.0.0:2301,tcp://127.0.0.1) [evoke]
conjur     89539   89291  0 15:21 ?        00:00:00 puma 5.3.1 (tcp://0.0.0.0:2301,tcp://127.0.0.1) [evoke]
```

Filtering entries that have the `evoke` word in them fixes the issue

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/bin/conjur-cli/commands/configuration/apply.rb
+++ b/bin/conjur-cli/commands/configuration/apply.rb
@@ -42,7 +42,7 @@ module Commands
       private
 
       def server_pid
-        cmd = "ps -ef | grep puma | grep -v grep | grep -v cluster | " \
+        cmd = "ps -ef | grep puma | grep -v grep | grep -v cluster | grep -v evoke | " \
               "grep conjur | awk '{print $2}' | tr -d '\n'"
         stdout, _ = @command_runner.capture2(cmd)
         stdout.to_i


bin/conjur-cli/commands/configuration/apply.rb
```diff
@@ -42,7 +42,7 @@ def call
       private
 
       def server_pid
-        cmd = "ps -ef | grep puma | grep -v grep | grep -v cluster | " \
+        cmd = "ps -ef | grep puma | grep -v grep | grep -v cluster | grep -v evoke | " \
```
micahlee, line 45: I think referring to the `evoke` name in Conjur seems like it should be a last resort, since this becomes an implicit dependency.

I would prefer us to pursue an alternative way to better identify the right Puma process(es) to terminate.

This is a helpful analysis, though! Thanks for digging into that. :)

Tovli: +1

orenbm: > I think referring to the evoke name in Conjur seems like it should be a last resort, since this becomes an implicit dependency.

that's a good point, thanks. feel free to close the PR or leave it open, whatever you think is best for when someone tries to properly fix this.

jvanderhoof:
@orenbm, first off, thanks for attempting to fix this!!!!  I believe we're better off implementing this in `conjurinc/evoke`. We're going to need to create a more generic interface for managing processes, one that can handle both Runit and Systemd, ideally without caring which process manager is used!

micahlee:
Hey @jvanderhoof , `runit` vs `systemd` is unrelated to this issue. Puma runs as a sub-process to the root service process that those control.

Thanks to @orenbm's debugging, I understand now that the issue here is that the `conjurctl configuration apply` command is trying to find a process with a tag like:

```
puma 4.3.8 (tcp://0.0.0.0:3000) [conjur-server]
```

However, the default tag (the `[conjur-server]` part) is just the directory name that puma is launched from, so this is different between open source, the appliance container, and the package-based appliance.

We get lucky in the appliance container, because `possum` sorts above `ui`. The filter is still incorrect, though:
```
puma 4.3.8 (tcp://localhost:5000) [possum]
puma 5.3.1 (tcp://localhost:3000) [ui]
```

I don't know what all of the puma process descriptions look like in the package-based install, but we clearly don't get lucky there.

I've proposed an alternative solution here, that uses a well-known process tag that is consistent across the different deployment environments: https://github.com/cyberark/conjur/pull/2291/files

Let me know what you think!

micahlee:
This change is finished in https://github.com/cyberark/conjur/pull/2291

Issue #2282, opened 2021-06-18
Possible Regression


hughsaunders:
## Summary
[conjur-ecs-deploy](https://github.com/cyberar/conjur-ecs-deploy) is a cloudformation template and set of scripts for deploying conjur containers to ECS/Fargate. It's ci pipeline deploys `conjur:edge` and then runs a few basic functionality tests, such as 500 requests in serial to the root url.
Recently this test started failing.

Here's the Jenkins build history for the main branch:
![Screenshot 2021-06-18 at 17 18 50](https://user-images.githubusercontent.com/253506/122590319-4b5d3b00-d059-11eb-9514-30e2c4b54fec.png)

Here's the [serial test log](https://jenkins.conjur.net/job/cyberark--conjur-ecs-deploy/job/main/92/artifact/ab_1_read.log/*view*/) for the last successful build:
```
This is ApacheBench, Version 2.3 <$Revision: 1843412 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking ecsdeploycimain92.ci.conjur.net (be patient)


Server Software:        
Server Hostname:        ecsdeploycimain92.ci.conjur.net
Server Port:            443
SSL/TLS Protocol:       TLSv1.2,ECDHE-RSA-AES128-GCM-SHA256,2048,128
Server Temp Key:        ECDH P-256 256 bits
TLS Server Name:        ecsdeploycimain92.ci.conjur.net

Document Path:          /secrets/conjur/variable/TestPolicy/TestVariable
Document Length:        Variable

Concurrency Level:      1
Time taken for tests:   7.704 seconds
Complete requests:      500
Failed requests:        0
Total transferred:      180000 bytes
HTML transferred:       18000 bytes
Requests per second:    64.90 [#/sec] (mean)
Time per request:       15.408 [ms] (mean)
Time per request:       15.408 [ms] (mean, across all concurrent requests)
Transfer rate:          22.82 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        4    5   0.4      4       8
Processing:     9   11   2.3     10      45
Waiting:        9   11   2.3     10      45
Total:         13   15   2.3     15      49
ERROR: The median and mean for the initial connection time are more than twice the standard
       deviation apart. These results are NOT reliable.

Percentage of the requests served within a certain time (ms)
  50%     15
[...]

PR #2294, opened 2021-06-25
Set Puma Requests Per Connection To Infinity

hughsaunders:
#### What's Changed?
* Increased the default value for puma's MAX_FAST_INLINE value from 10 to Infinity (& Beyond)
* Introduced an environment variable MAX_REQUESTS_PER_CONNECTION to allow users to set their own value.

#### Why have these changes been made?
Conjur recently upgraded Puma to 4.3.8 to mitigate [CVE-2021-29509](https://nvd.nist.gov/vuln/detail/CVE-2021-29509). Unfortunately the mitigation chosen by the upstream Puma developers is to close connections once 10 requests have been served. This leads to ~10% failure rate for clients sending large number of requests and re-using connections. Setting MAX_FAST_INLINE to Infinity brings back Puma's previous behaviour and the 10% error rate is reduced to 0 in the conjur-ecs-deploy test suite. This value is [acceptable](https://github.com/puma/puma/issues/2361#issuecomment-690332926) when Conjur is behind a proxy or load balancer, which it is in all recommended architectures. The new environment variables was introduced to cover the unlikely case that a user wants to expose Conjur directly. 

Resolves: cyberark/conjur#2282
Docs Request: https://github.com/cyberark/conjur-docs/issues/1676

#### Change log

#### Test coverage

#### Documentation
- [x] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or https://github.com/cyberark/conjur-docs/issues/1676

#### API Changes


micahlee (changes requested):
Thanks @hughsaunders! In bumping puma, we also need to update the version in the `NOTICES.txt` file here:
https://github.com/cyberark/conjur/blob/98502d847df676b27597880d5b8e77a6e74d4a88/NOTICES.txt#L22

and here:
https://github.com/cyberark/conjur/blob/98502d847df676b27597880d5b8e77a6e74d4a88/NOTICES.txt#L196

Do you mind adding that to the commit? 

Everything else LGTM!

micahlee (approved):
LGTM! Nice catch and fix.

alexkalish:
@hughsaunders: Two comments:
1. Moving forward, all configuration env variables should be prefixed with `CONJUR_`, so that the target of the value is clearly Conjur.  Would you mind quickly making that change?
2. Ideally, this would get folded into the new configuration feature (i.e. focusing on the YAML config file) and not be provided as a raw env variable.  Given the limited audience for this, I don't think that is required here.

And a question!  Users of Conjur Enterprise would not want to modify this, correct?  Enterprise is always behind Nginx, so I assume connections are heavily reused.

Thanks!!

PR #2302, opened 2021-06-29
Add authn jwt algorthem tests

tzheleznyak:
### What does this PR do?
Updates gem version of jwt to 2.2.3 : https://github.com/jwt/ruby-jwt/releases/tag/v2.2.3
Adds tests for checking algorithems in jwt authenticator

ONYX 8858
ONYX 8859
ONYX 8852
ONYX 8851

#### Change log

#### Test coverage

#### Documentation

#### API Changes


cucumber/authenticators_jwt/features/authn_jwt_security.feature
```diff
@@ -0,0 +1,270 @@
+Feature: JWT Authenticator - Security
+
```
eladkug, line 2: Add description

eladkug (commented):
We can run several tests with one table so it will be readable, WDYT?

jonahx:
There are some CC errors worth fixing, but I thought at the code level this looks good.

I have some higher level, strategic comments, which hopefully will be helpful for future PRs but are probably "too big" to change for this one.

- `authn_jwt_security.feature` is thorough, but very long.  I think many of these test paths could be more easily tested with unit tests.  `AuthJwt::Authenticator` already has all its dependencies injected, so the unit tests corresponding to these scenarios would be easy to write.  Indeed, enabling unit tests is one of the main benefits of writing `AuthnJwt::Authenticator` this way.   Having a couple of end-to-end tests like these as a sanity check is a good idea, but for thoroughly testing all the possible code paths I much prefer unit tests.  Having to test "through" the outermost interface makes the tests slower and more complex than they need to be.
- We recently did a deep refactor of a lot of the `policy` tests, which allowed to get rid of almost all of the cucumber helper methods, so that each step definition is whole self-contained and readable, without having to jump to multiple helper method definitions.  The look like this:
```ruby
Then("the {string} should be:") do |field, expected_json|
  _, kind, id = @result.body['id'].split(':')
  @client ||= Client.for("user", "admin")
  @result = @client.fetch_resource(kind: kind, id: id)
  expect(@result.body[field]).to eq(JSON.parse(expected_json))
end
```
[More examples](https://github.com/cyberark/conjur/blob/master/cucumber/policy/features/step_definitions/resource_steps.rb)

The only additional object is a single [API client](https://github.com/cyberark/conjur/blob/master/cucumber/policy/features/support/client.rb), which is mostly self-explanatory and re-used across the definitions (rather than having multiple different helper functions for each step definition).  I thought the strategy greatly simplified the step defs.

cucumber/authenticators_jwt/features/authn_jwt_security.feature
tzheleznyak: Done

cucumber/authenticators_jwt/features/authn_jwt_security.feature
```diff
@@ -0,0 +1,120 @@
+Feature: JWT Authenticator - Security
+  Tests checking that JWT authenticator stands against different attacks and security risks.
```
Avi-Eyal, line 2: Add new line before the description

cucumber/authenticators_jwt/features/authn_jwt_security.feature
```diff
@@ -0,0 +1,120 @@
...
+  Checking different the authenticators with different algorithems siging the jwt token.
```
Avi-Eyal, line 3:
```suggestion
  Checking different authenticators with different algorithms signing the jwt token.
```

cucumber/authenticators_jwt/features/step_definitions/jwt_jwks_steps.rb
```diff
@@ -44,6 +101,24 @@
   )
 end
 
+Given(/I issue HMAC JWT token:/) do |token_body_string|
+  # token body has to be an object (not a string) for correct token creation
+  issue_jwt_hmac_token_token_with_rsa_key(
```
Avi-Eyal, line 106:
```suggestion
  issue_jwt_hmac_token_with_rsa_key(
```

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
```diff
@@ -37,12 +41,33 @@ def init_second_jwks_file_with_same_kid(first_file_name, second_file_name)
     )
   end
 
+  def init_ecdsa_jwks_file(file_name, key_type)
+    @default_file_name ||= file_name
+    @key_type = key_type
```
Avi-Eyal, line 46: maybe use  memoization here
```suggestion
    @key_type ||= key_type
```

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
```diff
@@ -91,10 +128,60 @@ def issue_jwt_token_not_memoized_key(token_body, algorithm = Algorithms::RS256)
...
+    token_parts = encoded_token.split(".")
+    headers = "{\"typ\":\"JWT\",\"alg\":\"#{alg_header}\",\"kid\":\"#{jwk_set[@default_file_name].kid}\"}"
+    token_parts[0] = Base64.urlsafe_encode64(headers).strip
+    @jwt_token = token_parts.join(".")
+  end
+
+  def issue_jwt_hmac_token_token_with_rsa_key(token_body)
+    @jwt_token = JWT.encode(
+      token_body,
+      rsa_keys[@default_file_name].to_s,
```
Avi-Eyal, line 165: I believe this is already a string (we dont convert to string in other places that do the same thing)
```suggestion
      rsa_keys[@default_file_name],
```

Avi-Eyal (approved):
LGTM but please wait for @sashaCher 's review

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
tzheleznyak: Its not string its object. HMAC should be string.

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
```diff
@@ -73,6 +98,18 @@ def issue_jwt_token_with_jwk(token_body, file_name, algorithm = Algorithms::RS25
...
   end
 
+  def issue_none_alg_jwt_token(token_body)
+    token_header_and_payload = JWT.encode(
+      token_body,
+      nil,
+      'none',
+      {
+        kid: jwk_set[@default_file_name].kid
+      }
```
sashaCher, line 108: It's unnecessary to put `kid` into `none` signed token

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
```diff
@@ -73,6 +98,18 @@ def issue_jwt_token_with_jwk(token_body, file_name, algorithm = Algorithms::RS25
...
+    )
+    @jwt_token = "#{token_header_and_payload}invalid_singature"
```
sashaCher, line 110: `none` signed token does not contain signature part at all
There's no need in extra manipulation.
In all cases output of `JWT.encode` is good to go.

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
```diff
@@ -91,10 +128,60 @@ def issue_jwt_token_not_memoized_key(token_body, algorithm = Algorithms::RS256)
...
+  def issue_rsa_jwt_token_with_alg_header(token_body, alg_header, algorithm = Algorithms::RS256)
+    encoded_token = JWT.encode(
+      token_body,
+      rsa_keys[@default_file_name],
+      algorithm
+    )
+    token_parts = encoded_token.split(".")
+    headers = "{\"typ\":\"JWT\",\"alg\":\"#{alg_header}\",\"kid\":\"#{jwk_set[@default_file_name].kid}\"}"
+    token_parts[0] = Base64.urlsafe_encode64(headers).strip
+    @jwt_token = token_parts.join(".")
```
sashaCher, line 159: You manipulate token after signature. Means it's invalid by definition.
Wrong algorithm has to be signed.

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
```diff
@@ -91,10 +128,60 @@ def issue_jwt_token_not_memoized_key(token_body, algorithm = Algorithms::RS256)
...
+  end
+
+  def issue_jwt_hmac_token_with_rsa_key(token_body)
+    @jwt_token = JWT.encode(
+      token_body,
+      rsa_keys[@default_file_name].to_s,
```
sashaCher, line 165: Are you sure that it's a right manipulation?

sashaCher (changes requested):
1. I think that unnecessary code should not be a part of this PR
2. Can we reduce this PR scope only to RSA algs sanity check and to handle algorithm substitution use-cases in the different one.

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
tzheleznyak: But this is what i want to test here

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
tzheleznyak: This is the algorithem substitution siging with one and giving another not?

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
tzheleznyak: Lets discuss it

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
tzheleznyak: Lets discuss it

tzheleznyak:
@sashaCher 
Removed code of elaptic curve and HMAC from PR

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
sashaCher: I'm not sure that this's the purpose of the test

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
sashaCher: I don't know what's the output of to_s in the case...

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
```diff
@@ -91,6 +104,29 @@ def issue_jwt_token_not_memoized_key(token_body, algorithm = Algorithms::RS256)
     )
   end
 
+  def issue_rsa_jwt_token_with_alg_header(token_body, alg_header, algorithm = Algorithms::RS256)
```
sashaCher, line 107: Let's remove the method. It has 0 references.

cucumber/authenticators_jwt/features/support/jwt_jwks_helper.rb
```diff
@@ -91,6 +104,29 @@ def issue_jwt_token_not_memoized_key(token_body, algorithm = Algorithms::RS256)
...
+  end
+
+  def issue_jwt_hmac_token_with_rsa_key(token_body)
+    @jwt_token = JWT.encode(
+      token_body,
+      rsa_keys[@default_file_name].public_key.to_s,
+      Algorithms::HS256,
+      {
+        kid: jwk_set[@default_file_name].kid
+      }
```
sashaCher, line 126: I think that `kid` is unnecessary there

tzheleznyak: Talked about it

sashaCher (approved):
LGTM

PR #2308, opened 2021-07-11
load policy validation design

nessiLahav:
### What does this PR do?
Add load policy validation design proposal
[link to document ](https://github.com/cyberark/conjur/blob/load-policy-design/design/proposals/load_policy/load_policy_validation.md)


design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Useful Links
+- [Feature doc](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-8566)
+- [UX research page](https://ca-il-confluence.il.cyber-ark.com.mcas.ms/display/rndp/JWT+Authenticator+-+load+policy+validation)
+- [JWT authenticator stage 1.1 feature doc](https://ca-il-confluence.il.cyber-ark.com.mcas.ms/display/rndp/JWT+Authenticator+-+re-scope#JWTAuthenticatorrescope-Stage1.1-SimplegenericJWTauthenticatorwithtokenschema)
+
+
+## Issue Description
+In misconfiguration errors, for example:
+hosts with empty JWT annotation value, we are failing at JWT authentication request which is bad UX.
```
orenbm, line 17: i would emphasize that we are returning a 401 response in such a case which is a client error but the error is not in the request itself rather than in the policy load. I would also add that we accept loading a policy that is worthless and failing it would help the customer to create valid policies

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,148 @@
...
+- create temp schema
+- merge new temp schema to main schema 
+- the validation logic will start after the schema merged to main
+
+2. new ValidatePolicyConstraints command class - is responsible to validate all the policy entities 
+   (host,user,webservice) which were sent by the user in load policy request.
+   
+    Note: in this design we will validate only hosts 
+   
+3. new CreatePolicyEntityValidatorsFactory command class - is responsible to create a list of validators as the amount
```
orenbm, line 65: the suffix `Factory` is redundant. `CreatePolicyEntityValidators` is enough.

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,148 @@
...
+3. new CreatePolicyEntityValidatorsFactory command class - is responsible to create a list of validators as the amount 
+   of the different policy entities types in the records input.
+   
+    Note: in this design we will create only `ValidateHostConstraints` with all the hosts ids in his belly 
+   
+4. new ValidateHostConstraints command class - is responsible to validate a list of host ids which are initialized in his creation.
+   the flow is to fetch the host annotations from the DB (from the main schema after the merge) and to 
+   use the `CreateHostValidatorsFactory` to create hosts authenticators validators.
+   
+5. new CreateHostValidatorsFactory command class - is responsible to create a list of validators as the amount
```
orenbm, line 74: same here. you can use `CreateHostValidators`

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,148 @@
...
+   of the different hosts authenticators types
+   
+    Note: 
+    - in this design we will create only `ValidateJwtHostConstraints` which is validator for hosts with JWT annotations
+    - this class can be refactored to use reflection if in the future we will want the separate the authenticator logic 
+    to a different gems 
+      
+6. new GetJwtHosts command class - is responsible to get a lists of host and annotations and to return only the hosts wjich are
```
orenbm, line 82: i am not sure that we need this as a command class, but rather as a method. what are the dependencies of this class?

InbalZilberman, line 82: a list of hosts and their annotations
wjich -> which
Can you explaining what is the module that runs the GetJwtHosts or all the functions you describe here please?

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,148 @@
...
+   
+    Note: 
+    - in this design we will create only `ValidateJwtHostConstraints` which is validator for hosts with JWT annotations
+    - this class can be refactored to use reflection if in the future we will want the separate the authenticator logic 
+    to a different gems 
+      
+6. new GetJwtHosts command class - is responsible to get a lists of host and annotations and to return only the hosts wjich are 
+related the JWT authenticator
+   
+7. new ValidateJwtHostConstraints command class - is responsible to to perform the validation logic as written in the requirements,
```
InbalZilberman, line 85: double to

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,148 @@
...
+for all annotations have values.
+   
+    Note: we will refactor the current code of the authentication flow to use this new command class as well 
+   
+
+   
+
+## Performance
+TBD
```
InbalZilberman, line 94: @nessiLahav I know you said several times that this has performance plenty that we will need to understand later on

InbalZilberman (commented):
I love how it is complex but elegant may the EE will be 5 :)

sashaCher:
1. What will be UX in conjunction with #2309 . Means when conjur admin adds new value to mandatory claims list all hosts "belong" to authenticator will became invalid. Am I understand right that from this moment any load policy contains those hosts will fail if it not adds new mandatory annotation?
2. Do you have in mind "general" annotation use case when it belongs to `authn-jwt` but does not belong to any existing `service-id`s particularly.
```
- !host
  id: myapp
  annotations:
    authn-jwt/namespace: mynamespace
```
If there are two services one with mandatory namespace and one without.

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,212 @@
...
+
+### Flow Diagrams
+In our use case, Host with JWT authenticator annotations validation:
+1. Conjur admin define JWT authenticator 
+2. Conjur admin load a policy of Host to be authenticated with JWT authenticator
+3. New policy validation layer is triggered and validate if the host is configured correctly 
+4. Host authenticate using JWT authenticator, due to above validation there are no configuration errors at the authenticate request stage
+
+### Sequence Diagrams
+![Load policy validation sequence diagram](load-policy-validation-sequence-diagram.png)
```
oburstein-hub, line 73: Cannot open sequence diagram. Also - can you please send me links to class diagram ?

nessiLahav: Push the link in the PR description

nessiLahav:
> 1. What will be UX in conjunction with [ONYX-10152 : Design Token Schema #2309](https://github.com/cyberark/conjur/pull/2309) . Means when conjur admin adds new value to mandatory claims list all hosts "belong" to authenticator will became invalid. Am I understand right that from this moment any load policy contains those hosts will fail if it not adds new mandatory annotation?
> 2. Do you have in mind "general" annotation use case when it belongs to `authn-jwt` but does not belong to any existing `service-id`s particularly.
> 
> ```
> - !host
>   id: myapp
>   annotations:
>     authn-jwt/namespace: mynamespace
> ```
> 
> If there are two services one with mandatory namespace and one without.

regarding 1: Correct, i mentioned it in `Backwards Compatibility` section

regrading 2: There is no general annotation for authn-jwt, but if we want to think about the future when we may have this requirement . I think it is implementation details of the `ValidateJwtHostConstraints`  depends on  the desired UX

design/proposals/load_policy/load_policy_validation.md
nessiLahav: I updated this section but still do not know to say how big will be the  performance plenty

design/proposals/load_policy/load_policy_validation.md
nessiLahav: In high level: in the flow i can get 10+ hosts in 1 policy fro the users as input.
i need to be able to create the Validator which will be know hot to handle each hosts based on his annotations keys (currently we are supporting only JWT) .

So GetJwtHosts helps me to filter the Hosts which have JWT annotations from the list

design/proposals/load_policy/load_policy_validation.md
nessiLahav: ?

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,148 @@
...
+   which will call to a new command class `ValidatePolicyConstraints` with tne input of the policy entities 
+   records (see example in the end the of the diagram) from the user body request.
+   
+Load policy current logic - 
+   for the load policy rest requests put/post/patch the current logic in high level is:
+    The current logic: 
+       
+- policy syntax check
+- load records from policy 
+- create temp schema
```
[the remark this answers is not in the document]
nessiLahav: Added thanks, under ` Policy validation entry point:`

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Issue Description
+In misconfiguration errors, for example:
+hosts with empty JWT annotation value, we are failing at JWT authentication request which is bad UX.
+
+## Solution
+For better UX we want to perform policy validation steps in order to identify configuration errors early as possible. 
+
+### Solution - scope
+This solution is creating a generic infrastructure for all policy entities validation.
```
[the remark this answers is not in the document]
nessiLahav: The infrastructure is built to support also authenticators of other entities in the future,

For example if you want to support Authenticator policy validation:
1. Create a new general entity validator : `ValidateAuthenticatorConstraints`
2. Update the factory to support it : `CreatePolicyEntityValidatorsFactory`
3 .....

design/proposals/load_policy/load_policy_validation.md
nessiLahav: Not sure if it will have dependencies, i marked it as a command class since it has only 1 public function.

design/proposals/load_policy/load_policy_validation.md
nessiLahav: Why? i think it is makes the class a bit more clear. 
by reading the word factory you are already guessing that the class is creating a different kinds of host validators based on some decision logic

design/proposals/load_policy/load_policy_validation.md
nessiLahav: answered below

design/proposals/load_policy/load_policy_validation.md
nessiLahav: thanks

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,257 @@
...
+## Useful Links
+- [Feature doc](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-8566)
+- [UX research page](https://ca-il-confluence.il.cyber-ark.com.mcas.ms/display/rndp/JWT+Authenticator+-+load+policy+validation)
+- [JWT authenticator stage 1.1 feature doc](https://ca-il-confluence.il.cyber-ark.com.mcas.ms/display/rndp/JWT+Authenticator+-+re-scope#JWTAuthenticatorrescope-Stage1.1-SimplegenericJWTauthenticatorwithtokenschema)
+
+
+## Issue Description
+In misconfiguration errors, for example:
+
+for hosts without JWT **mandatory** annotation,
```
hilagross, line 46: why only without mandatory fields?
We should validate all cases

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Solution
+For better UX we want to perform policy validation steps in order to identify configuration errors as early as possible. 
+
+### Solution - scope
+This solution is creating a generic infrastructure for all policy entities validation.
+And focusing on Host with JWT authenticator annotations as first use case to implement on this new suggested infrastructure.
+
+### UX
+#### Load JWT policy without mandatory JWT annotation
```
hilagross, line 60: same question - what about with mandatory fields?

hilagross: we should include mapping in here

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Audit
+* No need for new audit records 
+
+## Open Questions
+None.
+
+## Delivery Plan
+
+### 1st plan - 1 milestone Total EE 45
```
hilagross, line 213: what is the difference between 1st and 2nd plan?

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,285 @@
...
+#### Docs EE 3
+
+#### Demo EE 1
+
+### 2nd plan - 2 milestones Total EE 45
+
+#### M1 - Load policy infrastructure  SP 19
+
+**Implementations** - 16 SP
+* No need of 3.d
```
hilagross, line 250: impl - 16SP
the bullet says 'no need' and 3d
what needs to be done? why 16? why 3?

design/proposals/load_policy/load_policy_validation.md
nessiLahav: 1st plan - the whole feature 1 shot 
2nd plan - the 1st plan split into 2 milestones

design/proposals/load_policy/load_policy_validation.md
nessiLahav: this is the implementation tasks from plan 1 without 3.d task

design/proposals/load_policy/load_policy_validation.md
nessiLahav: sure this is just an example,
the use cases validation are not changing the high level design

design/proposals/load_policy/load_policy_validation.md
nessiLahav: As i see it this is an high level design and infrastructure to validate all current and future use cases of JWT so i didn't want to focus on specific use cases

design/proposals/load_policy/load_policy_validation.md
```diff
@@ -0,0 +1,285 @@
...
+3. New policy validation layer is triggered and validate if the host is configured correctly 
+4. Host authenticate using JWT authenticator, due to above validation there are no configuration errors at the authenticate request stage
+
+### Sequence Diagrams
+![Load policy validation sequence diagram](load-policy-validation-sequence-diagram.png)
+
+#### Sequence Diagrams - additional info
+1. Policy validation entry point:
+As described in the diagram the validation will be triggered as the final new stage of the current logic (read below),
+   which will call to a new command class `ValidatePolicyConstraints` with tne input of the policy entities
```
jvanderhoof, line 99:
```suggestion
   which will call to a new command class `ValidatePolicyConstraints` with the input of the policy entities 
```

jvanderhoof:
@nessiLahav, (cc: @rafis3, @jonahx, @oburstein-hub) Unfortunately this approach isn't going to work from a code organization perspective. What you're proposing is forcing the requirements of an authenticator into the policy engine. This will cause two problems:
1. Make the authenticators exponentially more complicated to separate from Conjur.
2. Substantially increase the complexity of the already complex policy engine by forcing one-off checks based on a mix of attributes and string values found in secrets.

We need to find an architecture that allows authenticators to validate the arbitrary choices made by a particular authenticator (in terms of policy). Alternatively, can we develop a generic mechanism for registering host "type" validations?  If a host becomes a "Gitlab JWT Host", or a "Kubernetes Host", we can create a hook for an authenticator to register its host format when it is loaded.  This may still be too much of a coupling to be acceptable, but I believe we need to explore other options.

oburstein-hub:
> @nessiLahav, (cc: @rafis3, @jonahx, @oburstein-hub) Unfortunately this approach isn't going to work from a code organization perspective. What you're proposing is forcing the requirements of an authenticator into the policy engine. This will cause two problems:
> 
> 1. Make the authenticators exponentially more complicated to separate from Conjur.
> 2. Substantially increase the complexity of the already complex policy engine by forcing one-off checks based on a mix of attributes and string values found in secrets.
> 
> We need to find an architecture that allows authenticators to validate the arbitrary choices made by a particular authenticator (in terms of policy). Alternatively, can we develop a generic mechanism for registering host "type" validations? If a host becomes a "Gitlab JWT Host", or a "Kubernetes Host", we can create a hook for an authenticator to register its host format when it is loaded. This may still be too much of a coupling to be acceptable, but I believe we need to explore other options.

Hi Jason

Since the estimation provided is 45 points I am not sure that all this development will be done at all
This should eliminate your concern

jvanderhoof (changes requested):
I'm formally blocking this PR to ensure it gets the proper conversation. In order to have this conversation, we need to understand the following:
- Who is accountable for technical direction within the Conjur Policy Engine?
- Who is accountable for product direction (to insure the proposed outcome aligns with the product vision).

I like the conversation started with this feature, but I have serious concerns with the proposed solution.

nessiLahav:
> I'm formally blocking this PR to ensure it gets the proper conversation. In order to have this conversation, we need to understand the following:
> 
> * Who is accountable for technical direction within the Conjur Policy Engine?
> * Who is accountable for product direction (to insure the proposed outcome aligns with the product vision).
> 
> I like the conversation started with this feature, but I have serious concerns with the proposed solution.

@jvanderhoof Thanks you are raising an excellent concerns, this is a very sensitive feature in the core of our product, and we should challenge it from all directions.

Regarding your concerns:
1. I agree but i think it is solvable, i mentioned it in a note in the sequence diagram, we can execute the validation logic by using reflection
2. It is increasing the complexity by the definition of the feature requirement, i have tried to simplify it in the design but you are correct this is a new engine.

I also agreed that we should think of long term architecture to solve the following technical depts :
1. How to manage configuration in Conjur ( we currently use secrets which is not their purpose) 
2. How to properly connect host to authenticator ( we currently use strings values in annotations)

I think that the above depts causes to the load policy validation logic to rely on a mix of attributes secrets which increases the complexity of this feature

Having say that, we should improve our product in steps, and i think that the load policy validation is in 1st priority for a better UX.
Do you agree ? If yes,  What would you change in this design ? When the idea is to first solve this issue but also not creating a new technical dept.

jvanderhoof:
@nessiLahav, 

> Having say that, we should improve our product in steps, and i think that the load policy validation is in 1st priority for a better UX.

This isn't a feature that brings "policy validation" to Conjur. This feature is bringing policy load feedback for the sole purpose of the JWT authenticator. IMO, this is an overdependence on policy for configuring authenticators. JWT is really the first authenticator for which we can't make Policy work effectively for.  Architecture fails when it can't support the needs of its use cases.

> If yes, What would you change in this design? When the idea is to first solve this issue but also not creating a new technical dept.

It feels like we need a new higher-level primate that reflects the needs of the system. Authenticators require a couple of things:

1. Connection information (how this authenticator connects to perform the validation)
2. Host information (how the remote identity is mapped to an internal identity)

If we create new primitives in Conjur policy, validation becomes part of the primitive itself. I do think we need to think hard about the setting and validation flow of all our authenticators. For example, is an authenticator is misconfigured if it can't successfully connect? Loading policy and setting various values via the API doesn't feel like the right approach when we think holistically.

nessiLahav:
> @nessiLahav,
> 
> > Having say that, we should improve our product in steps, and i think that the load policy validation is in 1st priority for a better UX.
> 
> This isn't a feature that brings "policy validation" to Conjur. This feature is bringing policy load feedback for the sole purpose of the JWT authenticator. IMO, this is an overdependence on policy for configuring authenticators. JWT is really the first authenticator for which we can't make Policy work effectively. Architecture fails when it can't support the use cases. It feels like our Policy and Variable based approach has reached its viable end.
> 
> > If yes, What would you change in this design? When the idea is to first solve this issue but also not creating a new technical dept.
> 
> It feels like we need a new higher-level primate that reflects the needs of the system. Authenticators require a couple of things:
> 
> 1. Connection information (how this authenticator connects to perform the validation)
> 2. Host information (how the remote identity is mapped to an internal identity)
> 
> If we create new primitives in Conjur policy, validation becomes part of the primitive itself. I do think we need to think hard about the setting and validation flow of all our authenticators. For example, is an authenticator is misconfigured if it can't successfully connect? Loading policy and setting various values via the API doesn't feel like the right approach when we think holistically.

@rafis3 @benoaviram @oburstein-hub I agree with Jason, 
how can this be promoted  ?

PR #2314, opened 2021-07-18
Add NonPermittedConstraint

tzheleznyak:
### What does this PR do?
Add NonPermittedConstraint to be able to check annotations not contain standard JWT claims 

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/app/domain/authentication/constraints/non_permitted_constraint.rb
+++ b/app/domain/authentication/constraints/non_permitted_constraint.rb
@@ -1,0 +1,21 @@
+module Authentication
+  module Constraints
+
+    # This constraint is initialized with an array of strings.
+    # They represent resource restrictions that are not allowed
+    # Calling `validate` enforces this constraint on the given list of resource restrictions.
+    # If there is annotation for one of these non permitted values proper error would be thrown
+    class NonPermittedConstraint
+
+      def initialize(non_permitted:)
+        @non_permitted = non_permitted
+      end
+
+      def validate(resource_restrictions:)
+        any_non_permitted_restrictions = resource_restrictions & @non_permitted
+
+        raise Errors::Authentication::Constraints::NonPermittedRestrictionGiven, any_non_permitted_restrictions unless any_non_permitted_restrictions.empty?
+      end
+    end
+  end
+end
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -540,6 +540,11 @@ module Errors
         code: "CONJ00069E"
       )
 
+      NonPermittedRestrictionGiven = ::Util::TrackableErrorClass.new(
+        msg: "Role can't have one of these none permitted restrictions {0-restrictions}",
+        code: "CONJ00069E"
+      )
+
       RoleMissingAnyRestrictions =  ::Util::TrackableErrorClass.new(
         msg: "Role must have at least one relevant annotation",
         code: "CONJ00099E"
--- a/app/domain/errors.rb
+++ b/app/domain/errors.rb
@@ -541,7 +541,7 @@ module Errors
       )
 
       NonPermittedRestrictionGiven = ::Util::TrackableErrorClass.new(
-        msg: "Role can't have one of these none permitted restrictions {0-restrictions}",
+        msg: "Role can't have one of these none permitted restrictions '{0-restrictions}'",
         code: "CONJ00069E"
       )
 
--- a/spec/app/domain/authentication/constraints/non_permitted_constraint_spec.rb
+++ b/spec/app/domain/authentication/constraints/non_permitted_constraint_spec.rb
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe(Authentication::Constraints::NonPermittedConstraint) do
+  let(:non_permitted_constraint) { ["iat", "nbf", "exp", "iss"] }
+  let(:raised_error) { ::Errors::Authentication::Constraints::NonPermittedRestrictionGiven }
+
+  subject(:constraint) do
+    Authentication::Constraints::NonPermittedConstraint.new(non_permitted: non_permitted_constraint)
+  end
+
+  context "when validating empty array" do
+    subject do
+      constraint.validate(resource_restrictions: [])
+    end
+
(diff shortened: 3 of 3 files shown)


app/domain/authentication/constraints/non_permitted_constraint.rb
```diff
@@ -0,0 +1,22 @@
...
+
+      def initialize(non_permitted:)
+        @non_permitted = non_permitted
+      end
+
+      def validate(resource_restrictions:)
+        any_non_permitted_restrictions = resource_restrictions & @non_permitted
+        if any_non_permitted_restrictions.length > 0
+          raise Errors::Authentication::Constraints::NonPermittedRestrictionGiven, any_non_permitted_restrictions
+        end
```
sashaCher, line 18: Do you consider to use !any_non_permitted_restrictions.empty? or something like this and to convert the raise t one-liner?

app/domain/authentication/constraints/non_permitted_constraint.rb
```diff
@@ -0,0 +1,22 @@
+module Authentication
+  module Constraints
+
+    # This constraint is initialized with an array of strings.
+    # They represent resource restrictions that are not allowed
+    # Calling `validate` enforces this constraint on the given list of resource restrictions.
+    # If there is annotation for one of these non permitted values proper error would be thrown
+    class NonPermittedConstraint
```
benoaviram, line 8: @tzheleznyak we will have more functions in this class? if not, why not making it command class?

tzheleznyak: To be in the same form as other constraints. If im turning this to CommandClass it can be in separate issue to refactor all validators . 

And it will influence also GCP, OIDC, K8S

spec/app/domain/authentication/constraints/non_permitted_constraint_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe(Authentication::Constraints::NonPermittedConstraint) do
+  let(:non_permitted_constraint) { ["iat", "nbf", "exp", "iss"] }
```
benoaviram, line 6: @tzheleznyak could you use the MANDATORY_CLAIMS_DENY_LIST const? 
it has more claims and single-source-of-truth

spec/app/domain/authentication/constraints/non_permitted_constraint_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  subject(:constraint) do
+    Authentication::Constraints::NonPermittedConstraint.new(non_permitted: non_permitted_constraint)
+  end
+
+  context "when validating empty array" do
+    subject do
+      constraint.validate(resource_restrictions: [])
+    end
+
+    it "raises an error" do
```
benoaviram, line 18: why?

spec/app/domain/authentication/constraints/non_permitted_constraint_spec.rb
tzheleznyak: Its a test so i don't think its really a problem. But already @nessiLahav  adding this to host so when he will add i will use it. Even in another pr i will do the usage. So there won't be two variables

spec/app/domain/authentication/constraints/non_permitted_constraint_spec.rb
tzheleznyak: Good catch thanks

tzheleznyak: The title needs to be change to not raise_error i put wrong title

spec/app/domain/authentication/constraints/non_permitted_constraint_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    subject do
+      constraint.validate(resource_restrictions: ["exp", "iat"])
+    end
+
+    it "raises an error" do
+      expect { subject }.to raise_error(raised_error)
+    end
+  end
+
+  context "when validating one permitted and one non permitted restriction" do
```
benoaviram, line 63: Maybe worth adding - 
1. restriction with name that contains the non permited (customer made) - i.e. "expref"
2. Make the first restriction permitted and the second non-permitted 
3. create a list with first restriction permitted, second is empty string, third is non-permitted

spec/app/domain/authentication/constraints/non_permitted_constraint_spec.rb
tzheleznyak: I forgot to change titles

spec/app/domain/authentication/constraints/non_permitted_constraint_spec.rb
tzheleznyak: 1. Name that contains inside something non permitted should be permitted in my opinion because there is no special treatment for strings that are different then the one in the const
2. Added it good point
3. This case is not possible because current load policy don't lets to load annitation without a name because its a broken yaml

benoaviram: Regarding 1 i agree with you, we need to make sure that this is allowed and will be allowed also in future changes in the feature

tzheleznyak: Ok now i got you. I will add it

benoaviram (dismissed):

app/domain/errors.rb
```diff
@@ -525,6 +525,11 @@ module Constraints
         code: "CONJ00069E"
       )
 
+      NonPermittedRestrictionGiven = ::Util::TrackableErrorClass.new(
+        msg: "Role can't have one of these none permitted restrictions {0-restrictions}",
```
nessiLahav, line 529: change {0-restrictions} to '{0-restrictions}'

nessiLahav (changes requested):
1 comment

nessiLahav (dismissed):
LGTM

nessiLahav (approved):

PR #2315, opened 2021-07-18
`aud` claim support

sashaCher:
### What does this PR do?

Enables `authn-jwt` to support `aud` claim according [RFC](https://datatracker.ietf.org/doc/html/rfc7519).

#### Change log

#### Test coverage

#### Documentation

#### API Changes


nessiLahav (changes requested):
Just add limitation that audience variable supports only singe value and not array

app/domain/authentication/authn_jwt/validate_and_decode/fetch_audience_value.rb
```diff
@@ -0,0 +1,79 @@
...
+          resource_class: ::Resource,
+          fetch_required_secrets: ::Conjur::FetchRequiredSecrets.new,
+          logger: Rails.logger
+        },
+        inputs: %i[authentication_parameters]
+      ) do
+        def call
+          @logger.debug(LogMessages::Authentication::AuthnJwt::FetchingAudienceValue.new)
+          
+          return empty_audience_value unless audience_resource_exists?
```
nessiLahav, line 18: Why? it should be similar to `FetchIssuerValue` command class

nessiLahav:
@sashaCher We need to decide what state if claim is mandatory to validate, the current behavior is:
iss,iat,nbf = if they exist in the token, we must to validate them

In this design you suggest otherwise:
If aud  exists in token is not must to validate
But if audience is configured it means it is must to be validated 

I think aud should act the same as other claims. (we can add in the future configuration of mandatory standard claim to validate)

@benoaviram @hilagross WDYT?

sashaCher:
@nessiLahav I've added limitation and emphasised that `aud` is mandatory once variable exists

sashaCher:
@nessiLahav 
1. It's a shame that we have no mechanism that allows customer to define claims mentioned by you as mandatory.
2. In any case we have to respect them if they are present
3. The behaviour should be consistent:
**Given:** As a authenticator owner I've defined the `audience` variable and provided it with valid value
**When** The authenticator receives a token without `aud` claim at all
**Then** The authentication failed

nessiLahav:
> @nessiLahav
> 
> 1. It's a shame that we have no mechanism that allows customer to define claims mentioned by you as mandatory.
> 2. In any case we have to respect them if they are present
> 3. The behaviour should be consistent:
>    **Given:** As a authenticator owner I've defined the `audience` variable and provided it with valid value
>    **When** The authenticator receives a token without `aud` claim at all
>    **Then** The authentication failed

I am OK with the decision that we are enforcing the audience by the policy but as you mention in bullet 1, i think the UX is problematic/confusing:

**exp** - is mandatory hardcoded 
**iss** - is mandatory if exists in token while `issuer` variable is optional 
**iat, nbf** - are mandatory if exist in token
**aud** - is mandatory if `audience` variable is configured
**jti** - is ignored

hilagross:
@nessiLahav @sashaCher 
if `audience` variable exists then it should must be in token and should be validated.
the variable itself should be optional, same as `iss`

I'm trying to understand if it's possible to verify `aud` claim that is in the token (same as other claims behaviour)

nessiLahav (approved):
LGTM

nessiLahav (approved):
LGTM

PR #2319, opened 2021-07-19
401 for empty username

hughsaunders:
### What does this PR do?
Previously Conjur returned 500 when trying to login with an empty
username. This commit alters Conjur to return 401 for an empty
username. This is consistent with other invalid usernames.

### What ticket does this PR close?
Related: cyberark/conjur#2318

#### Change log
- [ ] The CHANGELOG has been updated, or
- [x] This PR does not include user-facing changes and doesn't require a CHANGELOG update

#### Test coverage
- [x] This PR includes new unit and integration tests to go with the code changes, or
- [ ] The changes in this PR do not require tests

#### Documentation
- [ ] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or
- [x] This PR does not require updating any documentation

#### API Changes
- [ ] The [OpenAPI spec](https://github.com/cyberark/conjur-openapi-spec) has been updated to meet new API changes (or an issue has been opened), or
- [x] The changes in this PR do not affect the Conjur API

Issue #2318, opened 2021-07-19
500 on login with empty username


hughsaunders:
## Summary
When a user tries to login to conjur with an empty username, the server responds with 500. I would expect 401. 

## Steps to Reproduce
1. Deploy conjur via [conjur-quickstart](https://github.com/cyberark/conjur-quickstart).
2. Exec into the client container and try to login, press enter when prompted for a username and supply any value for the password: 

   ```bash
      15:08 $ docker exec -it conjur_client bash 
      root@99eb7ba03fbb:/# conjur authn login
      Enter your username to log into Conjur: 
      Please enter your password (it will not be echoed): 

      error: 500 Internal Server Error


## Expected Results
I expect Conjur to reutrn 401 with an empty username. Currently if the password is empty 401 is returned. I think it should be the same for an empty username.

## Actual Results (including error logs, if applicable)
Server returns 500.

Example server log:

```
[origin=172.21.0.7] [request_id=c5481771-9e44-4feb-b03b-d7c314deb303] [tid=50] Started GET "/authn/conjur/login" for 172.21.0.7 at 2021-07-19 15:14:18 +0000
[origin=172.21.0.7] [request_id=c5481771-9e44-4feb-b03b-d7c314deb303] [tid=50] Processing by AuthenticateController#login as */*
[origin=172.21.0.7] [request_id=c5481771-9e44-4feb-b03b-d7c314deb303] [tid=50]   Parameters: {:controller=>"authenticate", :action=>"login", :authenticator=>"authn", :account=>"conjur"}
[origin=172.21.0.7] [request_id=c5481771-9e44-4feb-b03b-d7c314deb303] [tid=50] CONJ00047I Login Error: #<Dry::Struct::Error: [Authentication::AuthenticatorInput.new] "" (String) has invalid type for :username violates constraints (format?(/\S+/, "") failed)>
[origin=172.21.0.7] [request_id=c5481771-9e44-4feb-b03b-d7c314deb303] [tid=50] /opt/conjur-server/app/controllers/concerns/basic_authenticator.rb:38:in `login_input'
[origin=172.21.0.7] [request_id=c5481771-9e44-4feb-b03b-d7c314deb303] [tid=50] /opt/conjur-server/app/controllers/concerns/basic_authenticator.rb:31:in `authenticator_login'
[...]

PR #2321, opened 2021-07-20
401 for empty username

hughsaunders:
### What does this PR do?
Previously Conjur returned 500 when trying to login with an empty
username. This commit alters Conjur to return 401 for an empty
username. This is consistent with other invalid usernames.

Related: cyberark/conjur#2318

#### Change log

#### Test coverage

#### Documentation

#### API Changes

This is my second attempt, previous PR was #2319.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Conjur now returns 401 (instead of 500) for authentication requests with an empty username. [cyberark/conjur#2381](https://github.com/cyberark/conjur/issues/2318)
+
 ### Security
 - Upgraded addressable in ./Gemfile.lock and ./docs/Gemfile.lock to 2.8.0 to resolve
   GHSA-jxhc-q857-3j6g [cyberark/conjur#2311](https://github.com/cyberark/conjur/pull/2311)
--- a/cucumber/api/features/login.feature
+++ b/cucumber/api/features/login.feature
@@ -81,6 +81,14 @@ Feature: Exchange a role's password for its API key
       cucumber:user:non-exist failed to login with authenticator authn service cucumber:webservice:conjur/authn: CONJ00007E 'non-exist' not found
     """
 
+  Scenario: Empty username results in 401 response
+    When I GET "/authn/cucumber/login" with username "" and password "My-Password1"
+    Then the HTTP response status code is 401
+
+  Scenario: Invalid username with non-alpha characters results in 401 response
+    When I GET "/authn/cucumber/login" with username "!@#$%^&*(){}" and password "My-Password1"
+    Then the HTTP response status code is 401
+
   Scenario: Wrong hostname cannot be used to obtain API key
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Conjur now returns 401 (instead of 500) for authentication requests with an empty username.
```
orenbm, line 10: can you please link to an issue? we do this for all CL entries

cucumber/api/features/login.feature
```diff
@@ -81,6 +81,10 @@ Feature: Exchange a role's password for its API key
       cucumber:user:non-exist failed to login with authenticator authn service cucumber:webservice:conjur/authn: CONJ00007E 'non-exist' not found
     """
 
+  Scenario: Empty username results in 401 response
```
orenbm, line 84: what about `/authenticate` requests? can we add a test in `authenticate.feature`?

eladkug: I would add uniq chars and space between two words as well

hughsaunders: Uniq chars? Eg single character user name? or a string of a single repeated character?

eladkug: something like` !@#$%^&*(){}` to make sure it doesnt fail as part of reading those chars

hughsaunders: Not sure the same test makes sense for authenticate as the route won't match with an empty string, so it will result in 404:

Correctly formed authenticate request (conjur-quickstart local docker instance):
```
root@4d8ed4895116:/# curl -i --header "Accept-Encoding: base64" --data "${key}" http://conjur/authn/conjur/admin/authenticateHTTP/1.1 200 OK
Content-Encoding: base64
Content-Type: text/plain; charset=utf-8
ETag: W/"76fc6a0836156143887cec2a33197620"
Cache-Control: max-age=0, private, must-revalidate
X-Request-Id: 7597cc83-a4e5-4553-b260-6a661b9b2d15
X-Runtime: 0.059081
Transfer-Encoding: chunked

eyJwcm... snip rest of token
```
Authenticate request with empty username:
```
root@4d8ed4895116:/# curl -i --header "Accept-Encoding: base64" --data "${key}" http://conjur/authn/conjur//authenticate
HTTP/1.1 404 Not Found
Content-Type: text/html; charset=utf-8
X-Request-Id: 25413780-9c20-4896-bc3e-1986b0cc6ca0
X-Runtime: 0.015727
```
In this case the error is 404, which is reasonable given the username is part of the route.

orenbm: you are correct, my bad

hughsaunders:
The latest SHA 6727528 has had a [green codeclimate build](https://codeclimate.com/github/cyberark/conjur/builds/7308), but due to ongoing codeclimate problems the status on the PR has not udpated. As there is a green code climate build, I propose we merge this anyway. 
codeclimate_green_build

jonahx (approved):

eladkug (approved):
LGTM

PR #2332, opened 2021-07-22
Add fix to validate restrictions tests

tzheleznyak:


cucumber/authenticators_jwt/features/authn_jwt_validate_restrictions.feature
```diff
@@ -99,10 +99,10 @@ Feature: JWT Authenticator - Validate restrictions
...
       id: myapp
       annotations:
-        authn-jwt/project-id: myproject
-        authn-jwt/aud: myaud
-        authn-jwt/raw/project-id: myproject
-        authn-jwt/raw/aud: wrong-aud
+        authn-jwt/project-id: valid
+        authn-jwt/ref: valid
+        authn-jwt/raw/project-id: valid
+        authn-jwt/raw/ref: invalid
```
sashaCher, line 105: It looks like the test loosing its point.
All annotations has the same value.
How will we identify the error?
Previous version is looking much better

Avi-Eyal (approved):
LGTM

PR #2338, opened 2021-07-25
Fix direct check when mapping bug

tzheleznyak:
Bugfix to make test ONYX-10889 work.

Currently when the mapping is given we can check a host annotation directly and according to mapping. The bugfix insures error being thrown if one of the host annotations is a mapped claim.

### DIFF
--- a/app/domain/authentication/authn_jwt/restriction_validation/create_constraints.rb
+++ b/app/domain/authentication/authn_jwt/restriction_validation/create_constraints.rb
@@ -10,9 +10,7 @@ module Authentication
       #   * NonPermittedConstraint - Checks there are no standard claims [exp,iat,nbf,iss] in the host annotations
       CreateConstrains = CommandClass.new(
         dependencies: {
-          non_permitted_constraint: Authentication::Constraints::NonPermittedConstraint.new(
-            non_permitted: CLAIMS_DENY_LIST
-          ),
+          non_permitted_constraint_class: Authentication::Constraints::NonPermittedConstraint,
           required_constraint_class: Authentication::Constraints::RequiredConstraint,
           multiple_constraint_class: Authentication::Constraints::MultipleConstraint,
           not_empty_constraint: Authentication::Constraints::NotEmptyConstraint.new,
@@ -20,7 +18,7 @@ module Authentication
           fetch_mapping_claims_class: Authentication::AuthnJwt::RestrictionValidation::FetchMappingClaims,
           logger: Rails.logger
         },
-        inputs: %i[authentication_parameters]
+        inputs: %i[authentication_parameters base_non_permitted_annotations]
       ) do
         # These is command class so only call is called from outside. Other functions are needed here.
         # :reek:TooManyMethods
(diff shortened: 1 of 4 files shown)


app/domain/errors.rb
```diff
@@ -584,7 +584,7 @@ module Constraints
       )
 
       NonPermittedRestrictionGiven = ::Util::TrackableErrorClass.new(
-        msg: "Role can't have one of these none permitted restrictions '{0-restrictions}'",
+        msg: "Role can't have standard claim or a mapped claim",
```
sashaCher, line 587: This message a bit problematic for me.
It's thrown by infrastructural class [NonPermittedConstraint](https://github.com/cyberark/conjur/blob/0084b9d7f2a6b09eaa75093dd994f38ccee3e52d/app/domain/authentication/constraints/non_permitted_constraint.rb#L17) but contains the name of concrete features in it.
What will happen if any other authenticaficator will need to use the class?
What will happen if other parameters except mapped claims will affect permitted restrictions list?

sashaCher: I prefer to keep the message as is as long it's an infrastructural message.
If we want to make things clear we should catch an exception and to have jet specific treatment for it.

app/domain/authentication/authn_jwt/restriction_validation/create_constraints.rb
```diff
@@ -79,8 +77,14 @@ def add_required_constraint
           @constraints.append(required_constraint)
         end
 
+        def non_permitted_constraint
+          @non_permitted_constraint ||= @non_permitted_constraint_class.new(
+            non_permitted: @base_non_permitted_annotations + mapping_claims.keys
```
sashaCher, line 82:
```suggestion
            non_permitted: @base_non_permitted_annotations + mapping_claims.values
```
Keys are annotations and values are claim names: `brach:ref`.

app/domain/errors.rb
hilagross: I think we should try to make it more understandable, the way it was it wasn't so clear

app/domain/authentication/authn_jwt/restriction_validation/create_constraints.rb
tzheleznyak: But i invert it

tzheleznyak: image

app/domain/errors.rb
```diff
@@ -542,6 +542,11 @@ module AuthnJwt
         msg: "Failed to parse mapping claims, {0-purpose} value '{1-claim-value}' appears twice or more",
         code: "CONJ00113E"
       )
+
+      RoleWithStandardOrMappedClaimError = ::Util::TrackableErrorClass.new(
+        msg: "Role can't have standard claim or a mapped claim. Error : '{0-error}'",
```
sashaCher, line 547:
```suggestion
        msg: "Role can't have registered claim or a mapped claim. Error : '{0-error}'",
```
https://datatracker.ietf.org/doc/html/rfc7519#section-4.1

sashaCher (commented):
Looks good.
Left one comment

sashaCher (approved):
LGTM
Message will be changed in docs PR

PR #2358, opened 2021-08-11
Add UTs for ExtractResourceRestrictions

nessiLahav:
### What does this PR do?
Add UTs for ExtractResourceRestrictions

#### Change log

#### Test coverage

#### Documentation

#### API Changes

### DIFF
--- a/spec/app/domain/authentication/resource_restrictions/extract_resource_restrictions_spec.rb
+++ b/spec/app/domain/authentication/resource_restrictions/extract_resource_restrictions_spec.rb
@@ -1,0 +1,99 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe(Authentication::ResourceRestrictions::ExtractResourceRestrictions) do
+  let(:authenticator_name) { 'authn-name' }
+  let(:service_id) { 'service-id' }
+  let(:role_name) { 'role-name' }
+  let(:account) { 'account' }
+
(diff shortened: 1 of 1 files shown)


spec/app/domain/authentication/resource_restrictions/extract_resource_restrictions_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+    allow(mocked_fetch_resource_annotations_instance_empty).to(
+      receive(:call).and_return({})
+    )
+
+    allow(mocked_fetch_resource_annotations_instance_valid).to(
+      receive(:call).and_return(
+        "#{authenticator_name}/#{service_id}/claim1" => "claim1-value",
+        "#{authenticator_name}/#{service_id}/claim2" => "claim2-value"
```
sashaCher, line 35: Do you want to add annotation belongs to other service id and to validate that it's ignored
And to add not related annotation at all

nessiLahav: @sashaCher This is the responsibility of this class `GetServiceSpecificRestrictionFromAnnotation` and we already have UT for it

sashaCher: From one side you're right.
From other you do not mock it so the tests are running the original class behavior and this is the place where UT starting to look a bit more "integrationnal". Means we can catch if the way we're integrating classes together is right.
CC: @eladkug

sashaCher (approved):
LGTM

PR #2360, opened 2021-08-16
Default http error code in authenticate controller is 401

sashaCher:
### What does this PR do?

Aligns error handling in authenticate controller. It returns http error code 401 by default.

While serving authentication request conjur approaches OIDC provider in order to fetch keys.
During key fetching procedure exceptions could occur.
Before the change some exceptions were translated into [Bad Gateway](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) and [Gateway Timeout](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504) http error codes.

Strictly speaking conjur does not act as a [gateway](https://en.wikipedia.org/wiki/Gateway_(telecommunications)) during OIDC like authentication flows, it does not routes/transfers an authenticated request to the OIDC provider rather just fetches from the provider keys.

During authentication a REST caller is unauthenticated, means anonymous, until authentication flow succeeded. It's a bad practice to signalize anonymous user about internal server state. `authn-jwt` especially emphasizes the point when `provider-uri` and `jwks-uri` will return different http codes in case of similar situation (timeout for example).

In any case:
image
means in both unauthenticated and gateway error codes REST caller will demand to approach an another person with appropriate permissions in order to understand what's the issue.

[ONYX- 9758](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-9758)

#### Change log

#### Test coverage

#### Documentation

#### API Changes
- [X] The [OpenAPI spec](https://github.com/cyberark/conjur-openapi-spec) has been updated to meet new API changes (or an issue has been opened), or

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fix bug of cache not working in authn jwt. [cyberark/conjur#2353](https://github.com/cyberark/conjur/pull/2353)
 
+### Changed
+- OIDC based authenticators no longer return Bad Gateway and Gateway Timeout http error codes.
+  Unauthorised is returned instead.
+  [cyberark/conjur#2360](https://github.com/cyberark/conjur/pull/2360)
+
 ## [1.13.0] - 2021-07-29
 
 ### Added
--- a/app/controllers/authenticate_controller.rb
+++ b/app/controllers/authenticate_controller.rb
@@ -202,17 +202,6 @@ class AuthenticateController < ApplicationController
     log_backtrace(err)
 
     case err
-    when Errors::Authentication::Security::AuthenticatorNotWhitelisted,
-      Errors::Authentication::Security::WebserviceNotFound,
-      Errors::Authentication::Security::RoleNotFound,
-      Errors::Authentication::Security::AccountNotDefined,
-      Errors::Authentication::AuthnOidc::IdTokenClaimNotFoundOrEmpty,
-      Errors::Authentication::Jwt::TokenVerificationFailed,
-      Errors::Authentication::Jwt::TokenDecodeFailed,
-      Errors::Conjur::RequiredSecretMissing,
-      Errors::Conjur::RequiredResourceMissing
-      raise Unauthorized
-
     when Errors::Authentication::Security::RoleNotAuthorizedOnResource
       raise Forbidden
 
@@ -222,15 +211,8 @@ class AuthenticateController < ApplicationController
     when Errors::Authentication::Jwt::TokenExpired
       raise Unauthorized.new(err.message, true)
 
-    when Errors::Authentication::OAuth::ProviderDiscoveryTimeout
-      raise GatewayTimeout
-
     when Errors::Util::ConcurrencyLimitReachedBeforeCacheInitialization
       raise ServiceUnavailable
-
-    when Errors::Authentication::OAuth::ProviderDiscoveryFailed,
-      Errors::Authentication::OAuth::FetchProviderKeysFailed
-      raise BadGateway
 
     when Errors::Authentication::AuthnK8s::CSRMissingCNEntry,
       Errors::Authentication::AuthnK8s::CertMissingCNEntry
--- a/cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
+++ b/cucumber/authenticators_azure/features/authn_azure_bad_configuration.feature
@@ -124,7 +124,7 @@ Feature: Azure Authenticator - Bad authenticator configuration leads to an error
     Errors::Authentication::Security::RoleNotAuthorizedOnResource
     """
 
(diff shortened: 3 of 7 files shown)


app/controllers/authenticate_controller.rb
```diff
@@ -222,16 +211,9 @@ def handle_authentication_error(err)
     when Errors::Authentication::Jwt::TokenExpired
       raise Unauthorized.new(err.message, true)
 
-    when Errors::Authentication::OAuth::ProviderDiscoveryTimeout
-      raise GatewayTimeout
```
orenbm, line 226: @sashaCher can you please elaborate on this change? why are we not returning a gateway timeout error in this case?
image

this seems a much better error than the generic 401

app/controllers/authenticate_controller.rb
```diff
@@ -222,16 +211,9 @@ def handle_authentication_error(err)
...
-
     when Errors::Util::ConcurrencyLimitReachedBeforeCacheInitialization
       raise ServiceUnavailable
 
-    when Errors::Authentication::OAuth::ProviderDiscoveryFailed,
-      Errors::Authentication::OAuth::FetchProviderKeysFailed
-      raise BadGateway
```
orenbm, line 233: @sashaCher can you please elaborate on this change? why are we not returning a gateway timeout error in this case?
image

this seems a much better error than the generic 401

orenbm:
@sashaCher the changes in this PR must have a Changelog entry as they are customer-facing.

orenbm (changes requested):
left some comments

tzheleznyak (dismissed):
LGTM

app/controllers/authenticate_controller.rb
sashaCher: Updated PR description

app/controllers/authenticate_controller.rb
sashaCher: Updated PR description

sashaCher:
@orenbm I've updated the pr description and changelog

app/controllers/authenticate_controller.rb
jonahx: @sashaCher Can you add the same context to the body of the git commit message?  That, imo, is the right place to capture it so it will live forever and be easily accessible to devs.  What I typically do is write my explanation in a commit message, then just copy that into the main PR description.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Fix bug of cache not working in authn jwt. [ONYX-11330](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-11330)
+### Changed
+- OIDC based authenticators not longer return Bad Gateway and Gateway Timeout http error codes.
+Unauthorised is returned instead. [cyberark/conjur#2360](https://github.com/cyberark/conjur/pull/2360)
```
orenbm, line 13:
```suggestion
- OIDC based authenticators not longer return Bad Gateway and Gateway Timeout http error codes.
   Unauthorised is returned instead. 
   [cyberark/conjur#2360](https://github.com/cyberark/conjur/pull/2360)
```

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Fix bug of cache not working in authn jwt. [ONYX-11330](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-11330)
+### Changed
```
orenbm, line 11:
```suggestion

### Changed
```

CHANGELOG.md
sashaCher: Done

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- OIDC based authenticators not longer return Bad Gateway and Gateway Timeout http error codes.
```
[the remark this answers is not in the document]
sashaCher: Conjur does not have full `oAuth` implementation and we don't use this term anywhere in our docs...

CHANGELOG.md
sashaCher: Done

orenbm (changes requested):
Thanks @sashaCher , i'll approve once you fix the commit message to include the description

sashaCher:
@orenbm @jonahx I'll try do my best but commit message length is limited to 72 symbols.
image

orenbm:
> @orenbm @jonahx I'll try do my best but commit message length is limited to 72 symbols.
> image

the subject should still be the same. just add an empty line before the commit description and you'll be good

sashaCher:
@orenbm @jonahx I've updated the commit message

orenbm (dismissed):

orenbm (approved):

tzheleznyak (approved):

PR #2362, opened 2021-08-17
Replace Jira links in CHANGELOG

jtuttle:
In an effort to keep our work publicly-accessible, we should not include links
to our private Jira stories in the CHANGELOG. I've replaced them with links to
pull requests since I don't believe we have GitHub issues to match the Jira
issues.


diverdane (approved):
LGTM!!!

PR #2364, opened 2021-08-18
ONYX-9742: Reduce _class dependencies from command classes

tzheleznyak:
### What does this PR do?
Refactor to authn-jwt according this rule of thamb
image

Need to decide if really needed

ONYX-9742

#### Change log

#### Test coverage

#### Documentation

#### API Changes


tzheleznyak:
This hurts preformance because:
1. Factory created object created onec in compile level
2. We can't do memoization inside tham because it will cause bugs
3. We have repeated requests to secrets like "proivider-uri"/"jwks-uri"/"token-app-propert"

So the slight quality improvment here not worth the mutiple requests.

Maybe another solution that requeires design changes/refactors needed to address both _class init problem and preformance. But the solution in this PR not good enogh so it will be closed

PR #2372, opened 2021-09-20
Simplify add prefix

jonahx:
For the most part, this should be a pure refactor -- same functionality with different implementation.  There is one exception: I removed a couple of logging statements for events that didn't seem log-worthy to me.  But I can easily add those back if Hila or anyone else has objections.

## Big Lessons
	
- Avoid putting functional logic inside of command classes.
- Instead, pull it out into pure functions, because those are much easier to test (no mocks!)
	- The specific example in this PR was pulling the core logic from `AddPrefixToIdentity` into a simple function.  This makes the UTs drastically simpler.  The remaining validations on `identity_prefix` and `identity` were superfluous because they are already tested in the only client code  of that class (`IdentityFromDecodedTokenProvider`), which makes more sense anyway.
- Then let your command classes do almost nothing except the side effectful stuff -- let them delegate to pure functions for logic of any substantial complexity.
- So the UTs for command classes (or any side-effectful classes for that matter) should be almost nothing _but_ tests involving mocks. 

## Names and Abstraction
	
In general, we should try to keep the names as short as possible, and let context provide the details.  
	
We should also avoid introducing names (whether as constants, methods, or local variables) where they're not needed.  This is related to avoiding abstraction when it's not needed.   
	
Name-proliferation is most apparent in our existing code in the unit tests, where we sometimes begin a file  with 30 or more `let` statements -- a hundred lines of code just to define variables.  This is simply too much to keep track of, and is usually telling us that the tests should be structured differently.  See the new spec in the refactor for an example of that.
	
Another small example: we have a constant  `IDENTITY_PATH_CHARACTER_DELIMITER`. In addition to being an extremely long name, it's doubtful that this character will change, and so introducing an abstrction to protect against that change is unnecessary.  Instead, simply hardcode the `/` wherever it's used.  I didn't actually make this change yet in this PR.  But it probably should be made, and I noticed a number of other constants that are also likely unnecessary. 
	
Here's another more detailed example:
	
Consider the name `token_id_field_secret`....
	
This was the code's way of referring to the key of the claim that held the identity's value.  It _happens_ that we store that value as a Conjur secret.  But this doesn't have anything to do with what the value _is_, namely, a the key of a JWT claim.  So in the refactor, I've renamed it to the shorter, and I think clearer, `id_claim_key`.
	
Similarly, the name of the Conjur secret where we store that value is `token-app-property`.  This name was also confusing to me, but I couldn't figure out why it was called that, so I didn't change that one.

Another example is pattern is `FetchingIdentityByInterface` -- I don't think the `ByInterface` is adding much.  I think `FetchingIdentity` would convey the same information.  
	
In general, I think "Provider" or "Interface" rarely add clarity to names, but do make them harder to read.   If it seems they are important to a name, it's likely we should revisit the underlying concepts and try an entirely different name.

## Rules and when to break (or revisit) them
	
Before this PR, we had `fetch_identity_name_from_token`  and 
	
```ruby
def identity_name_from_token
  @identity_name_from_token || fetch_identity_name_from_token
end
```
I believe this grew from the (good) intention of following the rule whereby we name the memoized variable after the method name.  But in this case applying the rule made the code more cluttered, because it introduced an unnecessary method and made the distinction between the two methods confusing.  So we should add something like the following exception to our style guide:

> When a method with a side effect also sets the value of an instance variable, use a sensible name for the instance variable, rather than naming it after the method.

In our case the fix was:
```ruby
def fetch_id_from_token  
  return @id_from_token if @id_from_token
```
I renamed the method too for clarity, per the discussion above.


app/domain/authentication/authn_jwt/functions.rb
```diff
@@ -0,0 +1,24 @@
...
+      def path_join_two(prefix, main)
+        return main if String(prefix).strip.empty?
+        return prefix if String(main).strip.empty?
+
+        sep = IDENTITY_PATH_CHARACTER_DELIMITER
+        [prefix.sub(/#{sep}$/, ''), main.sub(/^#{sep}/, '')].join(sep)
+      end
+    end
+  end
+end
```
tzheleznyak, line 24: End of file

app/domain/authentication/authn_jwt/functions.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Authentication
+  module AuthnJwt
+    # Joins "prefix" and "main" as a path, with exactly one slash (/).  If
+    # prefix ends with a slash or host begins with one, they're removed to avoid
+    # duplicate slashes.
+    class << self
+      def path_join(*parts)
+        parts.reduce { |m, x| path_join_two(m, x) }
```
tzheleznyak, line 10: Maybe m,x can be replaced with text and concated? or other meaningful variable names?

spec/app/domain/authentication/authn-jwt/functions_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    'with a trailing slash on left part': [%w(path/ host), 'path/host'],
+    'with a leading slash on right part': [%w(path /host), 'path/host'],
+    'with slashes on both parts': [%w(path/ /host), 'path/host'],
+    'with multiple parts': [%w(a b c d), 'a/b/c/d'],
+    'with leading slash on first part': [%w(/a b c d), '/a/b/c/d'],
+    'with trailing slash on last part': [%w(a b c d/), 'a/b/c/d/'],
+    'with many parts and slashes': [%w(/a/b/c/ /d/e/ f/g), '/a/b/c/d/e/f/g'],
+  }
+
+  examples.each do |description, (input, output)|
```
tzheleznyak, line 17: I liked this way of writing short test file. Thank you.
CC : @cyberark/cerberus

app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
```diff
@@ -9,108 +9,78 @@ module IdentityProviders
           fetch_identity_path: Authentication::AuthnJwt::IdentityProviders::FetchIdentityPath.new,
           fetch_authenticator_secrets: Authentication::Util::FetchAuthenticatorSecrets.new,
           check_authenticator_secret_exists: Authentication::Util::CheckAuthenticatorSecretExists.new,
-          add_prefix_to_identity: Authentication::AuthnJwt::IdentityProviders::AddPrefixToIdentity.new,
           logger: Rails.logger
         },
         inputs: %i[jwt_authenticator_input]
       ) do
-        extend(Forwardable)
-        def_delegators(:@jwt_authenticator_input, :service_id, :authenticator_name, :account)
```
tzheleznyak, line 18: Why removing the deligator?

app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
```diff
@@ -9,108 +9,78 @@ module IdentityProviders
...
+            LogMessages::Authentication::AuthnJwt::CheckingIdentityFieldExists.new(id_claim_key)
+          )
+
+          @id_from_token = @jwt_authenticator_input.decoded_token[id_claim_key]
 
-          @logger.debug(LogMessages::Authentication::AuthnJwt::CheckingIdentityFieldExists.new(token_id_field_secret))
-          @identity_name_from_token = decoded_token[token_id_field_secret]
-          if @identity_name_from_token.blank?
-            raise Errors::Authentication::AuthnJwt::NoSuchFieldInToken, token_id_field_secret
+          # Equivalent to Rails blank? but without a Rails dependency.
```
tzheleznyak, line 57: Conjur is Rails application what the problem depanding on rails ?

tzheleznyak:
image

app/domain/authentication/authn_jwt/functions.rb
```diff
@@ -0,0 +1,24 @@
...
+      end
+
+      private
+
+      def path_join_two(prefix, main)
```
tzheleznyak, line 15: Maybe there is a File library in ruby that has option to join file paths and it can be used to join path here ?

app/domain/authentication/authn_jwt/functions.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
```
tzheleznyak, line 1: This is a good util functions that maybe can be used in other parts of conjur also and moving it out of the jwt authenticator will benefit our code base. I can see cases where we would like to concat paths in the future in other parts of the Conjur OSS.

app/domain/authentication/authn_jwt/functions.rb
dataplex: File paths depend on operating system though , and most filesystems are forgiving of repetitive path separators, whereas I don't believe Conjur would inherit that same forgiveness when executing database queries or parsing relationships. That said, if there isn't a utility class in the PolicyEngine gem, it may be time to add one starting with this function. I like that it is SRP and much more testable.

jvanderhoof: @tzheleznyak, Ruby has `File.join`, but it is not as flexible as this implementation.  Join does not handle nil or empty strings well and does not strip trailing slashes.

app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
jonahx: @tzheleznyak Nothing under `app/domain` should have any dependency on Rails.  Indeed, the whole point of that directory was to decouple from Rails.

From [`app/domain` README](https://github.com/cyberark/conjur/blob/master/app/domain/README.md#introduction):

> You can think of app/domain as a haven where pure Conjur business logic (the Conjur "domain") lives as plain ruby code, unpolluted by framework-specific concepts like controllers or routes. The code in this folder should have no dependencies on Rails itself (if you find any, they are likely mistakes we should fix).

One of the major downsides of Rails is the way that code from anywhere can implicitly access code from anywhere else -- creating a network of hidden dependencies.  You can't know what some piece of code depends on without reading through all of the code and manually checking all the references.   In the `app/domain` code, we want all the dependencies to be explicit, and so we have to opt out of the "Rails soup". 

We've even talked about pulling the authenticators out into their own gems, to _enforce_ this separation.  Which would be great.  But doing that adds complexity in other ways, so we haven't done yet.  So until we have a better solution, we just have to rely on developers knowing that this is a rule.

app/domain/authentication/authn_jwt/identity_providers/identity_from_decoded_token_provider.rb
jonahx: Good question, I meant to address this in my writeup.

Over time, I've come to the conclusion that using `Forwardable` is usually not worth it.  I have a  few reasons for this:

1. The "benefit" is minimal.  Ostensibly, the benefit is to make the calling site cleaner.  You can just write `my_method` instead of `@my_object.my_method`.  This is a minor benefit, unless you are calling `my_method` many times, which we usually aren't.
2. It adds a layer of abstraction for the reader.  If you've jumped into the code from somewhere else, you'll have to do an extra lookup to determine how the method is defined.  What's worse, you can't do this jump automatically, because....
3. It breaks tooling.  You can't use your IDE's "Jump to definition..." feature for methods defined this way.  Granted, one of the downsides of the CommandClass is that the same is true of jumping to dependencies, but there's no need to compound the problem.  Also, methods defined in this way don't show up in the "structure" view, which is a really nice way to understand classes:

image

app/domain/authentication/authn_jwt/functions.rb
jonahx: Absolutely, but I prefer to wait until there's a proven need for re-use before moving the file around to a new place.  That is, you might look at this, say, "Hey, this is general utility function, it doesn't belong here, it belongs on the root lib folder" or something similar.  But unless we already have a module  where it belongs, this usually ends up being a premature optimization, which has some costs.  Instead, I like to keep it at first close to where it's used.  Later, when we find ourselves _actually_ wanting to use it somewhere else, that's when I'd pull it out.

app/domain/authentication/authn_jwt/functions.rb
jonahx: @jvanderhoof @tzheleznyak You know, I think you're right about `File.join`.   Even though my implementation handles nils, I think we may want the client code to handle that anyway.  I'm going to try that tomorrow.  If it works, we'll be able to delete my code too!  More code deleted, the better.

tzheleznyak:
@jonahx it looks good just let me know what is decided here and i approve
https://github.com/cyberark/conjur/pull/2372#discussion_r711964300

jvanderhoof (approved):

PR #2415, opened 2021-11-05
Move GCP and OIDC tests out of the parallel run for consistent perfor…

ismarc:
…mance

### Desired Outcome

Move performance sensitive tests outside of the parallel block so that performance is reliable.

### Implemented Changes

Lift OIDC tests from the dynamically generated block of tests to its own separate stage, and include GCP in the grouping of stages but not in parallel.  This does not include changes that would test it on EE-FIPS and similar changes would need to be made there.

### Connected Issue/Story

Resolves ???

CyberArk internal issue link: [??]()

### Definition of Done
Tests that have performance sensitive components have reliable and repeatable test results
#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

- [X] No behavior was changed with this PR

#### Security

- [X] There are no security aspects to these changes

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -324,185 +324,204 @@ pipeline {
                 }
               }
             }
-            /**
-            * GCP Authenticator -- Token Stashing -- Stage 1 of 3
-            *
-            * In this stage, a GCE instance node is allocated, a script runs
-            * and retrieves all the tokens that will be used in authn-gcp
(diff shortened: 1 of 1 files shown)


nessiLahav (approved):
Looking good!!

Do you think we should also separate the performance tests to a different feature file?

PR #2421, opened 2021-11-12
Add check to not allow password updates for hosts

foxefj-cyber:
### Desired Outcome

Fix to not allow hosts to update a password since they aren't supposed to have passwords.

### Implemented Changes

Added check if the resource of the role is a host - then throw error

### Connected Issue/Story

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

- [X] This PR changes product behavior and has been reviewed by a PO, or

#### Security

### DIFF
--- a/spec/support/authentication.rb
+++ b/spec/support/authentication.rb
@@ -35,12 +35,18 @@ end
 
 shared_context "create user" do
   def create_user(login)
-    Role.create(role_id: "rspec:user:#{login}").tap do |role|
+    id = "rspec:user:#{login}"
+    user_role = Role.create(role_id: id).tap do |role|
       options = { role: role }
       options[:password] = password if defined?(password) && password
       Credentials.create(options)
       role.reload
     end
+    Resource.create(resource_id: id, owner_id: id).tap do |resource|
+      resource.reload
+    end
+
+    return user_role
   end
 
   let(:login) { "default-login" }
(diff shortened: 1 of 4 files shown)


jvanderhoof (approved):
This looks great @foxefj-cyber. Thanks for adding all the unit tests!!!

spec/support/authentication.rb
```diff
@@ -50,3 +56,39 @@ def create_user(login)
...
+    create_host(host_login)
+  }
+  let(:host_api_key) { the_host.credentials.api_key }
+end
+
+shared_context "host authenticate Basic" do
+  let(:params) { { account: host_account, authenticator: authenticator } }
+  let(:basic_auth_header) {
+    basic = Base64.strict_encode64(['host/' + host_login, basic_password].join(':'))
+    puts(basic)
```
jonahx, line 88: Looks like a debug statement you forgot to remove?

spec/support/authentication.rb
```diff
@@ -50,3 +56,39 @@ def create_user(login)
...
+    end
+
+    return host_role
+  end
+
+  let(:host_login) { "default-host-login" }
+  # 'let!' always runs before the example; 'let' is lazily evaluated.
+  let!(:the_host) {
+    create_host(host_login)
+  }
```
jonahx, line 80: 2 small things here:
- Formatting nit: If it can fit on 1 line within 80 chars, use 1 line.
- I would in general avoid all `!` methods in ruby, including this rspec DSL one.  If you need something done sequentially, move the statement into the test case itself.  The will make the logic more explicit.

spec/support/authentication.rb
```diff
@@ -50,3 +56,39 @@ def create_user(login)
...
+      Credentials.create(options)
+      role.reload
+    end
+    Resource.create(resource_id: id, owner_id: id).tap do |resource|
+      resource.reload
+      host_role.reload
+    end
+
+    return host_role
+  end
```
jonahx, line 74: This is a larger issue you couldn't possibly be expected to know about, especially as it's used in our own code bases, but I would consider `shared_context` in rspec to be an anti-pattern.  It is basically an rspec DSL reimplementation of a core ruby language feature, namely making a method or object available to some code via `require`.  This is perhaps a longer conversation but I'll just say that for now.

spec/support/authentication.rb
foxefj-cyber: removed

spec/support/authentication.rb
foxefj-cyber: sorry - was following the existing code in the test.  moved into the actual spec file

spec/support/authentication.rb
foxefj-cyber: i agree - this should probably be a larger conversation.

jvanderhoof (approved):

PR #2432, opened 2021-12-03
Add ability to send hostname for TLS SNI extension

foxefj-cyber:
### Desired Outcome

Ability to define the SNI certificate hostname so the proper certificate is returned from the Kubernetes API Server.

### Implemented Changes

particular, consider:*

Added change to send the hostname during SSL connection as well as adding testing.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-14386](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-14386)

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [X] Docs (e.g. `README`s) were updated in this PR

#### Behavior

- [X] This PR changes product behavior and has been reviewed by a PO, or

#### Security

### DIFF
--- a/app/domain/authentication/authn_k8s/k8s_object_lookup.rb
+++ b/app/domain/authentication/authn_k8s/k8s_object_lookup.rb
@@ -25,6 +25,10 @@ module Authentication
         @cert_store = OpenSSL::X509::Store.new
         @cert_store.set_default_paths
         ::Conjur::CertUtils.add_chained_cert(@cert_store, ca_cert)
+
+        return unless ENV.key?('SSL_CERT_DIRECTORY')
+
+        load_additional_certs(ENV['SSL_CERT_DIRECTORY'])
       end
 
       def bearer_token
@@ -183,6 +187,7 @@ module Authentication
           KubeClientFactory.client(
             api: 'apis/apps.openshift.io', version: 'v1', host_url: api_url,
             options: options
+
           )
         ]
       end
@@ -194,6 +199,16 @@ module Authentication
           yield
         rescue KubeException
           raise unless $!.error_code == 404
+        end
+      end
+
+      def load_additional_certs(ssl_cert_directory)
+        # allows us to add additional CA certs for things like SNI certs
+        if Dir.exist?("#{ssl_cert_directory}/ca")
+          Dir["#{ssl_cert_directory}/ca/*"].each do |file_name|
+            ::Conjur::CertUtils.add_chained_cert(@cert_store, File.read(file_name)) if
+              File.exist?(file_name)
+          end
         end
       end
     end
--- a/app/domain/authentication/authn_k8s/web_socket_client.rb
+++ b/app/domain/authentication/authn_k8s/web_socket_client.rb
@@ -26,7 +26,8 @@ module Authentication
                                 uri.port || (uri.scheme == 'wss' ? 443 : 80))
         if %w[https wss].include?(uri.scheme)
           ctx = OpenSSL::SSL::SSLContext.new
-          ctx.ssl_version = options[:ssl_version] || 'SSLv23'
+          ssl_version = options[:ssl_version] || 'SSLv23'
+          ctx.ssl_version = ssl_version
           ctx.verify_mode = options[:verify_mode] || OpenSSL::SSL::VERIFY_NONE # use VERIFY_PEER for verification
           cert_store = options[:cert_store]
           unless cert_store
@@ -34,8 +35,13 @@ module Authentication
             cert_store.set_default_paths
           end
           ctx.cert_store = cert_store
+
           @socket = ::OpenSSL::SSL::SSLSocket.new(@socket, ctx)
+          if ssl_version != 'SSLv23'
+            @socket.hostname = options[:hostname] || uri.host
+          end
           @socket.connect
+          @socket.post_connection_check(@socket.hostname) if @socket.hostname
         end
         @handshake = ::WebSocket::Handshake::Client.new(url: url, headers: options[:headers])
         @handshaked = false
--- a/cucumber/kubernetes/features/sni.feature
+++ b/cucumber/kubernetes/features/sni.feature
@@ -1,0 +1,12 @@
+Feature: A permitted Conjur host can authenticate with a valid resource restrictions
+  that is defined in the id and the kubernetes host has a corresponding SSL certificate
+
+  @sni_fails
+  Scenario: Authenticate as a Pod.
+    When I authenticate with authn-k8s as "pod/inventory-pod" without cert and key
+    Then the HTTP status is "401"
+
+  @sni_success
+  Scenario: Authenticate as a Pod.
+    Given I can login to pod matching "app=inventory-pod" to authn-k8s as "*/*"
+    Then I can authenticate pod matching "pod/inventory-pod" with authn-k8s as "*/*"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Nothing should go in this section, please add to the latest unreleased version
   (and update the corresponding date), or add a new version.
 
+## [1.17.1] - 2022-02-09
+
+### Added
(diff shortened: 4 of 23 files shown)


app/domain/authentication/authn_k8s/k8s_object_lookup.rb
```diff
@@ -25,6 +25,13 @@ def initialize(webservice = nil)
         @cert_store = OpenSSL::X509::Store.new
         @cert_store.set_default_paths
         ::Conjur::CertUtils.add_chained_cert(@cert_store, ca_cert)
+
+        # allows us to add additional CA certs for things like SNI certs
+        if ENV['SSL_CERT_DIRECTORY'] and File.exists?(ENV['SSL_CERT_DIRECTORY']) and File.exists?(ENV['SSL_CERT_DIRECTORY'] + '/ca')
```
jvanderhoof, line 30: To resolve the Code Climate errors (and to make it easier to test), what do you think about refactoring this to something like:

```ruby
        return unless ENV.key?('SSL_CERT_DIRECTORY')
        
        load_additional_certs(
          ssl_cert_directory: ENV['SSL_CERT_DIRECTORY'],
          cert_store: @cert_store,
          cert_utils: ::Conjur::CertUtils
        )
```
We can then break the work into a method like:
```ruby
      def load_additional_certs(ssl_cert_directory:, cert_store:, cert_utils:)
        certificate_directory = "#{ssl_cert_directory}/ca"
        if Dir.exist?(certificate_directory)
          Dir.entries(certificate_directory).each do |file_name|
            if File.exist?(file_name)
              cert_utils.add_chained_cert(cert_store, File.read(file_name))
            else
              # Log error message a particular file is missing
            end
          end
        else
          # Log error message that the environment variable SSL_CERT_DIRECTORY is missing
        end
      end
```

andytinkham (changes requested):
This work should have had a security review meeting before any changes were made. I have concerns about pulling in any cert in a directory. We'll need to meet to talk through some threat modeling. @foxefj-cyber - please book some time for us to talk through this.

jvanderhoof:
@andytinkham, I know you're just making sure we're not accidentally introducing a security concern with this: 
> This work should have had a security review meeting before any changes were made. 

Although it may have been helpful to have you in a meeting before @foxefj-cyber did any work, in reality, you'll need to be involved in most enhancement/bug fix conversation, which doesn't feel like the best use of your time. I do think certificate related work would certainly benefit from your insight (and seems to be the majority of our security related issues). Moving forward, what do you think about flagging all work that touches certificates as requiring early security architect collaboration?

@andytinkham, instead of pulling all certificates from a folder, what do you think about specifying a comma (or other separator) delineated list of certificates?  It would allow support for multiple certificate chains while ensuring only the specified certs were loaded.

Thoughts?

andytinkham:
> Although it may have been helpful to have you in a meeting before @foxefj-cyber did any work, in reality, you'd need to be involved in most enhancement/bug fix conversation, which doesn't feel like the best use of your time. I do think certificate related work would certainly benefit from your insight (and seems to be the majority of our security related issues).

> Moving forward, what do you think about flagging all work that touches certificates as requiring early security architect collaboration?

For new features, we should be doing security review & threat modeling discussions by default all the time with the occasional path to exception being determining we don't need one (as opposed to the occasional path being to do it). For bug fixes, I'd say if it touches certificates, TLS, encryption, authentication, authorization, or direct handling of secrets values, let's have at least a quick chat at the beginning - we can determine if more is needed after that chat. My guess is most of the time that will be sufficient.

> @andytinkham, instead of pulling all certificates from a folder, what do you think about specifying a comma (or other separator) delineated list of certificates? It would allow support for multiple certificate chains while ensuring only the specified certs were loaded.

Maybe. @foxefj-cyber & I will talk tomorrow. I'm still at the level of "Should I have concerns here?" not "I definitely have concerns here". Once I get more background, I can weigh in on this idea. Definitely seems worth talking about risks & tradeoffs between the 2 methods.

Thanks!

andytinkham (dismissed):
@foxefj-cyber @micahlee @jvanderhoof & I talked through things and I agree there is no further security issue here. Good to go from a security standpoint.

jvanderhoof (approved):
LGTMed!

micahlee (approved):

app/domain/authentication/authn_k8s/k8s_object_lookup.rb
```diff
@@ -25,6 +25,10 @@ def initialize(webservice = nil)
         @cert_store = OpenSSL::X509::Store.new
         @cert_store.set_default_paths
         ::Conjur::CertUtils.add_chained_cert(@cert_store, ca_cert)
+
+        return unless ENV.key?('SSL_CERT_DIRECTORY')
+
+        load_additional_certs(ENV['SSL_CERT_DIRECTORY'])
```
doodlesbykumbi, line 31: I might be missing something but is the singular call to `add_chained_cert` above not enough ? Wouldn't all the certs be stored in the Conjur variable or the cert file on disk when running inside a pod ?

Separately, I'm wondering about the documentation around the envvar `SSL_CERT_DIRECTORY`, I'm not familiar with it yet I see it being used in a few repos and without looking at the code here I wouldn't immediately know how it is used... and I'm not sure if its meaning in the context of the code here is shared across the different places it is used.

app/domain/authentication/authn_k8s/web_socket_client.rb
```diff
@@ -26,16 +26,22 @@ def connect(url, options = {})
...
             cert_store.set_default_paths
           end
           ctx.cert_store = cert_store
+
           @socket = ::OpenSSL::SSL::SSLSocket.new(@socket, ctx)
+          if ssl_version != 'SSLv23'
+            @socket.hostname = options[:hostname] || uri.host
+          end
           @socket.connect
+          @socket.post_connection_check(@socket.hostname) if @socket.hostname
```
doodlesbykumbi, line 44: I would like to understand this better, since it's a bit more involved than my simple implementation of always setting the hostname.
1. Why are we only excluding this version ? Is it the case that SNI is only expected to be relevant on `ssl_version > SSLv23` ?
2. Can you explain the logic around the `post_connection_check`

cucumber/kubernetes/features/sni.feature
```diff
@@ -0,0 +1,12 @@
...
+
+  @sni_fails
+  Scenario: Authenticate as a Pod.
+    When I authenticate with authn-k8s as "pod/inventory-pod" without cert and key
+    Then the HTTP status is "401"
+
+  @sni_success
+  Scenario: Authenticate as a Pod.
+    Given I can login to pod matching "app=inventory-pod" to authn-k8s as "*/*"
+    Then I can authenticate pod matching "pod/inventory-pod" with authn-k8s as "*/*"
```
doodlesbykumbi, line 12: I think the unit tests provide amazing validation and coverage for supporting SNI. I'm curious why we are adding a cucumber test which to me seems to be testing the same thing as the unit tests but paying a significantly higher price i.e. deploying a whole Conjur master for each one of these tests ?

doodlesbykumbi (commented):
Hey @foxefj-cyber. This feedback is a bit late but I wanted to make sure it was covered at some point.

cucumber/kubernetes/features/sni.feature
foxefj-cyber: these are just to validate that the entire solution works with SNI - purely for validation.

app/domain/authentication/authn_k8s/web_socket_client.rb
foxefj-cyber: SNI is an extension in TLS only so we don't want to bother adding the hostname if they are using SSLv23 for some reason.

the post_connection_check will throw an error if we expected a certain SNI certificate but we didn't get it - this makes sure that if we are expecting SNI we get the proper certificate

doodlesbykumbi: On board with (1). For (2) I was sure I was able to get errors thrown with just calls to `#connect`, I'll reconfirm.

PR #2451, opened 2021-12-29
Delete policy versions

eranha:
### Desired Outcome
Limit the number of Policy Versions

### Implemented Changes
Old version beyond the limit of 20 records are deleted.

*Describe how the desired outcome above has been achieved with this PR.
PolicyVersion model after save deleted old version beyond the limit of 20 records.

### Connected Issue/Story

[ONYX-13335 Default number of policy version](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-13335)
[Task ONYX-15328](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-15328)

### Definition of Done
merged.*
- [x] Upon policy version creation versions limit is checked and enforced
- [x] Deleted versions are written to log at info level
- [x] The versions limit can be set using an environment variable

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/app/models/policy_version.rb
+++ b/app/models/policy_version.rb
@@ -84,6 +84,58 @@ class PolicyVersion < Sequel::Model(:policy_versions)
   def before_save
     require 'digest'
     self.policy_sha256 = Digest::SHA256.hexdigest(policy_text)
+    delete_stale_versions
+  end
+
+  def delete_stale_versions
+    log_policy_version_candidates_for_deletion
+    delete_stale_versions
+  end
+
+  def delete_stale_versions(limit = policies_version_limit)
+    Sequel::Model.db[<<-SQL, resource_id, limit, resource_id].delete
+    WITH
+      "ordered_versions" AS
+        (SELECT * FROM "policy_versions" WHERE ("resource_id" = ?) ORDER BY "version" DESC LIMIT ?),
+      "delete_versions" AS
+        (SELECT * FROM "policy_versions" LEFT JOIN "ordered_versions" USING ("resource_id", "version") 
+       WHERE (("ordered_versions"."resource_id" IS NULL) AND ("resource_id" = ?)))
+    DELETE FROM "policy_versions"
+    USING  "delete_versions" 
+    WHERE "policy_versions"."resource_id" = "delete_versions"."resource_id" AND
+      "policy_versions"."version" = "delete_versions"."version"
+    SQL
+  end
+
+  def log_policy_version_candidates_for_deletion
+    stale_policy_versions.all.each do | policy_version |
+      Rails.logger.info( "Policy version: #{policy_version} will be deleted" )
+    end
+  end
+
+  def stale_policy_versions(limit = policies_version_limit)
+    Sequel::Model.db[<<-SQL, resource_id, limit, resource_id]
+    WITH
+      "ordered_versions" AS
+        (SELECT * FROM "policy_versions" WHERE ("resource_id" = ?) ORDER BY "version" DESC LIMIT ?),
+      "delete_versions" AS
+        (SELECT * FROM "policy_versions" LEFT JOIN "ordered_versions" 
+       USING (
+         "resource_id", 
+         "version", 
+         "created_at", 
+         "finished_at", 
+         "client_ip", 
+         "role_id", 
+         "policy_text", 
+         "policy_sha256"
+       ) 
+       WHERE (("ordered_versions"."resource_id" IS NULL) AND ("resource_id" = ?)))
+    SELECT * FROM "policy_versions"
+    JOIN  "delete_versions" USING ("resource_id", "version") 
+    WHERE "policy_versions"."resource_id" = "delete_versions"."resource_id" AND
+      "policy_versions"."version" = "delete_versions"."version"
+    SQL
   end
 
   def before_update
--- a/app/models/policy_version.rb
+++ b/app/models/policy_version.rb
@@ -84,55 +84,55 @@ class PolicyVersion < Sequel::Model(:policy_versions)
   def before_save
     require 'digest'
     self.policy_sha256 = Digest::SHA256.hexdigest(policy_text)
-    delete_stale_versions
   end
 
-  def delete_stale_versions
-    log_policy_version_candidates_for_deletion
-    delete_stale_versions
+  def after_save
+    log_versions_off_limit
+    delete_versions_off_limit
+  rescue => e
+    Rails.logger.error("Error while ensuring policy '#{id}' versions retention limit: #{e.inspect}")
   end
 
-  def delete_stale_versions(limit = policies_version_limit)
-    Sequel::Model.db[<<-SQL, resource_id, limit, resource_id].delete
+  def log_versions_off_limit
+    versions_off_limit.all.each do |policy_version|
+      Rails.logger.debug(
+        "Deleting policy version: #{policy_version.slice(
+          :version,
+          :resource_id,
+          :role_id,
+          :created_at,
+          :client_ip
+        )}]"
+      )
+    end
+  end
+
+  def versions_off_limit
+    Sequel::Model.db[<<-SQL, resource_id, 20, resource_id]
+       WITH
+      "ordered_versions" AS (SELECT * FROM "policy_versions" WHERE ("resource_id" = ?) ORDER BY "version" DESC LIMIT ?)
+      SELECT * FROM "policy_versions" LEFT JOIN "ordered_versions" 
+      USING (
+        "resource_id", 
+        "version",
+        "role_id",
+        "created_at",
+        "client_ip"
+      )
+      WHERE (("ordered_versions"."resource_id" IS NULL) AND ("resource_id" = ?))
+    SQL
+  end
+
+  def delete_versions_off_limit
+    Sequel::Model.db[<<-SQL, resource_id, policies_version_limit, resource_id].delete
     WITH
       "ordered_versions" AS
         (SELECT * FROM "policy_versions" WHERE ("resource_id" = ?) ORDER BY "version" DESC LIMIT ?),
       "delete_versions" AS
         (SELECT * FROM "policy_versions" LEFT JOIN "ordered_versions" USING ("resource_id", "version") 
-       WHERE (("ordered_versions"."resource_id" IS NULL) AND ("resource_id" = ?)))
+         WHERE (("ordered_versions"."resource_id" IS NULL) AND ("resource_id" = ?)))
     DELETE FROM "policy_versions"
     USING  "delete_versions" 
-    WHERE "policy_versions"."resource_id" = "delete_versions"."resource_id" AND
-      "policy_versions"."version" = "delete_versions"."version"
-    SQL
-  end
-
-  def log_policy_version_candidates_for_deletion
-    stale_policy_versions.all.each do | policy_version |
-      Rails.logger.info( "Policy version: #{policy_version} will be deleted" )
-    end
-  end
-
-  def stale_policy_versions(limit = policies_version_limit)
-    Sequel::Model.db[<<-SQL, resource_id, limit, resource_id]
-    WITH
-      "ordered_versions" AS
-        (SELECT * FROM "policy_versions" WHERE ("resource_id" = ?) ORDER BY "version" DESC LIMIT ?),
-      "delete_versions" AS
-        (SELECT * FROM "policy_versions" LEFT JOIN "ordered_versions" 
-       USING (
-         "resource_id", 
-         "version", 
-         "created_at", 
-         "finished_at", 
-         "client_ip", 
-         "role_id", 
-         "policy_text", 
-         "policy_sha256"
-       ) 
-       WHERE (("ordered_versions"."resource_id" IS NULL) AND ("resource_id" = ?)))
-    SELECT * FROM "policy_versions"
-    JOIN  "delete_versions" USING ("resource_id", "version") 
     WHERE "policy_versions"."resource_id" = "delete_versions"."resource_id" AND
       "policy_versions"."version" = "delete_versions"."version"
     SQL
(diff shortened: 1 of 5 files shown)


app/models/policy_version.rb
```diff
@@ -86,6 +86,61 @@ def before_save
...
+  rescue => e
+    Rails.logger.error("Error while ensuring policy '#{id}' versions retention limit: #{e.inspect}")
+  end
+
+  def ensure_versions_limit
+    log_versions_off_limit
+    delete_versions_off_limit
+  end
+
+  def log_versions_off_limit
```
orenbm, line 100: what's versions off limit? are these the versions that are beyond the defined limit? e.g version 21?

orenbm: also - what do we log for each one?

app/models/policy_version.rb
```diff
@@ -86,6 +86,61 @@ def before_save
     self.policy_sha256 = Digest::SHA256.hexdigest(policy_text)
   end
 
+  def after_save
+    ensure_versions_limit
+  rescue => e
+    Rails.logger.error("Error while ensuring policy '#{id}' versions retention limit: #{e.inspect}")
+  end
+
+  def ensure_versions_limit
```
orenbm, line 95: this method is redundant imo. you can implement `log_versions_off_limit` and `delete_versions_off_limit` in `after_save` and it will be clearer without letting the user figure out what this ensure is

orenbm:
Nice work @eranha !

Are you planning to add tests to this PR?

app/models/policy_version.rb
```diff
@@ -86,6 +86,61 @@ def before_save
     self.policy_sha256 = Digest::SHA256.hexdigest(policy_text)
   end
 
+  def after_save
+    ensure_versions_limit
+  rescue => e
+    Rails.logger.error("Error while ensuring policy '#{id}' versions retention limit: #{e.inspect}")
```
InbalZilberman, line 92: @eranha  is this a log? If so please put Shuli as reviewer
Also for errors i think you need to have an error code and so on.
what is the reason for the error? what do you expect the user to do?

app/models/policy_version.rb
eranha: Yes for both questions

app/models/policy_version.rb
eranha: This is not expected to happen, but I don't want an error to fail the policy load.

app/models/policy_version.rb
```diff
@@ -86,6 +86,50 @@ def before_save
...
+  def log_versions_off_limit
+    versions_off_limit.all.each do |policy_version|
+      Rails.logger.debug("Deleting policy version: #{policy_version.except(:policy_text, :policy_sha256, :finished_at)}")
+    end
+  end
+
+  def versions_off_limit
+    Sequel::Model.db[<<-SQL, resource_id, policies_version_limit, resource_id]
+       WITH
+      "ordered_versions" AS (SELECT * FROM "policy_versions" WHERE ("resource_id" = ?) ORDER BY "version" DESC LIMIT ?)
```
orenbm, line 105: in the following method you indented this like this. please align between the 2
```suggestion
    Sequel::Model.db[<<-SQL, resource_id, policies_version_limit, resource_id]
    WITH
      "ordered_versions" AS (SELECT * FROM "policy_versions" WHERE ("resource_id" = ?) ORDER BY "version" DESC LIMIT ?)
```

app/models/policy_version.rb
```diff
@@ -86,6 +86,50 @@ def before_save
     self.policy_sha256 = Digest::SHA256.hexdigest(policy_text)
   end
 
+  def after_save
+    log_versions_off_limit
```
orenbm, line 90: how expensive is this log? i see that you make another call to the DB for this

cucumber/api/features/policy_versions_retention.feature
```diff
@@ -0,0 +1,24 @@
...
+    """
+  Scenario: I update a policy multiple times so it exceeds the default policy versions limit and get the default
+            limited number of versions in response when I retrieve the policy resource
+    Given I successfully POST 21 times "/policies/cucumber/policy/policy_test_versions" with body:
+    """
+      - !user bob
+    """
+    Then the HTTP response status code is 201
+    And I successfully GET "/resources/cucumber/policy/policy_test_versions"
+    Then there are 20 "policy_versions" in response
```
eladkug, line 20: what about the logs? we suppose to see on policy is deleted as well

shulifink:
@eranha  - what do you want me to review?

config/initializers/limits.rb
```diff
@@ -4,3 +4,8 @@
 def secrets_version_limit
   ( ENV['SECRETS_VERSION_LIMIT'] || 20 ).to_i
 end
+
+def policies_version_limit
+  ( ENV['POLICIES_VERSION_LIMIT'] || 20 ).to_i
```
eladkug, line 9: 20 need to be as variable otherwise we will need to change it every time.

app/models/policy_version.rb
```diff
@@ -86,6 +86,58 @@ def before_save
...
+    delete_versions_off_limit
+  rescue => e
+    Rails.logger.error("Error while ensuring policy '#{id}' versions retention limit: #{e.inspect}")
```
eranha, line 93: @shulifink

app/models/policy_version.rb
```diff
@@ -86,6 +86,58 @@ def before_save
...
+  end
+
+  def log_versions_off_limit
+    versions_off_limit.all.each do |policy_version|
+      Rails.logger.debug(
+        "Deleting policy version: #{policy_version.slice(
```
eranha, line 99: @shulifink

cucumber/api/features/policy_versions_retention.feature
```diff
@@ -0,0 +1,29 @@
...
+  Each update of a policy creates a policy version record and policy log records. This feature limits
+  the number of versions in the database to default value of 20 version, older versions are deleted.
+
+  Background:
+    Given I am the super-user
+    And I successfully POST "/policies/cucumber/policy/root" with body:
+    """
+      - !policy policy_test_version
+    """
+  Scenario: I update a policy multiple times so it exceeds the default policy versions limit and get the default
```
orenbm, line 12: please add empty line before this

cucumber/api/features/policy_versions_retention.feature
```diff
@@ -0,0 +1,29 @@
...
+    And I successfully GET "/resources/cucumber/policy/policy_test_version"
+    Then there are 20 "policy_versions" in response
+    And The following appears in the log after my savepoint:
+    """
+    Deleting policy version: {:version=>1, :resource_id=>"cucumber:policy:policy_test_version",
+    """
+
+
+
+
```
orenbm, line 29: please have only one newline

cucumber/api/features/step_definitions/response_steps.rb
```diff
@@ -12,6 +12,10 @@
   expect(@result.map{|r| r['id']}).to include(*@resources.values.map{|r| r.id})
 end
 
+Then(/^there are (\d+) "([^"]*)" in response$/) do |occurrences, key|
```
orenbm, line 15:
```suggestion
Then(/^there are (\d+) "([^"]*)" in the response$/) do |occurrences, key|
```

app/models/policy_version.rb
jvanderhoof: @eranha, what about changing these names to something like `log_versions_to_expire` and `remove_expired_policy_versions` for these two method names? `log_versions_off_limit` and `delete_versions_off_limit` don't give me a lot of context to what these methods might do.

app/models/policy_version.rb
```diff
@@ -86,6 +86,58 @@ def before_save
...
+          :resource_id,
+          :role_id,
+          :created_at,
+          :client_ip
+        )}]"
+      )
+    end
+  end
+
+  def versions_off_limit
```
jvanderhoof, line 110: What do you think about changing this name to `current_versions`?  It looks like it returns the last 20 versions in the `policy_versions` table.  `versions_off` doesn't immediately give me much insight into what it might return.

app/models/policy_version.rb
jvanderhoof: Isn't `versions_off_limit` the current set of policy versions we want to keep?

jvanderhoof:
@eranha, the delete functionality here might make a lot more sense to implement using the Command class. It doesn't really directly relate to the purpose of the current model, which is to persist data to the database. It'll also be easier to test.

app/models/policy_version.rb
eranha: latest_version?

app/models/policy_version.rb
eranha: no its the ones we delete

app/models/policy_version.rb
eranha: Great names

eranha:
> @eranha, the delete functionality here might make a lot more sense to implement using the Command class. It doesn't really directly relate to the purpose of the current model, which is to persist data to the database. It'll also be easier to test.

@jvanderhoof Can you please elaborate? A new Command class or an existing one? called from within PolicyVersion?

app/models/policy_version.rb
shulifink: @eranha  - can you show an example of the error with all the variables filled in?

app/models/policy_version.rb
shulifink: @eranha - can you show an example of the error with all the variables filled in?

shulifink (commented):
@eranha   can you please give me examples of these errors with all the variables filled in?

jvanderhoof:
@eranha, something like this (pseudo-code):
```rb
class DeleteExpiredPolicyVersions
  extend CommandClass::Include
  
  command_class(
    dependencies: { sequel: Sequel::Model, retention_limit: 20},
    inputs: %i[]
  ) do
    def call
      log_versions_to_expire
      remove_expired_versions
    rescue => e
      Rails.logger.error("Error while ensuring policy '#{id}' versions retention limit: #{e.inspect}")
    end

    private

    def remove_expired_versions
      @sequel.db[<<-SQL, resource_id, @retention_limit, resource_id].delete
      WITH
        "ordered_versions" AS
          (SELECT * FROM "policy_versions" WHERE ("resource_id" = ?) ORDER BY "version" DESC LIMIT ?),
        "delete_versions" AS
          (SELECT * FROM "policy_versions" LEFT JOIN "ordered_versions" USING ("resource_id", "version") 
          WHERE (("ordered_versions"."resource_id" IS NULL) AND ("resource_id" = ?)))
      DELETE FROM "policy_versions"
      USING  "delete_versions" 
      WHERE "policy_versions"."resource_id" = "delete_versions"."resource_id" AND
        "policy_versions"."version" = "delete_versions"."version"
      SQL
    end

    def log_versions_to_expire
      ...
    end

    ...
  end
end
``` 
This can be called from the Policy Controller, at the end of the policy save. Optionally, it could be included in the `after_save` block.

It'll let us keep the Policy Version class more of a simple data binding object rather than have it become concerned with lifecycle cleanup. A command class would also simplify the process of adding an external cleanup command (cron, evoke, etc.) if we ever chose to go that route.

config/initializers/limits.rb
orenbm: i think it can stay as 20 as this is the place where we set the variable

app/models/policy_version.rb
eranha: @shulifink 
` Error while ensuring policy 'cucumber:policy:root' versions retention limit: #<Sequel::DatabaseError: PG::SyntaxError: ERROR:  syntax error at or near "WITH_"`

cucumber/api/features/policy_versions_retention.feature
```diff
@@ -0,0 +1,26 @@
...
+    """
+
+  Scenario: I update a policy multiple times so it exceeds the default policy versions limit and get the default
+            limited number of versions in response when I retrieve the policy resource
+    Given I save my place in the log file
+    And I successfully POST 21 times "/policies/cucumber/policy/policy_test_version" with body:
+    """
+      - !user bob
+    """
+    Then the HTTP response status code is 201
```
orenbm, line 20: This is not what you test here imo. you test that we remove the 21st version so i'd change this to be Given (and) and let the next Then be the first one.
```suggestion
    And the HTTP response status code is 201
```

orenbm: anyway you shouldn't have 2 `Then`s

app/models/policy_version.rb
```diff
@@ -86,6 +86,58 @@ def before_save
...
+          :created_at,
+          :client_ip
+        )}]"
+      )
+    end
+  end
+
+  def expired_versions
+    Sequel::Model.db[<<-SQL, resource_id, policies_version_limit, resource_id]
+      WITH
```
orenbm, line 112: can you align the indentations of this query and the next one?

app/models/policy_version.rb
shulifink: Error while enforcing version retention limit for policy 'cucumber:policy:root'.

app/models/policy_version.rb
```diff
@@ -86,6 +86,58 @@ def before_save
...
+    log_versions_to_expire
+    remove_expired_versions
+  rescue => e
+    Rails.logger.error("Error while ensuring policy '#{id}' versions retention limit: #{e.inspect}")
+  end
+
+  def log_versions_to_expire
+    expired_versions.all.each do |policy_version|
+      Rails.logger.debug(
+        "Deleting policy version: #{policy_version.slice(
```
shulifink, line 99: "Delete policy version..."  is fine - no need for example.

shulifink (changes requested):

shulifink (approved):

orenbm (dismissed):

orenbm (approved):

orenbm (approved):

Issue #2381, opened 2021-10-06
Docker container fails to restart


Infinoid:
## Summary

The [`cyberark/conjur` docker container](https://hub.docker.com/r/cyberark/conjur/) does not restart gracefully.  It leaves a stale pidfile behind, and then refuses to start.

## Steps to Reproduce

1.  Follow the [quickstart setup instructions](https://www.conjur.org/get-started/quick-start/oss-environment/).
2.  Restart the server container: `docker restart conjur_server`

(Restarting the docker host machine is also sufficient to reproduce the problem.)

## Expected Results

Conjur server serves conjurs after the restart.

## Actual Results (including error logs, if applicable)

Server did not restart properly.  Clients get "connection refused" attempting to contact the server.

`docker logs conjur_server` contains an error message about how the PID file already exists.

This log contains output from both the first run and the second (failed) one:

```
authn-local is listening at /run/authn-local/.socket
=> Booting Puma
=> Rails 5.2.6 application starting in production 
=> Run `rails server -h` for more startup options
[19] Puma starting in cluster mode...
[19] * Puma version: 5.3.2 (ruby 2.5.8-p224) ("Sweetnighter")
[19] *  Min threads: 5
[19] *  Max threads: 5
[19] *  Environment: development
[19] *   Master PID: 19
[19] *      Workers: 2
[19] *     Restarts: (✔) hot (✖) phased
[19] * Preloading application
[19] * Listening on http://0.0.0.0:80
[19] Use Ctrl-C to stop
CONJ00038I OpenSSL FIPS mode set to true
Loaded configuration:
- trusted_proxies from defaults
- authenticators from defaults
[19] - Worker 0 (PID: 24) booted in 0.0s, phase: 0
Loaded configuration:
- trusted_proxies from defaults
- authenticators from defaults
[19] - Worker 1 (PID: 28) booted in 0.0s, phase: 0
error: SIGTERM
A server is already running. Check /opt/conjur-server/tmp/pids/server.pid.
=> Booting Puma
=> Rails 5.2.6 application starting in production 
=> Run `rails server -h` for more startup options
Exiting
authn-local is listening at /run/authn-local/.socket
```

Search the above for `server.pid`.

## Reproducible

I don't know if it's 100%, but it occurs at least 50% of the time for me.  Happens often for me, for the past year or more, whenever system updates on the docker host machine require a reboot.

## Version/Tag number

[...]

PR #2498, opened 2022-02-16
Remove the stale puma pidfile before starting the server process

ajay-vel:
#### Desired Outcome
- This changes ensures that the stale pidfile is removed before starting the conjur server process
This issue addresses https://github.com/cyberark/conjur/issues/2381

#### Implemented Changes
- A new method is added to conjur-cli to check if the stale puma pid file exists and deletes it if it exists.

#### Connected Issue/Story
- Issue https://github.com/cyberark/conjur/issues/2381

#### Definition of Done
- A new method checks if the stale pidfile exists and it is removed before the starting the conjur server. 

#### Changelog
- The CHANGELOG has been updated with the bug that was fixed.

#### Test coverage
- This PR includes two new unit tests to go with the code
  changes

#### Documentation
- This PR does not require updating any documentation

#### Behavior
-  No behavior was changed with this PR

#### Security
- There are no security aspects to these changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - IAM Authn bug fix - Take rexml gem to production configuration [#2493](https://github.com/cyberark/conjur/pull/2493)
+- Previously, a stale puma pid file would prevent the Conjur server from starting 
+  successfully. Conjur now removes a stale pid file at startup, if it exists.
+  [#2498](https://github.com/cyberark/conjur/pull/2498)
 
 ### Security
 - Updated nokogiri to 1.13.3 to resolve CVE-2022-23308 and CVE-2021-30560
--- a/bin/conjur-cli/commands/server.rb
+++ b/bin/conjur-cli/commands/server.rb
@@ -31,6 +31,9 @@ module Commands
       create_account
       load_bootstrap_policy
 
+      # Remove a stale puma PID file, if it exists
+      cleanup_pidfile
+
       # Start the Conjur API and service
       # processes
       fork_server_process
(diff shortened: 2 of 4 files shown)


bin/conjur-cli/commands/server.rb
```diff
@@ -85,6 +88,16 @@ def load_bootstrap_policy
       ) || exit(($CHILD_STATUS.exitstatus))
     end
 
+    def cleanup_pidfile
+      Process.fork do
```
doodlesbykumbi, line 92: Fly by comment... Could this not be achieved without forking the process ?

bin/conjur-cli/commands/server.rb
```diff
@@ -85,6 +88,16 @@ def load_bootstrap_policy
       ) || exit(($CHILD_STATUS.exitstatus))
     end
 
+    def cleanup_pidfile
```
micahlee, line 91: I would suggest adding a larger comment above the method describing why this is needed (the error observed), and why this solution fixes it.

bin/conjur-cli/commands/server.rb
```diff
@@ -85,6 +88,16 @@ def load_bootstrap_policy
...
+      Process.fork do
+        puts("Cleaning up the pidfile")
+
+        exec("
+          rm -f /opt/conjur-server/tmp/pids/server.pid 
+        ")
```
micahlee, line 97: I would suggest first checking if the stale PID file exists, and only print the message and delete it if it does. This will provide more information later if we're reviewing logs for a customer, for example.

We also don't need to exec to delete the file, we should do that just within ruby:

```ruby
pid_file_path = '/opt/conjur-server/tmp/pids/server.pid'
if File.exist?(pid_file_path)
  puts("Removing existing PID file: #{pid_file_path}")
  File.delete(pid_file_path)
end
```

micahlee (changes requested):

micahlee:
Hey @ajay-vel , another thing I just realized is that this PR description is not the standard template. Can you please use that instead?

There are some important questions there for both you and the reviewer to consider for each PR. For example, this change does need entry in the CHANGELOG.

Thanks!

```md
### Desired Outcome

### Implemented Changes

particular, consider:*

### Connected Issue/Story

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

```

micahlee:
One other thing I remembered seeing the PR checklist again is that we should add tests for this change. They can probably go in the existing spec suite here: https://github.com/cyberark/conjur/blob/master/spec/conjurctl/server_spec.rb

We should verify the behavior:
- when the PID file does exist
- when the PID file does not exist

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - IAM Authn bug fix - Take rexml gem to production configuration [#2493](https://github.com/cyberark/conjur/pull/2493)
+- Bug fix that removes stale puma pidfile[#2381](https://github.com/cyberark/conjur/issues/2381)
```
micahlee, line 50: Couple of things here:
- Ideally, a "fixed" changelog entry should contain both the problem and what was changed to fix it.
- A small formatting note, there should be space before the PR link.
- The link should be to the PR containing the change, rather than the issue.

This would come together as something like:
```markdown
- Previously, a stale puma pid file would prevent the Conjur server from starting 
  successfully. Conjur now removes a stale pid file at startup, if it exists.
  [#2498](https://github.com/cyberark/conjur/pull/2498)
```

micahlee (commented):
Hey @ajay-vel , thanks for updating the tests! One last comment on the changelog, and this should be good to go!

micahlee (approved):
LGTM, thanks for the effort getting this wrapped up!

PR #2501, opened 2022-02-24
Add Prometheus Metrics endpoint and client store

gl-johnson:
### Desired Outcome

Create the groundwork for Conjur Telemetry and metrics collection using Prometheus. For more details on this feature, see the [Solution Design.](https://github.com/cyberark/conjur/blob/metrics-solution-design/design/telemetry/metrics_solution_design.md#solution-design---conjur-metrics-telemetry)

### Implemented Changes
Added middleware (ConjurExporter) to expose a /metrics endpoint to be used by Prometheus, as well as initialization methods to create a client store for metric collection, and relevant unit tests.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-16193](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-16193)

### Definition of Done

- [x] Create a Prometheus client store
- [x] Expose the Prometheus client store via the GET /metrics scrape target
- [x] Add test cases to ensure the scrape target works as expected

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -24,6 +24,12 @@ require 'conjur_audit'
 # Must require because lib folder hasn't been loaded yet
 require './lib/conjur/conjur_config'
 
+# Require prometheus dependencies and metrics module
+# so that a clean data store can be initialized
+# This should be done dynamically depending on whether
+# metrics are enabled in the future
+require './lib/monitoring/prometheus'
+
 module Conjur
   class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
@@ -74,5 +80,8 @@ module Conjur
     # We create this in application.rb instead of an initializer so that it's
     # guaranteed to be available for other initializers to use.
     config.conjur_config = Conjur::ConjurConfig.new
+
+    # Initialize metrics and clean existing data
+    Monitoring::Prometheus.setup
   end
 end
--- a/config/initializers/rack_middleware.rb
+++ b/config/initializers/rack_middleware.rb
@@ -27,6 +27,11 @@ Rails.application.configure do
   # to the start of the Rack middleware chain.
   config.middleware.insert_before(0, ::Rack::DefaultContentType)
 
+  # If using Prometheus telemetry, we want to ensure that the middleware
+  # which collects and exports metrics is loaded at the start of the 
+  # middleware chain to prevent any modifications to the incoming requests
+  config.middleware.insert_before(0, Monitoring::Middleware::PrometheusExporter, registry: Monitoring::Prometheus.registry, path: '/metrics')
+
   # Deleting the RemoteIp middleware means that `request.remote_ip` will
   # always be the same as `request.ip`. This ensure that the Conjur request log
   # (using `remote_ip`) and the audit log (using `ip`) will have the same value
--- a/Gemfile
+++ b/Gemfile
@@ -81,6 +81,8 @@ gem 'openid_connect'
 gem "anyway_config"
 gem 'i18n', '~> 1.8.11'
 
+gem 'prometheus-client'
+
 group :development, :test do
   gem 'aruba'
   gem 'ci_reporter_rspec'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -290,6 +290,7 @@ GEM
       ast (~> 2.4.1)
     pg (1.2.3)
     powerpack (0.1.3)
+    prometheus-client (3.0.0)
     pry (0.13.1)
       coderay (~> 1.1)
       method_source (~> 1.0)
@@ -505,6 +506,7 @@ DEPENDENCIES
   openid_connect
   parallel
   pg
+  prometheus-client
   pry-byebug
   pry-rails
   puma (~> 5.5, >= 5.6.2)
--- a/lib/monitoring/middleware/prometheus_exporter.rb
+++ b/lib/monitoring/middleware/prometheus_exporter.rb
@@ -1,0 +1,89 @@
+require 'prometheus/client'
+require 'prometheus/client/formats/text'
+
+module Monitoring
+  module Middleware
+    # Exporter is a Rack middleware that provides a sample implementation of a
+    # Prometheus HTTP exposition endpoint.
+    #
+    # By default it will export the state of the global registry and expose it
+    # under `/metrics`. Use the `:registry` and `:path` options to change the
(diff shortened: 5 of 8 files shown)


config.ru
```diff
@@ -3,4 +3,11 @@
 # This file is used by Rack-based servers to start the application.
 
 require ::File.expand_path('../config/environment', __FILE__)
+
+#require 'rack'
+require ::File.expand_path('../lib/monitoring/middleware/conjur_exporter.rb', __FILE__)
+
+#use Rack::Deflater
+use Monitoring::Middleware::ConjurExporter
```
doodlesbykumbi, line 11: Looks like we already have a dedicated place for applying middleware, see https://github.com/cyberark/conjur/blob/master/config/initializers/rack_middleware.rb#L24-L28. I think we would want to have similar code as we do for `::Rack::DefaultContentType` so that the exporter middleware runs before anything else.

lib/monitoring/metrics.rb
```diff
@@ -0,0 +1,51 @@
+require 'prometheus/client'
+require 'prometheus/client/data_stores/direct_file_store'
+
+module Monitoring
+  module Metrics
```
doodlesbykumbi, line 5: I think it might be good to name this Prometheus so that its clear what flavour of metrics you're getting.

lib/monitoring/middleware/conjur_exporter.rb
```diff
@@ -0,0 +1,93 @@
...
+        end
+
+        quality
+      end
+
+      def respond_with(format)
+        [
+          200,
+          { 'Content-Type' => format::CONTENT_TYPE },
+          [format.marshal(Monitoring::Metrics.registry)]
```
doodlesbykumbi, line 67: Thoughts about dependency injecting this via the constructor. I think it makes this easier to test and decouples it from the Metrics module. It looks like other parts of the code expect this to be passed in as an option

doodlesbykumbi (commented):
Left some comments

config.ru
gl-johnson: Good point - I imagine this is especially true if using middleware to collect metrics on incoming requests, in case something further down the middleware chain were to modify the request. Do you think the same precedence is needed for the exporter middleware?

lib/monitoring/middleware/conjur_exporter.rb
gl-johnson: Addressed in #[0417de0](https://github.com/cyberark/conjur/pull/2501/commits/0417de010938d472b8e2d16d4c142bb09e71928a)

gl-johnson:
@doodlesbykumbi Thanks for the feedback. Most/all of it should be addressed in [0417de0](https://github.com/cyberark/conjur/pull/2501/commits/0417de010938d472b8e2d16d4c142bb09e71928a), let me know anything doesn't look right

lib/monitoring/middleware/conjur_exporter.rb
```diff
@@ -0,0 +1,94 @@
...
+
+module Monitoring
+  module Middleware
+    # Exporter is a Rack middleware that provides a sample implementation of a
+    # Prometheus HTTP exposition endpoint.
+    #
+    # By default it will export the state of the global registry and expose it
+    # under `/metrics`. Use the `:registry` and `:path` options to change the
+    # defaults.
+    class ConjurExporter
```
doodlesbykumbi, line 12: Thinking we should rename this `PrometheusExporter` to more accurately capture what it represents

config/initializers/rack_middleware.rb
```diff
@@ -27,6 +27,11 @@
   # to the start of the Rack middleware chain.
   config.middleware.insert_before(0, ::Rack::DefaultContentType)
 
+  # If using Prometheus telemetry, we want to ensure that the middleware
+  # which collects and exports metrics is loaded near the start of the 
+  # middleware chain to prevent any modifications to the incoming requests
+  config.middleware.insert_before(1, Monitoring::Middleware::ConjurExporter)
```
doodlesbykumbi, line 33: Why 1, I would have thought 0 ? Are we relying on `::Rack::DefaultContentType` ? 

```suggestion
  config.middleware.insert_before(0, Monitoring::Middleware::ConjurExporter)
```

config/application.rb
```diff
@@ -74,5 +80,8 @@ class Application < Rails::Application
     # We create this in application.rb instead of an initializer so that it's
     # guaranteed to be available for other initializers to use.
     config.conjur_config = Conjur::ConjurConfig.new
+
+    # Initialize metrics and clean existing data
+    Monitoring::Prometheus.setup
```
doodlesbykumbi, line 85: ✅  Seems like a good location for this since right above it we get access to the config

lib/monitoring/middleware/conjur_exporter.rb
```diff
@@ -0,0 +1,94 @@
...
+      attr_reader :app, :path, :registry
+
+      FORMATS  = [::Prometheus::Client::Formats::Text].freeze
+      FALLBACK = ::Prometheus::Client::Formats::Text
+
+      def initialize(app, options = {})
+        @app = app
+        @path = options[:path] || '/metrics'
+        @port = options[:port]
```
doodlesbykumbi, line 21: We can probably remove anything related to the port option, it feels like an additional toggle that I don't think we'll be using at all.

lib/monitoring/middleware/conjur_exporter.rb
```diff
@@ -0,0 +1,94 @@
...
+        @registry = options[:registry] || Monitoring::Prometheus.registry
```
doodlesbykumbi, line 22: My thoughts are that it might be better for us to explicitly pass in these values via the constructor and not have defaults. For 2 reasons
1. Avoid the reader having to look at both the constructor and the call site to understand the options used. For the explicit case they just need to look at the call site
2. This might be a weaker reason, but... Avoid implicitly coupling default option values with the exporter definition. 

I realise this is subjective so it's your choice how to proceed.

doodlesbykumbi (commented):
Thanks for addressing those earlier comments @gl-johnson. Left some new comments

doodlesbykumbi (approved):
Thanks for addressing comments. LGTM!

PR #2506, opened 2022-03-02
Ofira block create user1

oburstein-hub:
### Desired Outcome

The direct outcome of this PR is to limit users creation to be allowed only under root if the environment variable CONJUR_USERS_IN_ROOT_POLICY_ONLY is true.
If the CONJUR_USERS_IN_ROOT_POLICY_ONLY is undefined user creation behaviour is allowed (does not change)

### Implemented Changes

Logics has beed added in types.rb under User verification - throwing exception when user creation fails

### Definition of Done

Validation fully implemented and testes

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


app/models/loader/types.rb
```diff
@@ -201,6 +201,12 @@ class User < Record
 
       # Below is a sample method verifying policy data validity
       def verify
+        user_creation_allowed = ENV['CONJUR_ALLOW_USER_CREATION']
```
jvanderhoof, line 204: might want to change this to something like:

```suggestion
        user_creation_allowed = ENV.fetch('CONJUR_ALLOW_USER_CREATION', 'true').downcase
```
to allow variations of false below.

jvanderhoof (changes requested):
Looks good. Please add some positive and negative tests for the `verify` method we changed. The integration test is helpful, but requires policy load actions which are quiet slow (relative to functional tests).

app/models/loader/types.rb
oburstein-hub: done

spec/models/loader/types.rb
```diff
@@ -0,0 +1,37 @@
+require 'spec_helper'
+
+describe Loader::Types do
+  context "Check user creation verification" do
+    def new_user()
```
jvanderhoof, line 5: Typically this is done with a `let` block instead of a method:
```ruby
let(:new_user) {  Loader::Types::User.new(id: 'alice') }
```

spec/models/loader/types.rb
```diff
@@ -0,0 +1,37 @@
+require 'spec_helper'
+
+describe Loader::Types do
+  context "Check user creation verification" do
```
jvanderhoof, line 4: As we're unit testing, it can be helpful to define the method before the context:
```rb
describe Loader::Types do
  describe '.check_user_creation_allowed' do
    context 'when user creation is disabled' do
       ...
```
Also, context should clearly describe the state we're testing

spec/models/loader/types.rb
```diff
@@ -0,0 +1,37 @@
...
+      user = new_user()
+      status='success'
+      begin
+        user.check_user_creation_allowed(res_id: 'alice@my_sub_tree')
+      rescue Exceptions::InvalidPolicyObject => exc
+        status='failure'
+      ensure
+        ENV['CONJUR_ALLOW_USER_CREATION'] = 'true'
+      end
+      expect(status).to eq('failure')
```
jvanderhoof, line 20: We can simplify the above by mocking the environment variable:
```ruby
allow(ENV).to receive(:[]).with('CONJUR_ALLOW_USER_CREATION').and_return('false')
```
and checking for he exception:
```ruby
expect { new_user.check_user_creation_allowed(res_id: 'alice@my_sub_tree') }.to raise_error(Exceptions::InvalidPolicyObject)
```

jvanderhoof (changes requested):
Better Spec: https://www.betterspecs.org/ is a great guide to RSpec best practices.  There's quiet a bit of simplification and cleanup we can do to ensure the tests are easily readable by other engineers.

dev/docker-compose.yml
```diff
@@ -43,6 +43,7 @@ services:
       # TODO: Where should we be running rspec tests from, ideally?
       # See https://github.com/DatabaseCleaner/database_cleaner#safeguards
       DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: "true"
+      CONJUR_ALLOW_USER_CREATION: false
```
jvanderhoof, line 46: As this is primarily for SaaS, let's default this to `true` if at all.

spec/models/loader/types.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+    it "should allow user creation not under root" do
```
jvanderhoof, line 22: As we're testing the same method (`check_user_creation_allowed`), this should be a context.

spec/models/loader_orchestrate_spec.rb
```diff
@@ -127,4 +127,24 @@ def verify_data(path)
       end
     end
   end
+
+  context "with non-root user policy" do
+    let(:base_policy_path) { 'empty.yml' }
+    it "loads the minimal policy" do
+      expect(resource_policy).to be
+      verify_data 'base/empty.txt'
```
jvanderhoof, line 135: What's happening here with this check?  If I'm reading it correctly, let's get parentheses to improve readability:

```ruby
expect(resource_policy).to be verify_data('base/empty.txt')
```

spec/models/loader_orchestrate_spec.rb
```diff
@@ -127,4 +127,24 @@ def verify_data(path)
...
+      begin
+        ENV['CONJUR_ALLOW_USER_CREATION'] = 'false'
+        replace_policy_with 'non_root_user.yml'
+      rescue Exceptions::InvalidPolicyObject => exc
+        status='failure'
+      ensure
+        ENV['CONJUR_ALLOW_USER_CREATION'] = 'true'
+      end
+      expect(status).to eq('failure')
+    end
```
jvanderhoof, line 148: Same suggestion as above, let's mock the environment variable and check for the expected exception.

app/models/loader/types.rb
```diff
@@ -199,8 +199,18 @@ def create!
     class User < Record
       def_delegators :@policy_object, :public_keys, :account, :role_kind, :uidnumber, :restricted_to
 
+      def check_user_creation_allowed(res_id:)
+        user_creation_allowed = ENV.fetch('CONJUR_ALLOW_USER_CREATION', 'true').downcase
+        if user_creation_allowed == 'false' && res_id.include?('@')  # not under root
+          message = "User creation is disallowed - please address administator"
```
jvanderhoof, line 205: This reads a bit rough.  What about something like:
```
User creation through policy is disabled. 
```

spec/models/loader/types.rb
oburstein-hub: done

spec/models/loader/types.rb
oburstein-hub: done

spec/models/loader/types.rb
oburstein-hub: done

dev/docker-compose.yml
oburstein-hub: OK - I will remove it from here at all in order to keep the existing conjur-oss functionality

spec/models/loader/types.rb
oburstein-hub: done

spec/models/loader_orchestrate_spec.rb
oburstein-hub: removed this code block - it is not nessesary

spec/models/loader_orchestrate_spec.rb
oburstein-hub: removing all this test

app/models/loader/types.rb
oburstein-hub: done

app/models/loader/types.rb
```diff
@@ -199,8 +199,18 @@ def create!
     class User < Record
       def_delegators :@policy_object, :public_keys, :account, :role_kind, :uidnumber, :restricted_to
 
+      def check_user_creation_allowed(res_id:)
+        user_creation_allowed = ENV.fetch('CONJUR_ALLOW_USER_CREATION', 'true').downcase
+        if user_creation_allowed == 'false' && res_id.include?('@')  # not under root
```
jvanderhoof, line 204: Although not conventional, I really don't agree with the change to prevent users from being created outside of the `root` policy.  As this breaks existing functionality, if we're going to do this, we need to do it in a very visible way and with alignment from PO/PM. Customer need to be aware it could break their existing workflows.

I know that users need to be part of the root policy in order to work with Idaptive. If we also want to limit created users to the `root` policy, what do you think about extending our settings to prevent creation of users outside of the `root` policy?

app/models/loader/types.rb
```diff
@@ -199,8 +199,18 @@ def create!
     class User < Record
       def_delegators :@policy_object, :public_keys, :account, :role_kind, :uidnumber, :restricted_to
 
+      def check_user_creation_allowed(res_id:)
```
jvanderhoof, line 202: Abbreviations make code harder to read.  Mind changing `res_id` to `resource_id` (assuming that's what you're abbreviating)?

app/models/loader/types.rb
jvanderhoof: Upon re-reading this code, I realize I misread the intent of this block.  Instead of including the check for a user in a non-root policy, how about returning, something like:

```ruby
def check_user_creation_allowed(resource_id:)
  return if resource_id.include?('@')

  if ENV.fetch('CONJUR_ALLOW_USER_CREATION', 'true').downcase == 'false'
    raise Exceptions::InvalidPolicyObject.new(
      res_id,
      message: 'User creation through policy is disabled.'
    )
  end
end
```
Personally, it helps me to see where the exits are without needing to think through the `&&` portion of the `if` statement.  it also highlights that we need separate unit tests for the possible results of `check_user_creation_allowed`:

1. User is outside the root policy => do nothing
2. User is inside the root policy and `CONJUR_ALLOW_USER_CREATION` is not set => do nothing
3. User is inside the root policy and `CONJUR_ALLOW_USER_CREATION` is not set to `true` => do nothing
4. User is inside the root policy and `CONJUR_ALLOW_USER_CREATION` is not set to false => throw exception

spec/models/loader/types.rb
```diff
@@ -0,0 +1,22 @@
...
+describe Loader::Types do
+
+  let(:new_user) {  Loader::Types::User.new(id: 'alice') }
+
+  describe '.check_user_creation_allowed' do
+    context "Check user creation verification" do
+
+      it "should not allow user creation not under root" do
+        allow(ENV).to receive(:[]).with('CONJUR_ALLOW_USER_CREATION').and_return('false')
+        user = new_user()
```
jvanderhoof, line 12: looks like `user` is never being used.  We can remove it.

spec/models/loader/types.rb
```diff
@@ -0,0 +1,22 @@
+require 'spec_helper'
+
+describe Loader::Types do
+
+  let(:new_user) {  Loader::Types::User.new(id: 'alice') }
+
+  describe '.check_user_creation_allowed' do
+    context "Check user creation verification" do
+
+      it "should not allow user creation not under root" do
```
jvanderhoof, line 10: Even better, if we use a subject we can remove a lot of this with something like:
```ruby
  describe '.check_user_creation_allowed' do
    context 'when the user is loaded in the "my_sub_tree" policy' do
      subject { Loader::Types::User.new(id: 'alice@my_sub_tree').check_user_creation_allowed }

      it { is_expected.not_to raise_error(Exceptions::InvalidPolicyObject)}
    end
  
    context 'when the user is loaded into the "root" policy' do
      subject { Loader::Types::User.new(id: 'alice') }
      allow(ENV).to receive(:[]).with('CONJUR_ALLOW_USER_CREATION').and_return(setting)

      context "when the CONJUR_ALLOW_USER_CREATION environment variable is false" do
        let(:setting) { 'false' }
        
        it { is_expected.to raise_error(Exceptions::InvalidPolicyObject)}
      end

      context "when the CONJUR_ALLOW_USER_CREATION environment variable is true" do
        let(:setting) { 'true' }
        
        it { is_expected.to raise_error(Exceptions::InvalidPolicyObject)}
      end
      ...
    end
  end

```

jvanderhoof:
@oburstein-hub, this statement:
>> The direct outcome of this PR is to limit users creation to be allowed only under root if the environment variable CONJUR_ALLOW_USER_CREATION is false.

doesn't make any sense to me. The name `CONJUR_ALLOW_USER_CREATION` insinuates it changes the ability to allow users to be created. If we want to prevent users from being created outside the `root` policy, how about changing the name to something like `CONJUR_USERS_IN_ROOT_POLICY_ONLY`, or something along those lines.

jvanderhoof:
If you rebase your changes against the `master` branch, Jenkins should run these tests for you.

app/models/loader/types.rb
oburstein-hub: Hi Jason
The required behavior is different then you describe
It is:
1. User is inside the root policy => do nothing
2. User is outside the root policy and CONJUR_ALLOW_USER_CREATION is not set => do nothing
3. User is outside the root policy and CONJUR_ALLOW_USER_CREATION is not set to true => do nothing
4. User is outside the root policy and CONJUR_ALLOW_USER_CREATION is not set to false => throw exception

Yet - I can align my code to the format you suggested

oburstein-hub: Regarding the influence of this change on conjur-oss behaviour - there should not be any influence there (you thought it will be because your description of login did not mach the actual implementation

app/models/loader/types.rb
oburstein-hub: done

spec/models/loader/types.rb
oburstein-hub: fixed

spec/models/loader/types.rb
```diff
@@ -0,0 +1,22 @@
+require 'spec_helper'
+
+describe Loader::Types do
+
+  let(:new_user) {  Loader::Types::User.new(id: 'alice') }
+
+  describe '.check_user_creation_allowed' do
+    context "Check user creation verification" do
```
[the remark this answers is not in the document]
oburstein-hub: done

oburstein-hub:
Found a problem with your original suggestion regarding substitution of ENV['CONJUR_ALLOW_USER_CREATION'] by 
ENV.fetch('CONJUR_ALLOW_USER_CREATION', 'true').downcase.
It returns different values in unittest run:

Please, see the prints:
ENV.fetch('CONJUR_ALLOW_USER_CREATION', 'true').downcase = true
ENV['CONJUR_ALLOW_USER_CREATION'] = false

Should I return it to be ENV['CONJUR_ALLOW_USER_CREATION']  - as it is the only one that works ?

@jvanderhoof

spec/models/loader/types.rb
oburstein-hub: Getting error on above:
 `allow` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).

oburstein-hub:
> CONJUR_USERS_IN_ROOT_POLICY_ONLY

changing

oburstein-hub:
> Just ran this locally playing around with some improvements to the tests and discovered:
> 
> ```ruby
> check_user_creation_allowed(resource_id: resourceid)
> ```
> 
> fails with the following:
> 
> ```shell
> Failure/Error: check_user_creation_allowed(resource_id: resourceid)
>      
>      NoMethodError:
>        undefined method `resourceid' for {:id=>"alice@my_sub_tree"}:Hash
> ```
> 
> Looks like the created user `Loader::Types::User.new(id: 'alice')` doesn't correctly extract the resource id. It may be that you need the account and types.

This is quite strange..
I ran this function both as part of unit test and in a conjur server run - and everything works OK ..

jvanderhoof (commented):
This PR needs to be rebased before the merge can occur, but the implementation looks solid.  Nice work!

jvanderhoof (dismissed):

PR #2512, opened 2022-03-08
Add Pub/Sub util and Collector middleware for metrics collection

john-odonnell:
### Desired Outcome

- Add dep-injectable Pub/Sub mechanism based on `ActiveSupport::Notifications`
- Add Collector middleware that can initialize metrics
- Metric updates through the Collector should be exposed through the Exporter from #2501 

### Implemented Changes

Adds:
- `lib/monitoring/pub_sub.rb`
  - Wraps `ActiveSupport::Notifications` to event subscription and publishing
  - Test cases:
    - Published events trigger execution of subscribed code blocks
    - Published events with extra provided content (payload) trigger execution of subscribed code blocks that can use payload content
- `lib/monitoring/middleware/conjur_collector.rb`
  - Custom collector middleware
  - Can initialize metrics and subscribe code blocks
  - Traces requests to Conjur and can provide identifying data to subscribed code blocks
  - The collector middleware is injected behind the exporter middleware, and does not trace requests to the metrics scrape endpoint
  - Test cases:
    - Metrics initialized by the collector can be triggered when corresponding events are published
    - Metrics updated through the collector are exposed through the registry linked to the Exporter middleware

### Connected Issue/Story

CyberArk internal issue link: [ONYX-16194](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-16194)

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


lib/monitoring/middleware/conjur_collector.rb
```diff
@@ -0,0 +1,65 @@
...
+
+module Monitoring
+  module Middleware
+    class ConjurCollector
+      attr_reader :app, :registry, :pubsub
+
+      def initialize(app, options = {})
+        @app = app
+        @registry = options[:registry]
+        @pubsub = options[:pubsub]
```
gl-johnson, line 12: I don't think there is a need to pass in a pubsub object here since in the rest of the class it looks like you are making static method calls on the Monitoring::PubSub module. Plus it would make the pubsub mechanism a bit more intrusive if we needed to rely on dependency injection for any modules/classes which need to be able to publish events.

lib/monitoring/middleware/conjur_collector.rb
```diff
@@ -0,0 +1,65 @@
+require 'benchmark'
+require ::File.expand_path('../../pub_sub.rb', __FILE__)
+
+module Monitoring
+  module Middleware
+    class ConjurCollector
```
gl-johnson, line 6: Kumbi mentioned this on my PR and I tend to agree - the naming of this class should probably reflect flavor of metrics we are implementing (Prometheus)

lib/monitoring/middleware/conjur_collector.rb
```diff
@@ -0,0 +1,65 @@
...
+        init_test_metric
+      end
+
+      def call(env)
+        trace(env) { @app.call(env) }
+      end
+
+      protected
+
+      def init_test_metric
```
gl-johnson, line 23: I'm wondering if this should be consolidated under prometheus.rb? There is already a test metric there (not using pub/sub), plus I think it helps to think about it in a more generic sense, because presumably there will be metrics collected for events beyond just incoming requests

lib/monitoring/middleware/conjur_collector.rb
john-odonnell: Once I take your advice below (moving metric init to `prometheus.rb`) the collector shouldn't be beholden to Prometheus. Should be able to trace and publish events with only subscribers needing to know how they're used.

lib/monitoring/prometheus.rb
```diff
@@ -1,5 +1,6 @@
 require 'prometheus/client'
 require 'prometheus/client/data_stores/direct_file_store'
+require 'monitoring/pub_sub'
 
 module Monitoring
   module Prometheus
```
jvanderhoof, line 6: Please try to avoid grouping methods using Modules.  I understand it's a common pattern in Conjur, but it greatly increases the complexity by decoupling methods from the place they are called. Methods grouped with modules are also significantly harder to test as we need to create a dummy class which includes the module in order to test the module methods.

Instead, I'd recommend creating a utility class which encapsulates the implementation details for for the publication. This is a super rough re-work of the code here, but I recommend something like:

```ruby
module Monitoring
  module Prometheus
    class Setup
      attr_reader :registry, :metrics_prefix

      def initialize(registry: ::Prometheus::Client::Registry.new, metrics_prefix: 'conjur_http_server')
        @registry = registry
        @metrics_prefix = metrics_prefix
        @metrics_dir_path = ENV['CONJUR_METRICS_DIR'] || '/tmp/prometheus'  
      end

      def setup
        clear_data_store
        configure_data_store
        init_metrics
      end

      protected

      def clear_data_store
        Dir[File.join(@metrics_dir_path, '*.bin')].each do |file_path|
          File.unlink(file_path)
        end
      end

      def configure_data_store
        ::Prometheus::Client.config.data_store = ::Prometheus::Client::DataStores::DirectFileStore.new(
          dir: @metrics_dir_path
        )
      end

      def init_metrics
        # Test a random gauge metric
        gauge = registry.gauge(:test_gauge, docstring: '...', labels: [:test_label])
        gauge.set(1234.567, labels: { test_label: 'gauge metric test' })

        # Test for pub/sub-based, collector-triggered metric updates
        PubSub.unsubscribe("collector_test_metric")
        request_counter = registry.counter(
          :collector_test_metric,
          docstring: '...',
          labels: [:code, :path]
        )
        PubSub.subscribe("collector_test_metric") do |payload|
          labels = {
            code: payload[:code],
            path: payload[:path]
          }
          request_counter.increment(labels: labels)
        end
      end
    end
  end
end
```

As I reorganized this, it's clear this is a command class (handles setup). Modules are really intended for shared methods (an alternative composition to inheritance).

lib/monitoring/pub_sub.rb
```diff
@@ -0,0 +1,22 @@
+module Monitoring
+    # Module PubSub wraps ActiveSupport::Notifications, providing PubSub
+    # plumbing to custom Controllers or Collectors.
+    module PubSub
```
jvanderhoof, line 4: It looks like you're using a module as a singleton here, which is a touch cleaner than creating a singleton class. We do need to consider a Singleton is really a global variable. We need to be careful to not create an invisible, tightly coupled dependency throughout our system.

As an alternative, this will allow you to test this singleton with unit tests, but also allow it to be used as singleton globally:
```ruby
module Monitoring
    # Module PubSub wraps ActiveSupport::Notifications, providing PubSub
    # plumbing to custom Controllers or Collectors.
    class PubSubBase
      def publish(name, payload = {}, &block)
        ActiveSupport::Notifications.instrument(name, payload, &block)
      end

      def subscribe(name)
        ActiveSupport::Notifications.subscribe(name) do |_, _, _, _, payload|
          yield payload
        end
      end

      def unsubscribe(name)
        ActiveSupport::Notifications.unsubscribe(name)
      end
    end

    class PubSub < PubSubBase
      include Singleton
    end
  end
end
```

spec/monitoring/middleware/conjur_collector_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'spec_helper'
+require 'rack/test'
+require 'monitoring/prometheus'
+require 'monitoring/middleware/conjur_collector'
+require 'monitoring/middleware/prometheus_exporter'
+
+describe Monitoring::Middleware::ConjurCollector do
```
jvanderhoof, line 7: If you haven't come across it yet, [BetterSpec](https://www.betterspecs.org/) is a great resource for RSpec best practices.

spec/monitoring/middleware/conjur_collector_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    Monitoring::Middleware::PrometheusExporter.new(app_with_collector, **exporter_options)
+  end
+
+  it 'returns the app response' do
+    get '/foo'
+    expect(last_response).to be_ok
+    expect(last_response.body).to eql('OK')
+  end
+
+  it 'metrics from traced requests update the global registry' do
```
jvanderhoof, line 33: This test looks a lot like an integration test.  It's testing that the `record` method correctly publishes an event (and the event is picked up in the registry).  Although this is fine (and certainly has its place in testing), we're skipping tests on all the methods in `Monitoring::Middleware::ConjurCollector`.  

Can you please include some more granular tests to verify happy and sad paths of the methods in `ConjurCollector`?

spec/monitoring/pub_sub_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rack/test'
+require 'spec_helper'
+require 'monitoring/pub_sub'
+require 'monitoring/prometheus'
+
+describe Monitoring::PubSub do
+  include Rack::Test::Methods
```
jvanderhoof, line 7: Same comment as above. This is really an integration test. Alternatively, you might want to mock the `ActiveSupport::Notifications` methods we're wrapping in `Monitoring::PubSub`. That'll allow you test individual methods more robustly.

spec/monitoring/pub_sub_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+
+      expect(publish_counter).to eql(0)
+      pubsub.publish(name)
+      expect(publish_counter).to eql(1)
+
+      pubsub.unsubscribe(name)
+      pubsub.publish(name)
+      expect(publish_counter).to eql(1)
+    end
```
doodlesbykumbi, line 57: You can avoid the counter by asserting on the arguments you expect the block to receive
```
       // the block argument on the block passed to expect is watched by rspec
        expect { |block| 
            name = "ok"

            // pass in the watched block
            pubsub.subscribe(name, &block)

            pubsub.publish(name, {e:1})
            pubsub.publish(name, {e:2})

            pubsub.unsubscribe(name)

            pubsub.publish(name, {e:3})
        }
        .to yield_successive_args({e:1}, {e:2})
```

spec/monitoring/metrics_spec.rb
```diff
@@ -1,15 +1,77 @@
...
+      :test_gauge,
+      docstring: '...',
+      labels: [:test_label]
+    ))
+
+    pubsub.subscribe("sample_test_gauge") do |payload|
+      metric = registry.get(:test_gauge)
+      metric.set(payload[:value], labels: payload[:labels])
+    end
+  end
```
doodlesbykumbi, line 18: I found myself wondering if we're doing too much here. I imagined it might be enough for the #setup unit tests to check if this method is called with the expected registry and pubsub.

I think I might have written in the issue or the design doc about the idea of being able to validate that pubsub events can result in changes on the registry. So I can see the value of those comprehensive tests on actual metric definitions (instead of on the test metric). My thoughts are we can keep this but strip it down to the bare minimum once we have an actual metric

lib/monitoring/pub_sub.rb
```diff
@@ -0,0 +1,33 @@
...
+require 'active_support/notifications'
+
+module Monitoring
+  # PubSub and PubSubBase wrap ActiveSupport::Notifications, providing pub/sub
+  # plumbing to custom controllers and collectors.
+
+  class PubSubBase
+    attr_reader :options
+
+    def initialize(notifications = ActiveSupport::Notifications)
```
doodlesbykumbi, line 11: I'm thinking this dependency injection of `notifications` might not be necessary since this particular version of pub/sub is fundamentally tied to `ActiveSupport::Notifications`. There's no other notifications object that will have the method signatures for `instrument`, `subscribe` etc.

doodlesbykumbi (commented):
Thanks for reworking this @john-odonnell. Left some comments mainly around addressing some small-ish testing considerations. Happy to have synchronous chat to wrap things up

doodlesbykumbi (commented):
Small comments to round off otherwise LGTM!

PR #2521, opened 2022-03-25
Configurable metrics collectrion

gl-johnson:
### Desired Outcome
Prometheus telemetry should be configurable via the Conjur configuration system which leverages [Anyway Config](https://github.com/palkan/anyway_config). This includes the following two options for enabling the telemetry feature (in order of precedence):
- As an environment variable: 
  - CONJUR_TELEMETRY_ENABLED=true
- In the configuration file (/etc/conjur/config/conjur.yml): 
  - telemetry_enabled: true

NOTE: If neither of the above are configured it will default to 'false'

As with all configurations, at this time, this must be set for every instance of the Conjur OSS Server. It’s not replicated across instances. After making one of the two changes above, simply apply the updated configuration using conjurctl configuration apply (or by restarting the Conjur processes).

At this point, Conjur will be able to collect metrics and tally them in Prometheus client store. Also, the /metrics endpoint is enabled and will no longer report a 401.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -24,12 +24,6 @@ require 'conjur_audit'
 # Must require because lib folder hasn't been loaded yet
 require './lib/conjur/conjur_config'
 
-# Require prometheus dependencies and metrics module
-# so that a clean data store can be initialized
-# This should be done dynamically depending on whether
-# metrics are enabled in the future
-require './lib/monitoring/prometheus'
-
 module Conjur
   class Application < Rails::Application
     # Settings in config/environments/* take precedence over those specified here.
@@ -81,7 +75,5 @@ module Conjur
     # guaranteed to be available for other initializers to use.
     config.conjur_config = Conjur::ConjurConfig.new
 
-    # Initialize metrics and clean existing data
-    Monitoring::Prometheus.setup
   end
 end
--- a/config/initializers/rack_middleware.rb
+++ b/config/initializers/rack_middleware.rb
@@ -27,11 +27,6 @@ Rails.application.configure do
   # to the start of the Rack middleware chain.
   config.middleware.insert_before(0, ::Rack::DefaultContentType)
 
-  # If using Prometheus telemetry, we want to ensure that the middleware
-  # which collects and exports metrics is loaded at the start of the 
-  # middleware chain to prevent any modifications to the incoming requests
-  config.middleware.insert_before(0, Monitoring::Middleware::PrometheusExporter, registry: Monitoring::Prometheus.registry, path: '/metrics')
-
   # Deleting the RemoteIp middleware means that `request.remote_ip` will
   # always be the same as `request.ip`. This ensure that the Conjur request log
   # (using `remote_ip`) and the audit log (using `ip`) will have the same value
--- a/spec/lib/monitoring/middleware/prometheus_exporter_spec.rb
+++ b/spec/lib/monitoring/middleware/prometheus_exporter_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
+require 'monitoring/middleware/prometheus_exporter'
 
 describe Monitoring::Middleware::PrometheusExporter do
-  include Rack::Test::Methods
 
   # Reset the data store
   before do
@@ -16,17 +16,21 @@ describe Monitoring::Middleware::PrometheusExporter do
 
   let(:options) { { registry: registry, path: path} }
 
+  let(:env) { Rack::MockRequest.env_for }
+
   let(:app) do
-    app = ->(_) { [200, { 'Content-Type' => 'text/html' }, ['OK']] }
-    described_class.new(app, **options)
+    app = ->(env) { [200, { 'Content-Type' => 'text/html' }, ['OK']] }
   end
+
+  subject { described_class.new(app, **options) }
 
   context 'when requesting app endpoints' do
     it 'returns the app response' do
-      get '/foo'
+      env['PATH_INFO'] = "/foo"
+      status, _headers, _response = subject.call(env)
 
-      expect(last_response).to be_ok
-      expect(last_response.body).to eql('OK')
+      expect(status).to eql(200)
+      expect(_response.first).to eql('OK')
     end
   end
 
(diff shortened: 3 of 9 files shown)


dev/docker-compose.yml
```diff
@@ -178,6 +178,29 @@ services:
...
       - ../ci/jwt/:/usr/src/jwks/
 
+
+  prometheus:
+    image: prom/prometheus:v2.1.0
+    command:
+      - '--config.file=/etc/prometheus/prometheus.yml'
+      - '--storage.tsdb.path=/prometheus'
+      - '--web.console.libraries=/usr/share/prometheus/console_libraries'
+      - '--web.console.templates=/usr/share/prometheus/consoles'
```
doodlesbykumbi, line 188: Unless I'm missing something let's try and maintain the scope of the PR. I think changes to this file, and other's under the dev folder related to running prometheus locally, are not needed for this PR.

config/initializers/rack_middleware.rb
```diff
@@ -30,7 +30,9 @@
   # If using Prometheus telemetry, we want to ensure that the middleware
   # which collects and exports metrics is loaded at the start of the 
   # middleware chain to prevent any modifications to the incoming requests
-  config.middleware.insert_before(0, Monitoring::Middleware::PrometheusExporter, registry: Monitoring::Prometheus.registry, path: '/metrics')
+  if config.conjur_config.telemetry_enabled
```
doodlesbykumbi, line 33: Let's have a test case to validate that when telemetry is not enabled then a called to `GET /metrics` returns 404 (or the appropriate response).

spec/lib/monitoring/middleware/prometheus_exporter_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'monitoring/middleware/prometheus_exporter'
```
doodlesbykumbi, line 2: It's a nice-to-have, and not a requirement. I would be nice, if possible, to get rid of the `spec_helper` require here and instead have everything be loaded explicitly. 

I'm also wondering about `Rack::Test::Methods` vs whatever is used when testing rails controllers. They may be the same. If they're not I wonder what the difference is because we would want to test the `PrometheusExporter` in the context of rails and not in isolation.

config/application.rb
```diff
@@ -81,7 +81,9 @@ class Application < Rails::Application
     # guaranteed to be available for other initializers to use.
     config.conjur_config = Conjur::ConjurConfig.new
 
-    # Initialize metrics and clean existing data
-    Monitoring::Prometheus.setup
+    #Initialize metrics and clean existing data
+    if config.conjur_config.telemetry_enabled
```
doodlesbykumbi, line 85: Nice! Very straightforward

doodlesbykumbi: Fly by comment. I'm wondering if it's a good idea to move all the prometheus/telemetry-setup related things to their own initializer

doodlesbykumbi (commented):
Left some comments

spec/lib/monitoring/middleware/prometheus_exporter_spec.rb
gl-johnson: spec_helper can be removed if the following are explicitly loaded the exporter spec:
`require File.expand_path("../../../../../config/environment", __FILE__)`
`require 'rspec/rails'`
although the impact on loading times seems fairly small to me (13.1 -> 11.8 seconds on my machine), plus it ditches the SimpleCov functionality. Unclear to me if that's needed.

Rack::Test::Methods I guess is a more Rack-specific wrapper for testing API calls. It does not seem to be functionally much different than using the Rails request spec, i.e:
`describe Monitoring::Middleware::PrometheusExporter, type: :request do`
The latter works just fine so we can use that to be consistent with controllers.

config/initializers/rack_middleware.rb
gl-johnson: This behaves a bit strangely since the endpoint is exposed via middleware versus a Rails route. If the middleware isn't loaded (default behavior, metrics disabled) it will be a 401, which I believe is true for any requests that don't include an Auth header, not just /metrics.

Testing a 404 seems a bit trickier. I assume you would need to pass an Auth header to produce this, but rspec complains that there is no matching route and doesn't actually respond 404 gracefully. Plus it seems a bit weird to pass credentials for an endpoint which doesn't require them.

If metrics is enabled then the endpoint is exposed and should be expected to return 200, but this is also tricky in terms of rspec since the Prometheus components only have a chance to be initialized on app startup and presumably we can't reload the entire application for a single test.

doodlesbykumbi: Glad you brought up auth. We need to exclude the metrics endpoint from authorization, see https://github.com/cyberark/conjur/blob/1540a6fb3e515b44687ca3c69a237400d284d38a/config/initializers/rack_middleware.rb#L6-L22.

doodlesbykumbi: Glad you brought up auth. I think it somehow slipped through that we need to exclude the metrics endpoint from authorization, see https://github.com/cyberark/conjur/blob/1540a6fb3e515b44687ca3c69a237400d284d38a/config/initializers/rack_middleware.rb#L6-L22.

doodlesbykumbi: For testing the endpoint under different configuration options we might be able to lean on cucumber which I think runs the actual rails server.

spec/lib/monitoring/middleware/prometheus_exporter_spec.rb
```diff
@@ -25,8 +25,8 @@
     it 'returns the app response' do
       get '/foo'
 
-      expect(last_response).to be_ok
-      expect(last_response.body).to eql('OK')
+      expect(response.status).to eql(200)
```
doodlesbykumbi, line 28: Purely out of curiosity, how does `get` know to direct traffic to the `app` you defined. I haven't been able to find docs on this. I typically expect request specs to just work against the Rails application (minus the rack middleware).

gl-johnson: This may be a bit of misuse of a request spec, but to my understanding it does factor in the middleware layer. In this case the app gets defined by the lambda which is used to initialize to a new instance of the middleware (described_class).
`app = ->(_) { [200, { 'Content-Type' => 'text/html' }, ['OK']] }`
`described_class.new(app, **options)`

Requests make their way to the `call` method in the described class with the passed in lambda as the expected app response.

I'm pushing up a different version of the tests which I think is made a bit clearer by calling the middleware directly instead of leaning on request spec features. Plus it seems to be more inline with conventions I have seen for testing middleware. Let me know what you think.

PR #2522, opened 2022-03-28
Auth persist api

telday:
### Desired Outcome

Adds a new API collection which allows users to configure a policy/secret values for an authentication webservice with one HTTP request.

### Implemented Changes

Adds a new route to Conjur (POST /authenticators/:account/persist) which accepts a JSON object in the body:
```json
{
    "authenticator": "<authenticator type>",
    "service-id": "<authenticator service  id>",
    <variables>
}
```

### Connected Issue/Story

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

- [x] Security architect has reviewed the changes in this PR,


app/controllers/authenticators_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+
+  def conjur_policy_branch_name
+    "#{account}:policy:conjur"
+  end
+
+  def policy_text(loaded_policy_details)
+    loaded_policy_details[:policy].values[:policy_text]
+  end
+
+  def policy_resource_exists(policy_branch_name)
```
jvanderhoof, line 63: in ruby, functions which result in a boolean value typically get a question mark:
```ruby
policy_resource_exists?
```
to improve readability.

app/controllers/authenticators_controller.rb
```diff
@@ -0,0 +1,102 @@
...
+    Resource[format("%s:policy:%s", params[:account], policy_branch_name)] ? true : false
+  end
+
+  def auth_policy_branch
+    branch_format_string = "conjur/%s/%s"
+    format(branch_format_string, authenticator, service_id)
```
jvanderhoof, line 69: [NIT] - this is more commonly achieved with interpolation:
```ruby
"conjur/#{authenticator}/#{service_id}"
```

app/domain/authentication/auth_data.rb
```diff
@@ -0,0 +1,15 @@
+
+module AuthData
```
jvanderhoof, line 2: I think it's fine to leave this as a module, but as a note, this is actually a great candidate for inheritance.

app/domain/authentication/auth_data.rb
```diff
@@ -0,0 +1,15 @@
...
+    def auth_name
+      @json_data["authenticator"]
+    end
+
+    def service_id
+      @json_data["service_id"]
+    end
+
+    def auth_parameters
```
jvanderhoof, line 11: [NIT] - it may be worth changing these methods with `auth_` to just `parameters` or `name` as they have the "authenticator" context from the module/injected class.

app/domain/authentication/authn_k8s/initialize.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      def json_parameter_names
+        [ 'service-account-token', 'ca-cert', 'api-url' ]
+      end
+
+      def json_present?
+        !auth_parameters.keys.empty?
+      end
+
+      validates(
```
jvanderhoof, line 30: What's the value in skipping validation if the JSON params are not present?

app/domain/authentication/authn_k8s/initialize.rb
```diff
@@ -0,0 +1,100 @@
...
+      command_class(
+        dependencies: {
+          conjur_ca_repo: Repos::ConjurCA,
+          secret: Secret,
+          logger: Rails.logger
+        },
+        inputs: %i[conjur_account service_id auth_data current_user]
+      ) do
+        def call
+          loaded_secrets = @auth_data.auth_parameters&.each do |key, value|
```
jvanderhoof, line 77: `each` doesn't return anything.  It just operates on the object (and in this case, the variable `loaded_secrets` will be the result of the last `@secret.create(resource_id: variable_id, value: value)` call).

app/domain/authentication/authn_k8s/initialize.rb
```diff
@@ -0,0 +1,100 @@
...
+
+          ca_repo_id = format("%s:webservice:conjur/%s/%s", @conjur_account, @auth_data.auth_name, @service_id)
+          auth(@current_user, :update, Resource[ca_repo_id])
+          @conjur_ca_repo.create(ca_repo_id)
+          loaded_secrets
+        end
+
+        private
+
+        def logger
```
jvanderhoof, line 93: [NIT] - it looks like `logger` is never called.

jvanderhoof (dismissed):
Great work on this @telday!

app/domain/authentication/authn_k8s/initialize.rb
jvanderhoof: Can we pass the environment variable into this context?

telday: We could always grab the value of the variable and use it, but that would also then force the user to have pre-configured the environment variable before calling the endpoint or it would error out. Is that what we want?

telday: Looks like this was a misread of the documentation on my part. The variables actually do need to be set in Conjur itself. I updated the class to reflect this.

app/controllers/authenticators_controller.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+# REST endpoints for persisting authenticator configuration details in Conjur
+class AuthenticatorsController < RestController
+  include FindResource
+  include AuthorizeResource
+  
+  before_action :current_user
+  before_action :find_or_create_conjur_policy_branch
```
jvanderhoof, line 9: Instead of using a before action, just call `find_or_create_conjur_policy_branch` instead of using `@conjur_policy_branch`.  Calling the method directly removes some indirection.

jvanderhoof:
@telday, I was playing with this a bit on Friday and before we get this in, we need to standardize the URLs to conform both to our current authenticator standards, but also REST best practices.  Can you please break each authenticator "create" into the following convention:

```
POST https://{Conjur-server-hostname}/authn-{type}/{service-id}/{account}
```

The JSON body should be updated to remove the `service_id` from the doc (as it's now in the URI).

app/domain/authentication/persist_auth_factory.rb
```diff
@@ -0,0 +1,55 @@
...
+        Authentication::PersistAuth.new(
+          auth_initializer: auth_initializer(authenticator).new,
+          auth_data_class: auth_data_class(authenticator)
+        )
+      end
+
+      private
+
+      def auth_data_class(authenticator)
+        case authenticator
```
jvanderhoof, line 17: With some slight renaming, we can greatly simplify this:

```ruby
def auth_data_class(authenticator)
   "Authentication::#{authenticator.underscore.camelize}::AuthenticatorData".classify
rescue 
  raise ArgumentError("AuthenticatorData class is not implemented for authenticator #{authenticator}")
end
```

**Note**: The above renames `K8sAuthenticatorData`/`AzureAuthenticatorData`/etc. to `AuthenticatorData`

app/domain/authentication/persist.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        @auth_initializer.(conjur_account: @conjur_account, service_id: @service_id, auth_data: auth_data, current_user: @current_user)
+
+        policy_details
+      end
+
+      private
+
+      def auth_policy(auth_data:, service_id:)
+        @auth_policy ||= ApplicationController.renderer.render(
```
jvanderhoof, line 66: When using Rails Controller functionality in business logic, we need to perform dependency injection to ensure it's obvious it's a dependency.

app/domain/authentication/persist.rb
```diff
@@ -0,0 +1,85 @@
...
+          resource: @resource,
+          current_user: @current_user,
+          client_ip: @client_ip,
+          auth_policy: auth_policy(auth_data: auth_data, service_id: @service_id)
+        )
+
+        @auth_initializer.(conjur_account: @conjur_account, service_id: @service_id, auth_data: auth_data, current_user: @current_user)
+
+        policy_details
+      end
```
jvanderhoof, line 61: Digging into this method, it looks like there's a fair amount of "method wrapping" going on (a method which wraps another single method). This increases cognitive complexity and makes the coder harder to read. What about pushing it all up into the `call` method with something like:

```ruby
def call
  auth_data = @auth_data_class.new(@request_data)
  raise ArgumentError, auth_data.errors.full_messages unless auth_data.valid?

  @auth_initializer.(conjur_account: @conjur_account, service_id: @service_id, auth_data: auth_data, current_user: @current_user)

  @policy_loader.(
    delete_permitted: false,
    action: :update,
    resource: @resource,
    policy_text: ApplicationController.renderer.render(
      template: "policies/#{auth_data.auth_name}",
      locals: { service_id: @service_id, auth_data: auth_data }
    ),
    current_user: @current_user,
    client_ip: @client_ip
  )
end
```

telday: I moved the logic out of the individual functions however I cannot change the order of the initializer and policy load calls because the initailizer relies on the policy being loaded

jvanderhoof (dismissed):
@telday, I think this is in a good place.  Mind cleaning up the commit history so we can get this merged?

config/routes.rb
```diff
@@ -46,6 +46,10 @@ def matches?(request)
         put  '/:authenticator/:account/api_key'  => 'credentials#rotate_api_key'
 
         post '/authn-k8s/:service_id/inject_client_cert' => 'authenticate#k8s_inject_client_cert'
+
+        constraints authenticator: /|authn-azure|authn-iam|authn-k8s|authn-ldap|authn-gcp|authn-oidc/ do
+          post '/:authenticator(/:service_id)/:account' => 'authenticators#persist_auth'
```
jvanderhoof, line 51: Instead of including the `service_id` in the path, let's move it to the body, like this:

```
POST /authn-{authenticator}/{account}}
 {
 "service_id": "<service-id>"
 ...
 }
```
This will give the endpoint a more RESTful convention. For context, when we add them:
```
# Update - future effort
PUT /authn-{authenticator}/{service-id}/{account}}
 {
 ...
 }
# Delete - future effort
DELETE /authn-{authenticator}/{service-id}/{account}}
```

telday: Okay, I just moved the service id out of the url and into the body of the request. Should be all set now

PR #2539, opened 2022-04-14
Collect API metrics

gl-johnson:
### Desired Outcome

For this story we need to register REST API metrics onto the Prometheus client store. Consult the [solution design](https://github.com/cyberark/conjur/blob/040f5041fadf085e47087e4ef8286bc66e72ec74/design/telemetry/metrics_solution_design.md#conjur-http-requests) for details around the metric metadata. The general idea is to

1. Collect the metrics through custom Rack middleware. The prometheus client gem provides an [example Rack middleware](https://github.com/prometheus/client_ruby/blob/master/lib/prometheus/middleware/collector.rb) for tracing HTTP requests. This idea was used in the hackathon and expanded upon in the POC for the soluition design, so for the implementation we can crib from `lib/prometheus/custom_collector.rb` in the [diff from the POC branch](https://github.com/cyberark/conjur/compare/metrics-poc-kt#diff-963b0f7173b1dd7863e1c623cccdc5e03fba88c1ca33e1d3730ae571da2913ce).
2. Publish an event (e.g. `conjur.request`) with enough data that the subscriber can use to make the metrics updates.
3. The Prometheus subscriber for these metrics will interpret that event and make appropriate metric updates to the store. The act of adding the metrics onto the store, combined with the prior work to setup the Prometheus exporter should result in the metric being present in the scrape target endpoint `GET /metrics`.

NOTE: In the Rack middleware we need to ensure that our measurements take into account all other Rack middleware applied to a request. This means the metrics collection Rack middleware must be [first in the list](https://blog.appoptics.com/monitoring-rails-get-hidden-metrics-rails-app/#:~:text=Note%20that%20the%C2%A0Librato%3A%3ARack%C2%A0middleware%20is%20the%20very%20first%20item%20in%20the%20list%2C%20while%20my%20Rails%20app%20is%20the%20last).

A/C

e2e test cases:

lightweight test: requests made on the REST API are reflected in the response of the scrape target endpoint
pub/sub plumbing is unit tested so that when any relevant events are dispatched the expected changes are reflected on the store

### DIFF
--- a/lib/monitoring/middleware/prometheus_collector.rb
+++ b/lib/monitoring/middleware/prometheus_collector.rb
@@ -1,0 +1,60 @@
+require 'benchmark'
+require_relative '../operations.rb'
+
+module Monitoring
+  module Middleware
+    class PrometheusCollector
+      attr_reader :app
+
+      def initialize(app, options = {})
+        @app = app
+        @pubsub = options[:pubsub]
+      end
+
+      def call(env) # :nodoc:
+        trace(env) { @app.call(env) }
+      end
+
+      protected
+
+      # Trace HTTP requests
+      def trace(env)
+        response = nil
+        duration = Benchmark.realtime { response = yield }
+        record(env, response.first.to_s, duration)
+        return response
+      rescue => exception
+        operation = find_operation(env['REQUEST_METHOD'], env['PATH_INFO'])
+        @pubsub.publish(
+          "conjur.request_exception", 
+          operation: operation,
+          exception: exception.class.name,
+          message: exception
+        )
+        raise
+      end
+
+      def record(env, code, duration)
+        operation = find_operation(env['REQUEST_METHOD'], env['PATH_INFO'])
+        @pubsub.publish(
+          "conjur.request", 
+          code: code,
+          operation: operation,
+          duration: duration
+        )
+      rescue
+        # TODO: log unexpected exception during request recording
+        nil
+      end
+
+      def find_operation(method, path)
+        Monitoring::Metrics::OPERATIONS.each do |op|
+          if op[:method] == method && op[:pattern].match?(path)
+            return op[:operation]
+          end
+        end
+        return "unknown"
+      end
+    end
+  end
+end
--- a/config/initializers/prometheus.rb
+++ b/config/initializers/prometheus.rb
@@ -1,7 +1,21 @@
-  # If using Prometheus telemetry, we want to ensure that the middleware
+if Rails.application.config.conjur_config.telemetry_enabled 
+  require 'monitoring/prometheus'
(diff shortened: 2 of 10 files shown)


lib/monitoring/middleware/prometheus_collector.rb
```diff
@@ -0,0 +1,52 @@
...
+      rescue => exception
+        Monitoring::PubSub.instance.publish(
+          "conjur.request_exception", 
+          exception: exception
+        )
+        raise
+      end
+
+      def record(env, code, duration)
+        Monitoring::PubSub.instance.publish(
```
doodlesbykumbi, line 33: Thinking it might good dependency inject the pubsub. It would make unit testing a bit simpler.

lib/monitoring/middleware/prometheus_collector.rb
```diff
@@ -0,0 +1,52 @@
...
+          "conjur.request", 
+          code: code,
+          method: env['REQUEST_METHOD'],
+          path: strip_ids_from_path(env['PATH_INFO']),
```
doodlesbykumbi, line 37: In the solution design we landed on having the labels `code` and the open-spec `operation` identifier, see https://github.com/cyberark/conjur/blob/metrics-solution-design/design/telemetry/metrics_solution_design.md#conjur-http-requests. There's a reference implementation from the hackathon at  https://github.com/cyberark/conjur/blob/bdfc77a03cefd1e6667a56c10111bbefc23b61ac/lib/monitoring/metrics/endpoints.rb#L64.

doodlesbykumbi (commented):
Left some comments

lib/monitoring/middleware/prometheus_collector.rb
```diff
@@ -0,0 +1,76 @@
...
+          "conjur.request_duration", 
+          duration: duration,
+          operation: operation
+        )
+      rescue
+        # TODO: log unexpected exception during request recording
+        nil
+      end
+
+      def strip_ids_from_path(path)
```
doodlesbykumbi, line 59: This is something that the solution design unfortunately doesn't cover.

Do we still need this? For `find_operation`, if we get a match we don't care about the path and if we don't get a match then I want to think the path isn't our responsibility. That last part feels a bit negligent so maybe its best to just have the operation be `unknown`, since even if we decided to take on the responsibility the possibilities of paths are endless.

gl-johnson: Good question. To me, it seems helpful in the current implementation where we are fetching the operation from a file where there is no guarantee that it will always be in sync with the API spec. But if we are able to utilize something dynamic like `Rails.application.routes.recognize_path` then I think that need goes away. I'll do some investigating

PR #2556, opened 2022-05-13
Add authn OIDC endpoint

mFelgate:
### Desired Outcome

### Implemented Changes

particular, consider:*

### Connected Issue/Story

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+module Authenticators
+  module Oidc
+    class ProvidersController < RestController
+      include FindResource
+      include AssumedRole
+      include CurrentUser
+
+      def index
+        # Rails 5 requires parameters to be explicitly permitted before converting
+        # to Hash.  See: https://stackoverflow.com/a/46029524
+        allowed_params = %i[account]
+        
+        render(
+          json: Contexts::Authenticators::AvailableAuthenticators.new.call(
+            role: role,
+            account: params[:account]
+          )
+        )
+        
+      end
+
+      # The v5 API currently sends +acting_as+ when listing resources
+      # for a role other than the current user.
+      def role
+        assumed_role(params[:role].presence) || assumed_role(params[:acting_as].presence)
+      end
+    end
+  end
+end
--- a/app/db/repository/authenticator_repository.rb
+++ b/app/db/repository/authenticator_repository.rb
@@ -5,13 +5,39 @@ module DB
         @resource_repository = resource_repository
       end
 
-      def find(type:, account:, service_id:)
-        return nil unless exists?(
-          type: type,
-          account: account,
-          service_id: service_id
-        )
+      def find_all(type:, account:)
+        @resource_repository.where(
+          Sequel.like(
+            :resource_id,
+            "#{account}:webservice:conjur/authn-#{type}%"
+          )
+        ).map do |webservice|
+          load_authenticator(account: account, id: webservice.id.split(':').last, type: type)
+        end
+      end
 
+      def find(type:, account:,  service_id:)
+        webservice =  @resource_repository.where(
+          Sequel.like(
+            :resource_id,
+            "#{account}:webservice:conjur/authn-#{type}/#{service_id}%"
+          )
+        ).first
+        unless webservice
+          return
+        end
+
+        load_authenticator(account: account, id: webservice.id.split(':').last, type: type)
+      end
+
+      def exists?(type:, account:, service_id:)
+        @resource_repository.with_pk("#{account}:webservice:conjur/authn-#{type}/#{service_id}") != nil
+      end
+
+      private
+
+      def load_authenticator(type:, account:, id:)
+        service_id = id.split('/')[-1].underscore.to_sym
         variables = @resource_repository.where(
           Sequel.like(
             :resource_id,
@@ -21,18 +47,15 @@ module DB
 
         args_list = {}.tap do |args|
           args[:account] = account
-          args[:service_id] = service_id
+          args[:service_id] = id.split('/')[-1].underscore.to_sym.to_s
           variables.each do |variable|
             next unless variable.secret
+
             args[variable.resource_id.split('/')[-1].underscore.to_sym] = variable.secret.value
           end
         end
 
         "Authenticator::#{type.camelize}Authenticator".constantize.new(**args_list)
-      end
-
-      def exists?(type:, account:, service_id:)
-        return @resource_repository.with_pk("#{account}:webservice:conjur/authn-#{type}/#{service_id}") != nil
       end
     end
   end
--- a/cucumber/policy/features/support/client.rb
+++ b/cucumber/policy/features/support/client.rb
@@ -109,6 +109,14 @@ class Client
     resource(uri("policies", "policy", id)).post(policy, auth_header)
   end
 
+  # Authenticator methods
+  #
+
+  # :reek:DataClump
+  def fetch_authenticators
+    resource(uri('authn-oidc', 'providers')).get(auth_header)
+  end
+
   # Resource methods
   #
   def fetch_resource(kind:, id:)
@@ -164,3 +172,4 @@ class Client
     )
   end
 end
+0
--- a/spec/app/db/repository/authenticator_repository_spec.rb
+++ b/spec/app/db/repository/authenticator_repository_spec.rb
@@ -3,12 +3,12 @@
 require 'spec_helper'
 
 RSpec.describe('DB::Repository::AuthenticatorRepository') do
-  let(:repo) { DB::Repository::AuthenticatorRepository.new() }
+  let(:repo) { DB::Repository::AuthenticatorRepository.new }
 
   describe('exists method') do
     context "with missing parameters" do
       it "returns false with no parameters" do
-        expect(repo.exists?(type: nil, account: nil, service_id: nil)).to eq (false)
+        expect(repo.exists?(type: nil, account: nil, service_id: nil)).to eq(false)
       end
 
       it "returns false with no type parameter" do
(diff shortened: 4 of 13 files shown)


app/controllers/authenticate_controller.rb
```diff
@@ -3,6 +3,60 @@
 class AuthenticateController < ApplicationController
   include BasicAuthenticator
   include AuthorizeResource
+  include CurrentUser
+  include FindResource
+  include AssumedRole
+
+  def list_authenticators
```
jvanderhoof, line 10: This method feels kind of weird here in that it's specific to OIDC, but the name doesn't give that indication.  As it's OIDC specific, what do you think about moving this into an OIDC specific controller?  

Historically we don't do this, but the `AuthenticateController` is a mess and handles a lot of authenticator specific behavior via methods.

app/controllers/authenticate_controller.rb
```diff
@@ -3,6 +3,60 @@
...
+    rescue ApplicationController::Forbidden
+      raise
+    rescue ArgumentError => e
+      raise ApplicationController::UnprocessableEntity, e.message
+    end
+
+    render(json: scope)
+  end
+
+  def authenticators(role, repo, handler, account:, service_id: nil)
```
jvanderhoof, line 31: Mind updating the positional arguments to be named arguments? They have the benefit of conveying context to the  calling code.
```suggestion
  def authenticators(role:, repository:, handler:, account:, service_id: nil)
```

app/db/repository/authenticator_repository.rb
```diff
@@ -5,34 +5,74 @@ def initialize(resource_repository: ::Resource)
...
             args[variable.resource_id.split('/')[-1].underscore.to_sym] = variable.secret.value
           end
         end
 
         "Authenticator::#{type.camelize}Authenticator".constantize.new(**args_list)
       end
 
+      def resources(role:)
+        unless role
+          return @resource_repository
```
jvanderhoof, line 50: Won't this bypass our security?  If the role isn't present, it'll return ALL resources instead of those which are visible to the requester.

app/db/repository/authenticator_repository.rb
```diff
@@ -5,34 +5,74 @@ def initialize(resource_repository: ::Resource)
...
+        end
+
+        @resource_repository.visible_to(role)
+      end
+
+      def fetch_authenticators(account:, type:, service_id:, role:)
+        puts account, type, service_id, role
```
jvanderhoof, line 57: Please remove debugging statements.

app/db/repository/authenticator_repository.rb
```diff
@@ -5,34 +5,74 @@ def initialize(resource_repository: ::Resource)
...
             next unless variable.secret
+
             args[variable.resource_id.split('/')[-1].underscore.to_sym] = variable.secret.value
           end
         end
 
         "Authenticator::#{type.camelize}Authenticator".constantize.new(**args_list)
       end
 
+      def resources(role:)
```
jvanderhoof, line 48: If we remove the behavior of a blank role (which I believe we need to, see below), this method becomes a wrapper method for `visible_to(role)`. I think we're better off removing it and pushing the `visible_to(role)` call down the `fetch_authenticators` method.

app/db/repository/authenticator_repository.rb
```diff
@@ -5,34 +5,74 @@ def initialize(resource_repository: ::Resource)
         @resource_repository = resource_repository
       end
 
-      def find(type:, account:, service_id:)
+      def find_all(type:, account:, role: nil)
```
jvanderhoof, line 8: If `role` is `nil`, will that bypass our security checks for permission?

jvanderhoof:
@mFelgate, the repository object should focus on all the CRUD actions of an authenticator. We can combine the `find` and `find_all` with something like:

```ruby
module DB
  module Repository
    class AuthenticatorRepository
      def initialize(resource_repository: ::Resource)
        @resource_repository = resource_repository
      end

      def find_all(type:, account:, role:)
        @resource_repository.visible_to(role).where(
          Sequel.like(
            :resource_id,
            "#{account}:webservice:conjur/authn-#{type}/%"
          )
        ).each do |webservice|
          load_authenticator(account: account, id: webservice.id.split(':').last, role: role, type: type)
        end
      end

      def find(type:, account:, service_id:, role:)
        webservice = @resource_repository.visible_to(role).where(
          Sequel.like(
            :resource_id,
            "#{account}:webservice:conjur/authn-#{type}/#{service_id}"
          )
        ).first
        load_authenticator(account: account, id: webservice.id.split(':').last, role: role, type: type)
      end

      private
      
      def load_authenticator(type:, account:, id:, role:)
        @resource_repository.visible_to(role).where(
          owner_id: "#{accout}:policy:#{id}"
        ).eager(:secrets).all.each do |resource|
          next unless resource.id.split(':')[1] == 'variable'

          args_list = {}.tap do |args|
            args[:service_id] = service_id
            args[:account] = account
            variables.each do |variable|
              args[variable.resource_id.split('/')[-1].underscore.to_sym] = variable.secret.value
            end
          end
          "Authenticator::#{type.camelize}Authenticator".constantize.new(**args_list)
        end
      end
   end
end
```
The argument `resource_repository` is present for dependency injection, to enable the Sequel connector to be mocked.  This will allow us unit test the above repository object.

app/controllers/authenticate_controller.rb
```diff
@@ -3,6 +3,9 @@
 class AuthenticateController < ApplicationController
   include BasicAuthenticator
   include AuthorizeResource
+  include CurrentUser
```
jvanderhoof, line 6: I don't think these are needed anymore.

app/controllers/oidc_controller.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class OidcController < RestController
```
jvanderhoof, line 3: Let's get this into an `Authenticators` namespace to denote its purpose.  This is accomplished by moving it into the folder `app/controllers/authenticators` and moving the above to:

```ruby
module Authenticators
  class OidcController < RestController
    ...
```

app/controllers/oidc_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  def query_role
+    params[:role].presence || params[:acting_as].presence
+  end
+
+  def options(allowed_params)
+    params.permit(*allowed_params)
+          .slice(*allowed_params).to_h.symbolize_keys
+  end
+
+  def authenticators(role:, repository:, handler:, account:, service_id: nil)
```
jvanderhoof, line 41: Let's push this logic up to a business object.  It's super hard to test here and it's causing this method to do two distinctly separate tasks (return a list of provider, returning a full provider).

app/controllers/oidc_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+          name: authn.name,
+          login_url: handler.generate_login_url(authn)
+        }
+      end
+    end
+
+    authn = repository.find(role: role, account: account, type: "oidc", service_id: service_id)
+    return {} unless authn
+
+    handler.generate_login_url(authn)
```
jvanderhoof, line 58: Do we have a case where we're requesting the login for a single OIDC service id?  if not, can we get rid of this?

mFelgate: it was in the design doc for where you add a service_id to the end or the url

app/controllers/oidc_controller.rb
mFelgate: i like it, i made the changes to this new service.

app/controllers/providers_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+            account: params[:account]
+          )
+        )
+        
+      end
+
+      # The v5 API currently sends +acting_as+ when listing resources
+      # for a role other than the current user.
+      def query_role
+        params[:role].presence || params[:acting_as].presence
```
jvanderhoof, line 26: This method feels a little soft from a functional perspective. What do you think about having it focus on resolving the role of the requestor (either defined or "acting as").  If we shift `assumed_role` into this method:

```suggestion
      def role
        assumed_role(params[:role].presence || params[:acting_as].presence)
      end
```
we have method which handles the resolution of the role depending on the provided parameters.

app/controllers/providers_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      def options(allowed_params)
```
jvanderhoof, line 29: Is this called anywhere now?

spec/app/db/repository/authenticator_repository_spec.rb
```diff
@@ -42,6 +42,210 @@
...
+        ::Role.create(
+          role_id: "rspec:policy:conjur/authn-oidc/abc123"
+        )
+        ::Resource.create(
+          resource_id: "rspec:webservice:conjur/authn-oidc/abc123",
+          owner_id: "rspec:policy:conjur/authn-oidc/abc123"
+        )
+      end
+
+      it "returns a v2 authenticator with all properties set" do
```
jvanderhoof, line 81: [Note] Looks like these tests might be related to Eric's original work.  I don't see the V2 data object in this PR.

mFelgate: yeah. i think the authenticator class he created handles that

cucumber/policy/features/support/client.rb
```diff
@@ -109,6 +109,18 @@ def replace_policy(id:, policy:)
...
 
+  # Authenticator methods
+  #
+
+  # :reek:DataClump
+  def fetch_authenticators
+    resource(uri('authn-oidc', 'providers')).get(auth_header)
+  end
+
+  def fetch_authenticator(id:)
```
jvanderhoof, line 120: Do we still need this method?

mFelgate: oh shoot, missed that

PR #2565, opened 2022-06-01
Add annotation based user search

mFelgate:
### Desired Outcome

### Implemented Changes

particular, consider:*

### Connected Issue/Story

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


app/domain/authentication/handler/authentication_handler.rb
```diff
@@ -130,7 +130,14 @@ def validate_client_ip(client_ip_address, conjur_role)
...
       def fetch_conjur_role(authenticator, identity)
-        return @role_repository_class.from_username(authenticator.account, identity)
+        annotation = Annotation.fetch_annotation(
+          account: authenticator.account,
+          value: identity,
+          name: authenticator.authenticator_name
+        ).first
+        return @role_repository_class.from_username(authenticator.account, identity) unless annotation
+
+        @role_repository_class[role_id: annotation.resource_id]
```
micahlee, line 140: How come the default/base implementation for fetching a role is using annotations? I would expect this to be solely based on the ID, and I'm not sure why this change is necessary

app/domain/authentication/handler/oidc_authentication_handler.rb
```diff
@@ -59,6 +59,21 @@ def type
...
+      def generate_token(account, identity)
+        @token_factory.signed_token(
+          account: account,
+          username: conjur_identity(account: account, id: identity)
+        )
+      end
+
+      def conjur_identity(account:, id:)
+        annotation = Annotation.fetch_annotation(account: account, value: id).first
+        return id unless annotation
```
micahlee, line 72: This definitely needs a good comment describing what's going on here.

This is still giving precedence to annotations over role ID's, correct?

mFelgate: from what i understood yes. i asked while talking about this feature with jason if we wanted to default to role_id, then use annotation if it didn't match and Jason suggested the other way around. This wasn't about a specific section of code tho but more of a high level discussion

micahlee: Hey @jvanderhoof , can you give any additional context to why we would give precedence for annotation values over resource IDs when resolving OIDC authentications?

The risk I'm considering is that this allows a host to set an annotation and impersonate an existing role with the host ID used for the mapping. Also, it's generally been an implied design principle that annotations are a weaker security control than IDs or variables. So taking an annotation as the higher precedent over the ID doesn't seem to jive with that either.

I'm probably missing some background or context that's important here, though.

CC: @andytinkham

app/domain/authentication/handler/authentication_handler.rb
jvanderhoof: I agree with the order of selection concern. Let's use the following for the resolution order:

1. Find user by ID - essentially, this requires users to be in the root
2. Find user by annotation (`authn-oidc/identity`) - allows users to exist outside the root policy, but runs the risk of two users with the same OIDC identity

Defaulting to User ID as the default lookup maintains a similar behavior as other authenticators, but still allows for annotation based lookup.

app/domain/authentication/handler/oidc_authentication_handler.rb
jvanderhoof: @micahlee, as mentioned above, with annotations, I'm hoping to provide a path for getting users with email addresses (common OIDC pattern) mapped to Conjur users (where we use the "unfortunate" `@` to delineate policy scope for a user - `my.user@my-policy`).

Annotations have emerged as a common pattern for attributing external ID to internal ID. I agree it's less secure. We should absolutely keep pushing on this to understand how best to manage to "Users in root" requirement and tie internal/external mapping.

app/models/authenticator/authenticator.rb
```diff
@@ -15,6 +15,10 @@ def initialize(account:, service_id:, required_request_parameters: nil)
       end
     end
 
+    def authenticator_name
+      return self.service_id
```
jvanderhoof, line 19: In Ruby the last line is returned, so the `return` is redundant here.  

Alternatively, we can just add an alias:
```ruby
module Authenticator
  class Authenticator
    alias authenticator_name, service_id
    ...
  end
end
```

app/domain/authentication/handler/authentication_handler.rb
```diff
@@ -130,7 +130,14 @@ def validate_client_ip(client_ip_address, conjur_role)
       end
 
       def fetch_conjur_role(authenticator, identity)
-        return @role_repository_class.from_username(authenticator.account, identity)
+        annotation = Annotation.fetch_annotation(
```
jvanderhoof, line 133: Instead of including the `Annotation` dependency here, I'd recommend injecting it into the `Authentication::Handler::AuthenticationHandler` initialization method.  

This will allow you to write tests for this handler without loading anything into the database.

app/domain/authentication/handler/authentication_handler.rb
```diff
@@ -130,7 +130,14 @@ def validate_client_ip(client_ip_address, conjur_role)
       end
 
       def fetch_conjur_role(authenticator, identity)
```
jvanderhoof, line 132: I think we should consider moving this into either a `Role` or a `User` repository. This allows us to decouple the process of finding the relevant user from the authentication process (which this class does).

We could do something like this (_warning pseudocode_) :
```ruby
module DB
  module Repository
    class RoleRepository
      def initialize(role: ::Role, annotation: ::Annotation)
        @role = role
        @annotation = annotation
      end

      def find(account:, identifier:, annotation: nil, type: 'user')
        role = find_by_id(account: account, identifier: identifier, type: type)
        if role.blank? && annotation.present?
          role = find_by_annotation(account: account, identifier: identifier, type: type, annotation: annotation)
        end
        role
      end

      # The methods below are public to simplify testing, but only `find` should be part of the primary interface.

      def find_by_id(account:, identifier:, type: 'user')
        @role[[account, type, identifier].join(':')]
      end

      def find_by_annotation(account:, identifier:, annotation:, type: 'user')
        resource_id = @annotation
          .where(name: annotation)                    # exact match on annotation name
          .ilike(value: identifier)                   # case insensitive match on annotation value
          .like(resource_id: "#{account}:#{type}:%")  # verify account and type
          .first
          .try(:resource_id)
        
        return nil unless resource_id

        @role[resource_id]
      end
    end
  end
end
```

Alternatively, we can extend `Annotation` to include a method to handle finding a resource given an annotation name and value.  In this case, the `Annotation` method should return the `resource_id` to align with the `Role#roleid_from_username` convention.

cucumber/authenticators_oidc/features/authn_oidc.feature
```diff
@@ -29,7 +29,10 @@ Feature: OIDC Authenticator - Hosts can authenticate with OIDC authenticator
         privilege: [ read, authenticate ]
         resource: !webservice
 
-    - !user alice
+    - !user
+      id: alice
+      annotations:
+       authn-oidc/keycloak: alice.somebody@cyberark.com
```
jvanderhoof, line 35: I was thinking about the annotation more along the lines of:
```yml
- !user
      id: alice
      annotations:
       authn-oidc/identity: alice.somebody@cyberark.com
```
which would allow the user to be selected given a JWT with the subject `alice.somebody@cyberark.com`. This lines up with the way annotations are used for mapping in other authenticators: [Azure](https://docs.cyberark.com/Product-Doc/OnlineHelp/AAM-DAP/Latest/en/Content/Operations/Services/azure_authn.htm#ConfiguretheAzureAuthenticator).

app/domain/authentication/handler/authentication_handler.rb
mFelgate: do you want a role repo or would this be fine to just add to the role model?

jvanderhoof: The problem is `Annotation`.  Adding `find_by_annotation` to Role feels a bit weird as it forces `Annotation` as a dependency on `Role`. 

It might make more sense to add a `find_by_key_and_value` type method to `Annotation`.

mFelgate: in that case should it be more like the original, but just have role_id be the default

adamouamani (commented):
I have two questions: 
1) can this annotation be anything (ie: not just email) so long as it matches user defined in JWT token
2) what happens if duplicate annotations for different users - this seems like a security issue 
3) can hosts have this annotation?

spec/app/db/repository/role_repository.spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  let(:roles) do
+    DB::Repository::RoleRepository.new(
+      role: role,
+      annotation: annotation,
+    )
+  end
+
+  describe('.find') do
```
micahlee, line 51: The only nit from reading better spec again is that this should follow the Ruby conventions for referring to a method:

```suggestion
  describe('#find') do
```

PR #2567, opened 2022-06-03
Add gemfile comments and fix OCP test script

micahlee:
### Implemented Changes
This is a follow-up PR to https://github.com/cyberark/conjur/pull/2564

Specifically it:
- Adds comments to the gemfile to explain the use of `require: false` for simplecov
- Fixes the OCP test entrypoint in the same way GKE was fixed in the prior PR

### Connected Issue/Story

N/A

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/ci/test_suites/authenticators_k8s/test_oc_entrypoint.sh
+++ b/ci/test_suites/authenticators_k8s/test_oc_entrypoint.sh
@@ -294,7 +294,7 @@ function run_cucumber() {
     -r ./cucumber/authenticators_k8s/features/support/world.rb \
     -r ./cucumber/authenticators_k8s/features/support/hooks.rb \
     -r ./cucumber/authenticators_k8s/features/support/conjur_token.rb \
-    $cucumber_args ./cucumber/authenticators_k8s/features" | cucumbercmd -i bash || true
+    $cucumber_args ./cucumber/authenticators_k8s/features" | cucumbercmd -i bash
 }
 
 main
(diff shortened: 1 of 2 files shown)


ci/test_suites/authenticators_k8s/test_oc_entrypoint.sh
```diff
@@ -294,7 +294,7 @@ function run_cucumber() {
     -r ./cucumber/authenticators_k8s/features/support/world.rb \
     -r ./cucumber/authenticators_k8s/features/support/hooks.rb \
     -r ./cucumber/authenticators_k8s/features/support/conjur_token.rb \
-    $cucumber_args ./cucumber/authenticators_k8s/features" | cucumbercmd -i bash || true
+    $cucumber_args ./cucumber/authenticators_k8s/features" | cucumbercmd -i bash
```
andytinkham, line 297: A quick search shows the `|| true` syntax occurs a couple other places- in dev/start, in ci/test, in ci/test_suites/authenticators_k8s/build_locally.sh, and in ci/test_suites/authenticators_k8s/entrypoint.sh.

I think the syntax is basically saying "Don't ever treat this as a failure". None of the other places I listed are cucumber calls, and for the ones in ci/test_suites, it makes sense as a "keep going even if cleanup fails" thing since it's paired with docker rm. In ci/test, the same argument might apply - it's paired with docker_diagnostics and docker-compose down, which probably aren't essential to the end result.

For the first one though, in dev/start, it's paired with creating the cucumber account (`docker-compose exec conjur conjurctl account create cucumber || true`. Is that another place that we should remove the || true? Seems like having the cucumber account is a critical step to running the tests...

andytinkham (approved):

PR #2574, opened 2022-06-12
Added configuration for default value for maximum number of results to listing requests

aloncarmel111:
### Desired Outcome
Limiting resources return when listing Resources API request

### Implemented Changes
Controller was changed
ConjurConfig was changed
Design - https://ca-il-confluence.il.cyber-ark.com/display/AAM/Conjur+List+Resources+Limit+Design

### Connected Issue/Story

Resolves #[CONJSE-1261](https://ca-il-jira.il.cyber-ark.com:8443/browse/CONJSE-1261)

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -11,6 +11,19 @@ class ResourcesController < RestController
     options = params.permit(*allowed_params)
       .slice(*allowed_params).to_h.symbolize_keys
 
+    # If a maximum limit is configured, we need to verify that the requested
+    # limit does not exceed the maximum.
+    # The maximum limit should not be apply if this is a count request.
+    if conjur_config.api_resource_list_limit_max.positive? && !params[:count]
+      # If no limit is given, default the limit to the configured maximum
+      options[:limit] = conjur_config.api_resource_list_limit_max.to_s unless options[:limit]
+
+      unless options[:limit].to_i <= conjur_config.api_resource_list_limit_max
+        raise ApplicationController::UnprocessableEntity, \
+              "'Limit' parameter must not exceed #{conjur_config.api_resource_list_limit_max}"
+      end
+    end
+
     if params[:owner]
       ownerid = Role.make_full_id(params[:owner], account)
       (options[:owner] = Role[ownerid]) || raise(Exceptions::RecordNotFound, ownerid)
@@ -102,5 +115,9 @@ class ResourcesController < RestController
         error_message: err.message
       )
     )
+  end
+
+  def conjur_config
+    Rails.application.config.conjur_config
   end
 end
--- a/lib/conjur/conjur_config.rb
+++ b/lib/conjur/conjur_config.rb
@@ -23,6 +23,9 @@ module Conjur
     attr_config(
       # Read TRUSTED_PROXIES before default to maintain backwards compatibility
       trusted_proxies: (ENV['TRUSTED_PROXIES'] || []),
+      # The maximum number of results for listing requests. The default value
+      # is 0 and means that there is no limit.
+      api_resource_list_limit_max: 0,
       user_authorization_token_ttl: 480, # The default TTL of User is 8 minutes
       host_authorization_token_ttl: 480, # The default TTL of Host is 8 minutes
       authenticators: []
--- a/spec/controllers/resources_controller_spec.rb
+++ b/spec/controllers/resources_controller_spec.rb
@@ -1,0 +1,118 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+DatabaseCleaner.strategy = :truncation
+
+describe ResourcesController, type: :request do
+
+  before do
+    Slosilo["authn:rspec"] ||= Slosilo::Key.new
+  end
+
+  let(:current_user) { Role.find_or_create(role_id: 'rspec:user:admin') }
+
+  describe '#post' do
+
+    before(:each) do
+      Rails.application.config.conjur_config.api_resource_list_limit_max = 0
+      load_variables()
+    end
+
+    let(:resources_url) do
+      '/resources/rspec'
+    end
+
+    let(:policies_url) do
+      '/policies/rspec/policy/root'
+    end
+
+    let(:token_auth_header) do
+      bearer_token = Slosilo["authn:rspec"].signed_token(current_user.login)
+      token_auth_str =
+        "Token token=\"#{Base64.strict_encode64(bearer_token.to_json)}\""
+      { 'HTTP_AUTHORIZATION' => token_auth_str }
+    end
+
+    def list_resources_without_limit()
+      get(resources_url, env: token_auth_header())
+    end
+
+    def list_resources_with_limit(limit)
+      get(resources_url, env: token_auth_header(), params: {:limit => limit})
+    end
+
+    def count_resources_without_limit()
+      get(resources_url, env: token_auth_header(), params: {:count => "true"})
+    end
+
+    def count_resources_with_limit(limit)
+      get(resources_url, env: token_auth_header(), params: {:limit => limit, :count => "true"})
+    end
+
+    def load_variables()
+      payload = '[!variable preexisting\n!variable preexisting1]'
+      put(policies_url, env: token_auth_header.merge({ 'RAW_POST_DATA' => payload }))
+    end
+
+    context 'with default configuration' do
+      it "should list all resources when limit not defined" do
+        list_resources_without_limit()
+        expect(response.code).to eq("200")
+        expect(JSON.parse(response.body).size).to eq(2)
+      end
+
+      it "should list resources according to query param limit when limit defined" do
+        list_resources_with_limit(1)
+        expect(response.code).to eq("200")
+        expect(JSON.parse(response.body).size).to eq(1)
+      end
+    end
+
+    context 'with custom configuration' do
+      it "should list resources according to custom configuration when limit not defined" do
+        Rails.application.config.conjur_config.api_resource_list_limit_max = 1
+        list_resources_without_limit()
+        expect(response.code).to eq("200")
+        expect(JSON.parse(response.body).size).to eq(1)
+      end
+
+      it "should list resources according to query param limit when custom configuration exceeds limit" do
+        Rails.application.config.conjur_config.api_resource_list_limit_max = 2
+        list_resources_with_limit(1)
+        expect(response.code).to eq("200")
+        expect(JSON.parse(response.body).size).to eq(1)
+      end
+
+      it "should throw error when limit exceeds custom configuration" do
+        Rails.application.config.conjur_config.api_resource_list_limit_max = 1
+        list_resources_with_limit(2)
+        expect(response.code).to eq("422")
+      end
+    end
+
+    context 'when validating count request' do
+      it "should count all resources when custom configuration defined" do
+        Rails.application.config.conjur_config.api_resource_list_limit_max = 1
+        count_resources_without_limit()
+        expect(response.code).to eq("200")
+        expect(response.body).to eq("{\"count\":2}")
+      end
+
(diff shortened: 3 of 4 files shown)


lib/conjur/conjur_config.rb
```diff
@@ -23,6 +23,7 @@ class ConjurConfig < Anyway::Config
     attr_config(
       # Read TRUSTED_PROXIES before default to maintain backwards compatibility
       trusted_proxies: (ENV['TRUSTED_PROXIES'] || []),
+      max_resources_limit: 0,
```
micahlee, line 26: What do you think about:
```
api_resource_list_limit_max
```
instead?

The thought process here is that each subsequent word in the name should provide a meaningful subdivision of the configuration space. This makes it easier to scan the file, and to potentially sort it alphabetically in a meaningful way.

`api` -> `resource` -> `list` -> `limit` -> `max`

Also interested in what others think from the @cyberark/conjur-core-team as we expand these configurations and identify some conventions for how that should be named? The goal is that they're consistent and predictable.

micahlee: Also, not sure if this is ready for final review or not, @aloncarmel111 , but this default value should include a comment explaining what the config setting is and what the default means. Thanks!

aloncarmel111: Updated the parameter name to `api_resource_list_limit` and added comment.

lib/conjur/conjur_config.rb
```diff
@@ -23,6 +23,7 @@ class ConjurConfig < Anyway::Config
     attr_config(
       # Read TRUSTED_PROXIES before default to maintain backwards compatibility
       trusted_proxies: (ENV['TRUSTED_PROXIES'] || []),
+      api_resource_list_limit: 0, # The maximum number of results for listing requests. The default value is 0 and means that there is no limit.
```
micahlee, line 26: Was there a reason for leaving the `_max` off of the suggestion? Just curious.

Also, please place the comment on the line above the config definition. See the `trusted_proxies` example above.

app/controllers/resources_controller.rb
```diff
@@ -11,6 +11,13 @@ def index
     options = params.permit(*allowed_params)
       .slice(*allowed_params).to_h.symbolize_keys
 
+    if Rails.application.config.conjur_config.api_resource_list_limit.positive?
+      options[:limit] = Rails.application.config.conjur_config.api_resource_list_limit.to_s unless options[:limit]
+
+      raise ApplicationController::UnprocessableEntity, "Limit value should not exceed #{Rails.application.config.conjur_config.api_resource_list_limit}" \
+      unless options[:limit].to_i <= Rails.application.config.conjur_config.api_resource_list_limit
```
micahlee, line 18: Given how long this line is, please use 
```ruby
unless ....
   raise
end
```

instead.

lib/conjur/conjur_config.rb
aloncarmel111: updated the comment,
seems to me redundant to use both `max` and `limit` wdyt?

app/controllers/resources_controller.rb
aloncarmel111: done

app/controllers/resources_controller.rb
```diff
@@ -11,6 +11,13 @@ def index
...
+    end
+
```
micahlee, line 21: Generally, I would suggest creating a helper method in the controller (or one of the base classes):

```ruby
def conjur_config
    Rails.application.config.conjur_config
end
```
I don't love it, but repeating that method chain multiple times is worse, IMO.

And then leverage that to clean up the block and add further comments on the flow. For example:

```ruby
# If a maximum limit is configured, we need to verify that the requested
# limit does not exceed the maximum.
if conjur_config.api_resource_list_limit_max.positive?
  # If no limit is given, default the limit to the configured maximum
  options[:limit] = conjur_config.api_resource_list_limit_max.to_s unless options[:limit]

  unless options[:limit].to_i <= conjur_config.api_resource_list_limit_max
    raise ApplicationController::UnprocessableEntity, \
          "'limit' parameter must not exceed #{conjur_config.api_resource_list_limit_max}"
  end
end
```

lib/conjur/conjur_config.rb
micahlee: `limit` is the name of the parameter. What we are configuring is the maximum value allowed in the parameter.

app/controllers/resources_controller.rb
```diff
@@ -11,6 +11,14 @@ def index
     options = params.permit(*allowed_params)
       .slice(*allowed_params).to_h.symbolize_keys
 
+    if Rails.application.config.conjur_config.api_resource_list_limit.positive?
+      options[:limit] = Rails.application.config.conjur_config.api_resource_list_limit.to_s unless options[:limit]
+
+      unless options[:limit].to_i <= Rails.application.config.conjur_config.api_resource_list_limit
+        raise ApplicationController::UnprocessableEntity, "Limit value should not exceed #{Rails.application.config.conjur_config.api_resource_list_limit}"
+      end
+    end
```
micahlee, line 20: Generally, I would suggest creating a helper method in the controller (or one of the base classes):

```ruby
def conjur_config
    Rails.application.config.conjur_config
end
```
I don't love it, but repeating that method chain multiple times is worse, IMO.

And then leverage that to clean up the block and add further comments on the flow. For example:

```ruby
# If a maximum limit is configured, we need to verify that the requested
# limit does not exceed the maximum.
if conjur_config.api_resource_list_limit_max.positive?
  # If no limit is given, default the limit to the configured maximum
  options[:limit] = conjur_config.api_resource_list_limit_max.to_s unless options[:limit]

  unless options[:limit].to_i <= conjur_config.api_resource_list_limit_max
    raise ApplicationController::UnprocessableEntity, \
          "'limit' parameter must not exceed #{conjur_config.api_resource_list_limit_max}"
  end
end
```

lib/conjur/conjur_config.rb
```diff
@@ -23,6 +23,8 @@ class ConjurConfig < Anyway::Config
     attr_config(
       # Read TRUSTED_PROXIES before default to maintain backwards compatibility
       trusted_proxies: (ENV['TRUSTED_PROXIES'] || []),
+      # The maximum number of results for listing requests. The default value is 0 and means that there is no limit
```
micahlee, line 26: This comment needs to be wrapped: 

```suggestion
      # The maximum number of results for listing requests. The default value
      # is 0 and means that there is no limit.
```

spec/controllers/resources_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        expect(response.code).to eq("200")
+        expect(response.body).to eq("{\"count\":1}")
+      end
+
+      it "should throw error when limit exceeds custom configuration" do
+        Rails.application.config.conjur_config.api_resource_list_limit = 1
+        count_resources_with_limit(2)
+        expect(response.code).to eq("422")
+      end
+    end
```
micahlee, line 83: This appears to introduce a bug. 

`limit` should never impact the count, even if they're inadvertently provided together. That's a gap in our validation now, but we shouldn't propagate the issue further. If `count` is true, we should not be applying the default limit value.

We do want the ability to get the total number of resources in Conjur that match the other parameters.

All of these test cases should be:
```ruby
expect(response.body).to eq("{\"count\":2}")
```

And there should be an additional error case that `limit` and `count` should not be used together.

micahlee: That also means we do need test cases that retrieve the records and not just count them.

jvanderhoof: @aloncarmel111, I'm glad we're finally rolling out a limit on resources returned as a single result.  It's a very common pattern across APIs to solve the issue the security team identified. 

Instead of setting the default to `0` (no limit), I feel it makes sense to set it to something reasonable for everyone (ex. `100`). I don't love the idea of making it configurable, but it does allow customers who depend on getting all resources in a single request to continue to support this behavior (while understanding the downside of this decision).

aloncarmel111: We can align to other CyberArk, in PVWA the default limit is 50. and the maximum number that can be specified is 1000.
@jvanderhoof wdyt the expected behavior should be when `limit` and `count` used together?

lib/conjur/conjur_config.rb
aloncarmel111: updated

app/controllers/resources_controller.rb
aloncarmel111: done

app/controllers/resources_controller.rb
aloncarmel111: done

spec/controllers/resources_controller_spec.rb
aloncarmel111: @micahlee , Updated the tests to retrieve the records and not just count them.
I will create issue to decide the expected behavior when `limit` and `count` used together

aloncarmel111: Hi @micahlee, added validation for not applying the default limit value when **count** set to true 
also opened the following issue https://github.com/cyberark/conjur/issues/2579 
can you prioritize this?

aloncarmel111: @jvanderhoof, The maximum limit for Conjur cloud is set to 50. DAP and OSS are in other scale than Conjur cloud, I will open issue for this, we need to understand the exact limit for DAP and OSS,  because it's can impact existing customer.
CC @rrefael @netacoral

jvanderhoof: @aloncarmel111, if `limit` and `count` are used together, I'd expect `limit` to have no effect, as a single response is returned (the total number of that resource search).

spec/controllers/resources_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      it "should count all resources when custom configuration not defined" do
+        count_resources_without_limit()
+        expect(response.code).to eq("200")
+        expect(response.body).to eq("{\"count\":2}")
+      end
+
+      it "should count resources according to query param limit " do
+        count_resources_with_limit(1)
+        expect(response.body).to eq("{\"count\":1}")
+      end
```
micahlee, line 111: Thanks @aloncarmel111, would you please add a small comment to this test case to describe that this isn't the desired behavior, only the current behavior? And link to the bug, something like:

```suggestion
      # There is a currently a bug in the API when supplying both the `limit`
      # and `count` parameters. A count response shouldn't be affected by
      # the `limit` parameter. This should be changed when the bug is fixed here:
      # JIRA link...
      it "should count resources according to query param limit " do
        count_resources_with_limit(1)
        expect(response.body).to eq("{\"count\":1}")
      end
```

app/controllers/resources_controller.rb
```diff
@@ -11,6 +11,18 @@ def index
     options = params.permit(*allowed_params)
       .slice(*allowed_params).to_h.symbolize_keys
 
+    # If a maximum limit is configured, we need to verify that the requested
+    # limit does not exceed the maximum.
+    if conjur_config.api_resource_list_limit_max.positive? && !params[:count]
```
micahlee, line 16: This comment should be amended to also explain the `!params[:count]` portion. To explain that we don't want to apply a default limit if this is a count request.

spec/controllers/resources_controller_spec.rb
aloncarmel111: Added

micahlee:
This looks good! Thanks for the updates, @aloncarmel111.

Would you please squash down the refactor and fix commits? Then this should be good to go!

micahlee (approved):
LGTM, thanks @aloncarmel111 !

PR #2582, opened 2022-06-23
Allow k8s test to be run locally and clean up test issues

micahlee:
### Desired Outcome

The desired outcome of this PR is to resolve issues in the k8s authenticator test suite causing regular build failures.

A side effect of this PR is to make it easier to run the GKE tests from a local development environment.

### Implemented Changes

This PR:
- Adds the GKE dev environment to the summon secrets file and adds documentation for running the test locally.
- Updates deprecated usage of the gcloud CLI
- Fixes an instance of using the OpenShift client to apply manifests to GKE.

### Connected Issue/Story

In support of https://github.com/cyberark/conjur/pull/2565

#### Changelog

- [ ] The CHANGELOG has been updated, or
- [x] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

- [ ] This PR includes new unit and integration tests to go with the code
  changes, or
- [x] The changes in this PR do not require tests

#### Documentation

- [x] Docs (e.g. `README`s) were updated in this PR
- [ ] A follow-up issue to update official docs has been filed here: [insert issue ID]()
- [ ] This PR does not require updating any documentation

#### Behavior

- [ ] This PR changes product behavior and has been reviewed by a PO, or
- [ ] These changes are part of a larger initiative that will be reviewed later, or
- [ ] No behavior was changed with this PR

#### Security

- [ ] Security architect has reviewed the changes in this PR,
- [ ] These changes are part of a larger initiative with a separate security review, or
- [x] There are no security aspects to these changes

PR #2584, opened 2022-06-23
Use the ShardedThreadedConnectionPool to disconnect after policy load

ajay-vel:
### Desired Outcome
The desired outcome of this PR is to resolve the memory leak in the postgres container causing it to OOM.

### Implemented Changes
This PR:

Makes use of [ShardedThreadedConnectionPool](https://www.rubydoc.info/github/jeremyevans/sequel/Sequel/ShardedThreadedConnectionPool) in Sequel so that active connections can be closed.
Closing active connections from conjur ensures that memory consumption in postgres stays low and a new connection for conjur is established every-time on policy load instead of using the existing connection.

### - Connected Issue/Story
- In support of [ONYX-18692](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-18692)

### Changelog

        CHANGELOG update

### Test coverage
 
        changes, or

### Documentation

- [ ] Docs (e.g. READMEs) were updated in this PR
- [ ] A follow-up issue to update official docs has been filed here: [insert issue ID](https://github.com/cyberark/conjur/pull/2582)

### Behavior

- [x]  No behavior was changed with this PR

### Security
 
### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added configuration for default value for maximum number of results return to `/resources` request
   [cyberark/conjur#2510](https://github.com/cyberark/conjur/pull/2510)
 
+### Fixed
+- Previously, the temporary schemas used to modify Conjur policy
+  caused the Postgres database catalog cache to leak memory over time,
+  leading to an eventual crash. Now, we recycle the database
+  connection after modifying policy to free this cache and prevent
+  the memory leak from occurring.
+  [cyberark/conjur#2584](https://github.com/cyberark/conjur/pull/2584)
+
 ### Security
 - Update rack to 2.2.3.1 to resolve CVE-2022-3023
   [cyberark/conjur#2564](https://github.com/cyberark/conjur/pull/2564)
--- a/app/models/loader/create_policy.rb
+++ b/app/models/loader/create_policy.rb
@@ -17,6 +17,8 @@ module Loader
       @loader.delete_shadowed_and_duplicate_rows
 
       @loader.store_policy_in_db
+
+      @loader.release_db_connection
     end
 
     def new_roles
--- a/config/application.rb
+++ b/config/application.rb
@@ -47,7 +47,15 @@ module Conjur
       Sequel.extension(:core_extensions, :postgres_schemata)
       Sequel::Model.db.extension(:pg_array, :pg_inet)
     end
-
+    
+    #The default connection pool does not support closing connections.
+    # We must be able to close connections on demand to clear the connection cache
+    # after policy loads [cyberark/conjur#2584](https://github.com/cyberark/conjur/pull/2584)
+    # The [ShardedThreadedConnectionPool](https://www.rubydoc.info/github/jeremyevans/sequel/Sequel/ShardedThreadedConnectionPool) does support closing connections on-demand.
+    # Sequel is configured to use the ShardedThreadedConnectionPool by setting the servers configuration on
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added configuration for token TTL
   [cyberark/conjur#2510](https://github.com/cyberark/conjur/pull/2510)
 
+### Fixed
+- Memory leak in postgres by active conjur connections. Now active connections are closed on policy creations and updations.
```
micahlee, line 21: We should provide a little more details throughout this PR to help answer the "why" question for others in the future.

Specifically for the changelog, I would suggest something like:

```md
- Previously, the temporary schemas used to modify Conjur policy
  caused the Postgres database cache to leak memory over time,
  leading to an eventual crash. Now, we recycle the database
  connection after modifying policy to free this cache and prevent
  the memory leak from occurring.
  [Link to PR...]()
```

app/models/loader/create_policy.rb
```diff
@@ -17,6 +17,8 @@ def call
       @loader.delete_shadowed_and_duplicate_rows
 
       @loader.store_policy_in_db
+
+      @loader.db.disconnect
```
micahlee, line 21: Generally when you see nested member access like this (`x.y.z`), it's a code smell that we're breaking [Demeter's law](https://en.wikipedia.org/wiki/Law_of_Demeter).

In this case, what we should do is add a new method to the loader that describes in our problem domain what we're doing (e.g. `@loader.release_db_connection` or `@loader.recycle_db_connection`).

**Importantly,** that method needs an extensive header comment explaining why this manual disconnect is necessary with relevant links to the bug in postgres, the wiki explaining the cache, etc. Otherwise the context is going to be lost to history.

I would also suggest adding that same detail to the commit description so that this information is really easy to uncover in the future.

config/application.rb
```diff
@@ -47,7 +47,10 @@ class Application < Rails::Application
       Sequel.extension(:core_extensions, :postgres_schemata)
       Sequel::Model.db.extension(:pg_array, :pg_inet)
     end
-
+    # Configure Sequel to use the ShardedThreadedConnectionPool
+    # so that we can close connections on-demand.
+    config.sequel.servers = {}
```
micahlee, line 52: This comment needs to be expounded a bit to explain that:
- The default connection pool does not support closing connections
- We must be able to close connections on demand to clear the connection cache after policy loads (perhaps link to this PR in the comment as well)
- The ShardedThreadedConnectionPool does support closing connections on-demand (perhaps link to the [docs here](https://www.rubydoc.info/github/jeremyevans/sequel/Sequel/ShardedThreadedConnectionPool))
- Sequel is configured to use the ShardedThreadedConnectionPool by setting the `servers` configuration on the database connection (perhaps link to the [docs here](https://github.com/jeremyevans/sequel/blob/master/doc/sharding.rdoc#the-servers-database-option-))

micahlee (changes requested):

micahlee (commented):
This a couple of further follow ups. Also, I'm not entirely sure if it was accidental or intentional, but we should spell out 'policy' in the commit message (`plcy` -> `policy`).

If you're trying to use a shorter message, I would suggest changing the headline and adding a body like:
```
Disconnect the database connection after policy loads

... Add more detail about why (Postgres's catcache) and how (with the shardedconnection pool) in the commit message body...
```

micahlee (approved):
LGTM, thanks for adding the extra comments, @ajay-vel !

micahlee (approved):

PR #2594, opened 2022-07-04
Mandate specific annotation on host creation under condition (currently true for Conjur Cloud)

oburstein-hub:
### Desired Outcome

Adding mandatory ApiKEY annotation for host upon relevant env variable is set to false

### Implemented Changes

In order to implement it - we are going to add flag to conjur_config.rb. The flag is called 'authn_api_key_default' and its default value is 'true' in order to preserve existing functionality in conjur-oss.

Also - we are adding validation to Host objects in loader/types.rb. 

If the flag is provided as 'false' the validations are mandating annotation "authn/api-key" with value true

### Connected Issue/Story

CyberArk internal issue link: [[ONYX-22003](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-22003)]

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update. (OSS users are not impacted by this change)

#### Test coverage

  changes, or

#### Documentation

#### Behavior

- [X] This PR changes product behavior and has been reviewed by a PO, or (but only under a flag)

#### Security

### DIFF
--- a/app/models/loader/types.rb
+++ b/app/models/loader/types.rb
@@ -134,8 +134,36 @@ module Loader
     end
 
     class Host < Record
-      def verify; end
       def_delegators :@policy_object, :restricted_to
+
+      # This is a temporary policy validation check to ensure that we're not
+      # creating hosts that will fail API key-based authentication by default
+      # in the future.
+      def future_api_key_auth_will_fail?
+        # The default config value is to allow API key authentication, so if this is
+        # either the default or set to true, then future API key authentication will
+        # continue to work and we don't need to reject this policy.
+        return false if Rails.application.config.conjur_config.authn_api_key_default
+
+        # If the default API authentication config is to disallow it, and the host
+        # does not explicitly state the policy authors intentions with the
+        # `authn/api-key` annotation with value true, then we should reject this until the annotation
+        # is added to the policy object.
+        self.annotations&.[]("authn/api-key").nil? or self.annotations["authn/api-key"] == false
+      end
+
+      def verify
+        # If policy contains a host with annotation authn/api-key effectively false, either by explicit
+        # value or by default value, then policy load is blocked.
+        if future_api_key_auth_will_fail?
+          message = "API key authentication for hosts is disabled by default and " \
+              "will be removed in a future release. Add the 'authn/api-key' " \
+              "annotation to this host with the value 'true' to " \
+              "ensure authentication works as expected for this host in the " \
+              "future."
+          raise Exceptions::InvalidPolicyObject.new(self.id, message: message)
+        end
+      end
 
       def create!
         self.handle_restricted_to(self.roleid, restricted_to)
--- a/lib/conjur/conjur_config.rb
+++ b/lib/conjur/conjur_config.rb
@@ -28,6 +28,7 @@ module Conjur
       api_resource_list_limit_max: 0,
       user_authorization_token_ttl: 480, # The default TTL of User is 8 minutes
       host_authorization_token_ttl: 480, # The default TTL of Host is 8 minutes
+      authn_api_key_default: true,
       authenticators: []
     )
 
--- a/spec/models/loader/types.rb
+++ b/spec/models/loader/types.rb
@@ -85,3 +85,67 @@ describe Loader::Types::User do
     end
   end
 end
+
+describe Loader::Types::Host do
+  let(:host) do
+    host = Conjur::PolicyParser::Types::Host.new
+    host.id = resource_id
+    if api_key != ''
+      host.annotations =  { "authn/api-key" => api_key }
+    end
+    Loader::Types.wrap(host, self)
+  end
+
+  describe '.verify' do
+    context 'when CONJUR_AUTHN_API_KEY_DEFAULT is true' do
+      before do
+        allow(ENV).to receive(:[]).with('CONJUR_AUTHN_API_KEY_DEFAULT').and_return('true')
+        Rails.application.config.conjur_config.authn_api_key_default = true
+      end
+
+      context 'when creating host with api-key annotation true' do
+        let(:resource_id) { 'myhost@admin' }
+        let(:api_key) { true }
+        it { expect { host.verify }.to_not raise_error }
+      end
+
+      context 'when creating host with api-key annotation false' do
+        let(:resource_id) { 'myhost@cyberark' }
+        let(:api_key) { false }
+        it { expect { host.verify }.to_not raise_error(Exceptions::InvalidPolicyObject) }
+      end
+
+      context 'when creating host without api-key annotation' do
+        let(:resource_id) { 'myhost@cyberark' }
+        let(:api_key) { '' }
+        it { expect { host.verify }.to_not raise_error(Exceptions::InvalidPolicyObject) }
+      end
+    end
+
+    context 'when CONJUR_AUTHN_API_KEY_DEFAULT is false' do
+      before do
+        allow(ENV).to receive(:[]).with('CONJUR_AUTHN_API_KEY_DEFAULT').and_return('false')
+        Rails.application.config.conjur_config.authn_api_key_default = false
+      end
+
+      context 'when creating host with api-key annotation true' do
+        let(:resource_id) { 'myhost@admin' }
+        let(:api_key) { true }
+        it { expect { host.verify }.to_not raise_error }
+      end
+
+      context 'when creating host with api-key annotation false' do
+        let(:resource_id) { 'alice@cyberark' }
+        let(:api_key) { false }
+        it { expect { host.verify }.to raise_error }
+      end
+
+      context 'when creating host without api-key annotation' do
+        let(:resource_id) { 'alice@cyberark' }
+        let(:api_key) { '' }
+        it { expect { host.verify }.to raise_error }
+      end
+    end
+
+  end
+end


app/models/loader/types.rb
```diff
@@ -134,8 +134,24 @@ def verify; end
     end
 
     class Host < Record
-      def verify; end
-      def_delegators :@policy_object, :restricted_to
+      def_delegators :@policy_object, :restricted_to, :apikey
```
micahlee, line 137: It's not clear what adding `:apikey` is accomplishing here. Can you elaborate on that?

app/models/loader/types.rb
```diff
@@ -134,8 +134,24 @@ def verify; end
...
+
+      def api_key
```
micahlee, line 139: This doesn't actually return an API, but rather, whether they're enabled. So this should probably be named something like `api_key_enabled?` (The '?' at the end is a common Ruby idiom for methods that return a boolean value.

app/models/loader/types.rb
```diff
@@ -134,8 +134,24 @@ def verify; end
...
+        return annot unless annot.nil?
+
+        config = Rails.application.config.conjur_config.authn_api_key_default
+        return config unless config.nil?
+
+        true
+      end
+
+      def verify
+        if api_key == false
```
micahlee, line 150: This can change to:
```suggestion
        if !api_key_enabled?
```

app/models/loader/types.rb
```diff
@@ -134,8 +134,24 @@ def verify; end
...
+          message = "Annotation authn/api-key is effectively false for host."
+          raise Exceptions::InvalidPolicyObject.new(self.id, message: message)
```
micahlee, line 152: Is preventing policy from loading if the API key is disabled really what you intend here?

That seems unusual to me. If that is the intention, this should commented with more context.

micahlee (changes requested):
Hey @egvili,

I left some comments for you. Would you please also update the PR title and description to better reflect the current change and to insure that the checklist is ready to go for review?

imheresamir:
Please ensure the commit history is clean before merge, thanks

app/models/loader/types.rb
egvili: This is the intention. I see some other cases where `verify` raises an exception when verification fails.
I'll add comments to assure readers understand the meaning of this exception

app/models/loader/types.rb
egvili: It is no longer needed. I deleted it

app/models/loader/types.rb
egvili: Done

app/models/loader/types.rb
egvili: Done

egvili:
> Please ensure the commit history is clean before merge, thanks

I'll squash all the commits on merge, so they won't be as messy

app/models/loader/types.rb
```diff
@@ -134,9 +134,27 @@ def verify; end
...
       def_delegators :@policy_object, :restricted_to
 
+      def api_key_enabled?
+        annot = self.annotations&.[]("authn/api-key")
+        return annot unless annot.nil?
+
+        config = Rails.application.config.conjur_config.authn_api_key_default
+        return config unless config.nil?
+
+        true
```
micahlee, line 146: The config value should never be `nil`, as this same check is part of the config class itself. This can return the value directly without the additional condition check.

app/models/loader/types.rb
micahlee: I don't understand. Why are we preventing this policy from loading?

app/models/loader/types.rb
egvili: I went on the safe side, but I will remove the check if you're sure it's not needed

app/models/loader/types.rb
egvili: This validation is temporary until we complete the feature optional api key. When the optional api key feature is completed,  the behavior will be that only hosts with the annotation will have api-key. 
We want to create alignment between hosts created before the feature and after it, so for now we enforce them to have the annotation (as they do have api-key and the annotation should reflect that)

micahlee: I'm working on gathering some additional input on this from the team. I don't think we should be failing a policy load like this based on partial work, but I'm checking on second opinions.

It's also not clear how you intend to use this part of the change with this in place. I thought the intention was to start using this annotation in Conjur Cloud in anticipation for disabling API keys for those roles in the future. Is that not why this PR is needed ahead of the actual change to disable API key authentication?

Future work to disable API key authentication will have to be based on more than whether an API key is present for an existing host (e.g. it will need to check this annotation at the point of authentication). So I'm not sure the alignment you're describing is necessary or justifies this enforcement on policy load.

egvili: Hi @micahlee , 
Similar blocking of policy load already exists in several scenarios, for example:
1. policy includes host factory without layer
2. policy includes creation users in Conjur Cloud

The final intention is not to disable API key **authentication**, but rather to disable API key **creation**. Meaning, host without this annotation will be created without having API key at all (thereby it could not authenticate using it).
So, by applying this validation right now, we eliminate creation of hosts that have API key but don't have an annotation.

micahlee: Hey @egvili,

I'm not suggesting we shouldn't do policy validation in general. Consider, however, what this validation failure represents: You've tried to create a host that we expect not to support API key authentication _at some point in the future_, and so you can't create this host. We're using policy validation for something other than rejecting invalid policy.

When you go to implement the actual feature change, you already have to handle the case that there are hosts that existed before that you either need to remove their API keys, or otherwise prevent their use. So there is minimal, if any development value in preventing this policy from loading. And in fact, this becomes debt that we must track and ensure is removed when the rest of the change takes place.

I can see a usability benefit of letting a user know that we expect the behavior of something they've done will change soon. The piece I can't approve for this is that the policy load is rejected. This is using policy validation in a way that sets a poor precedent for future work. I can think of a couple of ways to meet in the middle:

- The simplest would be to change this so that it writes a warning to the Conjur server log rather than raises a validation exception.

- Next in complexity, but would be more visible, is to write the warning to the audit log instead of the server log.

- A most complex, but most visible, approach would be to distinguish between error validations and warning (or info) validations. This check would move to the latter category, allowing the policy to load, but returning the message to the client in the policy load response.

egvili: Hey @micahlee ,
I agree that this validation is unusual, as it enforces behavior in the present as a preparation for future use.
BUT we do that exactly so that we won't have to go and remove API keys from old hosts. When the feature is completed we will allow having API key for hosts, but only if they explicitly set this annotation. Thus, old hosts will be valid hosts that will not require any modification. 
If we change the exception to warning or something similar, then we will have invalid hosts in the day that api-key feature is completed. Moreover, fixing them is not just a matter of deleting their API key, since the user needs to create an alternative authentication method for them. We would really like to avoid our customers finding out that all of their hosts lost connection to Conjur.

imheresamir: > If we change the exception to warning or something similar, then we will have invalid hosts in the day that api-key feature is completed. Moreover, fixing them is not just a matter of deleting their API key, since the user needs to create an alternative authentication method for them.

It seems like the reason why there would be invalid hosts is because this annotation is being introduced before the full feature is complete. What is the driver behind merging this new annotation work into the main branch before the full feature is developed?

egvili: The driver to introduce this annotation now is that Conjur Cloud is about to go out to Beta and is going to have its first customers (and hosts). If this annotation is not part of Conjur when the Beta starts, then customers will create hosts without this annotation and that will  get into the corner where we'll have to address old hosts not having annotation although having API key.
This is why we push to do something out of the ordinary, as it could save us a whole lot of trouble moving forward.

egvili:
> Please ensure the commit history is clean before merge, thanks

@imheresamir , I will before the merge once the PR is approved

app/models/loader/types.rb
rrefael: @micahlee @imheresamir As @egvili mentioned above, the purpose here is to enable us to improve the security of Conjur Cloud in a later stage. Without entering this annotation now, we will not be able to identify which API keys of hosts we will be able to remove when hosts will not have API keys anymore.

This change is only intended to affect Conjur Cloud, the default behavior in OSS and Enterprise will remain the same. We don't have the ability not to create API key when host is created due to time consideration, and the annotation addition should be entered immediately to enable us going to the Beta.

micahlee: Hey @egvili and @rrefael, 

I have one concern to raise and then also another compromise to propose.

---

**First,** regarding these statements:

> Without entering this annotation now, we will not be able to identify which API keys of hosts we will be able to remove when host will not have API key anymore.

> If this annotation is not part of Conjur when the Beta starts, then customers will create hosts without this annotation and that will get into the corner where we'll have to address old hosts not having annotation although having API key.
>
> This is why we push to do something out of the ordinary, as it could save us a whole lot of trouble moving forward.

Handling roles that existed before this validation change went into effect must be in the future effort to disable API key authentication/creation. I understand you are focused on Conjur Cloud and have the opportunity for clean data there, however this change WILL live in OSS and Enterprise as well, and so it must properly handle the cases where it affects customers upgrading from prior versions who have roles that do have API keys and now must be disabled and/or removed.

We cannot disregard that flow or assume it won't be used.

---

**Second,** regarding a proposal forward for this. I can approve this with the following changes:

It sounds like the most important condition is if the config value (`authn_api_key_default`) is set to `false` and the annotation is missing, correct? If so, let's only fail policy on those conditions.

So instead of `api_key_enabled?`, the method should point to the actual desired outcome, with something like `future_api_key_auth_will_fail?`:

```ruby
# This is a temporary policy validation check to ensure that we're not
# creating hosts that will fail API key-based authentication by default
# in the future.
def future_api_key_auth_will_fail?
  # The default config value is to allow API key authentication, so if this is
  # either the default or set to true, then future API key authentication will
  # continue to work and we don't need to reject this policy.
  return false unless authn_api_key_default

  # If the default API authentication config is to disallow it, and the host
  # does not explicitly state the policy authors intentions with the
  # `authn/api-key` annotation, then we should reject this until the annotation
  # is added to the policy object.
  return self.annotations&.[]("authn/api-key").nil?
end
```

Then, when we raise the error, it should be very clear how to resolve the failure:

```ruby
…
def verify
…
  if future_api_key_auth_will_fail?
    message = "API key authentication for hosts is disabled by default and " \
              "will be removed in a future release. Add the 'authn/api-key' " \
              "annotation to this host with the value 'true' or 'false' to " \
              "ensure authentication works as expected for this host in the " \
              "future."

     raise Exceptions::InvalidPolicyObject.new(self.id, message: message)
  end
…
end
…
```

---

WDYT?

egvili: Hey @micahlee, 
I don't fully understand this statement:
_however this change WILL live in OSS and Enterprise as well, and so it must properly handle the cases where it affects customers upgrading from prior versions who have roles that do have API keys and now must be disabled and/or removed_
Is the concern is having this (inactive) code in OSS/Ent or you mean that Conjur OSS/Ent is going to adopt this feature in the future?
* If OSS/Ent is not planning to adopt this feature, then an upgrade should not have an effect on existing hosts.
* If OSS/Ent is planning to adopt the feature, it means that it will change `authn_api_key_default` to false and will start deleting API keys with hosts without annotation. Which will prevent existing hosts from authenticating and will still impact existing hosts.


BTW, in `return false unless authn_api_key_default`, shouldn't it be with `if` instead of `unless`?

Perhaps we should make a Teams call to better understand

micahlee (approved):
LGTM, thanks for making the updates @egvili !

micahlee (approved):

PR #2608, opened 2022-07-20
Follow GCP function redirects in tests

micahlee:
The `ci/test_suites/authenticators_gcp/validate_gcf_url_accessible.sh` script has started failing with a 302 response in the GCP test suite. We now follow the redirect when we receive this response, allowing the test suite to successfully complete.


imheresamir (approved):
LGTM, thanks!

PR #2618, opened 2022-08-01
Handle annotation if given as string (host factory)

egvili:
### Desired Outcome

Issue: when host is created by host factory annotation is passed as string, thus the comparison `self.annotations["authn/api-key"] == false` always returns false. 

### Implemented Changes

Issue addresses annotation as both boolean or string, including different cases.
In addition more tests were added to capture the create host from host factory flow

### Connected Issue/Story

CyberArk internal issue link: [ONYX-21404](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-21404)]

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

- [X] These changes are part of a larger initiative that will be reviewed later, or

#### Security

- [X] These changes are part of a larger initiative with a separate security review, or


spec/models/host_factory_spec.rb
```diff
@@ -88,5 +88,58 @@
...
+      end
+
+      context 'when CONJUR_AUTHN_API_KEY_DEFAULT is false' do
+        before do
+          allow(ENV).to receive(:[]).with('CONJUR_AUTHN_API_KEY_DEFAULT').and_return('false')
+          Rails.application.config.conjur_config.authn_api_key_default = false
+        end
+
+        after do
+          Rails.application.config.conjur_config.authn_api_key_default = true
```
doodlesbykumbi, line 121: You can avoid the after block in the 2 spec files in this PR by stubbing the `authn_api_key_default` in the before block.
```ruby
allow(Rails.application.config.conjur_config).to receive(:authn_api_key_default).and_return('false')
```

doodlesbykumbi: The good thing about stubbing is that it does the clean up for you

egvili: @doodlesbykumbi , It is currently stubbed in the before block, no? 
before do
     allow(ENV).to receive(:[]).with('CONJUR_AUTHN_API_KEY_DEFAULT').and_return('false')
     Rails.application.config.conjur_config.authn_api_key_default = false
 end
When I don't put the cleanup in the after block, other tests in host_factory_spec.rb occasionally failed with message related to the config set to false. Maybe the cleanup happens only at the end of the entire class?

doodlesbykumbi: @egvili I think ` Rails.application.config.conjur_config.authn_api_key_default = false` doesn't count as stubbing, since it explicitly setting the value. I might be missing something but I thought instead using `allow(..).and_return(...)` results in rspec intercepting the calls only for the duration of the test.

egvili: @doodlesbykumbi , I now understand what you meant. Please check now

spec/models/host_factory_spec.rb
```diff
@@ -88,5 +88,54 @@
         expect(host.role.memberships_as_member.map(&:role)).to eq(host_factory.role.layers)
       end
     end
+
+    describe 'verify create host given AUTHN_API_KEY config' do
+      context 'when CONJUR_AUTHN_API_KEY_DEFAULT is true' do
+        before do
+          allow(ENV).to receive(:[]).with('CONJUR_AUTHN_API_KEY_DEFAULT').and_return('true')
+          allow(Rails.application.config.conjur_config).to receive(:authn_api_key_default).and_return(true)
```
micahlee, line 96: I agree with @doodlesbykumbi  on doubling the config object itself. With this I don't believe you need the line above to double ENV, do you? (Same on line 116)

egvili: done

app/models/loader/types.rb
```diff
@@ -149,7 +149,7 @@ def future_api_key_auth_will_fail?
         # does not explicitly state the policy authors intentions with the
         # `authn/api-key` annotation with value true, then we should reject this until the annotation
         # is added to the policy object.
-        self.annotations&.[]("authn/api-key").nil? or self.annotations["authn/api-key"] == false
+        self.annotations&.[]("authn/api-key").nil? or self.annotations["authn/api-key"].to_s.casecmp?("false")
```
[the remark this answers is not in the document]
egvili: done

micahlee (approved):
LGTM! Thanks for the update 😄

PR #2621, opened 2022-08-03
Add unit tests for socket server and list providers

mFelgate:
### Desired Outcome

### Implemented Changes

particular, consider:*

### Connected Issue/Story

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


spec/app/domain/util/socket_server_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+describe Util::SocketService do
+
+  let(:service)  do
+    Util::SocketService.new(
+      socket: socket
+    )
+  end
+
+  describe('#run') do
+    context 'When a socket is started and the socket file already exists' do
```
jvanderhoof, line 14: All the `context` blocks that start with "when a socket is started" should go into separate blocks, like:
```ruby
context 'when a socket is started' do
  context 'when the socket file already exists' do
    it 'raises an error' do
      ...
    end
  end
  context 'when the directory can not be found' do
    ...
   end
end
```

spec/app/domain/util/socket_server_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      let(:socket) {'/run/authn-local/.socket'}
+      it 'raises and error' do
```
jvanderhoof, line 16:
```suggestion
      it 'raises an error' do
```

spec/app/domain/util/socket_server_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+          service.run do |passed_arguments|
+            Authentication::AuthnOidc::V2::Commands::ListProviders.new.call(
+              message: passed_arguments
+            )
+          end
+        end.to raise_error("Socket: /run/authn-local/.socket already exists")
+      end
+    end
+
+    context 'When a socket is started and directory ca not be found' do
```
jvanderhoof, line 27:
```suggestion
    context 'When a socket is started and directory can not be found' do
```

spec/app/domain/util/socket_server_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    context 'When a socket is started and given a code block' do
+      let(:socket) {'/run/authn-local/.socket100'}
+      it 'will handle the input with the code block' do
+        thread = Thread.new do
+          service.run do |passed_arguments|
+             passed_arguments
+          end
+        end
+        thread.join(10)
```
jvanderhoof, line 48: I ran into problems running this locally (server was running in the foreground.  I solved it with the following:
```suggestion
        # Allow the socket server time to start
        sleep(2)
```
`join` holds the thread open (for ten seconds in this case), which isn't what we want.  We need the server in the background so RSpec can run in the foreground.

spec/app/domain/commands/authentication/issue_token_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let(:issue_token)  do
+    Commands::Authentication::IssueToken.new()
+  end
+
+  def create_account
+    Account.create("test")
+  end
+
+  describe('#call') do
+    context 'When you issue token for an existing account' do
```
jvanderhoof, line 14: Same comment as below.  All contexts which start with `"when you issue token"` should go into a context with the more specific contexts inside.

app/domain/authentication/authn_oidc/v2/commands/list_providers.rb
```diff
@@ -10,7 +10,9 @@ module Commands
         ListProviders ||= CommandClass.new(
           dependencies: {
             json_lib: JSON,
-            logger: Rails.logger
+            logger: Rails.logger,
+            repo: DB::Repository::AuthenticatorRepository,
```
jvanderhoof, line 14: Good catch here.

app/domain/authentication/authn_oidc/v2/commands/list_providers.rb
```diff
@@ -10,7 +10,9 @@ module Commands
...
+            provider: Authentication::AuthnOidc::V2::Views::ProviderContext.new()
```
jvanderhoof, line 15: Good catch on the missed dependency injection.

app/domain/authentication/authn_oidc/v2/commands/list_providers.rb
jvanderhoof, line 14: The `AuthenticatorRepository` shouldn't be injected here because it's a dependency of the `AuthenticatorRepository`.  Instead, pass the initialized `AuthenticatorRepository`:

```suggestion
            authenticator_repository: nil,
```
now in `ui.rake`
```ruby
Authentication::AuthnOidc::V2::Commands::ListProviders.new(
        logger: logger,
        authenticator_repository: DB::Repository::AuthenticatorRepository.new(
          logger: logger,
          data_object: Authentication::AuthnOidc::V2::DataObjects::Authenticator
        )
      ).call(
        message: passed_arguments
      )
```

**Note**: in general, please strongly prefer descriptive, non-abbreviated variable and argument names. It makes the code easier to understand.

spec/app/domain/util/socket_server_spec.rb
mFelgate: it works locally for me, but doesnt join hold it open until its complete or it hits 10? i just didnt want to do a sleep cuz it might get flakey depending on how long it takes for the socket to open. and i had it failing sometimes with sleep 5

app/domain/authentication/authn_oidc/v2/commands/list_providers.rb
```diff
@@ -10,31 +13,24 @@ module Commands
...
-                data_object: Authentication::AuthnOidc::V2::DataObjects::Authenticator,
-                logger: @logger
-              ).find_all(
-                account: account,
-                type: 'authn-oidc'
-              )
+            authenticators = @authenticatorRepository.find_all(account: account, type: 'authn-oidc')
             @logger.debug("#{self.class}##{__method__} - #{authenticators.inspect}")
-            Authentication::AuthnOidc::V2::Views::ProviderContext.new(
-              logger: @logger
```
andytinkham, line 31: With moving the ProviderContext.new up to line 18, does this lose the inclusion of the logger?

spec/app/domain/authentication/authn-oidc/v2/commands/ListProviders_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      allow(double).to receive(:discovery_information).and_return(endpoint)
+    end
+  end
+
+  let(:endpoint) { double(authorization_endpoint: '"http://test"') }
+
+  let(:res) do
+    [{ name: "foo",
+       redirect_uri: "\"http://test\"?client_id=ConjurClient&response_type=code&scope=openid%20email%20profile" \
+         "&state=foostate&nonce=secret&redirect_uri=http%3A%2F%2Fconjur%2Fauthn-oidc%2Fcucumber%2Fauthenticate",
```
andytinkham, line 29: Should this test differentiate somehow for foononce and barnonce instead of just "secret" for both? Are there other tests that make sure the nonce is unique each time?

andytinkham: Also, if we're generating state in the UI now, it's probably not "foostate" and "barstate". Seems these tests might need some reworking after the events of the past week.

andytinkham (changes requested):
After looking at ListProviders_spec.rb, I realized these tests were designed before we shifted our state and nonce handling plans. I suspect they'll need to be reworked now before we can merge them in.

PR #2622, opened 2022-08-04
Install updated kernel-headers from CentOS repo

micahlee:
This PR installs an updated version of kernel-headers from the CentOS 8 repos in the Conjur UBI image to resolve CVE-2022-1012 and CVE-2022-32250.


imheresamir (approved):
LGTM

PR #2637, opened 2022-08-28
Ofira jwt OIDC header1

oburstein-hub:
### Desired Outcome

Adding capability to do OIDC authentication when the ID Token is passed in header

### Implemented Changes

The changes are done in controller and its helping classes relevant to OIDC
The changes are to handle requests that get ID Token in header instead of body of request
The request that comes without IDToken in body is trying to read Authorization token from header and 
checks if it has a "Bearer " prefix. After it the prefix is cut-out and the token after it is passed as an ID Toke
to the existing validation mechanism

### Connected Issue/Story

Resolves [ONYX-23732](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-23732)

### Definition of Done
OIDC authentication for UI is implemented and passed first testing

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

- [ x] This PR changes product behavior and has been reviewed by a PO, or

#### Security

### DIFF
--- a/app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
+++ b/app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
@@ -11,14 +11,14 @@ module Authentication
       inputs: %i[authenticator_input]
     ) do
       extend(Forwardable)
+      # The 'request' field in the 'authenticator_input' object will be used to extract the id token from the request header
       def_delegators(:@authenticator_input, :service_id, :authenticator_name,
-                     :account, :username, :webservice, :credentials, :client_ip,
+                     :account, :username, :webservice, :credentials, :client_ip, :request,
                      :role)
 
       def call
         validate_account_exists
         validate_service_id_exists
-        validate_credentials_include_id_token
         verify_and_decode_token
         validate_conjur_username
         input_with_username
@@ -36,15 +36,6 @@ module Authentication
         raise Errors::Authentication::AuthnOidc::ServiceIdMissing unless service_id
       end
 
-      def validate_credentials_include_id_token
-        id_token_field_name = "id_token"
-
-        # check that id token field exists and has some value
-        if decoded_credentials.fetch(id_token_field_name, "") == ""
-          raise Errors::Authentication::RequestBody::MissingRequestParam, id_token_field_name
-        end
-      end
-
       def verify_and_decode_token
         @decoded_token = @verify_and_decode_token.(
           provider_uri: oidc_authenticator_secrets["provider-uri"],
@@ -53,9 +44,45 @@ module Authentication
         )
       end
 
-      # The credentials are in a URL encoded form data in the request body
+      # The credentials are in a URL encoded form data in the request body or in the request header
       def decoded_credentials
-        @decoded_credentials ||= Hash[URI.decode_www_form(credentials)]
+        @decoded_credentials ||= begin
+          return token_from_body if token_from_body
+
+          return token_from_header if token_from_header
+
+          # If the token is not in the header or body, raise a missing param exception
+          raise Errors::Authentication::RequestBody::MissingRequestParam, 'id_token'
+        end
+      end
+
+      def token_from_header
+        @token_from_header ||= begin
+          return nil unless request&.headers&.key?("HTTP_AUTHORIZATION")
+
+          # Extract the token from the authorization header
+          id_token = request.headers['HTTP_AUTHORIZATION'].split(' ', -1)
+
+          # Verify the token is given as a bearer token
+          return nil unless id_token[0] == "Bearer" && id_token.length == 2
+
+          # Return the token formatted as it would be if extracted from the body
+          # as `application/x-www-form-urlencoded` data.
+          {
+            # URL decode the ID token from the header
+            "id_token" => URI.decode_www_form_component(id_token[1])
+          }
+        end
+      end
+
+      def token_from_body
+        @token_from_body ||= begin
+          # Parse the request body
+          credential_token = Hash[URI.decode_www_form(credentials)]
+
+          # Return the parsed body if it include the token
+          credential_token unless credential_token.fetch('id_token', "").empty?
+        end
       end
 
       def oidc_authenticator_secrets
--- a/cucumber/authenticators_oidc/features/support/authn_oidc_helper.rb
+++ b/cucumber/authenticators_oidc/features/support/authn_oidc_helper.rb
@@ -24,6 +24,14 @@ module AuthnOidcHelper
     post(path, payload)
   end
 
+  def authenticate_id_token_with_oidc_in_header(service_id:, account:, id_token: parsed_id_token)
+    service_id_part = service_id ? "/#{service_id}" : ""
+    path = "#{conjur_hostname}/authn-oidc#{service_id_part}/#{account}/authenticate"
+    headers = {}
+    headers["Authorization"] = "Bearer #{id_token}"
+    post(path, {}, headers)
+  end
+
   def authenticate_code_with_oidc(service_id:, account:, code: url_oidc_code, state: url_oidc_state)
     path = "#{create_auth_url(service_id: service_id, account: account, user_id: nil)}"
     get(url_with_params(path: path,code: code, state: state ))
--- a/spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
+++ b/spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
@@ -27,13 +27,14 @@ RSpec.describe(Authentication::AuthnOidc::UpdateInputWithUsernameFromIdToken) do
   # request mock
   ####################################
 
-  def mock_authenticate_oidc_request(request_body_data:)
+  def mock_authenticate_oidc_request(request_body_data: , request_headers:)
     double('AuthnOidcRequest').tap do |request|
       request_body = StringIO.new
       request_body.puts(request_body_data)
       request_body.rewind
 
       allow(request).to receive(:body).and_return(request_body)
+      allow(request).to receive(:headers).and_return(request_headers)
     end
   end
 
@@ -42,23 +43,43 @@ RSpec.describe(Authentication::AuthnOidc::UpdateInputWithUsernameFromIdToken) do
   end
 
   let(:authenticate_id_token_request) do
-    mock_authenticate_oidc_request(request_body_data: "id_token={\"id_token_username_field\": \"alice\"}")
+    mock_authenticate_oidc_request(request_body_data: "id_token={\"id_token_username_field\": \"alice\"}", request_headers: nil)
   end
 
   let(:authenticate_id_token_request_missing_id_token_username_field) do
-    mock_authenticate_oidc_request(request_body_data: "id_token={}")
+    mock_authenticate_oidc_request(request_body_data: "id_token={}", request_headers: nil)
   end
 
   let(:authenticate_id_token_request_empty_id_token_username_field) do
-    mock_authenticate_oidc_request(request_body_data: "id_token={\"id_token_username_field\": \"\"}")
+    mock_authenticate_oidc_request(request_body_data: "id_token={\"id_token_username_field\": \"\"}", request_headers: nil)
   end
 
   let(:authenticate_id_token_request_missing_id_token_field) do
-    mock_authenticate_oidc_request(request_body_data: "some_key=some_value")
+    mock_authenticate_oidc_request(request_body_data: "some_key=some_value", request_headers: nil)
   end
 
   let(:authenticate_id_token_request_empty_id_token_field) do
-    mock_authenticate_oidc_request(request_body_data: "id_token=")
+    mock_authenticate_oidc_request(request_body_data: "id_token=", request_headers: nil)
+  end
+
+  let(:authenticate_id_token_request_id_token_in_header_field) do
+    mock_authenticate_oidc_request(request_body_data: "id_token=", request_headers: {"HTTP_AUTHORIZATION" => "Bearer {\"id_token_username_field\":\"alice\"}"})
+  end
+
+  let(:authenticate_id_token_request_invalid_id_token_in_header_field) do
+    mock_authenticate_oidc_request(request_body_data: "id_token=", request_headers: {"HTTP_AUTHORIZATION" => "{\"id_token_username_field\":\"alice\"}"})
+  end
+
+  let(:authenticate_id_token_request_empty_id_token_in_header_field) do
+    mock_authenticate_oidc_request(request_body_data: "id_token=", request_headers: {"HTTP_AUTHORIZATION" => ""})
+  end
+
+  let(:authenticate_id_token_request_missing_id_token_in_body_field) do
+    mock_authenticate_oidc_request(request_body_data: "", request_headers: {"HTTP_AUTHORIZATION" => "Bearer {\"id_token_username_field\":\"alice\"}"})
+  end
+
+  let(:authenticate_id_token_request_contain_only_bearer_in_header_field) do
+    mock_authenticate_oidc_request(request_body_data: "", request_headers: {"HTTP_AUTHORIZATION" => "Bearer"})
   end
 
   #  ____  _   _  ____    ____  ____  ___  ____  ___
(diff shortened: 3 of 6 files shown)


app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -12,7 +12,7 @@ module AuthnOidc
     ) do
       extend(Forwardable)
       def_delegators(:@authenticator_input, :service_id, :authenticator_name,
-                     :account, :username, :webservice, :credentials, :client_ip,
+                     :account, :username, :webservice, :credentials, :client_ip, :request,
```
rafis3, line 15: Isn't the `request` and `credentials` overlapping? The `credentials` value is supposed to be taken from the body of the request, but then we send the entire request as another parameter, which contains the body as well.

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
```diff
@@ -42,23 +45,35 @@ def request_body(request)
   end
 
   let(:authenticate_id_token_request) do
-    mock_authenticate_oidc_request(request_body_data: "id_token={\"id_token_username_field\": \"alice\"}")
+    mock_authenticate_oidc_request(request_body_data: "id_token={\"id_token_username_field\": \"alice\"}", request_headers: {"CONTENT_TYPE" => "text/plain"})
```
rafis3, line 48: Why does the test need this new content type? Doesn't that mean the API is broken from how it used to work?

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
```diff
@@ -42,23 +45,35 @@ def request_body(request)
...
+    mock_authenticate_oidc_request(request_body_data: "some_key=some_value", request_headers: {"CONTENT_TYPE" => "text/plain"})
   end
 
   let(:authenticate_id_token_request_empty_id_token_field) do
-    mock_authenticate_oidc_request(request_body_data: "id_token=")
+    mock_authenticate_oidc_request(request_body_data: "id_token=", request_headers: {"CONTENT_TYPE" => "text/plain"})
+  end
+
+  let(:authenticate_id_token_request_id_token_in_header_field) do
+    mock_authenticate_oidc_request(request_body_data: "id_token=", request_headers: {"HTTP_AUTHORIZATION" => "Bearer {\"id_token_username_field\":\"alice\"}"})
```
rafis3, line 68: I think we are missing tests where the `id_token` key is not present in the body at all, that the body is totally empty

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
```diff
@@ -42,23 +45,35 @@ def request_body(request)
...
+  end
+
+  let(:authenticate_id_token_request_invalid_id_token_in_header_field) do
+    mock_authenticate_oidc_request(request_body_data: "id_token=", request_headers: {"HTTP_AUTHORIZATION" => "{\"id_token_username_field\":\"alice\"}"})
```
rafis3, line 72: In general, I see that the given token is expected to be in cleartext and not in base64. Will that work for our new flow we need to support?

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
```diff
@@ -247,6 +262,108 @@ def request_body(request)
           expect { subject }.to raise_error(::Errors::Authentication::RequestBody::MissingRequestParam)
         end
       end
+
+      context "with valid id token in header" do
```
rafis3, line 266: This test also puts the token in the body. The name above implies that it will check the token from the header but that won't be the case. Am I right?

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
```diff
@@ -247,6 +262,108 @@ def request_body(request)
...
+          ::Authentication::AuthnOidc::UpdateInputWithUsernameFromIdToken.new(
+            verify_and_decode_token: authenticate_id_token_request_id_token_in_header_field,
+            validate_account_exists: mock_validate_account_exists(validation_succeeded: true)
+          ).call(
+            authenticator_input: input_
+          )
+        end
+      end
+
+      context "with valid id token in header" do
```
rafis3, line 287: This context has the exact name as above. Can you update it to describe its true purpose?

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
```diff
@@ -247,6 +262,108 @@ def request_body(request)
...
+            authenticator_input: input_
+          )
+        end
+
+        it "returns the input with the username inside it" do
+          expect(subject.username).to eql("alice")
+        end
+      end
+
+      context "with invalid id token in header" do
```
rafis3, line 314: Also in body, not just in header. Can you please update the description to be more accurate?

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
```diff
@@ -247,6 +262,108 @@ def request_body(request)
...
+            authenticator_input: input_
+          )
+        end
+
+        it "raises a MissingRequestParam error" do
+          expect { subject }.to raise_error(::Errors::Authentication::RequestBody::MissingRequestParam)
+        end
+      end
+
+      context "with empty id token in header" do
```
rafis3, line 341: Same as above, can you please update this description also with the state of the body of this test?

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
```diff
@@ -247,6 +262,108 @@ def request_body(request)
...
+
+        subject do
+          input_ = Authentication::AuthenticatorInput.new(
+            authenticator_name: 'authn-oidc',
+            service_id: 'my-service',
+            account: 'my-acct',
+            username: nil,
+            credentials: request_body(authenticate_id_token_request_missing_id_token_field),
+            client_ip: '127.0.0.1',
+            request: authenticate_id_token_request_empty_id_token_in_header_field
```
rafis3, line 352: Can you also add a test that checks that the Authorization header exists, with a value of Bearer but nothing after it?
Another test that I would add, is that there is an Authorization header, and the first word is not Bearer, while the second word is a valid token (should fail)

rafis3 (changes requested):
@oburstein-hub thanks, please take a look at my comments

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
oburstein-hub: done

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -36,19 +36,35 @@ def validate_service_id_exists
         raise Errors::Authentication::AuthnOidc::ServiceIdMissing unless service_id
       end
 
+      class << self
```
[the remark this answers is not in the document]
oburstein-hub: done

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
oburstein-hub: fixed

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
oburstein-hub: added

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
oburstein-hub: yes - they both work - I have checked

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
oburstein-hub: We will check token from header. it is missing in bode in this case

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
oburstein-hub: deleted it

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
oburstein-hub: done

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
oburstein-hub: done

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
oburstein-hub: done

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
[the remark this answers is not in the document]
rafis3: We are still left with the `class << self` block. Why do we need it?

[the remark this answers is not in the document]
oburstein-hub: managed to drop it now

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
jvanderhoof: Doesn't the `Bearer` value come through as a Base64 encoded JWT?

jvanderhoof: Just saw the comment below.  Maybe one of these is Base64 encoded to ensure the typical flow is tested?

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
```diff
@@ -247,6 +268,141 @@ def request_body(request)
...
+            service_id: 'my-service',
+            account: 'my-acct',
+            username: nil,
+            credentials: request_body(authenticate_id_token_request_id_token_in_header_field),
+            client_ip: '127.0.0.1',
+            request: authenticate_id_token_request_id_token_in_header_field
+          )
+
+          ::Authentication::AuthnOidc::UpdateInputWithUsernameFromIdToken.new(
+            verify_and_decode_token: mocked_decode_and_verify_id_token,
```
jvanderhoof, line 287: The `mocked_decode_and_verify_id_token` mock accepts and parses JSON: https://github.com/cyberark/conjur/blob/master/spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb#L18. 

It looks like this will fail if we pass Base64 encoded JSON. When the bearer token is passed in the body, is it passed as JSON?

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -36,13 +35,13 @@ def validate_service_id_exists
...
       end
 
-      def validate_credentials_include_id_token
-        id_token_field_name = "id_token"
-
-        # check that id token field exists and has some value
-        if decoded_credentials.fetch(id_token_field_name, "") == ""
-          raise Errors::Authentication::RequestBody::MissingRequestParam, id_token_field_name
+      def authorization_header_token(headers)
+        @id_token = if headers&.key?("HTTP_AUTHORIZATION")
```
jvanderhoof, line 39: The `@` denotes an instance variable, which means it's available outside the method. Unless we need `@id_token` outside this method, please use the local variable `id_token`.

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
jvanderhoof: @oburstein-hub, avoiding a cascading change feels like the right tradeoff.  Mind adding a comment as to why we're adding the `request` object (so future engineers understand the decision)?

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -53,9 +52,15 @@ def verify_and_decode_token
         )
       end
 
-      # The credentials are in a URL encoded form data in the request body
+      # The credentials are in a URL encoded form data in the request body or in the request header
       def decoded_credentials
         @decoded_credentials ||= Hash[URI.decode_www_form(credentials)]
+        id_token_field_name = "id_token"
```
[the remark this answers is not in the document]
aloncarmel111: fixed

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
aloncarmel111: fixed

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
aloncarmel111: added comment

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
aloncarmel111: The token is base64 encoded, we implemented same test as the `id_token` present in the body, will create integration test for Conjur cloud UI in different story

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
aloncarmel111: yes, we implemented same test as the `id_token` in the body.

aloncarmel111: yes, we implemented same test as the `id_token` in the body.

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -53,9 +52,17 @@ def verify_and_decode_token
         )
       end
 
-      # The credentials are in a URL encoded form data in the request body
+      # The credentials are in a URL encoded form data in the request body or in the request header
       def decoded_credentials
```
micahlee, line 56: Hey @aloncarmel111 , this is some tricky logic to follow, what do you think about something like this instead?

```ruby
      # The credentials are in a URL encoded form data in the request body or in the request header
      def decoded_credentials
        @decoded_credentials ||= begin
          return token_from_header if token_from_header

          return token_from_body if token_from_body

          # If the token is not in the header or body, raise a missing param exception
          raise Errors::Authentication::RequestBody::MissingRequestParam, 'id_token'
        end
      end

      def token_from_header
        @token_from_header ||= begin
          return nil unless request&.headers&.key?("HTTP_AUTHORIZATION")

          # Extract the token from the authorization header
          id_token = request.headers['HTTP_AUTHORIZATION'].split(' ', -1)

          # Verify the token is given as a bearer token
          return nil unless id_token[0] == "Bearer" && id_token.length == 2

          # Return the token formatted as it would be if extracted from the body
          # as `application/x-www-form-urlencoded` data.
          {
            # URL decode the ID token from the header
            "id_token" => URI.decode_www_form_component(id_token[1])
          }
        end
      end

      def token_from_body
        @token_from_body ||= begin
          # Parse the request body
          credential_token = Hash[URI.decode_www_form(credentials)]

          # Return the parsed body if it include the token
          credential_token unless credential_token.fetch('id_token', '').empty?
        end
      end
```

micahlee (commented):
Hi @aloncarmel111 , most of this looks good, thanks! I had one suggestion for trying to make the logic a little easier to read/follower. Let me know what you think!

micahlee:
Thanks @aloncarmel111 , the only blocker I have left is to squash the PR review commits into the one `Support authorization token in header for OIDC authentication` commit before approving. 

It looks like there are still unresolved discussions around base64 encoding the header in the tests with Jason and Rafi. So you can wait to do that squash until that is resolved, and then ping me for approval. Thanks!

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -53,9 +44,45 @@ def verify_and_decode_token
...
+          return token_from_body if token_from_body
+
+          return token_from_header if token_from_header
+
+          # If the token is not in the header or body, raise a missing param exception
+          raise Errors::Authentication::RequestBody::MissingRequestParam, 'id_token'
+        end
+      end
+
+      def token_from_header
```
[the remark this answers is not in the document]
jvanderhoof: I like the move to break up these methods to better cover edge cases.  I'd suggest passing the header in as an argument:

```suggestion
      def token_from_header(header)
```

Passing the header into the method will greatly simplify writing tests (which it looks like need to be written).

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
```diff
@@ -53,9 +44,45 @@ def verify_and_decode_token
...
+          # Return the token formatted as it would be if extracted from the body
+          # as `application/x-www-form-urlencoded` data.
+          {
+            # URL decode the ID token from the header
+            "id_token" => URI.decode_www_form_component(id_token[1])
+          }
+        end
+      end
+
+      def token_from_body
```
jvanderhoof, line 78: Same suggestion as above, maybe move to something like:

```ruby
token_from_body(Hash[URI.decode_www_form(credentials)]

def token_from_body(credential_token)
  # Return the parsed body if it include the token
  credential_token unless credential_token.fetch('id_token', "").empty?
end
```

The above change makes the method stateless and should simplify testing the happy and sad paths.

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
jvanderhoof: Please add at least a unit test here as well. I'm not comfortable adding functionality without tests to back it up.  The Core team will be on the line for supporting this work.

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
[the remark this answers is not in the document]
micahlee: @jvanderhoof 
> I'd suggest passing the header in as an argument:

These are internal private methods, so we wouldn't expect them to have unit tests for themselves. Rather, `request` is the input that would need to be manipulated for the public interface of `UpdateInputWithUsernameFromIdToken`. Adding parameters to these methods means we're passing internal class state between internal class methods. No?

If we want to unit tests these separately, that should be factored out and injected as dependencies here instead. I don't know that that's warranted in this case, but I don't think we should add parameters to them as class methods now.

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
aloncarmel111: added integration test

app/domain/authentication/authn_oidc/update_input_with_username_from_id_token.rb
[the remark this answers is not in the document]
jvanderhoof: Thanks @micahlee. I'm ok moving forward with what we have here.

cucumber/authenticators_oidc/features/support/authn_oidc_helper.rb
```diff
@@ -24,6 +24,14 @@ def authenticate_id_token_with_oidc(service_id:, account:, id_token: parsed_id_t
     post(path, payload)
   end
 
+  def authenticate_id_token_with_oidc_in_header(service_id:, account:, id_token: parsed_id_token)
+    service_id_part = service_id ? "/#{service_id}" : ""
+    path = "#{conjur_hostname}/authn-oidc#{service_id_part}/#{account}/authenticate"
+    headers = {}
+    headers["Authorization"] = "Bearer #{id_token}"
```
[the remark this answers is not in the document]
jvanderhoof: Never mind.  I just found the JWT generation here: https://github.com/cyberark/conjur/blob/0d12713fe96fc83a2297f5a651f4f0e70b8a0bcd/cucumber/authenticators_oidc/features/support/authn_oidc_helper.rb#L100

spec/app/domain/authentication/authn-oidc/update_input_with_username_from_id_token_spec.rb
jvanderhoof: Thanks!

jvanderhoof (approved):
Great work. Please rebase and cleanup commit history (ideally a single commit).

aloncarmel111 (dismissed):

micahlee (dismissed):

oburstein-hub (commented):
Looks good

aloncarmel111 (approved):

aloncarmel111:
@szh they are false positive

PR #2640, opened 2022-08-31
Prevent status webservices from being returned on the providers endpoint

mFelgate:
### Desired Outcome

Duplictae authenticators show up in the providers endpoint due to not removing the status webservices from the list of providers. This pr removes web services that end in /status from the list of authenticators

### Implemented Changes

particular, consider:*

Added the regex used in InstalledAuthenticators to remove the status webservices  
That the tests work effectively and the code is reasonably made  
No 
### Connected Issue/Story

CyberArk internal issue link: [ONYX-25530](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-25530)

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - List resources request (`GET /resources`) now produce audit events.
-  ([cyberark/conjur#2652](https://github.com/cyberark/conjur/pull/2652)
+  [cyberark/conjur#2652](https://github.com/cyberark/conjur/pull/2652)
 
 ### Changed
 - AWS Access Key Rotation now preserves only one key
+
+### Fixed
+- Removed Status webservices from the list providers endpoint
+  [cyberark/conjur#2640](https://github.com/cyberark/conjur/pull/2640)
 
 ## [1.18.4] - 2022-09-11
 
--- a/app/db/repository/authenticator_repository.rb
+++ b/app/db/repository/authenticator_repository.rb
@@ -1,19 +1,27 @@
 module DB
   module Repository
     class AuthenticatorRepository
-      def initialize(data_object:, resource_repository: ::Resource, logger: Rails.logger)
+      def initialize(
+        data_object:,
+        resource_repository: ::Resource,
+        logger: Rails.logger,
+        enabled_authenticators: Rails.application.config.conjur_config.authenticators
+      )
         @resource_repository = resource_repository
         @data_object = data_object
         @logger = logger
+        @enabled_authenticators = enabled_authenticators
       end
 
       def find_all(type:, account:)
-        @resource_repository.where(
+        enabled_authenticator_types = @enabled_authenticators.select { |authenticator| authenticator.match("#{type}") }
+                .map { |authenticator| "#{account}:webservice:conjur/#{authenticator}" }
+       @resource_repository.where(
           Sequel.like(
             :resource_id,
             "#{account}:webservice:conjur/#{type}/%"
           )
-        ).all.map do |webservice|
+        ).all.select { |webservice|enabled_authenticator_types.include?(webservice.resource_id) }.map do |webservice|
           load_authenticator(account: account, id: webservice.id.split(':').last, type: type)
         end.compact
       end
--- a/spec/app/db/repository/authenticator_repository_spec.rb
+++ b/spec/app/db/repository/authenticator_repository_spec.rb
@@ -8,9 +8,16 @@ RSpec.describe('DB::Repository::AuthenticatorRepository') do
   let(:repo) do
     DB::Repository::AuthenticatorRepository.new(
       resource_repository: resource_repository,
-      data_object: Authentication::AuthnOidc::V2::DataObjects::Authenticator
+      data_object: Authentication::AuthnOidc::V2::DataObjects::Authenticator,
+      enabled_authenticators: enabled_authenticators
     )
   end
+
+  let (:enabled_authenticators) {
+    %w[authn-oidc/foo-abc123
+       authn-oidc/baz-abc123
+       authn-oidc/bar-abc123]
+  }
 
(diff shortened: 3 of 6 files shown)


app/db/repository/authenticator_repository.rb
```diff
@@ -8,11 +8,18 @@ def initialize(data_object:, resource_repository: ::Resource, logger: Rails.logg
...
+
         @resource_repository.where(
           Sequel.like(
             :resource_id,
             "#{account}:webservice:conjur/#{type}/%"
           )
+        ).where(
+          Sequel.like(
+            identifier,
+            %r{conjur/(authn(?:-[^/]+)?(?:/[^/]+)?)$}
```
micahlee, line 21: It would be worth adding a comment here explaining how this regex prevents the status resources from being included in the result. (No forward slashes are allowed after `authn`).

spec/app/db/repository/authenticator_repository_spec.rb
```diff
@@ -89,6 +89,21 @@
...
+
+          context 'when webservices status are presents' do
+            before(:each) do
+                ::Resource.create(
+                  resource_id: "rspec:webservice:conjur/authn-oidc/foo-abc123/status",
+                  owner_id: "rspec:policy:conjur/authn-oidc/foo-abc123"
+                )
+              end
+
+            it { expect(repo.find_all(type: 'authn-oidc', account: 'rspec').length).to eq(2) }
```
micahlee, line 101: How cumbersome would it be to actually verify the contents of the list, rather than just the count? 

Being explicit about the expected outcome would be a more valuable assurance than just the number of results.

imheresamir:
When you get a chance could you fill out the PR description @mFelgate ? Thanks

app/db/repository/authenticator_repository.rb
micahlee: Also, given this isn't an urgent change, would should improve this re-use beyond just copying the regex between this location and https://github.com/cyberark/conjur/blob/master/app/domain/authentication/installed_authenticators.rb#L31.

For example, it should probably be a constant in one class or the other, and just reference in the other location:

```rb
# installed_authenticators.rb

...
class InstalledAuthenticators
    ...
    AUTHN_RESOURCE_PREFIX = "conjur/authn-"
    AUTHN_STATUS_FILTER = %r{conjur/(authn(?:-[^/]+)?(?:/[^/]+)?)$}
    ...
    # filter out nested status webservice
    .map { |id| id[AUTHN_STATUS_FILTER, 1] } 
    ...
```

```rb
# authenticator_repository.rb
...
...
Sequel.like(
  identifier,
  # filter out nested status webservice
  Authentication::InstalledAuthenticators::AUTHN_STATUS_FILTER
)
...
```

micahlee: @mFelgate Final comment here, and perhaps this is the most important. This needs an integration test that actually runs this query and verifies the result, because I don't believe this is going to work as-is.

The `Sequel.like` method expects a SQL  LIKE statement, and not a regex:
https://www.rubydoc.info/github/jeremyevans/sequel/Sequel%2FSQL%2FStringExpression.like

spec/app/db/repository/authenticator_repository_spec.rb
mFelgate: would not work since we and not brining over the real resource_id in the repository, but a created one based on the service_id which is the same between the status and normal web-services.

CHANGELOG.md
```diff
@@ -35,6 +39,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Reduces debug log verbosity.
   [cyberark/conjur#2639](https://github.com/cyberark/conjur/pull/2639)
+- Remove status webservices from providers endpoints
```
jvanderhoof, line 42: Each changelog entry should go under a single header, not multiple.  I think this makes more sense under `Fixed`.

CHANGELOG.md
```diff
@@ -13,11 +13,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 ### Added
 - List resources request (`GET /resources`) now produce audit events.
-  ([cyberark/conjur#2652](https://github.com/cyberark/conjur/pull/2652)
+  [cyberark/conjur#2652](https://github.com/cyberark/conjur/pull/2652)
 
 ### Changed
 - AWS Access Key Rotation now preserves only one key
 
+### Fixed
+- Removed Status webservices from the list providers endpoint
```
jvanderhoof, line 22: I'd recommend something like:
```suggestion
- OIDC Provider endpoint no longer includes duplicates when Status is enabled.
```

jvanderhoof:
I'd actually recommend making the changes in the `DB::Repository::AuthenticatorRepository` class.  If we updated the initializer and the `find_all` method to something like:

```ruby
module DB
  module Repository
    class AuthenticatorRepository
      def initialize(data_object:, resource_repository: ::Resource, logger: Rails.logger, enabled_authenticators: Rails.application.config.conjur_config.authenticators)
        @resource_repository = resource_repository
        @data_object = data_object
        @logger = logger
        @enabled_authenticators = enabled_authenticators
      end

      def find_all(type:, account:)
        enabled_authenticator_types = @enabled_authenticators
          .select { |authenticator| authenticator.match(/^#{type}/) }
          .map { |authenticator| "#{account}:webservice:conjur/#{authenticator}" }

        @resource_repository.where(
          Sequel.like(
            :resource_id,
            "#{account}:webservice:conjur/#{type}/%"
          )
        )
          .all
          .select { |webservice| enabled_authenticator_types.include?(webservice.id) }
          .map do |webservice|
            load_authenticator(account: account, id: webservice.id.split(':').last, type: type)
          end
          .compact
      end
      ...
```
we can limit the available authenticators to those enabled AND strip out the Status webservices.

PR #2644, opened 2022-09-06
Remove code and state from the info logs of the url request

mFelgate:
#### What does this PR do?
adds code and state to the list of paramters that are filtered in the log

Run the okta cucumber tests and then make sure code and state are filtered in the development.log


telday (approved):
LGTM

PR #2660, opened 2022-10-06
Fixed flakey conjurctl unit test

telday:
### Desired Outcome
Fix flakeyness in the unit test for `conjurctl server` which specifies a user password from stdin and an account specified.

### Implemented Changes

The thread reading stdout from the `conjurctl server` command was greatly simplified to reduce possible race conditions and issues with open/closed IO pipes which occasionally caused errors.

In order to ensure the fix worked properly the following command was used to run the test 100 times in sequence within the Conjur dev environment:
`seq 100 | xargs -Iz bundle exec rspec spec/conjurctl/server_spec.rb:39` which produced no errors (previously it produced ~4 errors per 100 test runs).

### Connected Issue/Story

CyberArk internal issue link: [ONYX-23533](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-23533)

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -114,72 +114,32 @@ def with_background_process(cmd, &block)
     cmd,
     pgroup: true
   ) do |stdin, stdout_and_err, wait_thr|
-    # We don't need to interact with stdin for the background process,
-    # so close that stream immediately.
-    stdin.close_write
-
+    output = StringIO.new
     # Read the output of the background process in a thread to run
     # the given block in parallel.
     out_reader = Thread.new do
       Thread.current.abort_on_exception = true
-      output = StringIO.new
 
       loop do
-        begin
-          ready = IO.select(
-            [stdout_and_err], # Watch for reading
-            [], # Not watching any files for writing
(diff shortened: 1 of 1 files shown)


micahlee (commented):
Thanks for digging into this @telday!

I'm mixed on this PR. I understand you've observed improved test reliability with these changes, so I won't be a stickler on these, but I'm concerned that there isn't a lot fundamentally different here and I don't understand why or which of these changes make the impact. Do you have any more context you can share as to how these changes remove the race conditions you mention?

Specifically, I notice changes for:

- Leaving the stdin open
- Combining the `#closed?` and `#eof?` checks
- Moving the `#closed?` and `#eof?` checks to the start of the loop
- Switching from `IO#select` to `#wait_readable`
 
I would be helpful if you could expound a little upon what each of these is contributing to resolve the problem.

Thanks!

micahlee (approved):
Thanks for adding the additional detail! Those definitely help frame the changes. LGTM

PR #2671, opened 2022-10-21
Add callback extensions for Policy Load and Group Membership API

micahlee:
### Desired Outcome

The outcome of this PR is to expose Conjur application events through callback extensions that may be used in Conjur enterprise to trigger further logic based on these events (e.g. when a policy is loaded, when role membership is modified).

Further, as these are experimental capabilities, they are added behind feature flags to avoid them impacting general
use of Conjur.

### Implemented Changes

This PR:

- Adds the ability to configure Feature flags for gating application changes unless specifically enabled.
- Adds the ability to load allow-listed extensions for Conjur
- Adds an extension point for receiving callbacks from Conjur policy load lifecycle events
- Adds an extension point for receiving role member callbacks for changes through the Roles REST API.

### Connected Issue/Story

CyberArk internal issue link: [ONYX-26940](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-26940)

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] Docs (e.g. `README`s) were updated in this PR

#### Behavior

#### Security

### DIFF
--- a/lib/conjur/extension/implementation.rb
+++ b/lib/conjur/extension/implementation.rb
@@ -1,0 +1,99 @@
+# frozen_string_literal: true
+
+require_relative './implementation_method'
+
+module Conjur
+  module Extension
+    # Extension::Implementation is a proxy wrapper around a Conjur extension
+    # class. It handles instantiating an instance of the extension class and
+    # proxying method calls to it through `#call`.
+    class Implementation
+
+      def self.from_class(
+        implementation_class,
+        logger: Rails.logger,
+        **kwargs
+      )
+        # Instantiate the extension class
+        implementation = ImplementationMethod.new(
+          # For the constructor, we're looking for the method on the class
+          # itself, rather than an instance of it.
+          implementation_object: implementation_class,
+          method: :new,
+          # These are dependencies the extension may use in its
+          # implementation. At a minimum, extension initializers must include
+          # a **kwargs initializer argument to consume any arguments it
+          # doesn't declare specifically.
+          logger: logger,
+          **kwargs
+        ).call
+
+        # Wrap the extension object with an Implementation proxy interface
+        Implementation.new(
+          implementation_object: implementation,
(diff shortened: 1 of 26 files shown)


lib/conjur/extension/implementation.rb
```diff
@@ -0,0 +1,99 @@
...
+    class Implementation
+
+      def self.from_extension_class(
+        extension_class,
+        logger: Rails.logger,
+        **kwargs
+      )
+        # Instantiate the extension class
+        implementation = ImplementationMethod.new(
+          # For the constructore, we're looking for the method on the class
```
imheresamir, line 19:
```suggestion
          # For the constructor, we're looking for the method on the class
```

lib/conjur/extension/repository.rb
```diff
@@ -0,0 +1,183 @@
...
+      @loaded_extensions = []
+      @registered_extension_classes = []
+
+      class << self
+        # This are intentionally writable to support testing
+        # :reek:Attribute
+        attr_accessor :loaded_extensions, :registered_extension_classes
+
+        # :reek:LongParameterList is due to dependency injection
+        def register_extension(
```
imheresamir, line 33: Since this is the Extension Repository, would it make sense to drop the `extension` prefixes and suffixes from methods and parameter names in this class? Not a huge deal

imheresamir (approved):
Let a couple of nitpicky comments but otherwise looks great to me, thanks for this!

lib/conjur/extension/repository.rb
micahlee: Good call! I left `load_extension` since `#load` is an `Object` built-in (and used in that method). Attempting to rename it causes all kinds of issues... 😆

imheresamir (approved):

PR #2674, opened 2022-11-04
Add token revocation and session management to AuthnOIDC

john-odonnell:
### Desired Outcome

If we plan on distributing long-lived OIDC refresh tokens, we need more token management features. We should at least include token revocation for dangling tokens, but should also look into session termination.

### Implemented Changes

- Refactors the `Authentication::AuthnOidc::V2::Strategy` class into `::Strategies::AuthzCode` and `::Strategies::RefreshToken`. This will hopefully make it easier to:
  1. Implement an end-session strategy
  7. Refactor our current OIDC ID token authentication strategy
- Update `openid_connect` gem and Keycloak docker-compose service used `dev` and `ci` environments. Both were missing token revocation features.
- Update the OIDC client class:
  1. Add `.revoke` method, for invalidating a given refresh token
  2. Add `.end_session` method, for revoking dangling tokens and returning a URI to the configured OIDC provider's session management API
  8. If applicable, ensure that rotated refresh tokens are revoked after use
- Adds an API endpoint, `/authn-oidc/:account/:service_id/logout`, which employ a new Logout handler and strategy.

### Open Questions

- Adds a `Authentication::Handler::LogoutHandler` class. This was an attempt to give the AuthenticateController a path to objects initialized in the `AuthenticationHandler` class without legitimately authenticating and issuing a token. Wondering if there is a better solution to be found here.

### Connected Issue/Story

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] A follow-up issue to update official docs has been filed here: [insert issue ID]()

#### Behavior

#### Security


app/controllers/authenticate_controller.rb
```diff
@@ -136,39 +140,68 @@ def authenticate_jwt
...
     params[:authenticator] = "authn-oidc"
     decode_and_merge_request_body
-    
-    if params[:refresh_token] || params[:code]
-      params.permit!
+    params.permit!
+
+    if params[:code] || params[:refresh_token]
+      namespace = 'Authentication::AuthnOidc::V2::Strategies'
+      strategy = 'AuthzCode' if params[:code]
```
jvanderhoof, line 169: I'd actually recommend pushing the logic for choosing a strategy up into the `Authentication::AuthnOidc::V2::Strategy` class.  This will allow you to keep details of a particular strategy out of the controller (which ultimately should be super generic).  Something like:

```ruby
module Authentication
  module AuthnOidc
    module V2
      class Strategy
        def callback(args)
          if args.key?(:refresh_token)
            # run refresh token workflow
          elsif(args.key?(:code)
            # run code validation workflow
          end
          ...
        end
      end
    end
  end
end
```

app/controllers/authenticate_controller.rb
```diff
@@ -136,39 +140,68 @@ def authenticate_jwt
...
+    ) do |authenticator|
+      Authentication::AuthnOidc::V2::Strategies::Logout.new(
+        authenticator: authenticator
+      ).callback(params.to_hash.symbolize_keys)
+    end
+
+    render(json: { oidc_logout_uri: logout_uri })
+  end
+
   def authenticate_oidc
```
jvanderhoof, line 162: This is the original "OIDC" functionality which is really just a glorified JWT validation specific to OIDC. If you're going to hijack the `POST` route, I'd recommend splitting this up a bit differently, something like:

```ruby
  def authenticate_oidc
    # Parse JSON request body
    # If refresh token is present, do refresh token workflow
    # Else, do this:
      params[:authenticator] = "authn-oidc"
      input = Authentication::AuthnOidc::UpdateInputWithUsernameFromIdToken.new.(
        authenticator_input: authenticator_input
      )
      # We don't audit success here as the authentication process is not done
    rescue => e
      # At this point authenticator_input.username is always empty (e.g. cucumber:user:USERNAME_MISSING)
      log_audit_failure(
        authn_params: authenticator_input,
        audit_event_class: Audit::Event::Authn::Authenticate,
        error: e
      )
      handle_authentication_error(e)
    else
      authenticate(input)
    end
```

app/controllers/authenticate_controller.rb
```diff
@@ -14,7 +14,11 @@ def oidc_authenticate_code_redirect
     ).call(
       parameters: params.to_hash.symbolize_keys,
       request_ip: request.ip
-    )
+    ) do |authenticator|
+      Authentication::AuthnOidc::V2::Strategies::AuthzCode.new(
+        authenticator: authenticator
+      ).call(params.to_hash.symbolize_keys)
+    end
```
jvanderhoof, line 21: I'd actually recommend pushing the logic for choosing a strategy up into the `Authentication::AuthnOidc::V2::Strategy` class.  This will allow you to keep details of a particular strategy out of the controller (which ultimately should be super generic).  Something like:

```ruby
module Authentication
  module AuthnOidc
    module V2
      class Strategy
        def callback(args)
          if args.key?(:refresh_token)
            # run refresh token workflow
          elsif(args.key?(:code)
            # run code validation workflow
          end
          ...
        end
      end
    end
  end
end
```

app/controllers/authenticate_controller.rb
```diff
@@ -136,39 +140,68 @@ def authenticate_jwt
...
   def authenticate_oidc
     params[:authenticator] = "authn-oidc"
     decode_and_merge_request_body
-    
-    if params[:refresh_token] || params[:code]
-      params.permit!
+    params.permit!
+
+    if params[:code] || params[:refresh_token]
+      namespace = 'Authentication::AuthnOidc::V2::Strategies'
```
jvanderhoof, line 168: These should be pushed up into the Strategy as described above to keep the OIDC abstraction from bubbling up.

app/domain/authentication/authn_oidc/v2/client.rb
```diff
@@ -48,11 +49,46 @@ def get_token_with_code(code:, nonce:, code_verifier:)
...
+
+          end_session_uri(
+            id_token: id_token,
+            state: state,
+            redirect_uri: redirect_uri
+          )
+        end
+
+        def end_session_uri(id_token:, state:, redirect_uri:)
+          query = URI.encode_www_form(
```
jvanderhoof, line 82: `URI` should be injected to simplify testing.

app/domain/authentication/authn_oidc/v2/client.rb
```diff
@@ -135,7 +171,7 @@ def discovery_information(invalidate: false)
             skip_nil: true
           ) do
             @discovery_configuration.discover!(@authenticator.provider_uri)
-          rescue HTTPClient::ConnectTimeoutError, Errno::ETIMEDOUT => e
+          rescue Faraday::TimeoutError, Errno::ETIMEDOUT => e
```
jvanderhoof, line 174: Did the `OpenID-Connect` gem move from HTTPClient to Faraday?

app/domain/authentication/authn_oidc/v2/strategy.rb
```diff
@@ -2,52 +2,89 @@
...
-          authenticator:,
-          client: Authentication::AuthnOidc::V2::Client,
-          logger: Rails.logger
-        )
-          @authenticator = authenticator
-          @client = client.new(authenticator: authenticator)
-          @logger = logger
-        end
+      module Strategies
+        module Common
```
jvanderhoof, line 6: Please NEVER use modules to inject methods into downstream objects, particularly the initializer.  Modules make it nearly impossible to reason about where a classes methods come from.  

Instead either use inheritance (probably the correct solution for this section), or create a generic, stateless object "Utility" object with common methods.  Utility objects should follow adhere to functional programming best practices (stateless, no mutation, etc.).

app/domain/authentication/authn_oidc/v2/client.rb
john-odonnell: Yes, only [recently](https://github.com/nov/rack-oauth2/pull/91) in v2.0.0+

PR #2678, opened 2022-11-23
Adds PKCE Support to Conjur

jvanderhoof:
### Desired Outcome

This PR attempts to accomplish two objectives:

1. Segregate the PKCE Support using the recently added Feature Flag functionality.
2. Provide a pattern for abstracting and testing workflows managed by a feature flag.

### Implemented Changes

**PKCE Support**
- Primary Changes:
    - Adds a generated code challenge and code verify to provider API
    - OIDC authentication endpoint now requires code verify, which is used in the provider token request
- Secondary Changes:
    - A unique nonce is generated on each Provider API request
    - State is no longer generated on Provider API request.  A client should manage its own state value without Conjur being aware.
    - Adds required and optional fields to the `Authenticator` data object. This allows the repository to load only relevant variables.
    - Moves the required parameters check up to the Strategy.  This allows a handler to be unconcerned with authentication details.

**Feature Flag Patterning**
- Primary Changes:
    - Adds the ClimateControl gem and a small helper method (`with_feature_enabled`) to run RSpec tests with a feature toggled on (flags are off by default). This prevents us using environment variables on the container or test level.
       
        _Note: we still need to develop an approach for testing Cucumber tests as they operate against an external container._

Reviewers should step through commits.

### Connected Issue/Story

N/A

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] A follow-up issue to update official docs has been filed here: [ONYX-27159](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-27159)

#### Behavior

#### Security

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Updated nokogiri in root and docs Gemfile.lock files to resolve GHSA-qv4q-mr5r-qprj
   [cyberark/conjur#2684](https://github.com/cyberark/conjur/pull/2684)
 
+### Fixed
+- Previously, if an OIDC authenticator was configured with a `Status` webservice,
+  the OIDC provider endpoint would include duplicate OIDC authenticators. This
+  change resolves ONYX-25530.
+  [cyberark/conjur#2678](https://github.com/cyberark/conjur/pull/2678)
+
 ## [1.19.0] - 2022-11-29
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   change resolves ONYX-25530.
   [cyberark/conjur#2678](https://github.com/cyberark/conjur/pull/2678)
 
+### Added
+- Provides support for PKCE in the OIDC Authenticator code redirect workflow.
+  This is disabled by default, but is available under the
+  `CONJUR_FEATURE_PKCE_SUPPORT_ENABLED` feature flag.
+  [cyberark/conjur#2678](https://github.com/cyberark/conjur/pull/2678)
+
 ## [1.19.0] - 2022-11-29
 
 ### Added
--- a/app/db/repository/authenticator_repository.rb
+++ b/app/db/repository/authenticator_repository.rb
@@ -1,10 +1,16 @@
 module DB
   module Repository
     class AuthenticatorRepository
-      def initialize(data_object:, resource_repository: ::Resource, logger: Rails.logger)
+      def initialize(
+        data_object:,
+        resource_repository: ::Resource,
+        logger: Rails.logger,
+        pkce_support_enabled: Rails.configuration.feature_flags.enabled?(:pkce_support)
+      )
         @resource_repository = resource_repository
         @data_object = data_object
         @logger = logger
+        @pkce_support_enabled = pkce_support_enabled
       end
 
       def find_all(type:, account:)
@@ -65,7 +71,14 @@ module DB
             args[variable.resource_id.split('/')[-1].underscore.to_sym] = variable.secret.value
           end
         end
+
         begin
+          if @pkce_support_enabled
+            allowed_args = %i[account service_id] +
+                          @data_object.const_get(:REQUIRED_VARIABLES) +
+                          @data_object.const_get(:OPTIONAL_VARIABLES)
+            args_list = args_list.select{ |key, _| allowed_args.include?(key) }
+          end
           @data_object.new(**args_list)
         rescue ArgumentError => e
           @logger.debug("DB::Repository::AuthenticatorRepository.load_authenticator - exception: #{e}")
--- a/app/db/repository/authenticator_repository.rb
+++ b/app/db/repository/authenticator_repository.rb
@@ -14,7 +14,14 @@ module DB
             "#{account}:webservice:conjur/#{type}/%"
           )
         ).all.map do |webservice|
-          load_authenticator(account: account, id: webservice.id.split(':').last, type: type)
+          service_id = service_id_from_resource_id(webservice.id)
+
+          # Querying for the authenticator webservice above includes the webservices
+          # for the authenticator status. The filter below removes webservices that
+          # don't match the authenticator policy.
+          next unless webservice.id.split(':').last == "conjur/#{type}/#{service_id}"
+
+          load_authenticator(account: account, service_id: service_id, type: type)
         end.compact
       end
 
@@ -27,7 +34,7 @@ module DB
         ).first
         return unless webservice
 
-        load_authenticator(account: account, id: webservice.id.split(':').last, type: type)
+        load_authenticator(account: account, service_id: service_id, type: type)
       end
 
       def exists?(type:, account:, service_id:)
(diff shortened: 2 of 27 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Nothing should go in this section, please add to the latest unreleased version
   (and update the corresponding date), or add a new version.
 
+## [1.19.1] - 2022-11-23
+### Added
+- Provides support for PKCE in the OIDC Authenticator code redirect workflow.
+  This is disabled by default, but is available under the
+  `CONJUR_FEATURE_POLICY_LOAD_EXTENSIONS` feature flag.
```
micahlee, line 16: I think this should be `CONJUR_FEATURE_PKCE_SUPPORT_ENABLED`, right?

jvanderhoof: doh! good catch.

spec/app/db/repository/authenticator_repository_spec.rb
```diff
@@ -3,207 +3,221 @@
...
-    DB::Repository::AuthenticatorRepository.new(
-      resource_repository: resource_repository,
-      data_object: Authentication::AuthnOidc::V2::DataObjects::Authenticator
-    )
-  end
-
-  let(:arguments) { %i[provider_uri client_id client_secret claim_mapping nonce state] }
+  %w[true false].each do |enabled|
+    context "with flag #{enabled}" do
+      with_feature_enabled(:pkce_support) do
```
micahlee, line 8: It doesn't look like `enabled` is actually used here to turn it on or off.

spec/app/db/repository/authenticator_repository_spec.rb
```diff
@@ -3,207 +3,221 @@
...
+        let(:data_object) do
+          if Rails.configuration.feature_flags.enabled?(:pkce_support)
+            Authentication::AuthnOidc::PkceSupportFeature::DataObjects::Authenticator
+          else
+            Authentication::AuthnOidc::V2::DataObjects::Authenticator
+          end
```
micahlee, line 14: Here's an example of where I'd rather see this driven by test inputs, the `enabled` option above, rather than indirectly through the feature flag environment variable.

app/db/repository/authenticator_repository.rb
```diff
@@ -54,8 +54,17 @@ def load_authenticator(type:, account:, id:)
             args[variable.resource_id.split('/')[-1].underscore.to_sym] = variable.secret.value
           end
         end
+
         begin
-          @data_object.new(**args_list)
+          if Rails.configuration.feature_flags.enabled?(:pkce_support)
```
micahlee, line 59: Again, with testing in mind. I'd rather see either `configuration`, `feature_flags`,  or even `pkce_support` passed in as a dependency to the class, and use that when unit testing this, rather than setting the environment.

micahlee (commented):
Hey @jvanderhoof,

I left some thoughts for you regarding the testing approach for feature flags. This isn't something I feel super strongly about, so you can tell me if you rather keep it as it is.

Thanks!

(Also, I'm assuming all of the logic itself is previously reviewed, and we're only reviewing the feature flag part, is that correct?)

app/db/repository/authenticator_repository.rb
jvanderhoof: Brillant. I didn't even think about that as an option 😢.  That may also allow us to remove the toggle in testing...

spec/spec_helper.rb
```diff
@@ -150,3 +150,11 @@ def conjur_server_dir
   # Navigate from its directory (/bin) to the root Conjur server directory
   Pathname.new(File.join(File.dirname(conjurctl_path), '..')).cleanpath
 end
+
+# Toggle a feature flag on for a particular set of tests.
+def with_feature_enabled(feature, &block)
```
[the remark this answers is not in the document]
jvanderhoof: Handling the flag toggle with dependency injection means this can be removed.  It's MUCH cleaner and prevents us from needing to mutate the environment variable during testing.  I'm removing this and the gem.

micahlee:
Hey @jvanderhoof, I think you meant to rebase this PR, rather than merge master in:
image

Gemfile
```diff
@@ -80,7 +80,6 @@ gem 'i18n', '~> 1.8.11'
 group :development, :test do
   gem 'aruba'
   gem 'ci_reporter_rspec'
-  gem 'climate_control'
```
micahlee: Can we squash these changes down so that the gemfile changes don't manifest in this PR at all (added in an earlier commit and deleted here in the same PR).

app/db/repository/authenticator_repository.rb
```diff
@@ -14,7 +20,10 @@ def find_all(type:, account:)
             "#{account}:webservice:conjur/#{type}/%"
           )
         ).all.map do |webservice|
-          load_authenticator(account: account, id: webservice.id.split(':').last, type: type)
+          service_id = service_id_from_id(webservice.id)
```
micahlee, line 23: Could this be `service_id_from_resource_id` instead?

app/db/repository/authenticator_repository.rb
```diff
@@ -14,7 +20,10 @@ def find_all(type:, account:)
...
+          next unless webservice.id.split(':').last == "conjur/#{type}/#{service_id}"
```
micahlee, line 24: It's not clear what business logic this is performing. Can you add a comment elaborating what the outcome of this filtering is?

app/domain/authentication/authn_oidc/pkce_support_feature/client.rb
```diff
@@ -0,0 +1,117 @@
...
+          end
+          id_token = bearer_token.id_token || bearer_token.access_token
+
+          begin
+            attempts ||= 0
+            decoded_id_token = @oidc_id_token.decode(
+              id_token,
+              discovery_information.jwks
+            )
+          rescue Exception => e
```
micahlee, line 71: Code climate also flags this. Do we need to rescue `Exception` here, or can it be just `StandardError`? If `Exception` is intentional, mind adding a comment why that is?

app/domain/authentication/authn_oidc/pkce_support_feature/strategy.rb
```diff
@@ -0,0 +1,43 @@
...
+          authenticator:,
+          client: Authentication::AuthnOidc::PkceSupportFeature::Client,
+          logger: Rails.logger
+        )
+          @authenticator = authenticator
+          @client = client.new(authenticator: authenticator)
+          @logger = logger
+        end
+
+        # Don't love this name...
```
micahlee, line 15: Mind either removing this comment, or elaborate on it such that someone other than you (or future you!) could continue the train of thought in an actionable way?

micahlee (commented):
Thanks @jvanderhoof, I left a few more comments on the PR as a whole. Feel free to take them or leave them! 😃

CHANGELOG.md
```diff
@@ -9,7 +9,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Nothing should go in this section, please add to the latest unreleased version
   (and update the corresponding date), or add a new version.
 
-## [1.19.0] - 2022-11-29
+## [1.19.0] - 2022-12-06
```
micahlee, line 12: It looks like this should be bumped to `1.19.1` (or greater).

szh:
@jvanderhoof Is there an ETA for when this will be mereged?

micahlee (approved):
LGTM! Thanks, @jvanderhoof

PR #2683, opened 2022-12-06
AuthnOIDC V2: Customizable Token TTL

john-odonnell:
### Desired Outcome

Allow users to extend the TTL of tokens distributed by a certain AuthnOIDC V2 instance.

### Implemented Changes

When creating an AuthnOIDC V2 instance in policy, add a `token-ttl` variable in the parent policy to configure a custom token TTL. The value assigned to this variable must be a string representing an ISO 8601 Duration. By default, AuthnOIDC still distributes 8 minute tokens. This can be increased up to a 5 hour global maximum.

```yaml
- !policy
  id conjur/authn-oidc/${service_id}
  body:
    - !webservice

    - !variable token-ttl
    ...
```

Includes `token-ttl` as an `OPTIONAL_VARIABLE` on the `DataObjects::Authenticator` class.

Includes `MINIMUM_AUTHENTICATION_TOKEN_EXPIRATION` constant on the `TokenFactory` class, in order to eliminate any negative duraitons.

Includes RSpec and Cucumber test cases.

### Connected Issue/Story

CyberArk internal issue ID: ONYX-29302 & ONYX-29303

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] A follow-up issue to update official docs has been filed here: [insert issue ID]

#### Behavior

#### Security

- [x] Security architect has reviewed the changes in this PR,

### DIFF
--- a/app/db/repository/authenticator_repository.rb
+++ b/app/db/repository/authenticator_repository.rb
@@ -77,7 +77,7 @@ module DB
             allowed_args = %i[account service_id] +
                           @data_object.const_get(:REQUIRED_VARIABLES) +
                           @data_object.const_get(:OPTIONAL_VARIABLES)
-            args_list = args_list.select{ |key, _| allowed_args.include?(key) }
+            args_list = args_list.select{ |key, value| allowed_args.include?(key) && value.present? }
           end
           @data_object.new(**args_list)
         rescue ArgumentError => e
--- a/app/domain/authentication/authn_oidc/pkce_support_feature/data_objects/authenticator.rb
+++ b/app/domain/authentication/authn_oidc/pkce_support_feature/data_objects/authenticator.rb
@@ -5,7 +5,7 @@ module Authentication
         class Authenticator
 
           REQUIRED_VARIABLES = %i[provider_uri client_id client_secret claim_mapping].freeze
-          OPTIONAL_VARIABLES = %i[redirect_uri response_type provider_scope name].freeze
+          OPTIONAL_VARIABLES = %i[redirect_uri response_type provider_scope name token_ttl].freeze
 
           attr_reader(
             :provider_uri,
@@ -28,7 +28,8 @@ module Authentication
             redirect_uri: nil,
             name: nil,
             response_type: 'code',
-            provider_scope: nil
+            provider_scope: nil,
+            token_ttl: 'PT8M'
           )
             @account = account
             @provider_uri = provider_uri
@@ -40,6 +41,7 @@ module Authentication
             @name = name
             @provider_scope = provider_scope
             @redirect_uri = redirect_uri
+            @token_ttl = token_ttl
           end
 
           def scope
@@ -52,6 +54,13 @@ module Authentication
 
           def resource_id
             "#{account}:webservice:conjur/authn-oidc/#{service_id}"
+          end
+
+          # Returns the validity duration, in seconds, of an instance's access tokens.
+          def token_ttl
+            ActiveSupport::Duration.parse(@token_ttl)
+          rescue ActiveSupport::Duration::ISO8601Parser::ParsingError
+            raise Errors::Authentication::DataObjects::InvalidTokenTTL.new(resource_id, @token_ttl)
           end
         end
       end
--- a/app/domain/authentication/authn_oidc/v2/data_objects/authenticator.rb
+++ b/app/domain/authentication/authn_oidc/v2/data_objects/authenticator.rb
@@ -20,7 +20,8 @@ module Authentication
             redirect_uri: nil,
             name: nil,
             response_type: 'code',
-            provider_scope: nil
+            provider_scope: nil,
+            token_ttl: 'PT8M'
           )
             @account = account
             @provider_uri = provider_uri
@@ -34,6 +35,7 @@ module Authentication
             @name = name
             @provider_scope = provider_scope
             @redirect_uri = redirect_uri
+            @token_ttl = token_ttl
           end
 
           def scope
@@ -46,6 +48,13 @@ module Authentication
 
           def resource_id
             "#{account}:webservice:conjur/authn-oidc/#{service_id}"
+          end
+
+          # Returns the validity duration, in seconds, of an instance's access tokens.
+          def token_ttl
+            ActiveSupport::Duration.parse(@token_ttl)
+          rescue ActiveSupport::Duration::ISO8601Parser::ParsingError
+            raise Errors::Authentication::DataObjects::InvalidTokenTTL.new(resource_id, @token_ttl)
           end
         end
       end
--- a/app/domain/authentication/handler/authentication_handler.rb
+++ b/app/domain/authentication/handler/authentication_handler.rb
@@ -78,7 +78,8 @@ module Authentication
 
         TokenFactory.new.signed_token(
           account: parameters[:account],
-          username: role.role_id.split(':').last
+          username: role.role_id.split(':').last,
+          user_ttl: authenticator.token_ttl
         )
       rescue => e
         log_audit_failure(parameters[:account], parameters[:service_id], request_ip, @authenticator_type, e)
--- a/app/domain/token_factory.rb
+++ b/app/domain/token_factory.rb
@@ -10,6 +10,7 @@ class TokenFactory < Dry::Struct
   attribute :slosilo, ::Types::Any.default{ Slosilo }
 
   MAXIMUM_AUTHENTICATION_TOKEN_EXPIRATION = 5.hours
+  MINIMUM_AUTHENTICATION_TOKEN_EXPIRATION = 0
 
   def signing_key(account)
     slosilo["authn:#{account}".to_sym] || raise(NoSigningKey, account)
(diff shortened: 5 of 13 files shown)


app/domain/authentication/authn_oidc/pkce_support_feature/data_objects/authenticator.rb
```diff
@@ -28,7 +48,8 @@ def initialize(
             redirect_uri: nil,
             name: nil,
             response_type: 'code',
-            provider_scope: nil
+            provider_scope: nil,
+            token_ttl: nil
```
jvanderhoof, line 52: Instead of passing `nil` here, what about setting it to the default token TTL (`PT8M`).  That'll allow you to avoid the nil check below.

app/domain/authentication/authn_oidc/pkce_support_feature/data_objects/authenticator.rb
```diff
@@ -18,6 +18,26 @@ class Authenticator
...
+          #   token/ttl: 'PT30M'
+          OPTIONAL_ANNOTATIONS = %w[token/ttl]
+
+          # These constants are used by the Authenticator class to enforce
+          # restrictions on the custom token TTL, regardless of source.
+          #
+          # Currently, the TokenFactory class enforces default (same as below)
+          # and maximum (5 hours) TTL values. With these constants defined here,
+          # different authn flavors can be subject to unique restrictions.
+          TOKEN_TTL_MIN = 5.minutes
```
jvanderhoof, line 37: Instead of forcing the Annotation class to have an awareness of a Conjur token (and that token's TTL), let's push all this into the `TokenFactory` class. Something like this should allow us us to allow us to refactor the `TokenFactory` class to something like:

```ruby
class TokenFactory < Dry::Struct

  # TODO: this error should be moved to the main `errors.rb` file
  NoSigningKey = ::Util::ErrorClass.new(
    "Signing key not found for account '{0}'")

  attribute :slosilo, ::Types::Any.default{ Slosilo }

  MAXIMUM_AUTHENTICATION_TOKEN_EXPIRATION = 5.hours
  MINIMUM_AUTHENTICATION_TOKEN_EXPIRATION = 2.minutes

  def signing_key(account)
    slosilo["authn:#{account}".to_sym] || raise(NoSigningKey, account)
  end

  def signed_token(
    account:,
    username:,
    host_ttl: Rails.application.config.conjur_config.host_authorization_token_ttl,
    user_ttl: Rails.application.config.conjur_config.user_authorization_token_ttl
  )
    signing_key(account).issue_jwt(
      sub: username,
      exp: Time.now + offset(
        ttl: username.starts_with?('host/') ? host_ttl : user_ttl
      )
    )
  end

  def offset(ttl:)
    offset = parse_ttl(ttl: ttl)
    return MAXIMUM_AUTHENTICATION_TOKEN_EXPIRATION if offset > MAXIMUM_AUTHENTICATION_TOKEN_EXPIRATION
    return MINIMUM_AUTHENTICATION_TOKEN_EXPIRATION if offset < MINIMUM_AUTHENTICATION_TOKEN_EXPIRATION

    offset
  end

  def parse_ttl(ttl:)
    # If TTL is an integer
    return ttl.to_i if ttl.to_i.to_s == ttl.to_s

    # If TTL is ISO 8601 Duration
    begin
      ActiveSupport::Duration.parse(ttl).to_i
    rescue ActiveSupport::Duration::ISO8601Parser::ParsingError
      # Attempt to coerce string into integer
      ttl.to_s.to_i
    end
  end
end
```

app/domain/authentication/authn_oidc/v2/data_objects/authenticator.rb
```diff
@@ -47,6 +69,26 @@ def name
...
             "#{account}:webservice:conjur/authn-oidc/#{service_id}"
           end
+
+          # Returns the validity duration, in seconds, of an instance's access tokens.
+          def token_ttl
+            return TOKEN_TTL_DEFAULT unless @token_ttl
+
+            begin
+              duration = ActiveSupport::Duration.parse(@token_ttl) if @token_ttl
+            rescue ActiveSupport::Duration::ISO8601Parser::ParsingError
```
jvanderhoof, line 79: I'm in the process of adding a "Status" check for the new OIDC. I think we should move the ISO8601 check into that verification.

app/db/repository/authenticator_repository.rb
```diff
@@ -70,13 +80,21 @@ def load_authenticator(type:, account:, service_id:)
...
 
         begin
           if @pkce_support_enabled
             allowed_args = %i[account service_id] +
                           @data_object.const_get(:REQUIRED_VARIABLES) +
-                          @data_object.const_get(:OPTIONAL_VARIABLES)
+                          @data_object.const_get(:OPTIONAL_VARIABLES) +
+                          @data_object.const_get(:OPTIONAL_ANNOTATIONS).map do |arg|
+                            arg.sub('/', '_').to_sym
+                          end
```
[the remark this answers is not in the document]
john-odonnell: Using annotations for all configuration points might be a pleasant UX. In this case, should we disallow storing the `client_secret` as an annotation, and force it to be a variable? May not want to tie permissions over the client secret to permissions over the webservice.

app/domain/authentication/authn_oidc/v2/data_objects/authenticator.rb
john-odonnell: Good idea, the Status check should definitely be able to verify the custom TTL config. Given that the `Status` and `TokenFactory` classes both need to decode ISO 8601 strings, does it make sense to keep that conversion in a method on the `Authenticator` class?

Related comment from above:
> Instead of forcing the Annotation class to have an awareness of a Conjur token (and that token's TTL), let's push all this into the TokenFactory class.

The `Authenticator` class needs some concept of a custom TTL, even if only in its ISO 8601 form. Maybe it's best to move the max/min range into the `TokenFactory`, but keep ISO 8601 parsing as part of the `Authenticator` class.

app/domain/authentication/authn_oidc/pkce_support_feature/data_objects/authenticator.rb
andytinkham: One question on avoiding the nil check - could someone still explicitly pass in nil? Does that need to be caught?

app/db/repository/authenticator_repository.rb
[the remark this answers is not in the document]
andytinkham: @john-odonnell & I talked about this in our implementation security review. We ended up down several rabbit holes of impacts on UX that indicate there's likely a deeper discussion to be had here that's more than just the 2 of us, including things like additional complexity in mixed config (knowing which takes precedence and determining which is active when trying to debug malfunctioning policy, for example) and not encouraging/allowing customers to put secrets in policy (like here, if the OIDC client-secret variable was instead a hardcoded value in an annotation.  

We also spent some time talking about how some config pieces come from different people (an Okta admin vs. a Conjur admin, for example) who might have different degrees of comfort/familiarity and how some of those values might actually be better off with a little more friction in changing, while others could use less friction. For example, maybe changing the provider URL for an OIDC provider should be a little harder than rotating a client secret for that provider. We may want to make it so that some changes can get multiple people involved in a change (like we make PRs need a review from someone else) to avoid a single rogue person causing problems.

There's a lot to discuss here and a broader group needed for participating in this discussion, so for now we went with making TTL a variable as well for consistency. We'd like to have that bigger discussion though and after that it might make sense ot revisit this piece too.

app/domain/authentication/authn_oidc/v2/data_objects/authenticator.rb
andytinkham: Given the edge cases we were talking about, @john-odonnell, with DST changes and leap seconds and all that, if we do have to add code to handle those, having one place do the processing seems like a really good idea too.

andytinkham:
John & I did an implementation review. I'm in agreement on this feaure's implementation.

app/domain/authentication/handler/authentication_handler.rb
```diff
@@ -78,7 +78,8 @@ def call(parameters:, request_ip:)
 
         TokenFactory.new.signed_token(
           account: parameters[:account],
-          username: role.role_id.split(':').last
+          username: role.role_id.split(':').last,
+          user_ttl: authenticator.token_ttl
```
gl-johnson, line 82: Is it safe to assume `token_ttl` will always be defined here? I understand this handler class currently only supports the OIDC/pkce authenticators which have been updated accordingly but wasn't sure if the intent was to generalize it for other cases

gl-johnson (commented):
Dropped a couple minor questions but otherwise LGTM. Happy to approve when no longer a draft

app/domain/authentication/handler/authentication_handler.rb
john-odonnell: The intent is definitely to extend this pattern to other authenticators, so you're right - currently this is a safe assumption, but that may not always hold true. I think this may be acceptable - as long as we have one flavor of `Authenticator` that has to report a customized `token_ttl`, I'm not sure if this is avoidable.

app/domain/token_factory.rb
```diff
@@ -28,9 +29,16 @@ def signed_token(account:,
...
+    offset = parse_ttl(ttl: ttl)
+    return MAXIMUM_AUTHENTICATION_TOKEN_EXPIRATION if offset > MAXIMUM_AUTHENTICATION_TOKEN_EXPIRATION
+    return MINIMUM_AUTHENTICATION_TOKEN_EXPIRATION if offset < MINIMUM_AUTHENTICATION_TOKEN_EXPIRATION
 
-    MAXIMUM_AUTHENTICATION_TOKEN_EXPIRATION
+    offset
   end
 
+  def parse_ttl(ttl:)
+    return ttl.to_i if ttl.to_i.to_s == ttl.to_s
```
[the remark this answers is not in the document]
john-odonnell: This is from Jason's review - I added some comments back in for clarity. Essentially it's an extra bit of input cleansing to force strings into integers if the `signed_token` method gets string input for a custom TTL.

```rb
def parse_ttl(ttl:)
  # If TTL is an integer, return it
  return ttl.to_i if ttl.to_i.to_s == ttl.to_s
  # Attempt to coerce a string into integer
  ttl.to_s.to_i
end
```

gl-johnson (approved):
Thanks for the clarifications. LGTM!

PR #2758, opened 2023-03-24
Update comment for CVE-2023-0286 in .trivyignore

andytinkham:
### Desired Outcome

Put a permanent skip on CVE-2023-0286 in the .trivyignore. We have the fix, but trivy doesn't detect it.


imheresamir (approved):
LGTM

PR #2760, opened 2023-03-27
AuthnOIDC V2 compatibility updates

john-odonnell:
### Desired Outcome

Generalize AuthnOIDC V2, so a Conjur user could realistically use any popular OIDC provider.

The bugs fixed here were found when trying to use [PingFederate](https://docs.pingidentity.com/r/en-us/pingfederate-112/pf_pingfederate_landing_page) with Conjur and AuthnOIDC V2.

### Implemented Changes

- Hardcoded 'scope=true' query param attached to token requests has been removed. According to OIDC Core, [authz requests require a scope parameter](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest), but token requests, defined by [RFC 6749](https://www.rfc-editor.org/rfc/rfc6749#section-4.1.3), do not. Including them may cause defensive OIDC providers to reject a token request that includes an unsupported scope.
- OIDC client is no longer configured with an 'end_session_endpoint'. Not only does Conjur not implement [RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html), but this endpoint is not required by OIDC Core or [Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) specifications. When targeting a minimal OIDC provider, we may be operating on a nil value.

### Connected Issue/Story

CNJR-1009 and CNJR-1010

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/app/domain/authentication/authn_oidc/v2/client.rb
+++ b/app/domain/authentication/authn_oidc/v2/client.rb
@@ -31,15 +31,14 @@ module Authentication
               authorization_endpoint: URI(discovery_information.authorization_endpoint).path,
               token_endpoint: URI(discovery_information.token_endpoint).path,
               userinfo_endpoint: URI(discovery_information.userinfo_endpoint).path,
-              jwks_uri: URI(discovery_information.jwks_uri).path,
-              end_session_endpoint: URI(discovery_information.end_session_endpoint).path
+              jwks_uri: URI(discovery_information.jwks_uri).path
             )
           end
         end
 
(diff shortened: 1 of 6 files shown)


jvanderhoof (dismissed):
Looks good @john-odonnell.  Thanks for the research to help make this more generic!

jvanderhoof (approved):

PR #2766, opened 2023-04-04
WIP: Add proxy support

micahlee:
### Desired Outcome

### Implemented Changes

particular, consider:*

### Connected Issue/Story

CyberArk internal issue ID: [insert issue ID]

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


app/domain/authentication/authn_k8s/web_socket_client.rb
```diff
@@ -124,17 +91,118 @@ def close
...
           send(nil, type: :close)
         end
+
         @closed = true
         @socket&.close
         @socket = nil
+
         emit(:__close)
+
         Thread.kill(@thread) if @thread
```
jvanderhoof, line 101: Is this to ensure we can safely spawn a new thread in `begin_event_loop` below? 

It feels a bit disjointed to have the lifecycle of a variable spread across three different classes, but without changing the interface, there's not much we can do.

jvanderhoof (approved):
Nice work on this @micahlee. I always like seeing what these protocols look like under the hood.

app/domain/authentication/authn_k8s/web_socket_client.rb
micahlee: I don't think it so much for spawning a new thread, but just to make sure we clean up the thread already created.

I started down a path of refactoring more of the event loop, but I became concerned about verifying that it hadn't regressed in the timeframe we have for code freeze, so I backed out of that. There is definitely room to decouple some of these lifecycle and thread management routines, though. It's really difficult to try to add unit tests for this class the way it's written right now. The `_spec.rb` file for the websocket client is really more of an integration test right now. It gets the job done,  but is slower than it needs to be and doesn't really test all of the lifecycle edge cases.

Considering we've fixed 3+ bugs in this code over the past year, this wouldn't be a bad candidate for focusing on in a quality sprint at some point, to make this code easier to test and make changes to. I tried to do some of that with the OpenSSL setup and the proxy support, but the thread/event loop is still just kind of spaghetti code. CC: @adamouamani

PR #2806, opened 2023-05-17
Sofia/slosilo key for hosts and users

sofiadimant1:
### Desired Outcome

### Implemented Changes

particular, consider:*

### Connected Issue/Story

CyberArk internal issue ID: [insert issue ID]

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


app/models/account.rb
```diff
@@ -40,34 +49,38 @@ def create(id, owner_id = nil)
...
         $1
       end.delete_if do |account|
         account == "!"
       end
     end
   end
 
-  def token_key
-    Slosilo["authn:#{id}"]
+  def token_key_host
```
jvanderhoof, line 65: What do you think about combining these two methods with the above slosilo id methods to make them a bit more generic, something like:

```ruby
def token_key(role)
  Slosilo[token_id(role)]
end

def token_id(role)
  "authn:#{account}:#{role}"
end
```

app/models/account.rb
```diff
@@ -17,15 +17,24 @@ def find_or_create_accounts_resource
...
+      "authn:#{account}:host"
+    end
+
+    def user_slosilo_id(account)
+      "authn:#{account}:user"
+    end
+
     def create(id, owner_id = nil)
-      raise Exceptions::RecordExists.new("account", id) if Slosilo["authn:#{id}"]
+      raise Exceptions::RecordExists.new("account", id) if Slosilo[host_slosilo_id(id)] or Slosilo[user_slosilo_id(id)]
```
jvanderhoof, line 29: Alternatively, you can use the `token_key...` methods below (which return the key if present:

```ruby
raise Exceptions::RecordExists.new("account", id) if token_key('user') || token_key('host')
```

jvanderhoof: You likely know this, but this line will raise if either host or user keys are missing. It the migration the only bit ensuring both keys are present?

sofiadimant1: what do you mean? there is a rspec test to test it

jeniaSakirko (approved):

PR #2818, opened 2023-05-26
Support running cucumber tests in parallel:

neil-k-zero:
- Spawn multiple processes to execute cucumber tests in parallel. Each process contains it's own indipendent conjur instance and postgresql database to avoid colisions.
 - Dynamically alter Ruby ENV variables based on the process executing the cucumber feature tests.
 - Dockerfile has been altered to support replicated services for parallel tests (this has not been altered to be dynamic and consists of static changes).
 - Alter cucumber.yml profiles to support parallel_tests ruby gem cmd usage

The following tests have not been parallelized:
 - authenticators_k8s
 - rspec tests

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -90,6 +90,7 @@ group :development, :test do
   gem 'faye-websocket'
   gem 'net-ssh'
   gem 'parallel'
+  gem 'parallel_tests'
   gem 'pry-byebug'
   gem 'pry-rails'
   gem 'rails-controller-testing'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -309,6 +309,8 @@ GEM
       validate_url
       webfinger (>= 1.0.1)
     parallel (1.21.0)
+    parallel_tests (4.2.0)
+      parallel
     parser (3.0.3.2)
       ast (~> 2.4.1)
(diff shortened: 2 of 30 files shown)


neil-k-zero:
FYI, I had a sleep 15 second sleep in authenticators_k8s that really seemed to help with the flaky failures, might be worth looking into. https://github.com/cyberark/conjur/blob/e2999ba55ac62e22df6da3218cb5fc2376782983/ci/test_suites/authenticators_k8s/test_gke_entrypoint.sh#L348

neil-k-zero:
- Master branch builds are roughly 16 minutes faster.
- Non-Master branch builds are capable of running all tests, not just smoke tests, as fast as non-parallel builds take to run only smoke tests
- Parallel Non-Master branch smoke only builds won't see an increase in test performance due to k8s not being setup for parallel tests (it's possible to convert k8s tests to parallel, however, it hasn't been completed at this time)

micahlee (approved):
Thanks @neil-k-zero! 

I don't _love_ how we have resources repeated by copy/paste (e.g. in the docker compose file). However, I don't immediately know what to propose as an alternative and the benefit of running these in parallel is clear. LGTM!

PR #2827, opened 2023-06-13
Allow STS regional requests without 'Host' header

gl-johnson:
### Desired Outcome

Currently we assume that the signed request will feature a 'Host' header telling Conjur which STS endpoint to use.

This assumption is not valid for all HTTP clients (unless explicitly added to the request), including the Golang AWS SDK. Instead we should the following prioritized order in determining which endpoint to use:
1. Host header if it exists
2. AWS region (pulled from Authorization header)
3. Global endpoint (only if result from #2 is `us-east-1` and the regional endpoint failed)

### Implemented Changes

- Implemented above logic for determining STS endpoint + tests
- Minor fixes to dev environment

### Connected Issue/Story

CNJR-1904

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] A follow-up issue to update official docs has been filed here: CNJR-1907

#### Behavior

#### Security


app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -54,7 +54,27 @@ def extract_relevant_data(response)
 
       # Call to AWS STS endpoint using the provided authentication header
       def attempt_signed_request(signed_headers)
-        aws_request = URI("https://#{signed_headers['host']}/?Action=GetCallerIdentity&Version=2011-06-15")
+        sts_host = signed_headers['host'] || extract_sts_host(signed_headers)
```
jvanderhoof, line 57: I'd recommend we move the signed header resolution into the `extract_sts_host` method. This allows us to keep all logic related to 

```ruby
sts_host = signed_headers['host'] || extract_sts_host(signed_headers)
```
Then we push the logic for checking header value or attempting to extract the region from the credential:
```ruby
# Extract AWS region from the authorization header's credential string, i.e.:
# Credential=AKIAIOSFODNN7EXAMPLE/20220830/us-east-1/sts/aws4_request
def extract_sts_host(signed_headers)
  return signed_headers['host'] if signed_headers['host'].present?

  region = signed_headers['authorization'].match(%r{Credential=[^/]+/[^/]+/([^/]+)/})&.captures&.first
  raise(Errors::Authentication::AuthnIam::InvalidAWSHeaders, 'Failed to extract AWS region from authorization header') unless region

  "sts.#{region}.amazonaws.com"
end
```

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -54,7 +54,27 @@ def extract_relevant_data(response)
 
       # Call to AWS STS endpoint using the provided authentication header
       def attempt_signed_request(signed_headers)
```
jvanderhoof, line 56: Feels like with the multiple forms of region lookup, paired with fallback validation, we should move to region extraction and STS request validation.  What do you think about having `extract_sts_host` return the region?  Based on that response, we can build the STS url and validate it while falling back to the global endpoint if the regional validation fails.

jvanderhoof: I'd recommend something like:

```ruby
def aws_call(region:, headers:)
  host = if region == 'global'
    'sts.amazonaws.com'
  else
    "sts.#{region}.amazonaws.com"
  end
  aws_request = URI("https://#{host}/?Action=GetCallerIdentity&Version=2011-06-15")
  begin
    @client.get_response(aws_request, headers)
  rescue => e
    # Handle any network failures with a generic verification error
    raise(Errors::Authentication::AuthnIam::VerificationError, e)
  end
end

# Call to AWS STS endpoint using the provided authentication header
def attempt_signed_request(signed_headers)
  region = extract_sts_region(signed_headers)

  # Attempt check using the discovered region 
  response = aws_call(region: region, headers: signed_headers)
  return response if response.code.to_i == 200

  # If the discovered region is `us-east-1`, also check if the request
  # was made from the 
  if region == 'us-east-1'
    response = aws_call(region: 'global', headers: signed_headers)
    return response if response.code.to_i == 200
  end

  raise(Errors::Authentication::AuthnIam::VerificationError, 'Signature is invalid')
end

def extract_sts_region(signed_headers)        
  if signed_headers['host'].present? 
    if signed_headers['host'] == 'sts.amazonaws.com'
      'global'
    elsif match = signed_headers['host'].match(%r{sts.([\w\-]+).amazonaws.com})
      match.captures.first
    end
  else
    match = signed_headers['authorization'].match(%r{Credential=[^/]+/[^/]+/([^/]+)/})
    return match.captures.first if match

    raise(Errors::Authentication::AuthnIam::InvalidAWSHeaders, 'Failed to extract AWS region from authorization header')
  end
end
```

This separates the following into separate methods:
- Region extraction 
- Retry behavior on the global endpoint
- STS lookup

_Note_: the above is pseudo code, so please refactor as makes sense.

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -76,6 +97,25 @@ def response_from_signed_request(aws_headers)
...
+      # Credential=AKIAIOSFODNN7EXAMPLE/20220830/us-east-1/sts/aws4_request
+      def extract_sts_region(signed_headers)
+        host = signed_headers['host']
+
+        if host == 'sts.amazonaws.com'
+          return 'global'
+        end
+      
+        match = host&.match(%r{sts.([\w\-]+).amazonaws.com})
+        return match&.captures&.first if match
```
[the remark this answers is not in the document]
john-odonnell: Looking at this again, and even given my suggestion:
```ruby
return match.captures&.first if match
```

I wonder if we still need this Safe Navigation Operator.
- [`.match`](https://ruby-doc.org/core-3.1.1/String.html#method-i-match) returns `nil` if there are no matches found
- [`.match`](https://ruby-doc.org/core-3.1.1/String.html#method-i-match) returns a [`MatchData` object](https://ruby-doc.org/core-3.1.1/MatchData.html) for 1+ matches found

That means:
- `if match` catches cases with no matches
- Otherwise, there is guaranteed to be at least one match - `match.captures` will not be `nil`.

We might be able to do this instead, with no change in behavior and less `&`s:
```ruby
return match.captures.first if match
```

app/domain/authentication/authn_iam/authenticator.rb
```diff
@@ -54,13 +54,34 @@ def extract_relevant_data(response)
...
+      
+        # If the discovered region is `us-east-1`, fallback to the global endpoint
+        if region == 'us-east-1'
+          @logger.debug(LogMessages::Authentication::AuthnIam::RetryWithGlobalEndpoint.new)
+          fallback_response = aws_call(region: 'global', headers: signed_headers)
+          return fallback_response if fallback_response.code.to_i == 200
+        end
 
-          # Handle any network failures with a generic verification error
+        return response
```
john-odonnell, line 70: Wonder if in the case where `region == 'us-east-1'` and `fallback_response.code.to_i != 200` we want to return the `fallback_response` instead of the primary `response`.

gl-johnson: I wasn't too sure about this. There isn't a way to definitively know whether the regional endpoint or the global endpoint was used to generate the signed headers even after attempting both requests. AWS throws a fairly generic `The request signature we calculated does not match the signature you provided` error message either way

john-odonnell (approved):
LGTM!

PR #2842, opened 2023-06-30
Update CONJ00013E error message to apply to AuthnOIDC V1 and V2

john-odonnell:
### Desired Outcome

From CNJR-2106:
> CONJ00013E Claim 'login' not found or empty in ID token. This claim is defined in the id-token-user-property variable
>
> The docs have you define a non-existent claim in the claim-mapping original request to OKTA  (originating from the claim_mapping variable value), Okta returns a token without that claim mapping which is where the incorrect error message gets printed in Conjur.
>
> Steps to reproduce:
> 1. Docs have you define claim in "claim-mapping" variable - ie 'THIS-CLAIM-DOESNT-EXIST'
> 2. Okta returns a token without that claim mapping
> 3. An incorrect error message gets printed in Conjur
>
> Current Results:
> An incorrect error message gets printed in Conjur referencing id-token-user-property
>
> Expected Results:
> Error should reference variable : "claim-mapping" for OIDC v2

### Implemented Changes

- Update CONJ00013E to take a second argument, `{1-config-var}`, which changes the variable name displayed in the error message.
- Updates to fix dev environment

### Connected Issue/Story

CNJR-2106

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/ci/shared.sh
+++ b/ci/shared.sh
@@ -112,6 +112,16 @@ _run_cucumber_tests() {
     fi
   done
 
+  # Generate authn_local socket ENV variables based on the
+  # number of parallel processes.
+  for (( i=1; i <= ${#parallel_services[@]}; ++i )); do
+    if (( i == 1 )) ; then
+      sockets+=("AUTHN_LOCAL_SOCKET=/run/authn-local/.socket")
+    else
+      sockets+=("AUTHN_LOCAL_SOCKET${i}=/run/authn-local${i}/.socket")
+    fi
+  done
+
   # Add the cucumber env vars that we always want to send.
   # Note: These are args for docker compose run, and as such the right hand
   # sides of the = do NOT require escaped quotes.  docker compose takes the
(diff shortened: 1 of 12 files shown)


ci/shared.sh
```diff
@@ -112,6 +112,17 @@ _run_cucumber_tests() {
     fi
   done
 
+  # Generate authn_local socket ENV variables based on the
+  # number of parallel processes.
+  for (( i=1; i <= ${#parallel_services[@]}; ++i )); do
+    index=$(( i - 1 ))
```
gl-johnson, line 118: `index` unused?

Also if I'm not mistaken you could simplify the beginning of the for loop to:
`for i in "${!parallel_services[@]}"; do`

gl-johnson (commented):
Minor comment but otherwise LGTM. Thanks for fixing the dev environment!

ci/shared.sh
john-odonnell: > index unused?

Yep - good catch.

> Also if I'm not mistaken you could simplify the beginning of the for loop to:
`for i in "${!parallel_services[@]}"; do`

You're right, but we're using a counting-style 1-based indexing for the parallel services. The updated loop would start from 0, so I think I'll leave it as-is.

gl-johnson (approved):

PR #2854, opened 2023-07-10
ONYX-28233: Add telemetry

szh:
### Desired Outcome

Merge telemetry feature branch

The Telemetry feature is complete but it was done [on a feature branch](https://github.com/cyberark/conjur/tree/feature-conjur-telemetry). We will need to merge this back to master before we can truly consider it complete.

As part of the review, we should primarily do some sanity checking to verify that Telemetry is not on by default and is essentially invisible until enabled.

Once the cyberark/conjur changes are merged we will can also [merge the quick-start guide](https://github.com/cyberark/conjur-quickstart/pull/52).

### Implemented Changes

particular, consider:*

### Connected Issue/Story

CyberArk internal issue ID: ONYX-28233

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] Docs (e.g. `README`s) were updated in this PR

#### Behavior

#### Security

### DIFF
--- a/lib/monitoring/metrics/api_exception_counter.rb
+++ b/lib/monitoring/metrics/api_exception_counter.rb
@@ -16,7 +16,7 @@ module Monitoring
       end
 
       def update(payload)
-        metric = @registry.get(metric_name)
+        metric = registry.get(metric_name)
         update_labels = {
           operation: payload[:operation],
           exception: payload[:exception],
--- a/lib/monitoring/metrics/api_exception_counter.rb
+++ b/lib/monitoring/metrics/api_exception_counter.rb
@@ -16,7 +16,7 @@ module Monitoring
       end
 
       def update(payload)
-        metric = @registry.get(@metric_name)
+        metric = @registry.get(metric_name)
         update_labels = {
           operation: payload[:operation],
           exception: payload[:exception],
--- a/lib/monitoring/metrics/api_exception_counter.rb
+++ b/lib/monitoring/metrics/api_exception_counter.rb
@@ -1,0 +1,29 @@
+module Monitoring
+  module Metrics
+    class ApiExceptionCounter
+      attr_reader :registry, :pubsub, :metric_name, :docstring, :labels, :sub_event_name
+
+      def setup(registry, pubsub)
+        @registry = registry
+        @pubsub = pubsub
+        @metric_name = :conjur_http_server_request_exceptions_total
+        @docstring = 'The total number of API exceptions raised by Conjur.'
+        @labels = %i[operation exception message]
+        @sub_event_name = 'conjur.request_exception'
+
+        # Create/register the metric
+        Metrics.create_metric(self, :counter)
+      end
+
+      def update(payload)
+        metric = @registry.get(@metric_name)
+        update_labels = {
+          operation: payload[:operation],
+          exception: payload[:exception],
+          message: payload[:message]
+        }
+        metric.increment(labels: update_labels)
+      end
+    end
+  end
+end
--- a/lib/monitoring/metrics/api_exception_counter.rb
+++ b/lib/monitoring/metrics/api_exception_counter.rb
@@ -1,7 +1,7 @@
 module Monitoring
   module Metrics
     class ApiExceptionCounter
-      attr_reader :registry, :pubsub, :metric_name, :docstring, :labels, :sub_event_name, :throttle
+      attr_reader :registry, :pubsub, :metric_name, :docstring, :labels, :sub_event_name
 
       def setup(registry, pubsub)
         @registry = registry
--- a/lib/monitoring/metrics/api_exception_counter.rb
+++ b/lib/monitoring/metrics/api_exception_counter.rb
@@ -1,7 +1,7 @@
 module Monitoring
   module Metrics
     class ApiExceptionCounter
-      attr_reader :registry, :pubsub, :metric_name, :docstring, :labels, :sub_event_name
+      attr_reader :registry, :pubsub, :metric_name, :docstring, :labels, :sub_event_name, :throttle
 
       def setup(registry, pubsub)
         @registry = registry
--- a/lib/monitoring/metrics/authenticator_gauge.rb
+++ b/lib/monitoring/metrics/authenticator_gauge.rb
@@ -3,13 +3,18 @@ module Monitoring
     class AuthenticatorGauge
       attr_reader :registry, :pubsub, :metric_name, :docstring, :labels, :sub_event_name
 
+      def initialize(installed_authenticators: Authentication::InstalledAuthenticators)
+        @metric_name = :conjur_server_authenticator
+        @docstring = 'Number of authenticators enabled'
+        @labels = %i[type status]
+        @sub_event_name = 'conjur.policy_loaded'
+
+        @installed_authenticators = installed_authenticators
+      end
+
       def setup(registry, pubsub)
         @registry = registry
         @pubsub = pubsub
-        @metric_name = :conjur_server_authenticator
-        @docstring = 'Number of authenticators enabled'
-        @labels = [:type, :status]
-        @sub_event_name = 'conjur.policy_loaded'
         
         # Create/register the metric
         Metrics.create_metric(self, :gauge)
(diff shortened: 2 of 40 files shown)


lib/monitoring/metrics/api_exception_counter.rb
```diff
@@ -0,0 +1,29 @@
...
+        update_labels = {
+          operation: payload[:operation],
+          exception: payload[:exception],
+          message: payload[:message]
+        }
+        metric.increment(labels: update_labels)
+      end
+    end
+  end
+end
```
jvanderhoof, line 29: What about breaking these up into using inheritance?  If we structure these metrics like this:

```ruby
module Monitoring
  module Metrics
    class BaseMetric
      attr_reader :registry, :pubsub, :metric_name, :docstring, :labels, :sub_event_name

      def setup(registry, pubsub)
        @registry = registry
        @pubsub = pubsub

        # Create/register the metric
        Metrics.create_metric(self, :counter)
      end

      def update(payload)
        metric = @registry.get(@metric_name)
        update_labels = {
          operation: payload[:operation],
          exception: payload[:exception],
          message: payload[:message]
        }
        metric.increment(labels: update_labels)
      end
    end

    class ApiExceptionCounter < BaseMetric
      def initialize
        @metric_name = :conjur_http_server_request_exceptions_total
        @docstring = 'The total number of API exceptions raised by Conjur.'
        @labels = %i[operation exception message]
        @sub_event_name = 'conjur.request_exception'
      end
    end
  end
end
```

It'll allow us to reduce the metric classes to their bare minimum definition without breaking the interface (`setup` and `update` methods).

jvanderhoof: _Note_: Just noticed that the `update_labels` variable is unique.  As the rest of the call appears to be constant across implementations, I've updated my suggestion to include that in the child class.

lib/monitoring/metrics/authenticator_gauge.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      def update_configured_authenticators(metric)
+        configured_authenticators =  Authentication::InstalledAuthenticators.configured_authenticators
+        configured_authenticator_counts = get_authenticator_counts(configured_authenticators)
+        configured_authenticator_counts.each do |type, count|
+          metric.set(count, labels: { type: type, status: 'configured'})
+        end
+      end
+
+      def get_authenticator_counts(authenticators)
```
jvanderhoof, line 51: _Nit_: Alternatively, we can achieve this with:

```ruby
def get_authenticator_counts(authenticators)
  authenticators.each_with_object(Hash.new(0)) do |authenticator, rtn|
    type = authenticator.split('/')[0]
    rtn[type] += 1
  end
```

lib/monitoring/middleware/prometheus_collector.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      def record(env, code, duration, operation)
+        @pubsub.publish(
+          "conjur.request", 
+          code: code,
+          operation: operation,
+          duration: duration
+        )
+      rescue
+        # TODO: log unexpected exception during request recording
```
jvanderhoof, line 50: Please add a log message here (at the `DEBUG` level if you're concerned about log verbosity). Swallowing exceptions is super dangerous as it provides no indication that a failure has occurred.

lib/monitoring/middleware/prometheus_collector.rb
```diff
@@ -0,0 +1,64 @@
...
+        nil
+      end
+
+      def find_operation(method, path)
+        Monitoring::Metrics::OPERATIONS.each do |op|
+          if op[:method] == method && op[:pattern].match?(path)
+            return op[:operation]
+          end
+        end
+        return "unknown"
```
jvanderhoof, line 60: _nit:_ the last `return` is not required.  Ruby returns the last action of a method by default.

jvanderhoof:
@szh, awesome work on this feature.  My only real ask is to add some unit tests.  It looks like coverage is largely handled via controller integration tests. Your code looks well written and it looks like it'll be simple to add unit tests.

lib/monitoring/metrics/api_exception_counter.rb
```diff
@@ -0,0 +1,29 @@
...
+        @docstring = 'The total number of API exceptions raised by Conjur.'
+        @labels = %i[operation exception message]
+        @sub_event_name = 'conjur.request_exception'
+
+        # Create/register the metric
+        Metrics.create_metric(self, :counter)
+      end
+
+      def update(payload)
+        metric = @registry.get(@metric_name)
```
jvanderhoof, line 19: To fix the `assumes too much for instance variable '@metric_name'` CodeClimate warning, I believe you can refer to the method:

```ruby
 metric = @registry.get(metric_name)
```

lib/monitoring/metrics/authenticator_gauge.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      def update(*payload)
+        metric = @registry.get(@metric_name)
+        update_enabled_authenticators(metric)
+        update_installed_authenticators(metric)
+        update_configured_authenticators(metric)
+      end
+
+      def update_enabled_authenticators(metric)
+        enabled_authenticators = Authentication::InstalledAuthenticators.enabled_authenticators
```
jvanderhoof, line 29: I'd recommend using dependency injection to make `Authentication::InstalledAuthenticators` available inside the class.  This will greatly simplify unit testing.

Something like:

```ruby 
module Monitoring
  module Metrics
    class AuthenticatorGauge < Monitoring::Metrics::BaseMetric      
      def initialize(installed_authenticators: Authentication::InstalledAuthenticators)
        @metric_name = :conjur_server_authenticator
        @docstring = 'Number of authenticators enabled'
        @labels = [:type, :status]
        @sub_event_name = 'conjur.policy_loaded'

        @installed_authenticators = installed_authenticators
      end

      ...

    end
  end
end
```

lib/monitoring/metrics/api_exception_counter.rb
szh: @jvanderhoof this gets a bit more complicated with the metrics that aren't simple counters. I'm not sure we'll really gain much by refactoring this once we consider that there are only 2 counter metrics and the others will still need custom logic in the `update` method.

jvanderhoof: I agree @szh.  Let's keep these more verbose until we have a full set of potential metrics established.  If it makes sense to refactor to reduce code duplication, we can do that when we have a more complete picture of the range of metrics.

szh:
> @szh, awesome work on this feature. My only real ask is to add some unit tests. It looks like coverage is largely handled via controller integration tests. Your code looks well written and it looks like it'll be simple to add unit tests.

@jvanderhoof - @gl-johnson added some unit tests and this is now ready for your re-review.

szh:
Thanks to @doodlesbykumbi for restoring all that git commit history! 🥇

spec/lib/monitoring/middleware/prometheus_exporter_spec.rb
```diff
@@ -16,17 +16,21 @@
...
   context 'when requesting app endpoints' do
     it 'returns the app response' do
-      get '/foo'
+      env['PATH_INFO'] = "/foo"
+      status, _headers, _response = subject.call(env)
 
-      expect(last_response).to be_ok
-      expect(last_response.body).to eql('OK')
+      expect(status).to eql(200)
+      expect(_response.first).to eql('OK')
```
micahlee: Generally, a leading underscore indicates that the return value isn't used. Since we are using `_response` here, can we remove the underscore? (The same applies to other tests in this file as well)

micahlee (approved):
Hey @szh , this is going to be squash merged, right?

If so, this looks good to me! Thanks for the push to get it over the finish line!

szh:
@micahlee We were going to keep the git history actually. CC @doodlesbykumbi

micahlee:
Hey @szh (@doodlesbykumbi),

If we're keeping the commit history, I suggest we remove the following commits, at a minimum, as they represent a no-op once combined:

- https://github.com/cyberark/conjur/pull/2854/commits/89807795ba705d55d2fb410fec7b3145f1c3529e
- https://github.com/cyberark/conjur/pull/2854/commits/98ecbc3bb552ea91806fdf891f6897d9378092d0

Generally speaking, I would also expect linting fixes to be squashed back into the original commit, especially in cases where the PR introduces the files for the first time, as these commits on their own don't meaningfully contribute to the story of this PR:
- https://github.com/cyberark/conjur/pull/2854/commits/fc864fc36c6c8ba1779da5a5d21bc3efd1e9fd44
- https://github.com/cyberark/conjur/pull/2854/commits/c0e42f6cb603facf24b41333a631546614eaa102
- https://github.com/cyberark/conjur/pull/2854/commits/482316b08b74c068c52b37ec7f78b9e1d3bd3086

With those history modifications, I think this makes sense to include with the commit history. I think we must remove the two no-op commits to avoid those existing in master history. The others are a strong preference, but if it's too onerous I'm not strictly opposed to them.

jtuttle:
@micahlee I just made the suggested commit updates if you could re-review when you have a chance. Thanks!

micahlee (approved):
Awesome! Thanks for getting the commit history cleaned up a bit.

LGTM, @szh and @jtuttle !

PR #2865, opened 2023-07-21
Remove httpclient private certs

andytinkham:
### Desired Outcome

Remove test and sample private keys in the httpclient gem.

### Implemented Changes

Fixed the Docker call that was supposed to already be removing the *.pem files and added the capacity to delete *.key files too.

### DIFF
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -73,12 +73,12 @@ RUN INSTALL_PKGS="gcc \
     bundle --without test development && \
     # Remove the build packages
     yum remove -y $INSTALL_PKGS && \
-    yum -y clean all --enablerepo='*'
+    yum -y clean all --enablerepo='*' && \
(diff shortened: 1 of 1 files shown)


Dockerfile.ubi
```diff
@@ -78,7 +78,8 @@ RUN INSTALL_PKGS="gcc \
 COPY . .
 
 # removing CA bundle of httpclient gem
-RUN find / -name httpclient -type d -exec find {} -name *.pem -type f -delete \;
+RUN find / -name 'httpclient-*' -type d -exec find {} -name '*.pem' -type f -delete \; && \
+    find / -name 'httpclient-*' -type d -exec find {} -name '*.key' -type f -delete \;
```
micahlee, line 82: Can this be added to the `RUN` block above for bundle install, to consolidate our container build layers?

PR #2873, opened 2023-07-27
Add authn-oidc configurability solution design

gl-johnson:
Quick solution design addressing some requested improvements in configurability for the OIDC authenticator:
- Support for configurable `ca-cert` variable in authn-oidc config
- Support for HTTPS_PROXY environment variable
- Improved debug logging

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security


john-odonnell:
Given that this is the AuthnOIDC V2 process:

```
Client (UI/CLI)             Conjur                    OIDC Provider
|-------- providers ------->|                         |
|                           |- openid-configuration ->|
|                           |<------- response -------|
|<------- response ---------|                         |
|                           |                         |
|-------- authz ------------------------------------->|
|<------- authz token --------------------------------|
|                           |                         |
|------- authz token ------>|                         |
|                           |------ authz token ----->|
|                           |<----- id token ---------|
|<----- access token -------|                         |
|                           |                         |
```

and that we've experienced these cert issues:

1. during provider discovery, and
2. either:
   - when the provider is using custom certificates, or
   - when a proxy sits between Conjur and the provider

are there cases where the UI and CLI will need a similar solution to properly send the authz token request? If so, and if we're storing the custom cert in Conjur, how can we expose it the UI and CLI?

gl-johnson:
Good question. My initial impression would be that we do not need to add anything to the UI/CLI. In those cases the request originates in the browser rather than from the UI/CLI server and we can assume that a user of the OIDC provider will have the necessary certs installed on their machine. Will do some validation to be sure though.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,185 @@
...
+### Open Questions
+- Do we need validation that `ca-cert` is a valid cert before attempting to use it?
+- Should debug logging of OIDC be enabled by default when Conjur is running in debug?
+- Should OpenSSL debug logs be included?
+
+## Tasks
+### 1. Update the OpenIDConnect gem (1 pt)
+Upgrade our OpenIDConnect gem to >= 2.0.0 (when it switched from `httpclient` to `faraday`) and ensure nothing breaks.
+
+### 2. Create a temporary truststore wrapper method (2 pts)
```
jvanderhoof, line 154: As we're using environment variables, will there be any issues in regards to thread safety? If two different OIDC authenticators are configured and each uses different CAs, if authentication requests come in simultaneously, will the second `SSL_CERT_FILE` set be used in the first request?

gl-johnson: I had concerns about this as well, but based on discussion with @ismarc I was under the impression the change to the environment should be localized to the thread which is handling that particular authn-oidc request. However I'm not the most well-versed in how Rails handles concurrency so I'll look into it a bit further.

Worst case I imagine would be adding a mutex to make sure only one request is accessing the environment variable until cleanup has occurred (if acceptable from a performance standpoint)

jvanderhoof:
@gl-johnson, after some digging through the OpenID Connect gem, I think there's a simple option to enable CAs. OpenID Connect includes the ability to pass configuration to the underlying Faraday library: https://github.com/nov/openid_connect/blob/2fdafc3802aca1967790b079cb4e58ce5c4e9c93/spec/openid_connect_spec.rb#L45

_(note: docs in are terrible, but the code is well tested. I've found the tests more valuable for "how-to" guidance.)_

It looks like this exposes the [Faraday SSL config options](https://lostisland.github.io/faraday/#/customization/ssl-options).

First, a method to create temp files that exist for the block duration: 

```ruby
# Create a temp file which exists for the duration of the block
def temp_ca_certificate(certificate_content, &block)
  ca_certificate = Tempfile.new('ca_certificates')
  begin
    ca_certificate.write(certificate_content)
    ca_certificate.close
    block.call(ca_certificate)
  ensure
    ca_certificate.unlink   # deletes the temp file
  end
end
```

Then add a certificate to an X509 cert store (so we can handle CA chains):

```ruby
store = OpenSSL::X509::Store.new

if @authenticator.ca_cert.present?
  temp_ca_certificate(@authenticator.ca_cert) do |file|
    store.add_file(file.path)
  end
else
  # Auto-include system CAs unless a CA has been defined
  store.set_default_paths
end

OpenIDConnect.http_config do |config|
  config.ssl.cert_store = store
end
```

We should be able to add this where we configure an [OpenID Connect Client](https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_oidc/v2/client.rb#L21).

jvanderhoof:
As Faraday has a `proxy` configuration, it's also possible the Faraday configuration approach could be used to support the Faraday proxy configuration: https://lostisland.github.io/faraday/#/customization/proxy-options

gl-johnson:
Unfortunately the http_config can only be configured once. After it has been called it will not overwrite the existing config. See this open issue: https://github.com/nov/openid_connect/issues/84

The same goes for the SWD and Webfinger gems as well, so we'd end up having to modify all 3 gems to make it reconfigurable.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -75,21 +81,30 @@ It would be more consistent with other authenticator configs if authn-oidc were
...
+OpenSSL truststore in the container to include any custom CA certs. We can leverage this idea in Conjur 
+to do the same thing on the fly. A simple wrapper method which creates a temporary symlink in the OpenSSL 
+truststore to a tempfile containing the certificate content should be sufficient. Cleanup involves 
+removing the symlink, and ensuring that the tempfile has been cleaned up after code execution. Since we 
+are (temporarily) adding the cert to the existing truststore, it should not interfere with concurrent 
+HTTP connections which may rely on the default OpenSSL certs still being available.
 
 An example of what this wrapper method may look like:
 ```ruby
 def with_temporary_cert_store(cert_string)
```
jvanderhoof, line 94: We can sort it out later, but this approach won't work for CA certs that are chains (the certificate creation will fail).  For those, we need to write the cert to a temp file and import it (example I previously posted is part of the authn-jwt rewrite, which includes support for CA chains).

gl-johnson: There's already a [parse_certs method](https://github.com/cyberark/conjur/blob/322861bbc8897cd6dc2be0d50b81b6a678024473/app/domain/conjur/cert_utils.rb#L11) in the CertUtils module which I imagine can be repurposed to do exactly what you're describing. But yes we would need slightly modified logic to create separate temp files/symlinks for each cert that gets parsed from the chain

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
...
+breaks.
+
+### 2. Create a temporary truststore wrapper method (2 pts)
+Create a method which accepts a string (`ca_cert`) and temporarily updates the default OpenSSL truststore, and 
+yields to another method. Add tests to ensure expected state during code execution and cleanup has occured 
+afterwards.
+
+I would suggest adding it to the [CertUtils module](https://github.com/cyberark/conjur/blob/master/app/domain/conjur/cert_utils.rb).
+
+### 3. Add support for `ca-cert` variable in authn-oidc config (2 pts)
```
niteshtaneja, line 193: Is this the only visible user facing change needed to support ca-certs?

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
...
+#### Unit tests (authenticator data object)
+1. Default `ca-cert` value is nil
+1. Set value matches expected
+
+#### Cucumber tests
+One or more of our Keycloak OIDC scenarios should be updated to use a `ca-cert` value provided via Conjur 
+variable. It may turn out to be easier to update all keycloak examples to use the cert variable, and the case 
+where the variable is not set can be covered by the Okta examples.
+
+### Open Questions
```
niteshtaneja, line 176: The ideal response to all these questions would be a "Yes" I am assuming

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
...
+- Do we need validation that `ca-cert` is a valid cert before attempting to use it?
```
niteshtaneja, line 177: Do we have a sense of how much additional work it will be to add the validation? If its not a valid cert, whats the worst thats going to happen?

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
...
+- Should debug logging of OIDC be enabled by default when Conjur is running in debug?
```
niteshtaneja, line 178: Is that what we are doing?

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
niteshtaneja: For customers upgrading from v13.0, do they just need to add this variable in the auth policy for custom certs to work?

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
gl-johnson: In all likelihood this will be a non-issue now that I've thought about it more. We will use the OpenSSL Ruby library to parse the cert(s) from the `ca-cert` value, and it will produce an error we can easily handle if it is malformed in any way. 

The worst that will happen is the TLS connection would fail and they would have to fix the variable value.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
gl-johnson: I mentioned this to @jtuttle and it sounds like it is preferable to **not** introduce a new configuration, so yes it would be based on whether `CONJUR_LOG_LEVEL` is set to `debug`.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
gl-johnson: Yes, the `ca-cert` variable will be optional and is the only UX change.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
jvanderhoof: Really just making sure our eventual tests include examples with the configured CA certificate being a single certificate as well as a certificate chain.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
jvanderhoof: Agreed! Our docs provide guidance to toggle Conjur into and out of debug mode while troubleshooting an authenticator.

jvanderhoof:
I like the approaches summarized in this document. 

_I'm adding this comment here as it's related to the CA configuration, but this should NOT be part of this effort._

Longer term, we should seriously consider replacing the `OpenIDConnect` gem with an internal implementation. We make only two calls:

- Discovery: a simple URL get request for which we need to parse the response into JSON
- Exchange code for token:
    ```
    POST /token HTTP/1.1
    Host: server.example.com
    Content-Type: application/x-www-form-urlencoded
    Authorization: Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW

    grant_type=authorization_code&code=SplxlOBeZQQYbYS6WxSbIA
    &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb 
    ```

Once the Authn-JWT refactor is completed, we can use Authn-JWT strategy to validate the resulting OIDC JWT token.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
...
+## Implementation
+An existing workaround for the CA cert issue mentioned previously involves a user manually updating the 
+OpenSSL truststore in the container to include any custom CA certs. We can leverage this idea in Conjur 
+to do the same thing on the fly. A simple wrapper method which creates a temporary symlink in the OpenSSL 
+truststore to a tempfile containing the certificate content should be sufficient. Cleanup involves 
+removing the symlink, and ensuring that the tempfile has been cleaned up after code execution. Since we 
+are (temporarily) adding the cert to the existing truststore, it should not interfere with concurrent 
+HTTP connections which may rely on the default OpenSSL certs still being available.
+
+An example of what this wrapper method may look like:
```
andytinkham, line 92: How do we make sure that this wrapper method only injects the certificates we intend and isn't available for use by an attacker?

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
...
+1. WebFinger - used for dynamic discovery when the credential issuer is unknown to the client. **Not used in Conjur.**
+
+### Considerations around OIDC Discovery
+The SWD gem discussed previously is used by OpenIDConnect across many authenticators in Conjur (authn-oidc, 
+authn-jwt, authn-azure, authn-gcp). It performs a couple common tasks in Conjur by invoking the OIDC discovery 
+endpoint:
+  - validate authenticator status and provider connectivity
+  - fetch provider keys for decoding tokens
+
+This should not interfere with the proposed solution, but it is important to note since the modified code will be
```
andytinkham, line 151: You go from discussing the SWD gem which was only just referenced above to talking about modified code. Are you intending to modify the gem's code? More is needed about how this modified code could affect the other authns.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
andytinkham: We also need to explore what might happen with malicious certs, not just malformed.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
andytinkham: Are these log messages we control? Either way, we need to be certain we don't inadvertently cause disclosures in debug mode.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
andytinkham: Addressing this will be critical for security sign-off on the design and implementation.

andytinkham (changes requested):
I see no evidence of security being considered in this design and no security considerations section exists.  There are several pieces of the individual comments that need to be addressed before this design can be approved (see my other comments). Because this is a change to authenticator behavior and touching on certificate handling, we need to do a full security review and threat modelling session as well.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
adamouamani: @gl-johnson @jvanderhoof 

love the collaboration in this PR 👍  - I was wondering if the contents of the `ca-certs` variable (ie: cert or cert-chain) represent the "trust-store" only for the oidc provider or proxy and would this impact other calls or session refresh?  What if you had multiple providers (okta & identity)?  Additionally, if you have a customer cert and a proxy would proxy CA cert include the cert for the oidc provider?  Also, Could we leverage the system/os trust store/cert store instead of using an variable?  Also, are we trimming spaces and CRLF?

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
adamouamani: @gl-johnson There is a current issue in Jira that OIDC does not give back stack trace for error logging / troubleshooting
https://ca-il-jira.il.cyber-ark.com:8443/browse/CNJR-2105

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
+# OIDC Authenticator Configuration Improvements
+
+## Issue Description
+There have been multiple instances where users of the OIDC authenticator have experienced limitations resulting 
+from our use of the third-party [OpenIDConnect gem](https://github.com/nov/openid_connect). The issues mainly stem 
+from the gem limiting changes to the HTTP config used for outgoing connections. So far the following problems have 
+been identified:
+- Unable to configure CA certs from the authenticator config in Conjur
+  - The workaround for custom CA certs is to add them to the Conjur container OpenSSL truststore (`/etc/ssl/certs`)
+- The underlying HTTP client does not honor the `HTTPS_PROXY` environment variable
```
adamouamani, line 10: Are we supporting both `http_proxy` and `https_proxy`?  

Also, I believe these variables are typically lower case, can we check for uppercase or lowercase?

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
...
+- Limited ability to debug OIDC-related HTTP errors
```
adamouamani, line 11: Current issue with authn_oidc debugging errors and stack traces - https://ca-il-jira.il.cyber-ark.com:8443/browse/CNJR-2105

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Solution
+It would be more consistent with other authenticator configs if authn-oidc were to support a `ca-cert` variable 
+in the authenticator policy. This value would be used to inform the HTTP client which CA cert(s) to use to verify
```
adamouamani, line 15: Is this a single CA cert or Chain?  How would you handle multiple certs - would these need to be concatenated together?

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,228 @@
...
+  @discovered_provider = @open_id_discovery_service.discover!(@provider_uri)
+end
+```
+
+As for the `HTTPS_PROXY` issue, it appears to be fixed in a more recent version of OpenIDConnect gem which 
+migrated from the `httpclient` library to `faraday` for handling outbound connections. Therefore we should 
+simply be able to bump this gem version (assuming no new issues arise) in order to address this issue.
+
+The last major concern was around better debug logging. This can also be enabled from the Conjur side, 
+via a code snippet like below:
```
adamouamani, line 124: Are these environment variables and is this something we want to add to docs or kb?
Would OpenIDConnect.logger help resolve this issue - https://ca-il-jira.il.cyber-ark.com:8443/browse/CNJR-2105

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
gl-johnson: It should support both, which we can do via a pre-existing [parse_certs](https://github.com/cyberark/conjur/blob/48e95904a5ee8cda1503db9f5744ff6eefcecbdb/app/domain/conjur/cert_utils.rb#L11C10-L11C10) method. In the policy they would be concatenated together, then Conjur would separate them into individual certs for the purpose of adding them to the truststore.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
gl-johnson: That's a function of the underlying HTTP library. After making the gem update it will be Faraday with a Net::HTTP adapter. In my testing it honored both environment variables.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
gl-johnson: No gem code would be modified. This is in reference to [this generic OAuth code](https://github.com/cyberark/conjur/tree/master/app/domain/authentication/authn_oidc/v2/client.rb#L106C20-L106C20) which is used by the status webservice of multiple authenticators to verify connectivity to the configured provider. It would need to include the custom cert in the case of authn-oidc without impacting the other authenticators that also use it.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
gl-johnson: This is the implementation that would likely go in an initializer in Conjur.

For a user this would be enabled by setting `CONJUR_LOG_LEVEL` to `debug` per the docs https://docs.cyberark.com/AAM-DAP/13.0/en/Content/Operations/Services/logs.htm#Enabledebuglogmessages

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
gl-johnson: It will only going to inject the cert(s) provided via the authenticator policy, so unless an attacker has the ability to change variable values in the authenticator policy I would think it's a non-issue.

I can't reply to your other comments for some reason but I was curious about what a malicious cert would look like and if that is something we can realistically detect and handle.

Also just pushed an updated doc with a few security considerations I could think of.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
gl-johnson: These would be log messages outside our control. 

OpenIdConnect and SWD logs will contain debug messaging mainly about the HTTP requests to the OIDC provider. The OpenSSL logs may also be useful for lower level issues. When working on a customer case where they had TLS issues in authn-oidc, the OpenSSL logs helped us discover that the CA file they added manually to Conjur did not have the correct permissions.

At most I would expect these logs to include the filepath where OpenSSL expects the CA cert to be, and possibly a public key.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,250 @@
...
+
+**Log Verbosity and Information Disclosure**: Debugging TLS issues with an OIDC provider is overly difficult with the current debug logging. We would like to expose additional logging from the OpenIDConnect gem (which makes requests to the provider) and OpenSSL (which performs TLS verification based on the configured truststore). This requires that we do not disclose any sensitive information with the additional debug logs.
+
+**Mitigation**: Verify that logs for each of the components do not disclose sensitive data, including certificate keys, metadata, or filepaths.
+
+### Testing
+The testing will depend on the final implementation, but at the very least the wrapper method will need to be 
+well-covered by unit tests.
+
+#### Unit tests (temporary truststore modifications)
```
adamouamani, line 176: How will this be impacted by pending FIPS/OpenSSL3 changes?

adamouamani:
@gl-johnson What if customer is running Conjur with Rootless Podman?  They will probably not have permissions to update /etc/ssl/certs

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
```diff
@@ -0,0 +1,250 @@
...
+1. Malformed certificates produce the expected error message
+1. Temp file/dir/symlink exists while yielding
+1. Temp file/dir/symlink contains expected certificate content while yielding
+1. Temp file/dir/symlink is removed after yielding
+
+#### Unit tests (authenticator data object)
+1. Default `ca-cert` value is nil
+1. Set value matches expected value
+
+#### Cucumber tests
```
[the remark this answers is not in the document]
gl-johnson: To do this we would need an Identity or Okta instance using a private CA which seems like a large ask from infra. Our Keycloak server uses already uses a self-signed cert which is functionally the same as a private CA anyway.

I don't know that we necessarily need an automated test for the environment variables in this approach, since we'd be testing whether a standard Ruby library (Net::HTTP) handles them correctly as opposed to anything within Conjur. Similarly for the configurable `ca-cert` change, OpenSSL doesn't care whether a trusted cert belongs to a CA or a proxy server so we'd be double-testing in a sense. 

If we do need tests against a proxy server we would likely have to setup an nginx server with a self-signed cert as a forward proxy for Conjur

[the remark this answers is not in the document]
adamouamani: @gl-johnson do we currently have any e2e or integration tests for following:
> self-signed cert (keycloak or okra/identity)
> certificate chain (keycloak or okra/identity)
> https_proxy & HTTPS_PROXY
> concurrent writing of requests to cert store (are we making any backup of this file beforehand?  Is there any risk of file corruption and are we leveraging built in linux tools - update-ca-certs/trust)

adamouamani:
> @gl-johnson What if customer is running Conjur with Rootless Podman?  Will they have permissions to update /etc/ssl/certs

Actually for rootless podman we run as root in container so this is fine.  In k8s-follower we do not run as root - will this be an issue?

andytinkham (approved):
My concerns are all addressed.

design/authenticators/authn_oidc/oidc_configuration_enhancements.md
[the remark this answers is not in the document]
gl-johnson: Keycloak is already configured with a self-signed cert which will be easy to test against. For a cert chain we'd need to generate a CA/intermediate chain and reload Keycloak with an updated [server config](https://github.com/cyberark/conjur/blob/1c84fa10e7b3b6a8d754d1a3cce7875f5d5819ad/ci/oauth/keycloak/standalone.xml#L41) - although I think it is sufficient cover cert chains in unit tests. We can easily test that a cert chain is parsed correctly and that the cert store contains all certs in the chain.

Support for HTTPS_PROXY/https_proxy is not a standard, nor is it a function of Conjur - rather the underlying library that makes the outgoing connection. The previous library chose to use the same proxy (`http_proxy`) for both HTTP and HTTPS requests which I imagine is the case 99% of the time. I personally think it is overkill to have e2e tests requiring an additional service when we can easily confirm via some manual smoke testing.

[Ruby's Tempfile library](https://docs.ruby-lang.org/en/2.4.0/Tempfile.html) ensures uniqueness so concurrent read/write operations should not be possible as far as the cert itself is concerned. We will need to add some logic to ensure we handle collisions of the symlinks properly should they happen, but if anything they would link back to the same cert (possibly stored in multiple locations) and still allow the connection to succeed.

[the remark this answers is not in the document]
adamouamani: Thanks @gl-johnson for response 

I'm fine with #1 (keycloak) so long as we do individual cert and chain which I believe we are already doing/planning!! 👍   

For #2 https_proxy seems like it would be good candidate for a happy path e2e test - I understand this is a 3rd party library but would be good to have test that verifies this all works and if gem changes we will be notified of any breakage in pipeline vs manual testing from release to release.  let me know what you think..

For # 3 seems like this is handled already - do we have any tests to confirm there is no issue?

For K8s Follower - need to understand if this is valid scenario and how a rootless user would update certstore if connecting to K8s-follower for authn-oidc v2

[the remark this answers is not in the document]
gl-johnson: If it was something under our control I would agree, but ultimately the proxy determination in this case is handled by a standard library in Ruby. [This code is extremely unlikely to change.](https://ruby-doc.org/stdlib-3.0.2/libdoc/uri/rdoc/URI/Generic.html#method-i-find_proxy), plus I imagine we already make this assumption with every other network call that Conjur makes to an external provider. It works like so:
1. Construct env var name using scheme of the request, should result in either `http_proxy` or `https_proxy`
2. If HTTP, favor `http_proxy` if it exists, otherwise use `HTTP_PROXY`
3. If HTTPS, favor `https_proxy` if it exists, otherwise use `HTTPS_PROXY`

The older version of OpenIDConnect used [the not very well-maintained httpclient](https://github.com/nahi/httpclient/blob/4658227a46f7caa633ef8036f073bbd1f0a955a2/lib/httpclient.rb#L1071), which loaded the proxy settings itself and does not consider the scheme of the request when looking at environment variables.

As far as I can tell the K8s follower does not handle authn-oidc requests, but if you have information otherwise then I can look into it further.

szh:
Has this been fully implemented? Can we merge or close this PR?

PR #2895, opened 2023-08-09
Add flag to conjurctl server to skip migrations

telday:
### Desired Outcome
Adds new flag to the `conjurctl server` command which allows for skipping the migrations step even when the Sequel gem would normally think we need to. This change is required by a k8s follower using selective replication. When a k8s follower boots with selective replication enabled it performs a schema dump instead of a full database backup which means the `schema_migrations` and `schema_info` tables containing what migrations the database has had applied is missing and therefore the Sequel gem attempts to apply all of them. This then causes errors when the migration attempts to create schema elements already present in the database.

### Implemented Changes

If the new `--no-migrate` flag is used the migration step is skipped when booting Conjur.

### Connected Issue/Story

CyberArk internal issue ID: [CNJR-1599]

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/bin/conjur-cli/commands/server.rb
+++ b/bin/conjur-cli/commands/server.rb
@@ -1,16 +1,19 @@
 # frozen_string_literal: true
 
 require 'command_class'
+require 'sequel'
 
 # Required to use $CHILD_STATUS
 require 'English'
 
 require_relative 'db/migrate'
+require_relative 'connect_database'
 
 module Commands
   Server ||= CommandClass.new(
     dependencies: {
-      migrate_database: DB::Migrate.new
+      migrate_database: DB::Migrate.new,
+      connect_database: ConnectDatabase.new
     },
 
     inputs: %i[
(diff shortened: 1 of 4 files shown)


imheresamir:
@telday I think we may need some tests added to `spec/conjurctl/server_spec.rb`

bin/conjur-cli/commands/server.rb
```diff
@@ -19,14 +22,17 @@ module Commands
...
   ) do
     def call
       # Ensure the database is available
       # and the schema is up-to-date
       @migrate_database.call(
         preview: false
-      )
+      ) unless @no_migrate
+
+      @connect_database.call if @no_migrate
```
micahlee, line 35: Do we not want to keep this, regardless of the migration, to ensure the database is able to receive connections?

imheresamir (dismissed):
LGTM!

imheresamir (approved):

PR #2902, opened 2023-08-10
Support HTTPS_PROXY env variable in authn-oidc

gl-johnson:
### Desired Outcome

Fix support for `HTTPS_PROXY` variable when using authn-oidc. This issue was occuring due to the underlying HTTP client of the OpenIDConnect gem, and appears to be fixed in a more recent version when it switched to Faraday as its HTTP client.

### Implemented Changes

Updated OpenIDConnect gem to 2.x

### Connected Issue/Story

CNJR-2390

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/app/domain/authentication/authn_oidc/v2/client.rb
+++ b/app/domain/authentication/authn_oidc/v2/client.rb
@@ -104,7 +104,7 @@ module Authentication
             skip_nil: true
           ) do
             @discovery_configuration.discover!(@authenticator.provider_uri)
-          rescue HTTPClient::ConnectTimeoutError, Errno::ETIMEDOUT => e
+          rescue Errno::ETIMEDOUT => e
             raise Errors::Authentication::OAuth::ProviderDiscoveryTimeout.new(@authenticator.provider_uri, e.message)
           rescue => e
             raise Errors::Authentication::OAuth::ProviderDiscoveryFailed.new(@authenticator.provider_uri, e.message)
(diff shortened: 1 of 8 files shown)


app/domain/authentication/authn_oidc/v2/client.rb
```diff
@@ -104,7 +104,7 @@ def discovery_information(invalidate: false)
             skip_nil: true
           ) do
             @discovery_configuration.discover!(@authenticator.provider_uri)
-          rescue HTTPClient::ConnectTimeoutError, Errno::ETIMEDOUT => e
+          rescue Errno::ETIMEDOUT => e
```
john-odonnell, line 107: I was trying to figure out whether we should replace these `HTTPClient` errors with `Faraday` ones, but I'm not sure that it's necessary. The error returned by `discover!` can only be type [`OpenIDConnect::Discovery::DiscoveryFailed`](https://github.com/nov/openid_connect/blob/master/lib/openid_connect/discovery.rb#L4C19-L4C19), which is too non-specific, I think.

gl-johnson: Yep it's the same exception type for basically any network error that occurs. I suppose it would be doable to parse it out of the error message but that seemed a bit hacky and unnecessary

john-odonnell (approved):

PR #2919, opened 2023-08-24
AuthnOIDC client: wrap Discovery and Callback in optional cert logic

john-odonnell:
### Desired Outcome

From CNJR-2474:
> We need two wrapper methods to inject certs into the OpenSSL truststore before reaching out to the provider. One for `discover!` and one for `access_token!`. Include tests to ensure cleanup occurs and that hash collisions are handled correctly

### Implemented Changes

- Added class method `self.discover` to `Authentication::AuthnOIDC::V2::Client` class

  `self.discover` wraps `::OpenIDConnect::Discovery::Provider::Config.discover!` with commands to write & clean up a certificate to & from the default Conjur container certificate store.

  `self.discover` is a class method, because there are a few contexts outside this class where the underlying `discover!` method is used. Call it by running:

  ```rb
  Authentication::AuthnOIDC::Client.discover(...).
  ```

- Added instance method `callback_with_temporary_cert` to `Authentication::AuthnOIDC::V2::Client` class

  `callback_with_temporary_cert` wraps the `callback` method with commands to write & clean up a certificate to & from Conjur's default certificate store.

  Unlike `self.discover`, which wraps a single `::OpenIDConnect` method, `callback_with_temporary_cert` wraps the entire `callback` method, which includes multiple calls to the OIDC provider, all of which require the temporary certs to be successful:

  - `callback` uses the `discovery_information` method to get the provider metadata. This will be covered by using the new `self.discover`.
  - `callback` calls `::OpenIDConnect::Client.access_token!`
  - `callback` calls the `jwks` method on the `discovery_information` response object, which sends a request to the provider's JWKS endpoint.

- Added unit tests for each
  - Ensure requests that do not rely on a provided certificate pass as expected
  - Ensure temp cert is written when underlying method is called
  - Ensure temp cert is cleaned up when each method returns
  - Ensure providing a malformed cert raises and error
  - Ensure that if a symlink for the provided cert subject already exists, it is not overwritten or cleaned up when the function returns

### Open Questions for Reviewer

1. Right now, I have the AuthnOIDC client:

   - Verify the cert string isn't empty
   - Verify the target symlink doesn't already exist

   What else can we do to make this foolproof?

   - Verify cert CN/SAN against `provider_uri`?
   - Verify cert is not expired?

### Connected Issue/Story

CyberArk internal issue ID: CNJR-2474

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] A follow-up issue to update official docs has been filed here: [insert issue ID]

#### Behavior

#### Security

### DIFF
--- a/app/domain/authentication/authn_oidc/v2/client.rb
+++ b/app/domain/authentication/authn_oidc/v2/client.rb
@@ -110,6 +110,42 @@ module Authentication
             raise Errors::Authentication::OAuth::ProviderDiscoveryFailed.new(@authenticator.provider_uri, e.message)
           end
         end
+
+        # discover wraps ::OpenIDConnect::Discovery::Provider::Config.discover!
+        # with commands to write & clean up a certificate to & from the default
+        # Conjur container certificate store.
+        #
+        # The temporary certificate file name is "x.0", where x is the hash of
+        # the certificate subject name. If this file already exists in the
+        # default cert store, the original certificate is used.
+        #
+        # discover is a class method, because there are a few contexts outside
+        # this class where the underlying discover! method is used. Call it by
+        # running Authentication::AuthnOIDC::Client.discover(...).
+        def self.discover(
+          provider_uri:,
+          discovery_configuration: ::OpenIDConnect::Discovery::Provider::Config,
+          cert_dir: OpenSSL::X509::DEFAULT_CERT_DIR,
+          cert_string: nil
+        )
+          d = -> { discovery_configuration.discover!(provider_uri) }
+
+          return d.call if cert_string.blank?
+
+          cert = OpenSSL::X509::Certificate.new(cert_string)
+          symlink = File.join(cert_dir, "#{cert.subject.hash.to_s(16)}.0")
+          return d.call if File.exist?(symlink)
+
+          Dir.mktmpdir do |tmp_dir|
+            tmp_file = File.join(tmp_dir, 'ca.pem')
+            File.write(tmp_file, cert_string)
+            File.symlink(tmp_file, symlink)
+
+            d.call
+          ensure
+            File.unlink(symlink) if symlink.present? && File.symlink?(symlink)
+          end
+        end
       end
     end
   end
--- a/app/domain/authentication/authn_oidc/v2/client.rb
+++ b/app/domain/authentication/authn_oidc/v2/client.rb
@@ -97,14 +97,13 @@ module Authentication
           decoded_id_token
         end
 
-
         # callback_with_temporary_cert wraps the callback method with commands
-        # to write & clean up a certificate to & from Conjur's default
-        # certificate store.
+        # to write & clean up a given certificate or cert chain in a given
+        # directory. By default, Conjur's default cert store is used.
         #
-        # The temporary certificate file name is "x.0", where x is the hash of
-        # the certificate subject name. If this file already exists in the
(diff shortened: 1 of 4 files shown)


app/domain/authentication/authn_oidc/v2/client.rb
```diff
@@ -97,6 +97,46 @@ def callback(code:, nonce:, code_verifier: nil)
...
+          nonce:,
+          code_verifier: nil,
+          cert_dir: OpenSSL::X509::DEFAULT_CERT_DIR,
+          cert_string: nil
+        )
+          c = -> { callback(code: code, nonce: nonce, code_verifier: code_verifier) }
+
+          return c.call if cert_string.blank?
+
+          cert = OpenSSL::X509::Certificate.new(cert_string)
```
gl-johnson, line 125: Any idea if this constructor will work with a cert chain?

If not, ::Conjur::CertUtils has a `parse_certs` method which should return an array of Certificate objects

app/domain/authentication/authn_oidc/v2/client.rb
```diff
@@ -97,6 +97,46 @@ def callback(code:, nonce:, code_verifier: nil)
...
+          symlink = File.join(cert_dir, "#{cert.subject.hash.to_s(16)}.0")
+          return c.call if File.exist?(symlink)
```
gl-johnson, line 127: Initially I had planned to handle hash collisions by incrementing the file extension, i.e. `"#{hash}.0"` -> `"#{hash}.1"` -> `"#{hash}.2"`. Is it safe to assume that if the hashed name already exists it represents the same cert? 

I'm okay with the assumption if we can provide reasonable justification, but we may at least want to log a `debug` or `warn` message when it occurs

app/domain/authentication/authn_oidc/v2/client.rb
```diff
@@ -110,6 +150,42 @@ def discovery_information(invalidate: false)
...
+        # with commands to write & clean up a certificate to & from the default
+        # Conjur container certificate store.
+        #
+        # The temporary certificate file name is "x.0", where x is the hash of
+        # the certificate subject name. If this file already exists in the
+        # default cert store, the original certificate is used.
+        #
+        # discover is a class method, because there are a few contexts outside
+        # this class where the underlying discover! method is used. Call it by
+        # running Authentication::AuthnOIDC::Client.discover(...).
```
gl-johnson, line 164:
```suggestion
        # running Authentication::AuthnOIDC::V2::Client.discover(...).
```

gl-johnson (commented):
Overall looking good. The main thing would be adding cert chain support and a few tests around that if possible. I'm fairly sure they need to be parsed and written to the cert store separately

app/domain/authentication/authn_oidc/v2/client.rb
john-odonnell: This constructor does not work with a chain - updated to use `CertUtils.parse_certs`

app/domain/authentication/authn_oidc/v2/client.rb
john-odonnell: Updated to increment file extension in case of hash collisions

gl-johnson (commented):
LGTM. Nice work on the tests - that's super important for this strategy.

gl-johnson (approved):

PR #2921, opened 2023-08-24
Remove openid_connect gem test private key

andytinkham:
### Desired Outcome

Remove the private key that the openid_connect gem brings in with its test data.

### Implemented Changes

Added command to find any *.pem files in an openid_connect directory and delete them.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,12 +33,13 @@ COPY Gemfile \
 COPY gems/ gems/
 
 
-RUN bundle --without test development
+RUN bundle --without test development && \
+    # Remove private keys brought in by gems in their test data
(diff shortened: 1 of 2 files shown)


telday (commented):
Does this change also need to be included in `Dockerfile.fpm` as well?

andytinkham:
I looked at the Dockerfile.fpm file but it never runs bundle install any longer.  So as it stands, the gem is never installed that I see and thus there's no cert. A bigger question is whether Dockerfile.fpm should be running bundle install - if it should be, then the delete should be there, too.

telday (approved):
LGTM

PR #2932, opened 2023-08-30
Quote and wildcard gem directory when removing unneeded certs

andytinkham:
### Desired Outcome

The delete I added in #2921 didn't actually work because the find command was looking in sub-directories of the openid_connect-* dir, not the main dir. It then didn't find the private key and thus didn't delete anything.

### Implemented Changes

Add quotes and a wildcard to the directory name to find the top-level gem directory (and thus the *.pem file several layers underneath it).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,9 +35,9 @@ COPY gems/ gems/
 
 RUN bundle --without test development && \
     # Remove private keys brought in by gems in their test data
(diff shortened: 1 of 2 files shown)


andytinkham:
Dockerfile significantly changed as part of the FIPS stuff. I'll have to re-evaluate next week to determine if private keys are still present and try again.

PR #2933, opened 2023-08-31
Wrap discovery and callback calls

gl-johnson:
### Desired Outcome

Implement the wrapper methods introduced in https://github.com/cyberark/conjur/pull/2919 and the optional `ca-cert` config in https://github.com/cyberark/conjur/pull/2920 to allow custom certs to be provided in the OIDC authenticator config.

### Implemented Changes

- Added wrapper methods around:
  - OAuth discovery + JWKS retrieval
  - Identity resolution callback
  - OIDC configuration discovery
- Updated keycloak OIDC environment to use `ca-cert` variable instead of container truststore (includes integration tests)

### Connected Issue/Story

CNJR-2476

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

- [x] A follow-up issue to update official docs has been filed here: CNJR-2309

#### Behavior

#### Security

### DIFF
--- a/app/domain/authentication/o_auth/fetch_provider_keys.rb
+++ b/app/domain/authentication/o_auth/fetch_provider_keys.rb
@@ -29,8 +29,14 @@ module Authentication
       end
 
       def fetch_provider_keys
+        # We have to wrap the call to the JWKS URI separately to accomodate
+        # custom CA certs in authn-oidc
         jwks = {
-          keys: @discovered_provider.jwks
+          keys: Authentication::AuthnOidc::V2::Client.discover(
+            provider_uri: @provider_uri,
+            cert_string: @ca_cert,
+            jwks: true
+          )
         }
         algs = @discovered_provider.id_token_signing_alg_values_supported
         @logger.debug(LogMessages::Authentication::OAuth::FetchProviderKeysSuccess.new)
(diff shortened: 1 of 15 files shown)


app/domain/authentication/o_auth/fetch_provider_keys.rb
```diff
@@ -29,8 +29,14 @@ def discovered_provider
       end
 
       def fetch_provider_keys
+        # We have to invoke the custom discovery service to be able to fetch the jwks
+        # since it also requires the CA to invoke the jwks URI
         jwks = {
-          keys: @discovered_provider.jwks
```
gl-johnson, line 33: @john-odonnell This is still a WIP, but I'm curious if you have thoughts about this: turns out that `.jwks` makes a separate call to the provider's JWKS uri so the cert needs to be available for that as well.

As a workaround I added an optional `jwks` flag to the custom `Client.discover` which updates the lambda to include the `.jwks` call accordingly. Feels a bit hacky but better than the alternative of adding yet another wrapper imo. 

At the same time I'm wondering if is worth refactoring the wrapper functionality into a single method that can perform one of three operations: `[:callback, :discover, :jwks]`. Would require better validation of the inputs but may look cleaner than the duplicated code we have currently

john-odonnell: Otherwise, you're right - feels hacky, but I think this works for now. Makes me think that ditching the underlying OIDC client gem might make all this a bit cleaner, though.

john-odonnell (approved):
LGTM 👍

PR #2971, opened 2023-09-28
Design for single-use-token (SUT) authenticator

szh:
### Desired Outcome

Our first pass at a UI-PSM integration was rejected for security reasons and the more secure option likely requires us to create a new token authenticator for Conjur. Create a solution design for that and figure out how large of an effort it would be.

### Implemented Changes

Created a solution design document for a new single-use-token (SUT) authenticator

### Connected Issue/Story

CyberArk internal issue ID: CNJR-2514


jvanderhoof:
I don't love the term `authn-sut` (looks too much like a potentially derogatory word).  How about something like `authn-singleuse`, `authn-single-use` or `authn-token`?

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,283 @@
...
+
+### High-Level Design
+
+#### Workflows
+
+##### Basic SUT Workflow
+
+The following diagram shows the basic workflow for using a SUT to authenticate
+to Conjur. The first step is to call the
+`GET /authn-sut/<authenticator-name>/<account>/login` endpoint with the user's
```
jvanderhoof, line 78: What do you think about having this authenticator avoid the individual `service_id` pattern (as we do with `authn-gcp`). It doesn't seem like there will be any specific details that will necessitate having multiple "versions" of the single-use token authentication workflows.

The pattern we use in the GCP authenticator is to nest the key bits under the authenticator level:

```yml
- !policy
  id: conjur
  body: 
  - !policy:
    id: authn-sut
    body:
    - !webservice

    - !group 
      id: authenticatable
      annotations:
        description: Group with permission to authenticate using this authenticator

    - !permit
      role: !group authenticatable
      privilege: [ read, authenticate ]
      resource: !webservice

    - !webservice
      id: status
      annotations:
        description: Web service for checking authenticator status

    - !group
      id: operators
      annotations:
        description: Group with permission to check the authenticator status

    - !permit
      role: !group operators
      privilege: [ read ]
      resource: !webservice status
```

jvanderhoof:
Although it likely makes sense to limit the initial scope to Conjur Users, it's important to enable this functionality for Hosts as well. There are a number of interesting use cases that will be enabled with support for SUT for hosts.

For awareness, there is a push to remove the API key as an authentication mechanism. When that fully comes to fruition, I suspect that adding an `authn-apikey` authenticator is the best way to re-enable the use of API keys for authentication.  

As a final note, the decoupling of identity from authentication will likely require is to support "default" authentication mechanisms. It's quite possible we can support this behavior via Policy Factories, but we may need a more native solution at some point.

szh:
> I don't love the term `authn-sut` (looks too much like a potentially derogatory word). How about something like `authn-singleuse`, `authn-single-use` or `authn-token`?

I feel that token is too generic and confusing, since we already have auth tokens. Maybe authn-nonce or something similar?

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: I have no problem with that. I'll update the doc to remove the service-id pattern.

szh:
> Although it likely makes sense to limit the initial scope to Conjur Users, it's important to enable this functionality for Hosts as well. There are a number of interesting use cases that will be enabled with support for SUT for hosts.

Absolutely, I see no reason not to allow this for hosts.

> For awareness, there is a push to remove the API key as an authentication mechanism. When that fully comes to fruition, I suspect that adding an `authn-apikey` authenticator is the best way to re-enable the use of API keys for authentication.

To clarify, are you saying that API key authentication will no longer be enabled by default, and it'll need to be moved to an authn-apikey authenticator that will need to be enabled manually? How will this affect the single use token authenticator?

> As a final note, the decoupling of identity from authentication will likely require is to support "default" authentication mechanisms. It's quite possible we can support this behavior via Policy Factories, but we may need a more native solution at some point.

Again to clarify, do you mean we'll want to have this authenticator enabled by default without requiring the administrator to create a policy for it?

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,310 @@
...
+user's credentials.
+
+![3rd Party Client Workflow](./sut_3rd_party.png)
+
+##### PSM Integration Workflow
+
+The following diagram shows the workflow for using a SUT to authenticate to
+Conjur via the PSM integration. The first step is to call the
+`GET /authn-sut/<account>/login` endpoint with the user's
+credentials to generate a SUT. The SUT is then passed to the Conjur UI via a URL
```
niteshtaneja, line 105: Are there any security concerns around passing the SUT as a URL parameter?

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,310 @@
...
+parameter: `GET <conjur-ui-url>/ui/callback?sut=<sut>`. The Conjur UI then calls
+the `POST /authn-sut/<account>/authenticate` endpoint with
+the SUT to authenticate the user to Conjur and retrieve a short-lived auth
+token. The Conjur UI then starts a session for the user.
```
niteshtaneja, line 109: Whats the default TTL of the auth token? Configurable or fixed? - TBD?
What happens once it expires? Does the user gets logged out of the UI?

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,310 @@
...
+
+We cannot use JWTs for the SUT as we need the ability to revoke the SUT after it
+is used, a feature not easily implemented with JWTs. This would make it
+impossible to prevent replay attacks.
+
+##### Open Questions
+
+- Is it possible to restrict use of the SUT to a specific IP range (like with
+  hostfactory tokens) or other criteria to limit risk if the SUT is leaked?
+- Is it better to actually delete the SUT from the database once it is used or
```
niteshtaneja, line 220: same question for expired ones

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: It's not the most secure way of passing around sensitive data, which is why we're suggesting using a single-use, short-lived token instead of using the 8 minute auth token as was previously considered. This way even if it's leaked the window for exploitation is low. Currently PSM does not support POST requests so this the only way we can do it that will work for them.

design/authenticators/authn_sut/authn_sut_solution_design.md
niteshtaneja: I am assuming this is regardless of the session activity? As long as PSM is okay with it, we should be good.

szh: Correct.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,310 @@
...
+end
+```
+
+#### Authenticator
+
+The SUT authenticator will be implemented as a new authenticator plugin. The
+implementation will follow the existing authenticator plugin pattern, as
+documented in [AUTHENTICATORS.md](../AUTHENTICATORS.md)
+
+## Security
```
andytinkham, line 256: I'm thinking about taking a page from OIDC to add security here. What if PSM (or any client) had to generate a unique code and passed the hash of that code to the authn-sut/login endpoint. Then, PSM sends the actual code to Conjur UI. Conjur UI includes that along with the SUT to Conjur. Conjur applies the same hash (we'd have to pick a single algorithm - maybe SHA256) to the code and only honors the SUT if that matches the request code.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,310 @@
...
+endpoint to authenticate the user to Conjur and retrieve a short-lived auth
+token.
+
+![Basic SUT Workflow](./sut.png)
+
+##### 3rd Party Client Workflow
+
+The following diagram shows the workflow for using a SUT to authenticate to
+Conjur on behalf of a 3rd party client. The first step is to call the
+`GET /authn-sut/<account>/login` endpoint with the user's
```
andytinkham, line 90: Do we want to have registered client IDs and require the client to sign this request?  We'd have to have a keypair set up to allow the client to request a SUT as part of the configuration, but it would stop the endpoint being used by non-authorized services.

andytinkham (changes requested):

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,310 @@
...
+    annotations:
+      description: Group with permission to authenticate using this authenticator
+
+  - !permit
+    role: !group authenticatable
+    privilege: [ read, authenticate ]
+    resource: !webservice
+```
+
+and the `clients` group can be used to entitle a user(s) to use the
```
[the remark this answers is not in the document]
szh: Yes, I'll fix that.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,310 @@
...
+
+Assumptions: Defining the assumptions being made in preparing this Solution
+Design. Detailing why these assumptions are valid and/or acceptable.
+Prerequisites: Do we need additional tooling to develop this feature? Do we need
+additional tooling to test this feature in our pipeline? What needs to be in
+place to build this feature?
+
+## Out-of-scope
+
+- SUTs will not be able to be used to authenticate to Conjur via the CLI
```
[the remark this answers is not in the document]
szh: There's just no need for it. I'll clarify that as well.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,310 @@
...
+
+The SUT authenticator will be the mechanism used to authenticate the user to
+Conjur when they click "Connect" on a Conjur instance in PVWA. PSM will generate
+a SUT for the user using their stored username and password or API key. PSM will
+then pass the SUT to the Conjur UI via a URL parameter. The Conjur UI will use
+the SUT to authenticate the user to Conjur and start a session.
+
+## Assumptions and Prerequisites
+
+Assumptions: Defining the assumptions being made in preparing this Solution
```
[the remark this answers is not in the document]
szh: Ah yes, I left the template text here. I'll update this.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,310 @@
...
+##### Log Messages
+
+##### Error Log level
+
+| | Scenario | Log message |
+|--- |-----------------------------------------------------------------------  |----------------------------------------------------------------------------------------|
+| 1  | Authenticator is not enabled (in DB/ENV)                                | Authenticator '{0-authenticator-name}' is not enabled                                                 |
+| 2  | Webservice is not defined in a Conjur policy                            | Webservice '{0-webservice-name}' not found                                          |
+| 3  | User or host is not permitted to authenticate with the webservice               | '{0-role-name}' does not have 'authenticate' privilege on {1-service-name}             |
+| 4  | User or host is not defined in Conjur                                           | '{0-role-name}' wasn't found                                                           |
```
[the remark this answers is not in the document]
szh: These are log messages that are not returned to the user. Same as other authenticators such as https://github.com/cyberark/conjur/blob/master/design/authenticators/authn_azure/authn_azure_solution_design.md#error-log-level

design/authenticators/authn_sut/authn_sut_solution_design.md
andytinkham: Filling this out more explicitly - I'm modeling this on PKCE in OIDC. 
1) PSM generates a GUID for a specific single use token request.
2) PSM SHA256 hashes the GUID and sends the resulting hash value and the transformation method (currently only "sha256" supported, but this will make it easier for us to introduce post-quantum crypto handling or other algorithmss later) to the endpoint for the SUT.
3) The single use endpoint includes the received hash value and transformation methods as additional columns in the database where the token is stored. The token is returned to PSM.
4) PSM sends the token and the original (unhashed) GUID to Conjur UI / Conjur. The verification process includes a step that takes the GUID, applies the transformation method associated with the token, and verifies that it gets the same result as was written from the original request. If it does not get the same result, the token is rejected and declared invalid immediately. 

Key pieces:
- The GUID itself is never sent anywhere until it's used with the token to prevent interception.
- The GUID ties a single use token to a particular request, so the token can't be used elsewhere.
- The transformation method must be one way, so even if someone got a hold of the database, they couldn't reverse engineer the GUID and gain access to using the token. 
- Even though we only support one transformation method at the start, we have the mechanism in place to allow other choices in the future to make upgrades and introducting new functionality easier.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+
+We will need two new API endpoints for the SUT authenticator:
+
+##### `GET /authn-sut/<account>/login`
+
+This endpoint will be used to generate a SUT for a user. The user will provide
+their username and password (or API key) in the `Authentication` header using the
+standard format for HTTP Basic authentication.
+
+Additionally, the client will provide a Code Challenge and Code Challenge Algorithm
```
jvanderhoof, line 188: Headers are intended to provide metadata about the corresponding body. To me, the hashing algorithm makes sense, but as the code challenge value is critical to the overall authentication request, it should be part of the login body (with the other key bits: username/password or host/API key).

I suspect we're proposing header values to avoid the use of the `POST` action.  In this case, I'd argue that it's more appropriate to use `POST` because the result is the creation of a single-use token.  In the normal login flow, the result is that user's API key, which already exists.

szh: Yes, you got it exactly. Use of a header is intended to allow us to use a GET request to be consistent with the authn and authn-ldap routes. I'm fine with making it a POST request, just knowing that it may cause some confusion.

jvanderhoof: What do you think about changing this to a `POST`: 

`/authn-sut/<account>/login`

**Headers**
|Header|Sample Value|
|-|-|
|Code-Challenge-Algorithm|sha256|
|Authentication|Zm9vOmJhcg== (base64 encoding of "foo:bar")|

**Body**
Code-Challenge-Value

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+For more details on the code challenge, see the [Code Verifier and Code Challenge](#code-verifier-and-code-challenge) section.
+
+The endpoint will return a SUT that can be used to authenticate the user to Conjur.
+
+The format of the URL matches the existing authenticator endpoint for
+authn-ldap.
+
+##### `POST /authn-sut/<account>/authenticate`
```
jvanderhoof, line 203: This request should also include the hashing algorithm as an optional header

szh: This shouldn't be necessary to send in this request since it's stored in the database.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+### Low Level Design
+
+#### Data Model
+
+The SUT will be a cryptographically secure, random string. It will be generated
+by the Conjur server and stored in the database. The SUT will be associated with
+the user and expiration time. The SUT will be deleted once it is utilized
+to authenticate the user to Conjur.
+
+When we store the SUT in the database, we will store a salted hash of the SUT instead
```
jvanderhoof, line 234: Salting is a pattern to bring uniqueness to commonly used passwords. Essentially it removes the collision problem of hashed passwords. Given we'll be generating unique values to begin with, I don't believe the additional complexity of salting provides any additional security.

I'd recommend we take a similar approach for our single use tokens as we do when generating API keys:

```ruby
# Decomposed to expose the implementation details
require 'base32/crockford'

def random_token
  # note, this removes the down-casing to reduce collision potential
  OpenSSL::Random::random_bytes(32).unpack("N*").map{|i| Base32::Crockford::encode(i)}.join
end
```
If we store the above value hashed using SHA 256, there's essentially no way to get back to the original values even if the database was exposed.
I'd recommend we

szh: Good point. @andytinkham you originally [suggested](https://github.com/cyberark/conjur/pull/2971#discussion_r1349375507) salting. What do you say?

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+
+Initially we will support the SHA-256 algorithm for the code challenge. In the
+future we may want to support additional algorithms.
+
+The code challenge and code verifier flow are modeled after OIDC PKCE.
+
+##### Open Questions
+
+- Is it possible to restrict use of the SUT to a specific IP range (like with
+  hostfactory tokens) or other criteria to limit risk if the SUT is leaked?
```
jvanderhoof, line 283: We could achieve a similar outcome by inspecting IP range restrictions on the target role when the single-use token is exchanged for an auth token.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+- Should we provide a way for clients to register a key pair to use for
+  signing the SUT generation request? This would allow us to verify that the
+  request came from a trusted client.
+- Should we create a job to clean up expired SUTs?
```
jvanderhoof, line 287: Absolutely.  We can use Cron as it's already present in Conjur and Conjur Enterprise.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+
+The code challenge and code verifier flow are modeled after OIDC PKCE.
+
+##### Open Questions
+
+- Is it possible to restrict use of the SUT to a specific IP range (like with
+  hostfactory tokens) or other criteria to limit risk if the SUT is leaked?
+- Should we provide a way for clients to register a key pair to use for
+  signing the SUT generation request? This would allow us to verify that the
+  request came from a trusted client.
```
jvanderhoof, line 286: I really like this, but it'll certainly increase the complexity for the service generating tokens. We could do something like:

- Service which needs single-use tokens, generates a key pair.
- Public key is uploaded into the single-use token authenticator (stored in a variable)
- Service requests token as described.
- Conjur encrypts single-use token using public key.
- Service decrypts using private key.
- Service sends decrypted single-use token to Conjur.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+- Should we create a job to clean up expired SUTs?
+- Should we create a mechanism to revoke SUTs before they expire (other than
+  manually deleting them from the database)?
+
+##### Example Database Table
+
+| id | role_id | guid_hash | code_challenge | code_challenge_algorithm | expires_at |
+|----|---------|-----------|----------------|--------------------------|------------|
+| 1  | user:john.doe | lkduj...  | 3iod... | sha256 | 2023-09-28 12:08:23 |
+| 2  | user:demo.user | 2k1g2...  | kld80... | sha256 | 2023-09-28 12:09:45 |
```
jvanderhoof, line 296: A couple of comments

- [semantic] `role_id` should have an account in it.
- I'd recommend changing `guid_hash` to `token_hash` to avoid leaking the implementation details. (I'd recommend avoiding GUIDs in favor of base32 or base58 encoded strings of a pre-determined length).
- The column `id` needs to be designed with care to ensure customers can't exhaust the available values as the rows increment.  It might make more sense to use `token_hash` as the unique key (which guarantees uniqueness of available single-use tokens at any point in time).

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+Sequel.migration do
+  change do
+    create_table :authn_sut_tokens do
+      primary_key :id
+      foreign_key :role_id, :roles, type: String, null: false, on_delete: :cascade
+      String :guid_hash, null: false, unique: true
+      String :code_challenge, null: false
+      String :code_challenge_algorithm, null: false
+      DateTime :expires_at, null: false
+    end
```
jvanderhoof, line 311: This table is missing an index.  I'd recommend creating an index on the on `guid_hash` and `expires_at`.  This will allow us to run something like this with a level of efficiency:

```sql
select role_id, code_challenge, code_challenge_algorithm from authn_sut_tokens where guid_hash = '...' and expires_at >= NOW();
```

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+| Authenticate with SUT - Used SUT | Sad | Verify that a user cannot authenticate to Conjur with a used SUT | | |
+| Authenticate with SUT - Authenticator not defined in policy | Sad | Verify that a user cannot authenticate to Conjur with an authenticator that is not defined in policy | | |
+| Authenticate with SUT - Invalid authenticator | Sad | Verify that a user cannot authenticate to Conjur with an invalid authenticator | | |
+| Authenticate with SUT - Disabled authenticator | Sad | Verify that a user cannot authenticate to Conjur with a disabled authenticator | | |
+| Authenticate with SUT - Authenticator not enabled for user | Sad | Verify that a user cannot authenticate to Conjur with a disabled user | | |
+| Authenticate with SUT - Missing code verifier | Sad | Verify that a user cannot authenticate to Conjur without a code verifier | | |
+| Authenticate with SUT - Invalid code verifier | Sad | Verify that a user cannot authenticate to Conjur with an invalid code verifier | | |
+| Generate & Authenticate - Multiple SUTs | Sad | Generate two SUTs for a user. Verify that the first one cannot be used after the second is created. | | |
```
jvanderhoof, line 365: I'd like to see us minimize the use of Cucumber as a test mechanism in favor of Controller-level RSpec tests. Those will be much more efficient and achieve similar results.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+  - !group 
+    id: authenticatable
+    annotations:
+      description: Group with permission to authenticate using this authenticator
+
+  - !permit
+    role: !group authenticatable
+    privilege: [ read, authenticate ]
+    resource: !webservice
+```
```
jvanderhoof, line 167: I'd recommend we add a setting for defining the single-use token TTL. If not present we can use something reasonable, like 30 seconds.

design/authenticators/authn_sut/authn_sut_solution_design.md
jvanderhoof: Good call!

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+is the case, the endpoint will return a short-lived Conjur auth token.
+
+Example request body:
+
+```json
+{
+  "single_use_token": "lkduj...",
+  "code_verifier": "3iod..."
+}
+```
```
jvanderhoof, line 218: Mind adding the failure responses as well?

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,384 @@
...
+- !grant
+  role: !group conjur/authn-sut/authenticatable
+  member: !user <username>
+```
+
+#### API
+
+We will need two new API endpoints for the SUT authenticator:
+
+##### `GET /authn-sut/<account>/login`
```
jvanderhoof, line 182: Mind adding the failure responses as well?

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: Do you think this is necessary for the initial implementation? I added this under out-of-scope in an effort to keep this implementation small.

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: Is that necessary? It should be the same as any other authenticator and should not require any custom code.

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: Where is Cron used in Conjur?

design/authenticators/authn_sut/authn_sut_solution_design.md
jvanderhoof: I'd love to get feedback from @niteshtaneja about if we can defer.

design/authenticators/authn_sut/authn_sut_solution_design.md
jvanderhoof: It's really important we capture all the scenarios in these docs. Each authenticator has a specific set of bits that can go wrong that are unique to that authenticator, and it's important we capture that.  

For example, when running in debug mode, I'd expect to see that exchanging an expired token failed because the token was expired. It's helpful to think about that message before we write the implementation to ensure we don't leak security related data in log messages.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,409 @@
...
+
+For more details on the code challenge, see the [Code Verifier and Code Challenge](#code-verifier-and-code-challenge) section.
+
+The endpoint will return a SUT that can be used to authenticate the user to Conjur.
+
+The format of the URL matches the existing authenticator endpoints for authn and
+authn-ldap. However, the request action for authn-sut is `POST` instead of `GET`
+since we need to send a code challenge in the request body.
+
+##### `POST /authn-sut/<account>/authenticate`
```
jvanderhoof, line 225: As I was thinking more about the interface of the other authenticators, I realized they typically include the role ID: 

- `POST /authn-jwt/{service-id}/<account>/[{host-id}]/authenticate`
- `POST /authn-azure/{service-id}/{account}/{host-id}/authenticate`
- `POST /{authenticator}/{account}/{login}/authenticate`

I think it would be valuable to follow this convention and require the role identifier (host or username) in the URI.

design/authenticators/authn_sut/authn_sut_solution_design.md
jvanderhoof: Only a small portion of authentication failure is handled outside the authenticator.  There's no need to re-document those.  I'm thinking about errors specific to Single-Use tokens. If the above error messages are the final, what do you think about including a debug level and error code for each error? That'll at least give a sense of the expected errors and the communication back to the customer.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,409 @@
...
+
+This endpoint will be used to authenticate a user to Conjur using a SUT. The
+user will provide the SUT and code verifier in the request body.
+The endpoint will verify that the SUT is valid and has not expired, and that
+the code verifier matches the code challenge associated with the SUT. If this
+is the case, the endpoint will return a short-lived Conjur auth token.
+
+Example request body:
```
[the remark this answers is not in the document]
szh: I don't love the idea of passing what is meant to be a secret value in the URL. @andytinkham what are your thoughts?

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: Hmm. We don't use this format for authn or authn-ldap, which are the only two authenticators that also have a /login route and what I've been mostly basing this on. Perhaps it's because the username is part of the HTTP Basic auth header.

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: I'm not seeing where in the code we have custom errors for the other authenticators. Can you please give me some pointers so I have what to compare with?

design/authenticators/authn_sut/authn_sut_solution_design.md
andytinkham: As long as the token is of sufficient length, we should be OK not salting it.  Doing the same as the API keys should be OK.

design/authenticators/authn_sut/authn_sut_solution_design.md
jvanderhoof: Not login, but the default authenticator (role and API key) uses this pattern: https://docs.cyberark.com/conjur-enterprise/latest/en/Content/Developer/Conjur_API_Authenticate.htm?tocpath=Developer%7CREST%C2%A0APIs%7C_____3

szh: Ah I see, I missed that nuance. I'll update this.

jvanderhoof: I'd also make the argument that mandating the role identifier in the URI makes the implementation slightly more secure by forcing the requestor to know the role and the role's single-use token before the request. If the provided role identifier and the single-use token role don't match, the authentication should fail.

design/authenticators/authn_sut/authn_sut_solution_design.md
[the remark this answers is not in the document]
jvanderhoof: An excellent point. Reviewing the OpenID Connect PKCE flow, the submission is a POST.  I think it makes sense to proceed with the recommended JSON document below.

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: Good point. I've updated this.

design/authenticators/authn_sut/authn_sut_solution_design.md
niteshtaneja: yeah I had the same question but then Shlomo said that PSM team confirmed that the default TTL is good enough for them. Given that, it should be okay to defer it from v1 implementation.
Just to clarify though- the default is going to be 8 mins and NOT 30s,  correct?

szh: 30s is for the single use token which PSM will use to log in. Once PSM logs in, it will have the 8 minute auth token.

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: For the record, we discussed this and clarified that the intention is to have more detailed debug logs, not HTTP responses. I'm adding those now.

design/authenticators/authn_sut/authn_sut_solution_design.md
niteshtaneja: Sorry I mistook this to be making the auth token configurable. Yes, from what I remember from our conversations with the PSM team, a default 30s single-use token should work for them so we can skip making it configurable for the initial implementation.

design/authenticators/authn_sut/authn_sut_solution_design.md
jvanderhoof: My mistake. Cron is part of Conjur Enterprise, not Conjur. Two options that come to mind are:

- Remove expired single-use tokens during creation, exchange, or both.
- Create a service like the variable rotation service to run and periodically remove expired tokens

szh: OK got it. For now I'll update this doc to allow for some extra points for figuring that out during the implementation.

jvanderhoof (approved):
Looks good!

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,433 @@
...
+critical that the implementation is secure. Threat modeling should be performed
+before implementation begins to identify potential security issues and ensure
+that they are addressed. There should be an additional security review of the
+implementation before it is merged.
+
+Much of the security of this feature will be provided by the underlying
+authenticator architecture, so we only need to be concerned with the security of
+the SUT and the implementation of the SUT authenticator plugin.
+
+## Testing
```
adamouamani, line 355: @szh nice list of test scenarios.  I had a couple comments/questions..

1) what if ldap is enabled for Conjur-UI login by default or authn-oidc?
2) check that tokens are not logged (even though they are short lived)
3) how about authn-temp as credential name?
4) will there be a status endpoint for authn_sut?
5) could you define more than one authn_sut authenticator?

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,433 @@
...
+> into the main UI dashboard, without showing the initial login screen. More
+> details will be added in the solution design (TBD).
+>
+> The overall UX will remain the same where users will continue to use the
+> existing workflow of logging onto the PVWA, clicking connect on a defined
+> Conjur instance 1 and connecting through PSM straight into the Conjur UI
+> console.
+
+The SUT authenticator will be the mechanism used to authenticate the user to
+Conjur when they click "Connect" on a Conjur instance in PVWA. PSM will generate
```
andytinkham, line 42: Minor nit: PSM isn't generating the SUT. Conjur generates the SUT and passes it back to PSM.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,433 @@
...
+| 4  | User or host is not defined in Conjur                                           | '{0-role-name}' wasn't found                                                           |
+| 5  | Login error                               | Login error: ...                                     |
+| 6  | Missing SUT in /authenticate request body | Single use token not provided    |
+| 7  | SUT not found for user, or expired | Single use token is invalid    |
+| 8  | Missing code challenge in /login request body | Code challenge not provided    |
+| 9  | Invalid code challenge in /login request body | Code challenge not valid    |
+| 10 | Invalid code challenge algorithm in /login request headers | Code challenge algorithm not supported    |
+| 11 | Missing code challenge algorithm in /login request headers | Code challenge algorithm not provided    |
+| 12 | Missing code verifier in /authenticate request body | Code verifier not provided    |
+| 13 | Code verifier does not match code challenge | Code verifier does not match code challenge    |
```
andytinkham, line 143: I might make this error message something like "Invalid code verifier".  No need to disclose anything that could be construed as the verification method necessarily.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,433 @@
...
+
+###### Debug Log level
+
+|    | Scenario                                              | Log message                                                                            |
+|--- |-------------------------------------------------------|----------------------------------------------------------------------------------------|
+| 1  | After validating SUT | Single use token for {0-role-name} has been validated successfully                |
+| 2  | Attempt to use expired SUT | Single use token for {0-role-name} has expired    |
+| 3  | Attempt to use used SUT | Single use token for {0-role-name} has already been used    |
+| 4  | Attempt to use SUT for different user | Single use token for {0-role-name} is not valid for {1-role-name}    |
```
andytinkham, line 152: I wonder if this should be only a debug message. This might be an indicator of an active attack and thus something we should be surfacing to the audit DB and maybe even at a Warning log level.

design/authenticators/authn_sut/authn_sut_solution_design.md
```diff
@@ -0,0 +1,433 @@
...
+| Scenario | Path Type | Test Purpose | Failure Scenarios | Details |
+|----------|-----------|--------------|-------------------|---------|
+| Generate SUT | Happy | Verify that a SUT can be generated for a user | | |
+| Generate SUT - Invalid credentials | Sad | Verify that a SUT cannot be generated for a user with invalid credentials | | |
+| Authenticate with SUT | Happy | Verify that a user can authenticate to Conjur with a SUT | | |
+| Authenticate with SUT - Invalid SUT | Sad | Verify that a user cannot authenticate to Conjur with an invalid SUT | | |
+| Authenticate with SUT - Expired SUT | Sad | Verify that a user cannot authenticate to Conjur with an expired SUT | | |
+| Authenticate with SUT - Invalid code verifier | Sad | Verify that a user cannot authenticate to Conjur with an invalid code verifier | | |
+| Authenticate with SUT - Incorrect user | Sad | Verify that a user cannot authenticate to Conjur with a SUT for a different user | | |
+| Generate & Authenticate - Multiple SUTs | Sad | Generate two SUTs for a user. Verify that the first one cannot be used after the second is created. | | |
```
andytinkham, line 400: In the spirit of Jason's desire to minimize cucumber tests stated above, do we need all these Sad path tests here as well as in unit tests? 1 happy path test (generate the SUT and authenticate) and 1 sad path test to make sure errors make it to where they should be seem like enough given the set of unit tests.

andytinkham (approved):
I added some minor things and 1 thing (audit log & higher-than-debug logging of one user trying to use another user's token) that I think is important, but not enough to merit another cycle of review.  If you feel that merits changes that invalidate this approval, I'll do approve again quickly.

design/authenticators/authn_sut/authn_sut_solution_design.md
szh: I'll clarify these. I think authn-temp is too vague - it makes it sound like the authenticator itself is temporary.

szh: I'm actually not sure about authn-ldap - I'll need some clarification from someone more familiar with UI project.

PR #2972, opened 2023-09-28
Bump google/cloud-sdk to 448.0.0 in tests

andytinkham:
### Desired Outcome

Remove base images that contain vulnerable versions of libwebp to remove possibility of exploit of CVE-2023-5129 / CVE-2023-4863.

### Implemented Changes

Upgrade google/cloud-sdk to latest version (448.0.0)

### Connected Issue/Story

CyberArk internal issue ID: CONJSE-1717

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/conjur#2827](https://github.com/cyberark/conjur/pull/2827)
 
 ### Security
+- Upgrade google/cloud-sdk in ci/test_suites/authenticators_k8s/dev/Dockerfile/test
+  to use latest version (448.0.0)
+  [cyberark/conjur#2972](https://github.com/cyberark/conjur/pull/2972)
 - Support plural syntax for revoke and deny
   [cyberark/conjur#2901](https://github.com/cyberark/conjur/pull/2901)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   [cyberark/conjur#2827](https://github.com/cyberark/conjur/pull/2827)
 
 ### Security
+- Upgrade google/cloud-sdk in ci/test_suites/authenticators_k8s/dev/Dockerfile/test
+  to use latest version (448.0.0)
+  [cyberark/conjur#2972](https://github.com/cyberark/conjur/pull/2972)
```
micahlee, line 54: Usually the changelog is reserved for changes that have an end-user impact. Giving this is a container for tests, it's not actually needed here.

micahlee (approved):

CHANGELOG.md
andytinkham: Fair - are you suggesting I remove it, or just commenting for next time? (I included it largely because the CVE is getting publicity, but since it's only in test files, it probably doesn't matter either way.)

micahlee: No, I don't think you need another push just to remove it. If you end up making any other updates you could pull it out, but if this build is good, it's fine to stay. Thanks!

PR #2983, opened 2023-10-09
Add build tools to policy-parser image for compiling bigdecimal

hughsaunders:
Recently active support was released (7.1.0), this new version was picked up
by an unpinned dependency in conjur-policy-parser. Active support 7.1.0
adds a dependency on bigdecimal which has a native extension and couldn't be 
built within the policy parser test docker image due to lack of dev tools.

This commit adds the necessary dev tools and libs for big decimal's native
extensions to be compiled.

### DIFF
--- a/gems/policy-parser/Dockerfile.test
+++ b/gems/policy-parser/Dockerfile.test
@@ -1,4 +1,4 @@
-FROM cyberark/ubuntu-ruby-fips:latest
+FROM cyberark/ubuntu-ruby-builder:latest
 
 RUN mkdir /src
 WORKDIR /src
@@ -11,3 +11,5 @@ COPY lib/conjur-policy-parser-version.rb lib/conjur-policy-parser-version.rb
 ENV BUNDLER_VERSION=2.4.14
 RUN gem install bundler -v ${BUNDLER_VERSION} && \
     bundle install
+
+ENV OPENSSL_CONF=/usr/lib/ssl/openssl.cnf


gems/policy-parser/Dockerfile.test
```diff
@@ -7,6 +7,8 @@ COPY Gemfile Gemfile
 COPY conjur-policy-parser.gemspec conjur-policy-parser.gemspec
 COPY lib/conjur-policy-parser-version.rb lib/conjur-policy-parser-version.rb
 
+RUN apt-get update && apt-get install -y libffi-dev build-essential libyaml-dev
```
tarnowsc, line 10: Hi @hughsaunders, did you considered using the cyberark/ubuntu-ruby-builder:latest as a base image it comes with everything needed for installing the native extensions. Additionally it has FIPS mode disabled which makes bundler install faster (since bundler uses md5 based optimization while downloading gems).
To enable the FIPS mode, after installing all the gem, you could simple add:
```
ENV OPENSSL_CONF=/usr/lib/ssl/openssl.cnf
```

hughsaunders: Hey @tarnowsc, I didn't know of this image, thanks will try it 👍

hughsaunders: Pushed a commit with this suggestion

tarnowsc (approved):
LGTM

hughsaunders:
Rebuilding due to k8s failure

hughsaunders:
Yay the build passed, so this just needs a review from @cyberark/conjur-core-team then can be merged

PR #2988, opened 2023-10-11
AuthnOIDC V2: write custom certs to non-default directory

john-odonnell:
### Desired Outcome

Authn-OIDC with `ca-cert` variable fails on Conjur Enterprise.

```
Errno::EACCES: Permission denied @ rb_file_s_symlink - (/tmp/d20231006-289-9u7rno/conjur-oidc-client.0.pem, /usr/lib/ssl/certs/bb3ed97b.0)
```

When Authn-OIDC V2 is configured with a `ca-cert` variable, it writes the provided certificate content into a temporary file, and creates a symlink to the file in the Conjur container's default certificate directory, meaning it would be trusted for outbound requests to the configured OIDC provider. After a successful request, the symlinks are deleted, and the temp files destroyed.

In Conjur Enterprise, the Conjur server is running as a user `conjur`, which lacks write permissions on the container's default cert store.

### Implemented Changes

Instead of creating symlinks in the default cert store, instead create them in the Conjur project directory tree - the Conjur process is able to write to the `${conjur_root}/etc/certs` in both Conjur Open Source and Enterprise.

In the previous solution iteration, if a custom certificate was established in the `ca-cert` variable but connecting to the OIDC provider did not require that particular certificate, Authn-OIDC could fall back on any trusted certificate in the default cert store. No longer - if the `ca-cert` variable is set, it must contain the full certificate chain required for connecting to the OIDC provider. No certificates included in the system's default store will be considered.

### Connected Issue/Story

CyberArk internal issue ID: CNJR-2844

### Definition of Done
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

### DIFF
--- a/app/domain/authentication/authn_oidc/v2/client.rb
+++ b/app/domain/authentication/authn_oidc/v2/client.rb
@@ -18,6 +18,19 @@ module Authentication
           @logger = logger
         end
 
+        # Writing certificates to the default system cert store requires
+        # superuser privilege. Instead, Conjur will use ${CONJUR_ROOT}/tmp/certs.
+        def self.default_cert_dir(dir: Dir, fileutils: FileUtils)
+          if @default_cert_dir.blank?
+            conjur_root = __dir__.slice(0..(__dir__.index('/app')))
+            @default_cert_dir = File.join(conjur_root, "tmp/certs")
+          end
+
+          fileutils.mkdir_p(@default_cert_dir) unless dir.exist?(@default_cert_dir.to_s)
+
+          @default_cert_dir
+        end
+
         def oidc_client
           @oidc_client ||= begin
             issuer_uri = URI(@authenticator.provider_uri)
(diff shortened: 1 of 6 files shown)


app/domain/authentication/authn_oidc/v2/client.rb
```diff
@@ -18,6 +18,19 @@ def initialize(
...
+        def self.default_cert_dir(dir: Dir, fileutils: FileUtils)
+          if @default_cert_dir.blank?
+            conjur_root = __dir__.slice(0..(__dir__.index('/app')))
+            @default_cert_dir = File.join(conjur_root, "tmp/certs")
+          end
+
+          fileutils.mkdir_p(@default_cert_dir) unless dir.exist?(@default_cert_dir.to_s)
+
+          @default_cert_dir
+        end
```
jvanderhoof, line 32: Any particular reason we're using instance variables instead of variables scoped to the method?

Ah, I think I understand what you're trying to accomplish here.  We only want to set the `@default_cert_dir` if it hasn't yet been set.  This is in a singleton so you can pass it into the `callback_with_temporary_cert` method below.

jvanderhoof (approved):
I believe this code is good enough for now.  I'll work with @jtuttle, @andytinkham, and @adamouamani, and the rest of the C&I team on a refactor to cleanup the implementation and improve testability of the code.

